PR #316, opened 2011-03-08
Suggest to add previous and next page pattern config

daning:
n vimperator I can config previous and next page patterns, so I can add previous related words for other language, for example Chinese "下一页","上一页". But in vimium it sames that the pattern is written in code, which can't be configed by user.
I do some work and modify some code, now it can be configed.


balalaXD:
ipython
---
```
In [1]: u'\u00BB'                                                                                      
Out[1]: '»'

In [2]: u"\u00AB"                                                                                       
Out[2]: '«'
```

Issue #544, opened 2012-05-31
Allow remapping the space key in Vimium's settings


nXqd:
I want to map space to enterFindMode. I don't know what is the space character I can use to map. Do we have any documentation on key mapping ?
Maybe i'm missing something

kabeero:
fixed in #1675 

use
```
map <space> command
```

PR #665, opened 2012-09-25
Allow remapping the space key in Vimium's settings

ghost:
Implement #544


DenisGorbachev:
@edygar same here, have you found a workaround?

edygar:
@DenisGorbachev, only one: the surrender. I got used to `f.`

DenisGorbachev:
@philc Is it possible to reopen this issue as a bug, or is the surrender our only option? :)

smblott-github:
@mrmr1993. Not having `<space>` just looks like an oversight, to me.  Am I overlooking something?

(The original repo for this PR has been deleted.  I created and tested a dup here: 81a22885c14b3cb12dc11484b01a689f1e0c2e46 -- it's two lines! Seems fine.)

mrmr1993:
I'm not a fan of how it's done, but it looks like it should work fine. The help dialog is going to look like it isn't showing a key for those bindings (will just print a space), which I also don't like. Logging will also suffer from this issue. But it will work.

I'd rather we handle space as `<space>` everywhere for consistency, which will have similar weight as a change. This has the obvious issue with passkeys (that it can't be disabled), but that's easily fixed by the suggestion I made and implemented in one of the open PRs. I'm happy to implement this, if this is the way we want to go.

philc:
I think we treat space as `<space>`.

mrmr1993:
@philc no, we don't.

smblott-github:
It turns out `<space>` shows up fine on the help page.  I'm not sure we need an extra layer of key-handling complexity to fix just this, which looks like a simple oversight.

Antoine-Patel:
> It turns out `<space>` shows up fine on the help page.

It seems like it's not the case for me...

I have a `map <space>j LinkHints.activateMode` in my Vimium options. In Vimium Help, it shows as `j	:	Open a link in the current tab` instead of `<space>j	:	Open a link in the current tab`. It's the same for any `<space>...` binding. It's also the case for binding where `<space>` is not the first key, eg `g<space>`.

Here's a screenshot below. It looks like I mapped two times `j` (`Scroll down` and `Open a link in the current tab`), it's confusing. It also shows the problem for `map g<space> scrollToBottom`.

vimium-mapping-space-help-issue

I'm using chromium `Version 45.0.2454.101 Built on Ubuntu 14.04, running on LinuxMint 17.2 (64-bit)` with vimium `Version 1.53`.

mrmr1993:
> It turns out `<space>` shows up fine on the help page.

That has never been a possibility, given the implementation. I'm sorry I wasn't paying enough attention to this to stand my ground.

@smblott-github if I implement what I proposed in [this comment above](https://github.com/philc/vimium/pull/665#issuecomment-104009164), could you merge it? It shouldn't result in any change in LOC.

For an extra LOC I can also retain compatability with passkeys including spaces, although I still maintain that #1368 and related PRs are a far better use of the space character (and could be easily converted to in `lib/settings.coffee`).

smblott-github:
What problem are you intending to solve, @mrmr1993?

(And, yes, there is an issue with `<space>` not showing up properly on the help page.)

mrmr1993:
> What problem are you intending to solve, @mrmr1993?

I want to keep key names in settings and key names in code as consistent as possible, to avoid bugs like this and make debugging easier (it's easy to overlook a space).

mrmr1993:
I've implemented my solution in [this branch](https://github.com/philc/vimium/compare/master...mrmr1993:map-space?expand=1). I haven't tested it as I'm on mobile, but this is how I personally would want the implementation to look (except for the passkeys hack).

smblott-github:
That looks fine.

(Except I'm not sure what the implications of the passkeys bit are.)

mrmr1993:
I'm at a computer now. It turns out my solution in that branch doesn't work correctly; I'd assumed that space was triggered by `keydown`, when it's actually triggered by `keypress`. For what I wanted to do, we'd need larger changes to the key handling code. Sorry for the noise.

Issue #499, opened 2012-03-06
Escape key caught by Vimium


ripdajacker:
Vimium catches the keystroke for escape, even when out of input mode. The easiest way to show this is on Facebook chat, where you in the old version could press escape twice and close the chat.

yoshuawuyts:
God this is annoying, have run into it multiple times now. Would be cool if it could be fixed, without http://stackoverflow.com/questions/12927752/keydown-keyup-events-not-detecting-the-escape-key-being-pressed-on-input-text-fi I would've been lost.

smblott-github:
What behaviour would you like to see from `Esc`?

ZombieHippie:
Vimium does not allow propagation of any webpage's Escape keyup bound
events. If you are not engaged in a Vimium activity(like `f` for find link)
then there it only seems to inhibit the abilities of the user.

smblott-github:
> Vimium does not allow propagation of any webpage's Escape keyup bound
events

Really?

Test:
```coffeescript

watch = (type) ->
  document.addEventListener type, (event) ->
    para = document.createElement "p"
    para.innerHTML = type
    document.body.appendChild para

watch "keydown"
watch "keypress"
watch "keyup"
```

Running [here](http://smblott.computing.dcu.ie/escape.html).  Looks the same to me with or without Vimium.

To repeat...

> What behaviour would you like to see from Esc?

Perhaps it would be helpful if you could point us to a specific page, so we all have a specific use case in mind.

ZombieHippie:
I'm looking at your demo right now, this is specifically a case where it
breaks: http://codemirror.net/demo/sublime.html

It may have something to do with a focused input.

ZombieHippie:
It does have to do with an input being focused.
http://codepen.io/ZombieHippie/pen/zxoXwW

smblott-github:
I'm not sure what you want me to do.

I go to which page?
I do what?
What happens?
What do you want to happen?

ZombieHippie:
Sorry for lack of explanation :-)

If you go to my code pen http://codepen.io/ZombieHippie/pen/zxoXwW which I
include your escape.coffe code in, you will see differences between
pressing escape while the input is focused and when it isn't.

In the codepen click on the input to focus, then compare behaviors of
pressing Escape:

No Vimium: The input does not lose focus upon pressing escape.
Vimium: The input does lose focus upon pressing escape.

tested on Chrome 39

smblott-github:
[...]

PR #749, opened 2012-12-20
Asking for comments: passing ESC events through to the page

oconnor663:
Issue #499 mentions that Vimium breaks escape-to-close-chat-tabs on facebook.com. The easiest way to fix this is to simply allow those events to propagate normally to the page. The current behavior was chosen because users might want to leave the chat field without closing the tab (or similar on other sites), but &lt;C-[&gt; already exists as an alternative to ESC that serves that purpose.

We could add an advanced option to configure this behavior. Or we could add a third, non-vim-standard escape sequence with the pass through behavior. But I wanted to suggest this one line change for comments, since it's simple, and it's likely to be what users expect (which is to say, it's what I expect :)

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -386,7 +386,6 @@ onKeydown = (event) ->
       if (isEditable(event.srcElement))
         event.srcElement.blur()
       exitInsertMode()
-      DomUtils.suppressEvent(event)
 
   else if (findMode)
     if (KeyboardUtils.isEscape(event))


blindahl:
Nowadays the url seem to be https://hangouts.google.com/* for the chat windows.

smblott-github:
If you focus the chat window (click it) and then open the Vimium popup, then Vimium should work out the correct hostname.

blindahl:
Ah ok, didn't know that feature was available. :D Always entered the url manually in settings. Thanks!

PR #801, opened 2013-03-19
Ignore keyboard layout

SHTrassEr:
I use two keyboards layouts: English and Russian. In English layout all fine. But when i switch to Russian keyboard layout i can not longer scroll page or switch between tabs. The most interesting thing that I can still open links.

Of course, It is possible to add key mapping for Russian caracters, but it is inconvenient.

These changes add ignoring keyboard layouts.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -132,7 +132,6 @@ initializePreDomReady = ->
 #
 initializeWhenEnabled = ->
   document.addEventListener("keydown", onKeydown, true)
-  document.addEventListener("keypress", onKeypress, true)
   document.addEventListener("keyup", onKeyup, true)
   document.addEventListener("focus", onFocusCapturePhase, true)
   document.addEventListener("blur", onBlurCapturePhase, true)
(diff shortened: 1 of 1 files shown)


Smeagol07:
As Vim does not care bout the layouts as this extension shouldn't. It surely could be added as an option but not as default one, really.

slavaGanzin:
@Smeagol07 In vim you can easily achieve this with:
```
cnoreabbrev RU set keymap=russian-jcukenwin
cnoreabbrev ENG set keymap=
```

And it was a bit, you know, frustrating to find that. 

Also constant rewriting language maps in vimium isn't ok too. But I think your polish friends would appreciate(as I appreciate work of @pletinsky and @tmp6154) if you spend your time adding keymap here: https://github.com/philc/vimium/wiki/Key-Mappings

Smeagol07:
@slavaGanzin Polish keyboard layout is the same as US one, nothing to change.

smblott-github:
Closing.

Out of date.

mortdeus:
I have no idea why this was important to me, but I just remember it was. lol

PR #984, opened 2014-01-02
Strip javascript content from bookmarklets

scottperlitch:


smblott-github:
It's not obvious that this is the right thing to do (or, if it were, whether it's the right thing to do).

Thanks, @scottperlitch, but I think we should leave it.

PR #1004, opened 2014-03-02
New feature: edit the current URL in the omnibar

sukima:
Often times when I'm working on sites I want to add a query string or modify the port number. The go up one level or root of the URL commands are nice but they don't alow you to add or remove specifics on the current URL.

This adds two commands that act exactly like the normal onmibar open but populates the bar with the current URL.

I picked `ge` as the default and `gE` as the default for the same in a new tab.


leoj3n:
This is awesome! 💯

PR #1018, opened 2014-04-08
New feature: visual mode with keybinding support

ForSpareParts:
Starting with @saminiir's and @ramiroaraujo's work on visual mode, added:

- Rebindable keys (use "mapVisualMode key VisualMode.nameOfVisualModeCommand")
- The ability to use browser key commands while in visual mode (i.e. visual mode now *ignores* commands it doesn't recognize and gives Chrome the opportunity to handle them)
- Visual mode automatically disabling if you click the screen
- One simple unit test (most of the functionality would belong in dom tests, but they're not running right now)

Most of this was accomplished by adding visual mode switches to parts of the Commands system. Visual mode is a weird feature by vimium standards: "modes" are expected to have their own key handling (a la link hints or Vomnibar), but visual mode needs essentially the same key binding behavior as the normal mode, just with an entirely different set of key-to-command bindings. Visual mode commands are actually handled in vimium_frontend, with the exception of Escape to deactivate the mode (which can't be rebound).

So, the approach I've taken is essentially to treat visual mode bindings as a one-off "special case" of normal keybindings. If vimium were to have more modes that *also* needed their own bindings, this could become cumbersome, but for just two I think it's pretty straightforward. I'd welcome anyone's thoughts on that.

Again, thanks to saminiir and ramiroaraujo for getting the ball rolling on this awesome feature!

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -2,39 +2,71 @@ Commands =
   init: ->
     for command, description of commandDescriptions
       @addCommand(command, description[0], description[1])
+    for command, description of visualModeCommandDescriptions
+      @addCommand(command, description[0], description[1], visualMode = true)
 
   availableCommands: {}
+  availableVisualModeCommands: {}
   keyToCommandRegistry: {}
+  keyToVisualModeCommandRegistry: {}
 
   # Registers a command, making it available to be optionally bound to a key.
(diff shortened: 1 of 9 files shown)


smblott-github:
Closing, we have visual mode now.

Issue #175, opened 2010-06-29
number command predicate for <gg>


s9gf4ult:
If we type `<10-gg>` we will go in position 10% of the page. Like in vimperator works.
It would be nice in vimium.

smblott-github:
Done in #1023.

PR #1023, opened 2014-04-16
Add number modifier to `gg' scrollToTop command

mrmr1993:
As per issue #175, this implements a number modifier for the `scrollToTop` command, bound to `gg` by default.

Instead of the percentage interpretation suggested in #175, I opted for a multiple of the `scrollStepSize` option, so that it may be thought of as a line number.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -258,7 +258,7 @@ commandDescriptions =
   scrollLeft: ["Scroll left"]
   scrollRight: ["Scroll right"]
 
-  scrollToTop: ["Scroll to the top of the page", { noRepeat: true }]
+  scrollToTop: ["Scroll to the top of the page", { passCountToFunction: true }]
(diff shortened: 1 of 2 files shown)


smblott-github:
@mrmr1993.  Let's do this.

(And should we not do it for `G` too?)

mrmr1993:
> Let's do this.

Rebased.

> (And should we not do it for `G` too?)

The key handling code makes `G` and `1G` indistinguishable at the moment, so we can't match Vim's behaviour without some more major changes. I decided it wasn't worth the effort and so left it.

smblott-github:
snapshot

mrmr1993:
Sorry, missed that, didn't have time to test. Updated.

You might want to squash the new commit when merging (I'm not at a computer so can't).

smblott-github:
Looks like you may have to check this properly when you are at a computer.

mrmr1993:
Fixed, sorry about that.

PR #1032, opened 2014-04-22
Add mouseover and mouseout commands

### LINKED ISSUE #484: Suggestion: Mouse hovering
It would be awesome to have the ability to "hover" the mouse on a certain link. Lots of newer web pages have dropdowns/hover scripts that are only visible when you hover the mouse over a link (as an example, put your mouse over the "Watchers" button, and the "Watchers" tooltip expands from it.

mrmr1993:
Added LinkHints.activateModeToHover and LinkHints.unhoverLast to
perform hovering. These are unbound by default.

Also call LinkHints.unhoverLast automatically before hovering/
clicking the next link, to avoid permanently showing hover effects
(eg. IMDb watchlist grid view mode).

Fixes #484, #1089.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -106,6 +106,8 @@ Commands =
       "LinkHints.activateModeWithQueue",
       "LinkHints.activateModeToDownloadLink",
       "LinkHints.activateModeToOpenIncognito",
+      "LinkHints.activateModeToHover",
+      "LinkHints.unhoverLast",
       "goPrevious",
       "goNext",
       "nextFrame",
@@ -156,6 +158,8 @@ Commands =
     "Vomnibar.activateEditUrl",
     "Vomnibar.activateEditUrlInNewTab",
     "LinkHints.activateModeToOpenIncognito",
+    "LinkHints.activateModeToHover",
+    "LinkHints.unhoverLast",
     "goNext",
     "goPrevious",
     "Marks.activateCreateMode",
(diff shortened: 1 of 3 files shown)


mrmr1993:
Rebased onto current master, as requested by @nfischer in [this comment](https://github.com/philc/vimium/issues/1857#issuecomment-148887468) on #1857.

smblott-github:
This is in #1948.

mrmr1993:
> This is in #1948.

The unhover command isn't, which was the real motivation for this PR in the first place. And I still don't support #1948 as an alternative.

smblott-github:
> The unhover command isn't, which was the real motivation for this PR in the first place.

It is. In #1948, `Escape` after hover unhovers.

> And I still don't support #1948 as an alternative.

Fair enough.  I understand.

mrmr1993:
> It is.

No it's not (specifically thinking for #1857).

nfischer:
@mrmr1993 If we now have an explicit hover command, would it be possible to replace the click (from an f command) with a "click and mouse out"? That would fix #1857 hopefully (I haven't tested this), and we still have an explicit hover command for menus. Or is it necessary to also have a "click and don't mouse out" command?

I think the ideal fix for #1857 would be one that automatically does a mouse out after clicking, not requiring an explicit "zF" to unhover.

AndrewSouthpaw:
Since #1948 was eventually closed, any chance we could still get a PR like this merged in? I'm just dying for the ability to hover over links.

chriszrc:
I added the `LinkHints.activateModeToHover` shortcut and tried it out on this page:

https://www.amazon.com/

on the "Accounts & Lists" hover, but it seems like it's sending a click as well as hovering, so I don't get to access the hover menu. Am I missing something? I built the latest version from master as of yesterday-

PR #1041, opened 2014-04-25
Reload content scripts after install/update

mrmr1993:
This loads the content scripts for pages when the extension is
installed or updated (or reloaded in developer mode). This means that
the new version of Vimium will become active immediately after loading,
and so tabs won't have to be refreshed.

As a developer, this is amazing. Refreshing each tab individually after
each code change/extension reload was a pain.


gdh1995:
In fact, if the frontend ports disconnect, there are two cases:
* a developer view the background page from chrome://extensions/, and press `F5` to refresh
* a common user click reload / "turn off & turn on" in chrome://extensions/
These commits deal with the second case well (applause).
However, for the first case, we just need to re-connect those ports, which has been proved OKay by my customized version for months. A better way, isn't it?

gdh1995:
When a port disconnects, we can try `chrome.runtime.connect()` and catch its exception:
`connect()` succeeds if only the background page is refreshed, and throws an exception "try to connect extension null ..." if the extension has been reloaded.

smblott-github:
Closing in favour of #1521 (which is still @mrmr1993's work).

PR #1049, opened 2014-05-02
Open link in new window

### LINKED ISSUE #1047: <open>InNewWindow commands
We could use app mode for these ones.

mrmr1993:
This adds commands to open links in a new window and a new fullscreen window. This is to resolve #1047.


smblott-github:
@mrmr1993... #1948 (or something similar) is probably a better way to address this.

PR #1074, opened 2014-06-21
Implement back/forward and open in new tab

### LINKED ISSUE #1070: Use `ctrl+shift+h` be back and open new tab?
Hi, I thought I'd open this issue before I add the feature (unless anyone else wants to take it?).

Since `Shift+H` is the back command and `Shift+L` is the forward command, and since in chrome you can hold `Ctrl` and click back and it will open the previous page in a new tab (similarly for forward), I propose that `Ctrl+Shift+H` should be back-and-open-new-tab and `Ctrl+Shift+L` is forward-and-open-new-tab.

Do people agree with this? I'd like to have it included so am happy to add the feature, just want to check that it would be wanted before I start.

mrmr1993:
This resolves #1070.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -97,7 +97,7 @@ Commands =
       "Marks.activateGotoMode"]
     findCommands: ["enterFindMode", "performFind", "performBackwardsFind"]
     historyNavigation:
-      ["goBack", "goForward"]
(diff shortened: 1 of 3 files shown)


smblott-github:
@mrmr1993. This seems pretty niche.  And we get the same effect with *duplicate-then-back*.  Closing.

PR #1077, opened 2014-06-22
Show "Insert Mode" in the HUD for a textbox focused before the user acts

### LINKED ISSUE #946: Focus on input indicator
I often visit sites and wonder why I can't navigate around until I notice that I'm typing into an input the whole time.. How about a little indicator (just like the one, that says that one is in INSERT mode) that shows that one is currently focussing an input?

mrmr1993:
Fixes #946.


mrmr1993:
This is extremely stale, and its purpose already achieved by `GrabBackFocus`. Closing.

Issue #963, opened 2013-11-25
Visually distinguish common and advanced commands in the help dialog


mcint:
When the help dialog is expanded from common to advanced commands, the advanced commands are not visually distinguished, making it harder to learn about new commands in an already long list.

smblott-github:
Closing.
This is live #1078.

PR #1078, opened 2014-06-22
Add a label to advanced commands in the help dialog

mrmr1993:
This adds a label to advanced commands which are shown in the help dialog, to help distinguish them from normal commands, as per issue #963. The style isn't at all right, but hopefully this is a step in the right direction.


smblott-github:
@mrmr1993. This seems like a good idea. Any interest in reviving it?

smblott-github:
OK.  So the theory sounds good, @mrmr1993, but the reality ... let's just say it could be more attractive...

snapshot

If we do this, then we need a more subtle (yet clear) way of highlighting commands.  Perhaps just a Vimium-blue asterisk which - on hover - pops up *Advanced command", or something?

PR #1081, opened 2014-06-24
Move regex searches to a web worker to stop the browser from hanging

### LINKED ISSUE #957: Search is slow on pages with lots of text
The search feature hangs browser for a small amount of time on each addition of characters to search.

To reproduce:
1. Go to http://build.chromium.org/f/chromium/perf/dashboard/ui/changelog.html?url=/trunk/src&range=217147:224845&mode=html
2. Search for anything

Tested on:
Chrome version 30.0.1599.101,
Mac OS X 10.8.5.

mrmr1993:
This fixes #957 (currently closed anyway) by asynchronously performing regex searches of the page in a separate thread. It holds two main advantages over performing the regex synchronously:

* The UI doesn't hang while waiting for each search to complete, which is better UX.
* The workers can be terminated before they complete. When the search terms are updated, the previous search with the old terms is terminated and a new one is started, instead of wasting time waiting for the old one to complete first.

This has been tested on the [example](http://build.chromium.org/f/chromium/perf/dashboard/ui/changelog.html?url=/trunk/src&range=217147:224845&mode=html) from #957, and feels like a great improvement over before.

### DIFF
--- a/content_scripts/searchWorker.coffee
+++ b/content_scripts/searchWorker.coffee
@@ -1,0 +1,10 @@
+self.addEventListener "message", (event) ->
+  text = event.data.text
+  regex = event.data.regex
+
+  regexMatches = text.match regex
+
+  postData =
+    regexMatches: regexMatches
+  self.postMessage postData
+, false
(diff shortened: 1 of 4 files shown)


smblott-github:
@mrmr1993.  Is this solving a real problem?  I don't recall noticing any significant delays.

(There are considerably worse delays generating link hints on Google Plus after scrolling down some way.)

smblott-github:
Closing.
The need for this doesn't seem so strong, and therefore the extra complexity probably isn't worth it.

(It's probably also hopelessly out of date.)

PR #1082, opened 2014-06-25
ability to jump between scrolling jump history

fent:
Like in vim, Ctrl-O and Ctrl-I jump backward/forward respectively in jump history. "jump history" is basically any command that uses the `scroller.jumpTo()` function.

I've often needed to see something at the top/bottom of a tall page, but wanted to quickly go back to where I was. What I had to do was create a mark, jump, and then jump back to the mark. With this feature, jump history will already be recorded, having the user press less keys if they want to jump back.

Jump history is truncated if a user jumps back and then uses a `scroller.jumpTo()` command. History will also be updated if the user scrolls after they've jumped to a position and history, and then jumped again.

History is cleared if the activated element ever changes and is scrolled with a command. I wasn't sure how to keep history per element without editing the page structure. So it seemed like the best solution was to clear it.


fent:
I've merged the latest changes. Max scrolling history added, and mapped `g;` to `scrollBackward` too.

Not sure what to do about having `<c-o>` mapped since chrome uses it. Maybe not have it mapped by default?

Also made it so previous animations are cancelled when `Scroller.scrollTo()` is called. This makes more sense as `scrollTo()` is used for absolute positions, running another animation while, would displace the final position.

smblott-github:
We have this to some extent with #1714 (which is merged in #1716).

The additional functionality seems a bit niche; too niche to warrant new commands.

PR #1088, opened 2014-07-04
Always show which mode we are currently in

alarive:
I think the HUD should always display something if vimium is enabled for the current URL and the option "Hide the Heads Up Display (HUD)" is disabled.

I find it irritating that sometimes I can't know quickly if I can type a vimium command or not. 

In vim you don't have this issue because if no mode information is displayed you are in normal mode. However in vimium when you have no information about the current mode you can be:
* in normal
* in insert mode, if the page forces the autofocus of an input field for example
* in no mode at all, for example when you are in the chrome omnibar (in which case I would like the HUD to display something like ** UNFOCUSED **)

If I choose to display the HUD it is because I want to know what the vimium state is.


smblott-github:
This is out of date now.

Where we feel the need to provide visual feedback, we can add an `indicator` field to the relevant `Mode`.  Closing.

Issue #603, opened 2012-07-27
Proposal: Make link hints selectable


int3:
See https://github.com/philc/vimium/issues/507#issuecomment-7273412.

As an additional enhancement, we could display the destination of the currently selected link, which is a feature that people have been requesting for some time.

Issue #601, opened 2012-07-26
Add the ability to preview link URL without following it


santosh:
Currently when we press **f** or **F** we are given a tag of letter(s) on each link to follow them.

For a example, we have a link to [Github](https://github.com/). We can press **_f**_ followed by **_M**_ (let's just assume) to open the link in the same tab.

There should be a feature/command in which user should be able to just see where is that text is linked to (as same as when we hover the mouse over the link).

Note : When we **just** talk about Google Chrome (without extension). When you browse through keyboard and you have to cycle the link through **TAB**, you are able to see the link at the bottom left of the page.

PR #1089, opened 2014-07-07
Add command to show link URL without opening it (related to issues #601 and #603)

ehdr:
Hi @philc!

Implemented a command for showing the link URL (in the HUD) without opening it, possibly solving issues #601 and #603.

Useful when you want to check the target URL of a link before clicking, yet don't want to reach for the trackpad to hover it with the pointer.

Implemented as LinkHints.activateModeToShowLinkUrl. Also had to modify the HUD handling slightly in LinkHints, to support keeping the message visible after deactivating the mode.

Mapped to `sf` by default.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -17,6 +17,7 @@ Contributors:
   ConradIrwin
   Daniel MacDougall <dmacdougall@gmail.com> (github: dmacdougall)
   drizzd
+  Eric Hansander (github: ehdr)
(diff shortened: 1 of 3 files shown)


smblott-github:
This seems pretty niche; too niche to warrant a new command.

Perhaps it subsumed (to some extent) by the "Yank URL" option in #1948.

Closing.

PR #1111, opened 2014-08-02
Add moveTabToLeft / Right, moveTab wraparound

elevengu:
I've had this in my personal fork since early 2013, and I think it's much better design to match the wraparound behavior of nextTab and previousTab in this way. I've noticed that the vast majority of my use cases involve wrapping.

I also implemented moveTabToLeft and moveTabToRight, but even I don't have them mapped to anything. So I don't think it's necessary to merge that with upstream unless there's actual demand for it.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -17,6 +17,7 @@ Contributors:
   ConradIrwin
   Daniel MacDougall <dmacdougall@gmail.com> (github: dmacdougall)
   drizzd
+  elevengu
   gpurkins
   hogelog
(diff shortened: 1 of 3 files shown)


smblott-github:
The wrapping inconsistency between changing tabs and moving tabs is unfortunate, but it's what we have.  Any change to the status quo would likely be unwelcome to some.

As @mrmr1993 suggests, a large count can be used to get pretty much the same effect.

Closing.

PR #1140, opened 2014-08-21
Add exitInsertMode, allow binding single keys only

### LINKED ISSUE #301: Allow any key to mapped to "exit mode" instead of just <esc>

Thanks

mrmr1993:
This creates the command `exitInsertMode`. The command only responds to single key mappings, so that all keys except those which explicitly exit insert mode are passed to the underlying page.

This fixes #301.


mrmr1993:
Stale since modes landed.

PR #1217, opened 2014-11-02
Add commands for fullscreening, maximising, minimising and restoring a window

### LINKED ISSUE #1206: [feature request] full screen shortcut
Is it possible to add a shortcut to toggle fullscreen? f11 is too far up.

mrmr1993:
Resolves #1206.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -144,7 +144,12 @@ Commands =
       "restoreTab",
       "moveTabToNewWindow",
       "togglePinTab",
-      "closeTabsOnLeft","closeTabsOnRight",
+      "restoreWindow",
+      "minimizeWindow",
+      "maximizeWindow",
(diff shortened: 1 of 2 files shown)


joshuacox:
merge this in, this is definite;y needed

smblott-github:
From #1206...

> `F11` is too far up.

That's not a very strong case.

Would this stuff not be better handled by a window manager?

mrmr1993:
> Would this stuff not be better handled by a window manager?

Arguably yes, expect for fullscreening. I think we should be looking for ways to support and document more commands, rather than reasons to bin them. Just my €0.02.

smblott-github:
> expect for fullscreening

Yes, and we have `F11` for that.
Closing this.

(There's clearly a discussion needed around how to support more commands, present them to the user, and yet maintain a coherent set of commands.)

PR #1231, opened 2014-11-08
Tabs order

### LINKED ISSUE #1146: Feature idea: Switch between two tabs
Just thing I'm planning to add to jb55/vimium-plus#4, just wondering if upstream or anyone else would find this useful:

> A key to switch back and forth between the last tab you were at and back again. CTRL-W x is the vim analogy here, thinking with windows instead of tabs. I find myself wanting to flip back and forth between two tabs a lot, so this would be a nice addition.

smblott-github:
In tabs mode (and when the query is empty) order tabs by recency, so the previous tab is just one `Enter` away.

Fixes #1146 (well, indirectly).

### DIFF
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -259,7 +259,7 @@ class DomainCompleter
   # Suggestions from the Domain completer have the maximum relevancy. They should be shown first in the list.
   computeRelevancy: -> 1
 
-# TabRecency associates a timestamp with each tab id.
+# TabRecency associates a logical timestamp with each tab id.
 class TabRecency
   constructor: ->
     @timestamp = 1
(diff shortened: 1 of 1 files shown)


glacjay:
This feature seems broken now.

philc:
@glacjay This is working for me: when opening Vomnibar's tab search (`shift-T`), the tabs shown are listed in the order they were recently switched to. Note that there is a threshold by which we consider a tab visited: it has to have been visible for 500ms or more.

If you're still seeing this behaving incorrectly, please post more details.

glacjay:
@philc not always, but sometimes. Haven't found any pattern yet. Will report back when I find more details.

glacjay:
@philc My bad, my issue is another problem, which is: sometime, the order of Vomnibar's tab search list been reset completely. So maybe I should submit an issue instead. Thank you.

Issue #839, opened 2013-05-13
alt-f is received by both vimium and wikipedia's search box


emallove:
On some pages, e.g., http://en.wikipedia.org/wiki/Main_Page, `<Alt-f>` (aka _"Open multiple links in a new tab_") scrolls to the top of the page.

innaterebel:
<kbd>Alt</kbd>-<kbd>p</kbd> as well, pins the tab and triggers print/print friendly layout simultaneously on various wiki-structured sites.

AdrienLemaire:
@innaterebel Nov 2019, and the bug still exists. Was wondering why I didn't notice it before, but I know understand it's a wikipedia-related issue :)

gdh1995:
Just an AD: I have [Vimium C](https://github.com/gdh1995/vimium-c) and it can prevent access keys when commands are triggered, at almost no cost & without modifying DOM.

RealYukiSan:
How "Open multiple links in a new tab" works? I don't get the big picture

PR #1234, opened 2014-11-10
Fix for accesskeys issue #839

mrmr1993:
Add the method `DomUtils.suppressAccesskeyAction` to prevent the browser focusing elements with `accesskey` set. We need this since the browser currently ignores `preventDefault` for the `accesskey` action, causing the issue in #839.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -141,7 +141,7 @@ LinkHints =
   # of digits needed to enumerate all of the links on screen.
   #
   getVisibleClickableElements: ->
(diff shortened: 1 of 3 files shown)


gdh1995:
I modified Vimium and added global Chrome shortcuts to it. For example, I can use `Alt-C` or `Alt-V` to call `previousTab` or `nextTab`. But even then Chrome still triggers `<a accesskey="c">` - ` on a Wikipedia page:
* `<a accesskey>` is triggered on `keydown`
* For `C` key, window` gets a `KeyupEvent(keyCode=67)` only.

So, I think the only solution is that somebody adds an issue to Chromium and Chrome fixes such a bug.
* test on Chrome 43.

philc:
Closing due to inactivity and low priority.

Issue #1188, opened 2014-10-26
Per-url key mappings


mrmr1993:
Create a method for different mappings to be applied to different URLs.

This would make allow for a much easier way of implementing passkeys (and disabling Vimium for URLs via `unmapAll`?), and make it much easier to resolve issues like #1176.

### Suggested implementation:
Add an `ifUrlMatches` command to the "Custom key mappings" option in the options page which

1. accepts a regular expression (or whatever we use for excluded urls) as its argument, seperated from `ifUrlMatches` by a single space.
1. is termined by the `endif` statement.
1. obeys indentation if present. ie. if the next non-empty non-comment line after the command is indented, every line with the same or deeper indentation is also matched, and the block is terminated (even if there is no `endif`) when the indentation level falls below the initial level. [**NOT CERTAIN ABOUT THIS**: this could make matching `endif`s complicated, so a first draft may not want to include this.]

Eg.
```
unmap F
unmap <a-f>
unmap f
map <a-f> LinkHints.activateModeWithQueue
map F LinkHints.activateModeToOpenInNewForegroundTab
map f LinkHints.activateMode

#
# Block only using indents
#
ifUrlMatches https://www.facebook.com
  # Facebook provides its own j, k commands to scroll by post, use these instead.
  unmap j
  unmap k

#
# Block with indents and an endif
#
ifUrlMatches https://www.facebook.com/*
# Pass the escape key even when we handle it, so facebook's messanger closes when we blur it.
  unmap <esc>
  map <esc> exitInsertModeAndPassKey
endif

#
# Block with no indents, endifs, and a nested block (with indents and an endif)
#
ifUrlMatches http://www.foo.com/bar/*/overview.html
# Sometimes I forget and use their next page binding, but the animation is obnoxious. Override it.
map @ goNext
ifUrlMatches http://www.foo.com/bar/*/gallery/overview.html
  # Gallery overviews use h, l to show next/previous page of images.
  unmap h
  unmap l
  # Sometimes the image is wider than the screen, so we still want scrolling.
  unmap zh
  unmap zl
  map zh scrollLeft
  map zl scrollRight
  # @ is a useful binding on this page, so unmap it.
  unmap @
endif
endif
```

Thoughts?

PR #1236, opened 2014-11-10
Restructure help dialog

mrmr1993:
This PR:
* Unifies various components of command definitions into a single location, to make it hard to add commands incorrectly, enforced with tests.
* Moves the help dialog to an iframe.
* Makes the help dialog much more independent of content scripts and the background page.
* Switches from using raw html via `innerHTML` to using `<template>`s, helping with stability and customisability.
* Re-group commands in the dialog to make it feel less 'one-sided'.

This arose out of wanting commands better managed and available from the content scripts, as part of working towards #1188. The help dialog refactor came as a nice extra when trying to un-break it after the commands changes.

Thoughts, anyone?


smblott-github:
@mrmr1993.  Let's do this in stages...

- Move the help dialog to an iframe.
- Then rethink how the help dialog should work.
- And separately, how it should look.  The current design looks pretty dated, to me.

Also, we could recognise that the use cases for the help dialog are different on the options page from in general use.

I'm going to close this.  But this is an issue we very much have to discuss and address.

PR #1246, opened 2014-11-17
Handle contentEditable elements correctly

### LINKED ISSUE #888: Vimium does not disable it's shortcuts when editing an HTML element with "contenteditable" set to true
example: http://html5demos.com/contenteditable

### LINKED ISSUE #1136: Unable to focus some text input areas
I've noticed that I couldn't focus text input boxes on some websites. Using "f" and typing a link hint makes border of those boxes to highlight for a second, but they don't get a focus. "gi" command also doesn't work with all boxes (though it seems to work for some boxes, for which "f" approach doesn't work).

For example, UbuntuForums.org quick reply and reply text areas don't want to get focus (for these, neither "f", nor "gi" work).

These could be focused with Tab, but you couldn't use Escape to leave insert mode.

### LINKED ISSUE #1245: Quick reply broken on inbox.google.com
On `inbox.google.com` and with vimium enabled, choose some thread and then click in the quick reply box, at the bottom here:

snapshot

Start typing.  Vimium still intercepts keystrokes like `o`.

The pop-out reply works fine (much like gmail, I guess).

mrmr1993:
Continuation of #1050.

This PR makes link hints handle `contentEditable` elements correctly, and ensures we disable our shortcuts while such elements are active. This

* fixes #888 
* fixes #1136 
* fixes #1245.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -220,6 +220,9 @@ LinkHints =
     if (DomUtils.isSelectable(clickEl))
       DomUtils.simulateSelect(clickEl)
       @deactivateMode(delay, -> LinkHints.delayMode = false)
+    else if clickEl.contentEditable == "true"
+      DomUtils.focusContentEditable(clickEl)
+      @deactivateMode(delay, -> LinkHints.delayMode = false)
     else
       # TODO figure out which other input elements should not receive focus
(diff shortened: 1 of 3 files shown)


smblott-github:
The problems this PR fixes have already been addressed elsewhere, @mrmr1993, yes?

(Also, this PR is against a very old code base.)

Closing.  Correct me or replace with another PR, @mrmr1993, If I'm mistaken.

PR #1250, opened 2014-11-19
Handle contentEditable elements correctly, with normal mode

mrmr1993:
Continuation of #1246.

This adds the ability to execute commands with text selected in a `contentEditable` element (which technically is still selected) after pressing `<esc>` once. Pressing `<esc>` again or pressing a key not bound by Vimium fully blurs the element, by removing its selection.


mrmr1993:
This is stale and mostly redundant now #1413 & co. are in. Closing.

PR #1252, opened 2014-11-20
Port overlapping link hint rotation by Space from VimFx

### LINKED ISSUE #757: An idea about resolving overlap link hints!
When one or more link hints overlaps,
maybe we could use "-" key(or something else) then the topmost link hint keys to hide it.

For example,

> [bottom]"AA", "BB", "CC"[top]
> ----------------------------------------->

I want to click "AA", but "CC" hides other link hints,
so I enter "-" then "CC", now "CC" disappears,
then again, I enter "-" then "BB", now "BB" was removed,
the one what I really want, "AA", now is so clear!

This method does not effect existing link hint generating algorithms,
so it may have minor impact on existing code.

BTW, I have become a crazy fan just after I heard Vimium can turn Chrome into a vim style browser.
Now if the overlapping problem resolved, Vimium would be perfect.

kartynnik:
This commit basically ports the VimFx'es ability to rotate overlapping marker stacks by pressing [Shift-]Space. This is very useful in the cases where many little (e.g. iconic) links are stuffed together. Resolves #757.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -42,5 +42,6 @@ Contributors:
   Werner Laurensse (github: ab3)
   Timo Sand <timo.j.sand@gmail.com> (github: deiga)
   Shiyong Chen <billbill290@gmail.com> (github: UncleBill)
+  Yury Kartynnik <yury@kartynnik.info> (github: kartynnik)
 
 Feel free to add real names in addition to GitHub usernames.
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -15,6 +15,7 @@ OPEN_WITH_QUEUE = {}
 COPY_LINK_URL = {}
 OPEN_INCOGNITO = {}
 DOWNLOAD_LINK_URL = {}
+MARKER_MINIMUM_Z_INDEX = 2000000000
 
 LinkHints =
   hintMarkerContainingDiv: null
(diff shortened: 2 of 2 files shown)


smblott-github:
Two comments:

- How useful is this feature in practice?  When hints overlap, it's usually pretty difficult for the user to know what each hint actually activates.

- However, we could do this anyway.  It's a cool feature which former VimFX users would expect to have, and it's minimally disruptive.

I suspect this PR is against an hopelessly old code base; so I'll close it for now.  (Correct me if I'm mistaken.) A new PR would be welcome.

kartynnik:
This feature is pretty useful in the occasions when the web app interface is full of little icons, for example. Another claim in its favor is that you may try once and remember - if you happen to be forced to use some repetitive interface, being able to use Vimium to speed up your regular navigation is beneficial. I remember myself wishing several times that this feature is implemented in Vimium, but it was mainly back when GMail support was not that polished.

Yes, the code base is like over a year old, I didn't update the PR 'cause there was no significant reaction. But @mrmr1993 had merged it into his own fork, if he had kept it up to date through the code base changes, maybe we could have a more modern implementation of this.

If this is not the case, I'll try to reimplement the PR against the current upstream.

lydell:
As a VimFx user:

- Every now and then sites have terrible markup where you get two markers on top of each other, because some wrapper element was considered clickable (as a false positive). Activating the hint for the wrapper element does nothing. Then it’s nice being able to press `<space>` once (that’s the default shortcut to rotate markers in VimFx) to see the “real” marker.

- When it’s difficult to see what each hint actually activates (such as those tiny pagination numbers some forums use), you can press and hold shift (by default) in VimFx to make the hints semi-transparent. Then it is easier to work out which hint to type.

(Also, it’s VimFx, not VimFX, hehe :) )

PR #1258, opened 2014-11-22
Draw overlays for input elements before focusing in focusInput

### LINKED ISSUE #1257: focusInput command overlays incorrect for off-screen inputs
The `focusInput` command (`gi`) shows input indicators in the wrong position if the page has to scroll to the input we focus.

screenshot from 2014-11-22 15 35 55
screenshot from 2014-11-22 15 36 23

mrmr1993:
This prevents us from drawing the overlays in the wrong place if the window scrolls to focus the chosen input (eg. when it is on the boundary of the page). Fixes #1257.

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -93,6 +93,7 @@ div.internalVimiumInputHint {
   display: block;
   background-color: rgba(255, 247, 133, 0.3);
   border: solid 1px #C38A22;
+  pointer-events: none;
 }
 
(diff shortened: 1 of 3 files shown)


smblott-github:
This too is worth fixing, @mrmr1993, if we can get it right without too much disruption.

smblott-github:
Thinking about this...

Should the rectangles not be the size of the entire element, rather than just the visible part?

Whenever we draw a rectangle that isn't the size of the element it's going to be wrong when scrolled.

PR #1259, opened 2014-11-22
Perform background commands on the requesting tab by default

### LINKED ISSUE #1160: Close function delay causes wrong tab to close
When I type 'x' and then change tabs quickly using ctrl-shift-tab or shift-tab the new tab I'm working in ends up getting closed because the X takes so long to close. I'm surprised it 1) takes so long to close a tab and 2) it doesn't record the order in which keystrokes are pressed or it doesn't make note of which tab is currently open immediately to ensure closer of the correct tab.

The reason I use ctrl-shift-tab or shift tab rather than gt or gT is because it's faster to press all the keys at the same time rather than press two keys in sequence. It's also more universal.

Google Chrome	36.0.1985.125 (Official Build 283153) dev-m
OS	Windows
Vimium 1.45

mrmr1993:
This PR makes all commands in `BackgroundCommands` take the same arguments, and tweak the implementations to use the calling tab by default. Fixes #1160.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -214,10 +214,10 @@ chrome.tabs.onSelectionChanged.addListener (tabId, selectionInfo) ->
   if (selectionChangedHandlers.length > 0)
     selectionChangedHandlers.pop().call()
 
-repeatFunction = (func, totalCount, currentCount, frameId) ->
-  if (currentCount < totalCount)
+repeatFunction = (func, count, frameId) ->
+  if (0 < count)
     func(
-      -> repeatFunction(func, totalCount, currentCount + 1, frameId),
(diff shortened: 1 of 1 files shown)


smblott-github:
I haven't encountered #1160, but I guess it's possible.
 
 This seems like the correct thing to do, @mrmr1993.  However, I suspect the PR is out of date, now.  Is it worth revising?

mrmr1993:
I think it's better to do what the user expects, as far as possible. I'll rework this when i get some time later.

smblott-github:
If you look at this, @mrmr1993, then please look at it in two parts:

- Standardising the parameters. (PR, discuss, merge, then...)
- Changing to using the requesting tab.

That would certainly make things easier from my side (hence, more likely to be actioned).  Thanks.

PR #1262, opened 2014-11-24
Improve contentEditable support

### LINKED ISSUE #888: Vimium does not disable it's shortcuts when editing an HTML element with "contenteditable" set to true

### LINKED ISSUE #1136: Unable to focus some text input areas

mrmr1993:
Rebase of #1253 onto current master. From #1253:

> #1246, less the functionality which blurs `contentEditable` elements when `esc` is pressed (by clearing the document selection).

>This is functionally the same as #1247 except
* it detects a wider range of `contentEditable` elements.
* it fixes link hints for `contentEditable` elements.

>This
* fixes #888 
* fixes #1136

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -220,6 +220,9 @@ LinkHints =
     if (DomUtils.isSelectable(clickEl))
       DomUtils.simulateSelect(clickEl)
       @deactivateMode(delay, -> LinkHints.delayMode = false)
+    else if clickEl.contentEditable == "true"
+      DomUtils.focusContentEditable(clickEl)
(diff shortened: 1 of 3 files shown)


smblott-github:
This is out of date now, @mrmr1993, yes?

PR #1265, opened 2014-11-24
Ensure that we use `for own` rather than `for`

mrmr1993:
This stops us traversing object prototypes when we're only interested in their direct properties. Implementation of the suggested code improvement in [this comment](https://github.com/philc/vimium/pull/1261#issuecomment-64189963).


smblott-github:
We should do this, @mrmr1993.

mrmr1993:
Closing in favour of #1994.

PR #1268, opened 2014-11-24
Implement vomnibar open in new window commands

### LINKED ISSUE #1264: Suggestion - Open URL, bookmark, history entry in a new **window**
A lot of times I find that I want to open a history entry in a new window. So today I hit "O" and after the tab loads I hit "W" to move it to a new window.

Is there a way to combine those two into one keystroke?

mrmr1993:
This resolves #1264.


smblott-github:
#1848 (or similar) is probably the way to do this.

mrmr1993:
I don't think that #1848 is what you meant there..

smblott-github:
Ooops, #1948.

Issue #1103, opened 2014-07-27
It can't change keyboard in find mode.


qhhonx:
When I type '/' into find mode, and I want to search Chinese strings, but it fails to change current keyboard into Chinese.
**Platform**: 14.04, Chromium 34

Issue #1013, opened 2014-04-02
Suggestions for find mode


ghost:
Find mode has some problem that bothers me, it would be great that the next version improved:
1 Now the find mode only supports ASCII, hopes the utf-8 supported 
2 Now the find mode only supports entered manually, hopes paste from the clipboard supported

Issue #576, opened 2012-07-12
"\" closes search bar - therefore unable to use regex search. 


St-B:
Whenever I use "\" in Chrome (German localization) it has the same effect as hitting the ESC-key - the control loses focus or as in this case disappears. This renders me unable to use regex search. Would be glad if this could be fixed or the regex key could be customized.

smblott-github:
Closing.
Should be fixed with iframes.

Issue #498, opened 2012-03-06
Support the Chinese in bookmark search and find mode


s2marine:
Can the bookmark search and the find mode support the Chinese IME?

PR #1273, opened 2014-11-27
Move the HUD to an iframe and handle find mode in it

### LINKED ISSUE #556: Use an text box for key entry in Vimium's find mode

### LINKED ISSUE #1266: Find mode box not really visible
Hi

When I press / I get the little find box in the lower right corner but it is a little bit to low. See image below.

Is it possible to fix this somehow?

skarmavbild 2014-11-24 kl 14 45 22

mrmr1993:
This PR
* moves the HUD to an iframe
* moves find mode key handling to the iframe
* changes the normal HUD to a `contentEditable` element for find mode
* tidies up some frontend code.

In terms of issues, this
* fixes #1266 
* fixes (possibly) #576 
* fixes #556 (and so also #1013, #416, #498 and #1103 which were closed in favour of it).


smblott-github:
What is the relationship between this and #1401?  I assume you'd prefer to go with #1401?

mrmr1993:
Oops, that's right! This can be closed; #1401 is much the same but uses `UIComponent`.

PR #1282, opened 2014-12-04
Add option to disable auto-focus

### LINKED ISSUE #523: Add a preference to prevent pages from stealing Vimium's focus when they first load
I like to scroll through my tabs using J,K. This works fine, until a tab has a highlighted text area, which the browser starts typing into. Next, I have to press Esc and continue with J,K until i encounter the next highlighted text area.
If wanted to type into a text area, I would select it myself using the f key.
Please fix :)

mrmr1993:
This fixes #523.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -151,6 +151,7 @@ installedListeners = false
 initializeWhenEnabled = (newPassKeys) ->
   isEnabledForUrl = true
   passKeys = newPassKeys
+  FocusPrevention.tryUnfocus()
   if (!installedListeners)
     # Key event handlers fire on window before they do on document. Prefer window for key events so the page
     # can't set handlers to grab the keys before us.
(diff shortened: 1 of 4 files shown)


smblott-github:
Hi, @mrmr1993.  I was trying this out on [this site](https://www.google.ie/).  Seems to work when the page is first visited (e.g. in a new tab) and when the page is refreshed with `r`.  Unfortunately, though, not when the page is refreshed with `Ctrl-R`.  Any ideas?

mrmr1993:
@smblott-github is this behaviour the same on other websites with autofocus (eg. [Bing](http://www.bing.com/))?

At a guess, I'd say the new page catches the `keyup` event with `Ctrl-R`, but our round-trip to the background page with `r` takes enough time that the old page gets it instead. If this is the case, dropping `"keyup"` from `allowFocusAfterEvents` should do the trick.

There are some problems with the code that I've noticed while reading it, I'll add a few comments to the code page.

content_scripts/vimium_frontend.coffee
```diff
@@ -162,6 +164,46 @@ initializeWhenEnabled = (newPassKeys) ->
...
+    true
+  unfocus: (event) ->
+    return true unless isFocusable event.target
+    FocusPrevention.lastFocused = event.target
+    DomUtils.suppressPropagation event if FocusPrevention.tryUnfocus()
+
+  # Either add or remove the event listeners responsible for stopping autofocus, depending on add.
+  modifyListeners: (add) ->
+    changeEventListener = if add then "addEventListener" else "removeEventListener"
+    document[changeEventListener] "focus", @unfocus, true
```
mrmr1993: Should have a `blur` listener that does `DomUtils.suppressPropagation event` so the page doesn't see an asymmetry between `blur`s and `focus`es.

Also, the `focus` and `blur` listeners should set `@isStillFocused` to `true` and `false` respectively.

smblott-github:
> is this behaviour the same on other websites with autofocus (eg. Bing)?

I tried bing.  with both `r` and `Ctrl-R`, sometimes it works, sometimes it doesn't.  So, looks like a race condition.

smblott-github:
Closing.  Fixed in #1480.

PR #1288, opened 2014-12-08
Add option to disable passing unhandled keys to the underlying page

### LINKED ISSUE #1287: Modes to deal with Vimium vs site specific shortcuts
Is it possbile to implement a 'normal' mode and 'insert' mode concept, such that on sites like Gmail:

- normal mode blocks native gmail shortcuts/allows only vimium shortcuts
- insert mode allows the native gmail shortcuts to work and blocks Vimium shortcuts until you return to 'normal' mode. 

I could even see this extending to a 'desktop' Vimium application so you can navigate your OS without the mouse and using relavent Vim commands in different settings (things like entering text in text fields/boxes).

mrmr1993:
This resolves #1287.

### DIFF
--- a/background_scripts/settings.coffee
+++ b/background_scripts/settings.coffee
@@ -67,6 +67,7 @@ root.Settings = Settings =
     linkHintNumbers: "0123456789"
     filterLinkHints: false
     hideHud: false
+    captureAllKeys: false
     userDefinedLinkHintCss:
       """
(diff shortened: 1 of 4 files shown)


smblott-github:
In light of pass keys, I'm still concerned that the use case for this isn't sufficiently strong.

Closing.

Continue the discussion, @mrmr1993, if you think I have this wrong.

PR #1290, opened 2014-12-09
[WIP] Unify separate settings implementations, use chrome.storage everywhere

mrmr1993:
This PR

* combines the two settings implementations (content scripts vs. background scripts) into one settings 'library', usable everywhere;
* uses `chrome.storage` to sync changes automatically, which lets us stop constantly requesting settings from the background page;
* syncs multiple changes in one go, rather than one-by-one, which greatly reduces the number of write operations we need;
* moves code not directly related to settings management (eg. parsing search engine definitions) out to the appropriate places in the codebase;
* absorbs the `Sync` 'module' into settings.coffee; much of the `Sync` code directly uses/modifies variables from `Settings`, so they might as well be together.

Still todo:

- [ ] Fix tests.


smblott-github:
Here are some problems with the existing settings implementation:

1. Two implementations: `Settings` and `settings`.
1. A massive amount of unnecessary messaging for `settings` in the content scripts.
1. We don't store settings which have their default value.  This creates an unnecessary and problematic asymmetry.
1. `postUpdateHooks` should be removed, and replaced with a mechanism which keeps related logic close together. E.g.  All of the logic for `keyMappings`, should be in `main.coffee`.
1. `Settings` should contain **nothing** related to any specific setting (apart from default values).  E.g. the `searchEngines` update logic has no business being in `Settings.coffee`.
1. We're not taking advantage of the great plumbing which `chrome.storage` provides.
1. All of this business with JSON values is unnecessary.

What could we do about it?

- Nothing.  Certainly possible.
- The two commits: d478aa5c56acba34b7fb83a2c49adc48baf47889 and c5ad4790f42cbdfc62f0914fc0dc0ee250ab4033.  Also possible.  Minimally disruptive.  Only really addresses 2 and 6, above.
- This PR.  I think this PR is more complicated than it needs to be because it only addresses 1 and 2 (and then *bends over backwards* to accommodate some of the problems above).
- Re-do this PR from scratch, addressing all of the problems above.

If we were to do the latter, then we would:
- The primary storage location for all settings is `chrome.storage.local`.  Do not use `localStorage` at all.
- Stop JSON encoding all settings.
- Use `chrome.storage` for all messaging relating to settings changes.
- Store all settings, including those with the default value.
- Have just one implementation of settings: `Settings` (although we might alias it as `settings` in content scripts to reduce the number of changes and keep the diff clean).  The main purpose of `Settings` would then be to provide synchronous access to settings cached from `chrome.storage` (and list default values).
- Replace `postUpdateHooks` with a method `Settings.onChange(key, callback)`.  Then, we don't need to worry about changes propagating through from the options page (for example).  And related logic stays close together.
- Get the search-engine logic out of `settings.coffee`.
- Replace `sync.coffee` with a simple listener which copies changes (if appropriate) between the two `chrome.storage` storage areas.  This can be just a few lines of code.
- Migration code.

What do you think, @mrmr1993?  On the one hand, it's a lot of change for no user-facing benefit.  On the other, it addresses a significant amount of technical debt which we've built up.

mrmr1993:
> Re-do this PR from scratch, addressing all of the problems above.

I think this is the way to go, and am more than happy to work on it. All of your suggestions sound like a good overview of how I would want the settings library to work.

A few criticisms:

> * Replace `sync.coffee` with a simple listener which copies changes (if appropriate) between the two `chrome.storage` storage areas. This can be just a few lines of code.

`chrome.storage.sync` is caching, so once the initial load of data is done, the only difference with `chrome.storage.local` is the pushing and pulling of network updates. Since we already have updates pushed to this equivalent of `chrome.storage.local`, it seems wasteful to wrap it and ferry data to and from the actual `chrome.storage.local`. I suggest, instead, we just mark settings that ought be synced, and then manage those with `sync` and the rest with `local`.

> * Do not use `localStorage` at all.

The potential for race conditions on browser startup (eg. a tab opened between the extension loading and `chrome.storage` returning, thus enabling Vimium on a disabled URL/omitting passkeys) seems too great that we can do away with this without further discussion and investigation. I think we should use `localStorage` for the background page's settings cache until we can make it/confirm that it is safe to stop doing so.

smblott-github:
> I suggest, instead, we just mark settings that ought be synced, and then manage those with sync and the rest with local.

I agree.  That's simpler.  The only difference is that `chrome.storage.sync` can fail because of quota exceeded, whereas `local` cannot.  The approach you propose is better if we're confident that that will never happen. Otherwise, keeping everything in `local` and using `sync` only for propagation is better.

> I think we should use `localStorage` for the background page's settings cache until we can make it/confirm that it is safe to stop doing so.

I'm not such a fan of that.  It gives us two implementations again.  Suggest try without `localStorage` then, once we see what it looks like, we work through the possible start-up conditions. We'll never catch issues if we don't try it.

It would be great if you would go ahead with this.  Now is a good time.  Modes and visual mode are big recent changes which require shake-out time for bugs and feature tweaking.  We can use that same shake-out time for getting the settings implementation right.

mrmr1993:
> The approach you propose is better if we're confident that that will never happen.

We've been using `chrome.storage.sync ` as our main non-synchronous storage for quite some time now, and haven't seen any issues relating to this. We should keep an eye, but it seems that it has never happened before, so for now this assumption is safe.

> I'm not such a fan of that.

Nor am I. I was suggesting that it is a necessary evil, at least until we can confirm that we don't introduce any race conditions by removing it.

> It gives us two implementations again.

We can literally have:

```coffee
Settings =
  ...
  cache: if Utils.isBackgroundPage() then localStorage else {}
  # Settings are only available immediately on the background page.
  loaded: Utils.isBackgroundPage()
  ...
```

and treat `Settings.cache` as if a native object everywhere else. You could perhaps call this 2 implementations of the settings cache, but certainly not of settings, and it's so minimal that I don't think it's a major concern.


> We'll never catch issues if we don't try it.

Iit is trivial to add a flag to the conditional for `localStorage`, which we can toggle to disable it for testing. It will also be easy to remove (just drop those conditional branches), so I'm in favour of implementing with.

> It would be great if you would go ahead with this.

I'll get on this when I get home after the weekend.

smblott-github:
Yes, @mrmr1993.  I can buy all of that.  Let's try it like that.

mrmr1993:
> All of this business with JSON values is unnecessary.

@smblott-github I've just been thinking about this; we're going to completely break backwards compatibility of settings by doing this. Since a user could have multiple computers, one with an older version of Vimium (because eg. their Chrome hasn't been updated), we would break the settings badly on the older version. This isn't a call I want to be responsible for making; what do you think?

mrmr1993:
> All of this business with JSON values is unnecessary.

@smblott-github I've just been thinking about this; we're going to completely break backwards compatibility of settings by doing this. Since a user could have multiple computers, one with an older version of Vimium (because eg. their Chrome hasn't been updated), we would break the settings badly on the older version. This isn't a call I want to be responsible for making; what do you think?

smblott-github:
@mrmr1993. You're not worried about `localStorage`.  You wouldn't be worried about `chrome.storage.local` either.  Migrating both these cases is straightforward.  You're worried about settings propagated via `chrome.storage.sync`, yes?

I still favour storing everything in `chrome.storage.local` and using `sync` **only** for propagation.  (I know you mentioned alternatives, above, but this is a simple, uniform model, and a good starting point.)

Whatever approach you take, you're going to need migration code.  You can get over your concern with JSON and inconsistent Vimium versions, above, by naming settings in `sync` in such a way that they won't clash with settings names in legacy installations.  E.g. instead of setting `scrollStepSize` in `sync`, set `propagage-scrollStepSize`.  In fact, as a transitional measure, you could set a JSONified `scrollStepSize`, and a non-JSONified `propagate-scrollStepSize`.

Not perfect, but it would work.  And we *never* use `sync` from code, we *always* use `local` (or, actually, we always use `Settings`/`settings`).  Code should not be concerned with whether a setting is synced or not.

Your bigger point is probably that migration is tricky, which is correct.  But I don't think it's insurmountable.

mrmr1993:
I think I'm going to:
* internally use versioned settings names;
* expire non-current-version settings after 1 week;
* create a migration function that can do/undo every migration we've ever done. This way,
  - we can go from any version to the newest version, and
  - we can test PRs which update the settings in a non-backwards-compatible way, and then revert the change (so we can go back to `master`) by running the migration function from the background page.

Obviously, this expects a minor version bump (eg. 1.49.1) every time the settings are updated. Is all of this okay?

smblott-github:
> Is all of this okay?

I'd suggest getting the core functionality right first, then worrying about migration.  The migration can certainly be done.  If you want, try what you describe.  But I am wondering whether it might be more complex than necessary.

mrmr1993:
Closing this. Currently redoing it as a work in progress, starting with #1599.

PR #1291, opened 2014-12-09
Add numbered prefixes to LinkHints commands

### LINKED ISSUE #1289: [Feature request] Use number prefix for link hints to activate that number of links
Sometimes it is useful to perform a certain number of link hint operations consecutively, eg.

* copy urls of multiple links to the clipboard (so a download manager can add them to its list);
* open multiple links in an incognito window (although annoyingly each opens in a seperate window);
* (using PR #1032) hover over elements one after another to reveal extra information (eg. IMDb watchlist);
* (using PR #1049) opening multiple links in a new window (again, annoyingly, each is relegated to its own window).

At the moment we disable number prefixes for link hints, and to repeat link hints we have to re-enter the command's key mapping; it would be nice to enable them and use them for multiple links.

mrmr1993:
Resolves #1289.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -280,7 +280,7 @@ commandDescriptions =
   toggleViewSource: ["View page source", { noRepeat: true }]
 
   copyCurrentUrl: ["Copy the current URL to the clipboard", { noRepeat: true }]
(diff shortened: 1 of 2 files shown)


smblott-github:
We should do this, @mrmr1993.  The implementation might be a bit out of date, perhaps.

Issue #1274, opened 2014-11-27
Move key handling to the frontend


mrmr1993:
At present, we go via the background page to handle all mapped keys, which has a few issues and oddities:
- frontend commands are activated via the backend, so we have to follow/debug 3 codepaths for one event.
- frontend commands have no access to the key events, except via unpleasant hacks (eg. keyboard repeat for scrolling in #1238).
- multi-key mappings can be completed across tabs/windows, which has no obvious utility but is confusing.
- `completionKeys` aren't synced, so sometimes a multi-key mapping across two tabs/windows will fire, and sometimes it will not, depending on the current state of the second tab.
- we have to be be careful (and have duplicate logic) to ensure that we recognise handle or don't handle keystrokes consistently in the frontend and backend (eg. #1254)

I would like to rewrite the frontend to do all the main processing, reduce the impact of the code, add per url mappings (#1188) and add moded mappings (ie. `nmap`, `vmap`, `imap`, etc.). This would also break (nearly) every outstanding PR. Would such a change be likely to get merged? Does anyone have strong opinions relating to this?

I am motivated mainly by #61; to make an effective attempt at visual mode beyond what was done in #943, I would like to have more reliable, more concise and less complicated code for key handling.

PR #1315, opened 2014-12-15
Move key handling to the frontend

mrmr1993:
This is PR #1275, but as a pull request on `post-1.46`. This makes the diff much more manageable to look at. To quote #1275:

> This PR begins work on #1274.
* Key handling has been moved to a content script (`content_scripts/key_handling.coffee`).
* Key handling is now done synchronously as part of the event's key handler.
* Checks on whether we handle a key are all done using the same code. This ensures consistency (eg. now we don't pass `0` to the underlying page if we've handled it ourselves as part of a number prefix).
>
> ...
>
> Apart from bugfixes coming from unifying the frontend and backend key handling, this should function the same as before.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -4,16 +4,8 @@ currentVersion = Utils.getCurrentVersion()
 
 tabQueue = {} # windowId -> Array
 tabInfoMap = {} # tabId -> object with various tab properties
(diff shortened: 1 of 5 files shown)


mrmr1993:
This is stale, and #1775 does much the same thing, but better. Closing

Issue #1350, opened 2014-12-17
Include build date somewhere


smblott-github:
This is only an issue for developers.

It could be helpful to have the build date/time (and possibly even the last commit SHA, and the git branch) accessible in one of the extension pages, somewhere.  It can be difficult to keep track of which version you're running.

This need not be directly available to users.

smblott-github:
Closing.  See #1352.

PR #1352, opened 2014-12-17
Add a dev-build command to show the current branch/commit

mrmr1993:
Running the dev-build command adds the current branch under the title in the options page. Clicking this branch name shows the HEAD of the current branch.

This is intended to address #1350.

### DIFF
--- a/Cakefile
+++ b/Cakefile
@@ -55,6 +55,7 @@ task "dev-build", "compile all coffeescript files and add git information", ->
     branchDone and commitDone and appendToOptionsPage """
       branchName = #{currentBranch}
       commitData = #{commitSummary}
+      buildDate = "#{(new Date()).toISOString()}"
     """
 
   spawn = child_process.spawn "git", ["rev-parse", "--abbrev-ref", "HEAD"]
--- a/Cakefile
+++ b/Cakefile
@@ -40,8 +40,36 @@ visitDirectory = (directory, visitor) ->
     visitor(filepath)
 
 task "build", "compile all coffeescript files to javascript", ->
+  fs.unlink "pages/git-information.js", (->)
   coffee = spawn "coffee", ["-c", __dirname]
   coffee.on 'exit', (returnCode) -> process.exit returnCode
+
+appendToOptionsPage = (data) -> fs.writeFile "pages/git-information.js", data
+
+task "dev-build", "compile all coffeescript files and add git information", ->
+  invoke "build"
+
+  branchDone = commitDone = false
+  currentBranch = commitSummary = ""
+  tryAppendToOptionsPage = ->
+    branchDone and commitDone and appendToOptionsPage """
+      branchName = #{currentBranch}
+      commitData = #{commitSummary}
+    """
+
+  spawn = child_process.spawn "git", ["rev-parse", "--abbrev-ref", "HEAD"]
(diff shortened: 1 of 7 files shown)


smblott-github:
Yes please, @mrmr1993!

(Also, include the date of the build.)

mrmr1993:
Rebased and added date.

smblott-github:
Thanks, @mrmr1993.

Comments:
- Does this generate an error if `git-information.js` is absent (which is the common case)?
- It looks ugly, as is. (Sorry.)
- Why not put it at the bottom of the page?
- Why build the element in CS?  Instead, just have an element in `options.html` (at the bottom) which is only populated and visible if it is a dev built?

Cakefile
```diff
@@ -40,9 +40,38 @@ visitDirectory = (directory, visitor) ->
...
+  fs.unlink "pages/git-information.js", (->)
   coffee = spawn "coffee", ["-c", __dirname]
   coffee.on 'exit', (returnCode) -> process.exit returnCode
 
+appendToOptionsPage = (data) -> fs.writeFile "pages/git-information.js", data
+
+task "dev-build", "compile all coffeescript files and add git information", ->
+  invoke "build"
+
+  branchDone = commitDone = false
```
philc: This could use a one-line comment explaining what's happening: "we want to show the user what dev version they're running via the options page" and reference #1350.

smblott-github:
Thinking about this, @mrmr1993, ...

- It would be better if `git-information.js` was just *always* present (but usually empty).
- We add a `div`/`span` to both the help page (near the version info) and the options page (at the bottom) which is usually unpopulated, but is populated when the relevant variables are set.
- And we style it a bit.

(Also, is there away to avoid the rather ugly asynchronous build?)

mrmr1993:
> * It looks ugly, as is. (Sorry.)

I'm happy for you to style it however you feel is best. Layout's not really my thing.

> * Why not put it at the bottom of the page?

Not sure what you meant by bottom (the bar or the scrolling area), so I've left this to be done with styling.

> * Why build the element in CS? Instead, just have an element in `options.html` (at the bottom) which is only populated and visible if it is a dev built?

> * It would be better if `git-information.js` was just always present (but usually empty).
> * We add a `div`/`span` to both the help page (near the version info) and the options page (at the bottom) which is usually unpopulated, but is populated when the relevant variables are set.

Done.

> * And we style it a bit.

Not done.

> is there away to avoid the rather ugly asynchronous build?

Yes, done.

mrmr1993:
Full disclosure: currently, loading the help dialog on the options page causes the build details to be duplicated in the option page's build details, rather than inserting them into the help dialog. I've left this unfixed, expecting that we'll move the help dialog to an iframe soon.

smblott-github:
A couple of issues...

- Some of the build-info text seems to be displayed on non `dev-build` builds.

- Having separate build targets isn't great.  Also, it's unfortunate that we don't get correct build info with `autobuild`.  We could get around that by displaying the install date rather than the build date (because we can calculate that at install time).

- For me, the only info I care about is the build date.  It might be better to just *always* show the build date beside the version number on the help page.

In summary, I'm not sure this (in its current state) is as useful as I'd hoped it would be.

smblott-github:
Closing.
We have an approximation to this with #2014.

PR #1366, opened 2014-12-21
Match multiple exclusion rules

smblott-github:
This does the following:
- Changes the exclusion-rule logic such that, if multiple rules with pass keys match, then the keys taken is the union of the matches.  If _any_ matching rule has no pass keys, then Vimium is disabled.
- Substantially updates the page popup.  Basically, it now reuses the structure and styling from the options page (and we lose `popup.coffee`).
- Consistent look-n-feel wrt exclusion rules.

Example:

snapshot

Other points:
- The state that would apply _if the rules were to be save_ is given at the top of the popup.
- Patterns turn red when they don't match the current URL. (This looks better in practice than you might expect.)

Issues:
- ~~The _only_ way to delete rules is via the little cross.  So that workflow is not great.~~
- ~~@mrmr1993 wanted some user feedback as to what a red pattern means.  I can't figure out how to do that for now.~~
- The popup is wider than it was before.  Is that ok?
- (There must be more, I'll add them as I think of them.)

Fixes #1339.

Edit:

Additional notes:
- The CSS from the options page has been moved to its own file, and is now shared with the popup page.
- The HTML for exclusion rules has been moved to its own file, and is similarly shared.

Also: snapshot Fewer LOC!

### DIFF
--- a/pages/options.coffee
+++ b/pages/options.coffee
@@ -1,3 +1,4 @@
+
 $ = (id) -> document.getElementById id
 bgUtils = chrome.extension.getBackgroundPage().Utils
 bgSettings = chrome.extension.getBackgroundPage().Settings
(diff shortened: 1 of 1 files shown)


smblott-github:
Posting a few screenshots for use on the wiki...

snapshot

snapshot

snapshot

PR #1368, opened 2014-12-21
Support compound keys (<a-f>, <c-[>, etc.) in passkeys

mrmr1993:
This implements support for compound passkeys, which must be separated from other passkeys keys by a space. We require spaces since
* we can't map to the ` `(space) key, so it doesn't cause any problems, and
* we need to be able to have `<` and `>`, since some people have bindings using them.

This is a follow-up to an idea from the conversation on #1211.

@z0rch, @smblott-github, can you take a look, possibly in conjunction with #1140?

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -359,7 +359,7 @@ setPassKeys = (newPassKeys) ->
   passKeyGroups = passKeys.split " "
   splitPassKeys = []
   for passKeyGroup in passKeyGroups
-    namedKeyMatch = passKeyGroup.match namedKeyRegex
+    namedKeyMatch = passKeyGroup.search namedKeyRegex
     if namedKeyMatch # The current space delimited group is a single named key.
       splitPassKeys.push passKeyGroup # Push this to as a single key
     else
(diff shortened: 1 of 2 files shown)


mrmr1993:
@smblott-github this PR needs a rework, but is the idea a possibility? It's come back to mind based on #1560.

smblott-github:
@mrmr1993.  I'm not sure I'm seeing a use case compelling enough to warrant the additional code and (more importantly) UI complexity.  On most sites (GMail, etc), the important shortcuts are simple (ie. non modified), and most Vimium bindings are simple too.  It may be better to wait and see what use cases emerge.

mrmr1993:
> UI complexity

Surely the only UI change is adding a small note to the help text by exclusions/passkeys in the options page? Or am I missing something?

> It may be better to wait and see what use cases emerge.

Could those requests for `<esc>` to be passed to the page count? That was the original use case I had in mind when making this. `<c-[>` is still hardcoded to behave as `<esc>`, so this would provide an easy fix for those.

smblott-github:
Still not sure about this, @mrmr1993.

smblott-github:
Closing, @mrmr1993.

I still don't think we should do this.  Demand is weak.  And we have `passNextKey` now (which helps).

PR #1369, opened 2014-12-21
Support all mappable keys in passkeys and treat exit insert mode as a command

mrmr1993:
This PR combines #1140 and #1368, and adds a tweak to support passkeys for `exitInsertMode`. It does the majority of what is outlined in [this comment](https://github.com/philc/vimium/pull/1211#issuecomment-67776693) with a view to making #1211 mergable and enabled by default.

Specifically, this PR:
* adds passkey support for all mappable keys, with multi-character key representations recognised when delimited with spaces (via #1368).
* adds `exitInsertMode` and `exitInsertModeAndPassKey` commands (via #1140), which fixes #301 and starts work on resolving #1176.
* enable passkeys for the `exitInsertMode`/`exitInsertModeAndPassKey` commands.

Issue #1176, opened 2014-10-20
Pressing 'esc' no longer closes Facebook message windows


enoch:
Facebook message windows that are in focus (listening for text-input) are usually closed when 'esc' is pressed, but with vimium active hitting 'esc' only unfocuses the message window.

smblott-github:
#1985 will make this possible.

PR #1372, opened 2014-12-22
Complex passkeys, exit insert mode commands, and enable escape insert mode for embeds

### LINKED ISSUE #301: Allow any key to mapped to "exit mode" instead of just <esc>

mrmr1993:
This is a continuation of #1369, incorporating code from #1211. 

From #1369, this PR:
> * adds passkey support for all mappable keys, with multi-character key representations recognised when delimited with spaces (via #1368).
> * adds `exitInsertMode` and `exitInsertModeAndPassKey` commands (via #1140), which fixes #301 and starts work on resolving #1176.
> * enable passkeys for the `exitInsertMode`/`exitInsertModeAndPassKey` commands.

In addition, this PR:
* enables `exitInsertMode`/`exitInsertModeAndPassKey` for `<embed>` and `<object>` elements.

This new addition can be tested on [this webpage](http://www.bensilvis.com/?p=146).
* Pressing `esc` now blurs the embedded flash.
* To pass `esc` to the embedded flash, you can add `esc` as a passkey.
* `exitInsertModeAndPassKey` can also be mapped to `esc` to blur *and* pass the key (but is generally a bad idea until something happens with #1188).

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,11 @@ Please see [CONTRIBUTING.md](https://github.com/philc/vimium/blob/master/CONTRIB
 
 Release Notes
 -------------
+1.49 (2014-12-16)
+
+- An option to toggle smooth scrolling.
+- Make Vimium work on older versions of Chrome.
+
 1.46, 1.47, 1.48 (2014-12-15)
 
 - Site-specific excluded keys: you can disable some Vimium key bindings on sites like gmail.com, so you can use the key bindings provided by the site itself.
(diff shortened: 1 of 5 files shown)


dentafrice:
Any updates on if/when this could be merged?  I ran into the same issue when dealing with escaping from a modal that autofocuses an input.

smblott-github:
> I ran into the same issue when dealing with escaping from a modal that autofocuses an input.

Thanks @dentafrice.  I don't understand what this means.  What problem did you encounter?

PR #1383, opened 2014-12-28
Merge post 1.46

smblott-github:
1.49 has been out for a couple of weeks now and doesn't seem to contain any more show stoppers.

So... This PR merges some of the changes which are in `post-1.46`, but not yet in `master`.

Included:
- #1286, #1360, #1377 

Omitted:
- #1064 and #1349 (hide cursor on scroll).  This is not a high-value feature, and has so far proven unreliable in practice (works on some sites, not others; triggers chrome bug on some sites for some chrome versions).
- #1214 (vomnibar in iframe).  See discussion in #1382.  The version currently in `post-1.46` is a little laggy, and we have not yet resolved whether and how to fix it (one proposal is in #1382).

Going forward:
- I propose that we work primarily off `master`.
- However, I'll keep `post-1.46` around for a bit, in case we make progress on any of the remaining features.


smblott-github:
Closing, due to #1407.

PR #1384, opened 2014-12-28
Uniform treatment of search queries

### LINKED ISSUE #1334: Multi-word Searches for Custom Search Engine Broken
When performing a custom search with multiple words you get search results from the default search engine containing the custom search URL and search terms. For example, I have set up a custom search for YouTube (`yt: http://www.youtube.com/results?search_query=%s`). If Google is your default search engine, and do a search for `foo bar`, you get sent [here](https://www.google.com/search?q=http%3A%2F%2Fwww.youtube.com%2Fresults%3Fsearch_query%3Dfoo%20bar&gws_rd=ssl), instead of [here](https://www.youtube.com/results?search_query=foo+bar). I find that typing a `+` between words gets you the right result.

smblott-github:
Our approach to building search URLs for the default search engine and for custom search engines has been different.  See #1334 for a case which is broken.  In fact, if you define Wikipedia to be your default search engine, then our long-established existing implementation is broken too

This PR introduces a consistent approach to building search URLs:
- URL encode each "word"
- Join url-encoded words with `+` instead of `Space`.

This is how chrome handles custom search engines, and seems (?) to work in more cases.

Note... There's a long discussion in #1169 about custom search engines which are admitted by our current implementation but probably shouldn't be.  This PR breaks such custom search engines (e.g it breaks a "custom search engine" which calls arbitrary Javascript).

Replaces #1335.
Fixes #1334.

### DIFF
--- a/lib/utils.coffee
+++ b/lib/utils.coffee
@@ -96,8 +96,8 @@ Utils =
 
   # Creates a search URL from the given :query.
   createSearchUrl: (query) ->
-    # it would be better to pull the default search engine from chrome itself,
-    # but it is not clear if/how that is possible
+    # It would be better to pull the default search engine from chrome itself.  However, unfortunately chrome
+    # does not provide an API for doing so.
     Settings.get("searchUrl") + @createSearchQuery query
 
   # Converts :string into a Google search if it's not already a URL. We don't bother with escaping characters


mijoharas:
Do we know when the next release is likely to be? I just got annoyed enough by the broken multi-word searches to come over here and fix it, looks like it's already been fixed :smile: . Will this be in the next release?

smblott-github:
> Will this be in the next release?

I expect so.

> Do we know when the next release is likely to be?

I don't. @philc's the man for that. `master` is not really ready for a release.  It's seen some big changes which need some shake-out time.  However, it might be possible to cherry-pick some of smaller but bad-UX fices.

mijoharas:
Is there any help needed with that? any way to contribute, do we think it's worthwhile or should we just wait for master to stabilize? Any thoughts on that @philc ?

PR #1390, opened 2014-12-29
Set up Vomnibar iframe as a UIComponent

mrmr1993:
This PR uses the `UIComponent` framework set up in #1388 and #1385 to manage the Vomnibar iframe, and removes some unused code from the Vomnibar scripts.

@smblott-github can you take a look at this?

### DIFF
--- a/tests/dom_tests/dom_tests.html
+++ b/tests/dom_tests/dom_tests.html
@@ -36,13 +36,9 @@
     <script type="text/javascript" src="../../lib/clipboard.js"></script>
     <script type="text/javascript" src="../../content_scripts/ui_component.js"></script>
     <script type="text/javascript" src="../../content_scripts/link_hints.js"></script>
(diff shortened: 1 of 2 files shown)


smblott-github:
@mrmr1993.  We're catching a trailing `keyup` on `Escape` from the vomnibar:
- Go [here](http://unixpapa.com/js/testkey.html).
- `o`, `Esc`, `o`, `Esc`

You'll see the misplaced `keyup`.  This affects Google Inbox in practice.

Any idea how we might suppress it?

mrmr1993:
> Any idea how we might suppress it?

I would `KeydownEvents.push` a fake `esc` event in the close handler.

smblott-github:
> I would KeydownEvents.push a fake esc event in the close handler.

Hmmm.  Not great, but I don't have a better idea right now...

Issue #1359, opened 2014-12-19
Cursor not always hidden until after first move (post-1.46 only)


smblott-github:
@mrmr1993.  I merged #1348.  Thanks.

However, I'm seeing a minor issue on some pages.  The cursor hider works, but only *after* the cursor has been moved on the page.  When the page first loads and the cursor has not yet been moved, the cursor is not hidden on scroll.

Examples: [here](http://www.bbc.com/) and [here](http://www.irishtimes.com/).

(Note: this does not affect `master` or `1.49`).

PR #1391, opened 2014-12-30
Add setting to allow manually enabling CursorHider

mrmr1993:
After the discussion in #1348, cursor hiding was completely disabled. Since it still has some utility, I [proposed there](https://github.com/philc/vimium/issues/1359#issuecomment-67738687) that we should have some way of enabling it without having to modify the source code. This PR gives such a method.

With this PR, `CursorHider` can be enabled by running `Settings.set("enableCursorHider", true)` in the background page.


smblott-github:
The cursor hider is dead code, @mrmr1993.  I think we should remove it.

I you want it, then it's not too hard to add it to your own live build in such a way that it won't get stomped on by future merges (e.g. put the bulk of the code in a separate file).

(In the interest of equity, the edit mode code should be removed too.)

mrmr1993:
Agreed, the code should be killed. I can open a PR for that if you'd like.

smblott-github:
@mrmr1993.  Yes.

PR #1401, opened 2014-12-30
Move the HUD to a UIComponent iframe

### LINKED ISSUE #556: Use an text box for key entry in Vimium's find mode

mrmr1993:
This PR
- moves the HUD to an iframe, with its interface provided by `UIComponent`
- updates `Tween` to keep track of its own timers and to modify CSS rather than directly changing an element's style.
- uses the HUD iframe to provide an input (a `<span contentEditable="plaintext-only">` element) for find mode, which
  - supports copy/paste
  - accepts characters which are generated from multiple subsequent keypresses
  - provides a caret, so the user can modify a search without re-typing it

It also includes a couple of simple fixes:
- A typo where `vimiumUIComponentShowing` should have been `vimiumUIComponentVisible`.
- A place where the update from `handledKeydownEvents` to `KeydownEvents` had missed.

This fixes #556.

### DIFF
--- a/content_scripts/hud_controller.coffee
+++ b/content_scripts/hud_controller.coffee
@@ -75,7 +75,7 @@ root.HUD =
       del: handleDeleteForFindMode
       enter: handleEnterForFindMode
     findModeQuery.rawQuery = data.query
-    handlers[data.type]?()
+    handlers[data.type]()
     @hudUI.hide()
 
   isReady: -> document.body != null
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -3,9 +3,9 @@ class UIComponent
   iframePort: null
   showing: null
   loaded: false
-  queuedActions: []
 
   constructor: (iframeUrl, className, @handleMessage) ->
+    @queuedActions = []
     @iframeElement = document.createElement "iframe"
     @iframeElement.className = className
     @iframeElement.seamless = "seamless"
--- a/pages/HUD.coffee
+++ b/pages/HUD.coffee
@@ -33,7 +33,7 @@ enterFindMode = (data) ->
     # Strip newlines in case the user has pasted some.
     UIComponentServer.postMessage name: "search", query: inputElement.innerText.replace(/\r\n/g, "")
 
-  inputElement.addEventListener "keydown", (event) ->
+  document.addEventListener "keydown", (event) ->
     if KeyboardUtils.isEscape event
       eventType = "esc"
     else if event.keyCode in [keyCodes.backspace, keyCodes.deleteKey]


content_scripts/hud_controller.coffee
```diff
@@ -0,0 +1,123 @@
...
+    findModeQuery.rawQuery = data.query
+    handlers[data.type]?()
+    @hudUI.hide()
+
+  isReady: -> document.body != null
+
+  # A preference which can be toggled in the Options page. */
+  enabled: -> !settings.get("hideHud")
+
+class Tween
```
smblott-github: Having `Tween` out of the front end is already an improvement.  No need to move it to the HUD frame, for now.

smblott-github:
This looks really nice, @mrmr1993.

Unfortunately, there's an issue.  And I can't immediately track it down.

With this, we can use `n` to search through text areas.  First hit in a text area, and we're dropped into insert mode.  Can you double check, please?

mrmr1993:
Not at my computer, but it sounds like I might have missed a `findMode` when changing them all to `window.findMode`. I'll have a look when I get home.

mrmr1993:
Fixed the error, and rebased to fix the merge conflict.

smblott-github:
Hi @mrmr1993.

There might be a race condition:
- This doesn't work: ae0afcd68472feebd67929c9b521956f63744285.
- This does: 86a7ecc0f899a4b64903cb2c59c4e3dc2b34710b.

This could affect the upgrade notification.

Or perhaps I'm putting my test in an unreasonable place?

mrmr1993:
>There might be a race condition:

Not a race condition, just that the frame hasn't loaded yet, so the request goes nowhere. 

I originally wrote code in `UIComponent` that queued requests made before the frame loaded and ran them when we were ready. At the time it wasn't used anywhere, so I junked it to reduce code bloat. Shall I reintroduce it?

smblott-github:
> just that the frame hasn't loaded yet

Yes, I guessed that.  That's a race condition.

> Shall I reintroduce it?

Well, we *do* need to fix this.  My first test was on the upgrade notification, and it does fail too, currently.

Queuing requests is fine if there's only one queued request.  It's not obvious what you do if there's more than one.  Do you handle them all in turn?  Discard all by the last?  Do you collapse together requests of the same type?  Not sure.

Possibly *discard all but the last* is safe.  If the upgrade notification is trumped, it'll still be displayed the next time a page loads.

Or, might it be enough to treat the upgrade notification specially?

(Not sure about all of this.)

mrmr1993:
Queuing them all seems the best option, in case we send multiple requests that update different state in the frame. Means we have to be slightly more race-condition-aware when doing frame -> background communication, but that should be quite simple (closing old ports before we open new ones).

I'll push something in a bit.

mrmr1993:
Implemented request queuing, as described in my previous comment. I've also touched some (6 lines) Vomnibar code to make sure the tests still pass, and to stop us from handle stale requests.

content_scripts/hud_controller.coffee
```diff
@@ -0,0 +1,123 @@
...
+  hideUpgradeNotification: ->
+    @upgradeTween.fade 0, 150, => @upgradeUI.hide false
+
+  showFindMode: ->
+    clearTimeout @showForDurationTimerId
+    @hudUI.activate {name: "find"}
+    @hudTween.fade 1.0, 150
+    # Refocus the HUD if the user focuses this window.
+    window.addEventListener "focus", @focusFindModeHUD, false
+    document.documentElement.addEventListener "mouseup", @focusFindModeHUD, false
```
smblott-github: We bounce them back to the HUD.  Is this the best UX?  It's not what Chrome does.  Surely the user is always right, here.  If they clicks on a text area, for example, then they probably want to edit the text area.  Why should we stop them?

content_scripts/hud_controller.coffee
```diff
@@ -0,0 +1,123 @@
...
+  updateMatchesCount: (count) ->
+    @hudUI.postMessage {name: "updateMatchesCount", count}
+
+  hideFindMode: (data) ->
+    handlers =
+      esc: handleEscapeForFindMode
+      del: handleDeleteForFindMode
+      enter: handleEnterForFindMode
+    findModeQuery.rawQuery = data.query
+    handlers[data.type]?()
```
smblott-github: Do we always expect only the three listed cases, here?  If so, it might be better to drop the `?`.  It masks errors which we really should know about, and it should never happen on release anyway.

smblott-github:
Try running this [branch](https://github.com/smblott-github/vimium/tree/hud-to-frame-uicomponent-wierd).  You get the vomnibar on every page load.  Weird.  Whatever's causing that, could it be triggered in any other way?  E.g., when the upgrade notification shows?

mrmr1993:
> You get the vomnibar on every page load.

Oops! Rookie error: `queuedActions` should be set in the constructor, not on the prototype. I'll push a fix when I get home after the weekend.

content_scripts/hud_controller.coffee
mrmr1993: My thinking was to let the can change the selection, so they can continue seeing partial matches from the new selection. I've only personally used it a couple of times, but it's useful for skipping over a particularly match-dense section.

With #1424 (which I think we should definitely consider merging), the user will be able to use the focused element as soon as they exit find mode with `esc`. I don't think exiting find mode by clicking is the right UX (and nobody's complained that it wasn't possible up to now).

smblott-github:
Two issues, related.

First, as I noted elsewhere, I'm not sure bouncing users back to the HUD on "click" is the right UX.  Probably just hide the HUD, no?

Second (related), there's a glitch:
- Start finding: `/abc`
- Click on the document somewhere (you're bounced back to the HUD).
- Now, `Enter` and `Escape` don't work.  They're only reactivated after you type a normal key (`Backspace` doesn't do it).

However, if you address the first issue, then the second one goes away.

mrmr1993:
> I'm not sure bouncing users back to the HUD on "click" is the right UX. Probably just hide the HUD, no?

We've always ignored user clicks and left the HUD open; this has never been a problem before. 'Ain't broke, don't fix it, and all that.

> `Enter` and `Escape` don't work. They're only reactivated after you type a normal key (`Backspace` doesn't do it).

Good catch. I'll move the `keydown` listener to `documentElement`.

smblott-github:
> We've always ignored user clicks and left the HUD open...

Hmm.  Well it still seems like the wrong UX to me.  Do you have a view, @philc?  In find mode, if the user clicks on the page, should the HUD grab back the focus?  Or get out of the way (a la chrome)?

mrmr1993:
> Well it still seems like the wrong UX to me.

The current implementation let's you change the selection, so future changes to the search are run from the there instead. We lose this, and change existing UX, for no benefit as far as I can see. And it complicates the code (extra click handler).

smblott-github:
> The current implementation let's you change the selection

Could you flesh out the use case for that?

I'm on some big page like [this](http://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html), and I want do, ... what?  And I do it... how?

mrmr1993:
Specifically, I've used it for a GitHub code page where I was looking for a variable, selecting at the bottom of a function that used the variable a lot (to skip it) so it just jumped to the declaration I wanted as I typed the variable. Not very useful, but IMHO more so than being able to exit find mode with the mouse.

I can give a concrete example when I'm next at a computer in a couple days, if that would help.

mrmr1993:
I've made the small changes that I said I would make, and rebased onto the current master (so travis tests pass again).

@smblott-github suggest opening an issue if you are still concerned that clicking the document should exit the search. This PR isn't about that UI change, and so should keep the old behaviour.

smblott-github:
> This PR isn't about that UI change, and so should keep the old behaviour.

That's reasonable.

> @smblott-github suggest opening an issue if ...

Will do.

mrmr1993:
@smblott-github is there anything else that needs to be done here? It would be good to see this merged sooner rather than later.

philc:
Let's merge

smblott-github:
@mrmr1993.  I want to merge either this and then #1413, or both at the same time.  My working merge is [here](https://github.com/smblott-github/vimium/tree/modes-dev-with-HUD-as-component).  It contains this, #1413 and #1431.

But there's a problem.  See b92c208536b022df99581e40b69cfa94812f7289 (which is just a place holder for one possible fix).  From there:

> Search with "/".  If the selection ends up in an input element, then, because the focus is in the HUD frame, the input element doesn't appear to get a focus event.  So, when we leave the HUD, it's not focused.  So, the selection is not visible.

My idea was to get the selection, find any (the highest) enclosing element for which `isEditable` would be true, and activate/focus it.  Unfortunately, I can't figure out how to do that.

Do you have any ideas?

smblott-github:
To test the problem, with this branch (not merged with #1413):
- go [here](http://static.smblott.org/inputs.html) (my junk page for testing).
- `/sssw<Enter>`
- no selection!

smblott-github:
@mrmr1993.  Is this fixable?  It would be great to merge it, a substantial improvement to the find-mode UX.

It looks like I changed the junk/test page referenced above.  It looks like now you'd have to search for `three` -- basically any search which lands in a text input.

Don't worry about the merge or rebasing.  I can take care of that.

content_scripts/ui_component.coffee
```diff
@@ -39,13 +60,21 @@ class UIComponent
...
-  hide: (focusWindow = true)->
-    @iframeElement.classList.remove "vimiumUIComponentShowing"
+  hide: (focusWindow = true, forceImmediate = false) ->
+    unless @loaded or forceImmediate
+      @queueAction "hide", arguments
+      return
+
+    @iframeElement.classList.remove "vimiumUIComponentVisible"
     @iframeElement.classList.add "vimiumUIComponentHidden"
     window.focus() if focusWindow
```
mrmr1993: > Is this fixable?

@smblott-github changing this line to `(document.activeElement ? window).focus()` should do the right thing. It was a silly oversight, sorry.

We should make sure to guard against `document.activeElement` being `null`, since the [spec specifically says that it can happen](https://html.spec.whatwg.org/#focus-management-apis:document-8).

smblott-github: That works: a3909bafd3a38d2d4cd780c757a136e2e0f20dee.
Your work now copied [here](https://github.com/smblott-github/vimium/tree/mrmr1993-hud-to-frame-uicomponent).

smblott-github:
My initial attempt at a merge is [here](https://github.com/smblott-github/vimium/tree/mrmr1993-hud-to-frame-uicomponent-merge).  There are still some (quite significant) issues, see notes in 443b6ca5c68f7397d11d9f18269853fff696fc35.  However, these remaining issues should just be a matter of moving code around. I'll get to it in the coming days.

Thanks, @mrmr1993.

mrmr1993:
@smblott-github I looked at [this one](https://github.com/smblott-github/vimium/commits/mrmr1993-hud-to-frame-uicomponent-merge-v2) instead, after seeing your commit message in ffc6b9a4d2f53116f3d557b3c307d598dcf8c262. Hopefully that was the right thing to do!

Everything there looks fine to me, except for stray conflict markers in `tests/dom_tests/dom_tests.html`.

smblott-github:
> Hopefully that was the right thing to do!

Yes.

> Everything there looks fine to me...

Yes.  Except it's kinda broken.  I'm having difficulty with some issues to do with the focus, I think.  Here are some symptoms:
- `/something<Enter>`, change tab, change back, `o` -- you're still in find mode.
- refresh tab, `/`, up-arrow; history works. `Escape`.  Enter find mode again, and history finds the right elements, but the text does not appear in the HUD.
- `/something<enter>`.  The active element is `document.body`.  Change tab, change back.  The active element is the HUD.

I suspect these all stem from the same source, but can't find it.

I've pushed a broken commit to [here](https://github.com/smblott-github/vimium/commits/mrmr1993-hud-to-frame-uicomponent-merge-v2), so you can see where things stand.  There may be a `...-v3` branch coming!

(These issues were all introduced in the merge.)

smblott-github:
Another symptom:

- With this PR, when you hit `Enter` in find mode, the HUD is hidden immediately.
- With the broken merge, the HUD flickers (so it's hidden immediately, then re-shown and fades, so tween).

That may be more help for tracking it down.

content_scripts/hud_controller.coffee
```diff
@@ -0,0 +1,123 @@
...
+  showFindMode: ->
+    clearTimeout @showForDurationTimerId
+    @hudUI.activate {name: "find"}
+    @hudTween.fade 1.0, 150
+    # Refocus the HUD if the user focuses this window.
+    window.addEventListener "focus", @focusFindModeHUD, false
+    document.documentElement.addEventListener "mouseup", @focusFindModeHUD, false
+
+  focusFindModeHUD: ->
+    return unless findMode
```
mrmr1993: @smblott-github I think i remember this line being killed by the merge; seems like a likely candidate for the focus change issue.

smblott-github: > I think i remember this line being killed by the merge

@mrmr1993.  Yes.  That was it. Thanks.  I changed it to remove the `focus` and `click` listeners when find mode exits.

PR #1406, opened 2015-01-01
Modal browsing, and towards visual mode (a proof of concept)

smblott-github:
A major goal for the next release is visual mode.  This is a proof-of-concept illustrating how we can introduce a clean, modal browsing experience within our existing framework.  And hence, how we can move forward to visual mode (and others).

Design goals:

- Re-use existing code where possible.  Because, if it works, let's use it.

- To the greatest extent possible, limit changes to existing code.  Because changes introduce unintended side effects and bugs; changes always come with risk.

- Allow modal browsing.  While the immediate goal is visual mode, ultimately we may introduce other new modes too.  For example, a vim-like editing mode.  We need an architecture which allows us to move smoothly and orthogonally between modes.  Examples:
  - `normal` -> `visual`
  - `normal` -> `insert`
  - `normal` -> `hint`
  - `normal` -> `edit` -> `insert`
  - `normal` -> `edit` -> `visual`
  - `normal` -> `edit` -> `visual` -> `edit` -> `insert` (and so on)

This proof-of-concept does not introduce any new functionality.  It does:

- Introduce a class `Mode` for handling modes and mode transitions.  Basically, this implements a stack of modes using the existing `HandlerStack` infrastructure for key handling.

- Move normal Vimium key handling to a mode (`normal`).  So `normal` mode is handled just like any other mode.

- Move insert mode and passkeys handling to new modes (`insert` and `passkeys`). `normal` mode and `passkeys` mode are always on the mode stack.  They are never deactivated.  Link-hint mode is not yet incorporated.

- Provide visual feedback as to the current mode via a badge on the browser popup.

Feedback would be appreciated.

Edit.  I forgot to add...
- With `insert` and `passkeys` mode incorporated within the mode framework, much of the current `normal`-mode key handling can be simplified.  To keep the PR clean, I've left that out for the moment.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -20,8 +20,8 @@ isEnabledForUrl = true
 passKeys = null
 keyQueue = null
 # The user's operating system.
-currentCompletionKeys = null
-validFirstKeys = null
+currentCompletionKeys = ""
+validFirstKeys = ""
 
 # The types in <input type="..."> that we consider for focusInput command. Right now this is recalculated in
 # each content script. Alternatively we could calculate it once in the background page and use a request to
@@ -109,9 +109,29 @@ initializePreDomReady = ->
   settings.addEventListener("load", LinkHints.init.bind(LinkHints))
   settings.load()
 
-  nf = -> true
-  new Mode(onKeydown, onKeypress, onKeyup, nf)
+  # Install normal mode. This will be at the bottom of both the mode stack and the handler stack, and is never
+  # deactivated.
+  new Mode
+    name: "normal"
+    keydown: onKeydown
+    keypress: onKeypress
+    keyup: onKeyup
+
+  # Initialize the scroller. The scroller installs key handlers, and these will be next on the handler stack,
+  # immediately above normal mode.
   Scroller.init settings
+
+  handlePassKeyEvent = (event) ->
+    for keyChar in [ KeyboardUtils.getKeyChar(event), String.fromCharCode(event.charCode) ]
+      return handlerStack.passThrough if keyChar and isPassKey keyChar
+    true
+
+  # Install passKeys mode. This mode is never deactivated.
+  new Mode
+    name: "passkeys"
+    keydown: handlePassKeyEvent
+    keypress: handlePassKeyEvent
+    keyup: -> true # Allow event to propagate.
 
   checkIfEnabledForUrl()
 
@@ -137,7 +157,7 @@ initializePreDomReady = ->
     getScrollPosition: -> scrollX: window.scrollX, scrollY: window.scrollY
     setScrollPosition: (request) -> setScrollPosition request.scrollX, request.scrollY
     executePageCommand: executePageCommand
-    getActiveState: -> { enabled: isEnabledForUrl, passKeys: passKeys }
+    getActiveState: -> { enabled: isEnabledForUrl, passKeys: passKeys, badge: Mode.getBadge() }
     setState: setState
     currentKeyQueue: (request) -> keyQueue = request.keyQueue
 
@@ -171,10 +191,7 @@ initializeWhenEnabled = (newPassKeys) ->
     # Key event handlers fire on window before they do on document. Prefer window for key events so the page
     # can't set handlers to grab the keys before us.
     for type in ["keydown", "keypress", "keyup"]
-      do (type) ->
-        installListener window, type, (event) ->
-          console.log type
-          handlerStack.bubbleEvent type, event
+      do (type) -> installListener window, type, (event) -> handlerStack.bubbleEvent type, event
     installListener document, "focus", onFocusCapturePhase
     installListener document, "blur", onBlurCapturePhase
     installListener document, "DOMActivate", onDOMActivate
@@ -192,7 +209,7 @@ setState = (request) ->
 window.addEventListener "focus", ->
   # settings may have changed since the frame last had focus
   settings.load()
-  chrome.runtime.sendMessage({ handler: "frameFocused", frameId: frameId })
+  chrome.runtime.sendMessage({ handler: "frameFocused", frameId: frameId, badge: Mode.getBadge() })
 
 #
 # Initialization tasks that must wait for the document to be ready.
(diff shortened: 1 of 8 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -21,8 +20,8 @@ isEnabledForUrl = true
 passKeys = null
 keyQueue = null
 # The user's operating system.
-currentCompletionKeys = null
-validFirstKeys = null
+currentCompletionKeys = ""
+validFirstKeys = ""
```
mrmr1993: Is there a good reason to do this? At the moment, initialising to `null` makes it easy to debug the case where something has gone wrong with getting these from the background (since we get an error message).

Also, `currentCompletionKeys` and `validFirstKeys` are both arrays when initialised, so this is misleading.

mrmr1993:
This is a beautiful @smblott-github, works perfectly for me!

mrmr1993:
Been running this for a bit. You've changed the behaviour so we block every `keydown` / `keyup` / `keypress` event that isn't a passkey while we're in normal mode. This means we:

* block keys even if we can't map them (eg. `<enter>`, `<space>`, etc.), 
* have to be in insert mode to pass any key we can't describe with a single character (for passkeys).

Can you update this to match the old behaviour? This is a real pain for websites with shortcuts.

smblott-github:
> Can you update this to match the old behaviour?

Will do.  Remember, this is still in the proof-of-concept stage.  There are lots of small issues.

content_scripts/vimium_frontend.coffee
smblott-github: > Is there a good reason to do this?

This was changed only to fix the tests.  One of the DOM tests bubbles a key through the handler stack (which previously was empty, but now contains normal mode) without having done the necessary initialisation.

I'll go back and look at fixing the test properly.

smblott-github:
> Can you update this to match the old behaviour? 

Thanks, that was a bug.

mrmr1993:
I've been thinking about this again. I'm not sure the implementation here really adds anything apart from complexity:
* the majority of value of modes comes from commands, which this doesn't add to at all;
* `onDeactivate`/`onReactivate` are already provided by `exitSomethingMode`/`enterSomethingMode`;
* having separate `keyup`/`down`/`press` handlers is only really useful for insert mode.

I'd like to propose an alternative, based on [my comment](https://github.com/philc/vimium/pull/1140#discussion_r22199333) on #1140:
* a keydown/keypress handler
```coffeescript
onKeyDown = onKeyPress = (event) ->
  # Get a list of all the modes that currently apply (eg. find and normal; dialog showing and
  # normal; find, dialog and insert; etc.)
  activeModes = getActiveModes()
  command = KeyHandler.getCommand event, activeModes
  # shouldExecute is set based on whether this is the right event (keydown vs keypress) and whether
  # the current keyQueue completely matches the command, or only does so partially.
  executeCommand command if command.shouldExecute
  DomUtils.stopPropagation event if command.noPropagate
  unless command.passKey
    event.preventDefault()
    HandledKeys.push event
```
* a mode-aware command getter
```coffeescript
KeyHandler =
  ...
  getCommand: (event, modes) ->
    currentMode = modes.pop()
    return @noopCommand unless currentMode?
    keyChar = KeyboardUtils.getKeyChar event
    @keyQueue.push keyChar
    # Get a command matching the current keyQueue for mode if there is one, or null otherwise.
    command = Commands[currentMode].getCommandFromKeyQueue @keyQueue
    if command
      @keyQueue = [] if command.shouldClearKeyQueue
      command
    else if modeDetails[currentMode].fallThrough
      KeyHandler.getCommand event, modes
    else
      @noopCommand
```

This assumes that #1315 has been merged, but stubbing `Commands.mode.getCommandFromKeyQueue` to check `currentCompletionKeys` and `validFirstKeys` should work for now. @smblott-github, any thoughts?

smblott-github:
> really adds anything apart from complexity...

There's an element of truth to that, and I have some sympathy with that point if view.  Modes are just providing a structured wrapper around `handlerStack`.  However, I think there's more to it than that.  Mainly orthogonality.

- Look how easy it is to add a new mode: [here](https://github.com/smblott-github/vimium/blob/modes---proof-of-concept---dev/content_scripts/mode_visual.coffee) and [here](https://github.com/smblott-github/vimium/blob/modes---proof-of-concept---dev/content_scripts/vimium_frontend.coffee#L343).
- All of the code associated with a new mode is self contained.
- (It handles badges beautifully, even if many modes are active.)
- We have a framework for re-using one mode in the context of another (e.g. find mode in the context of visual mode, or visual or insert mode in the context of an editing mode.)  That's orthogonality.  Just say `new VisualMode`, and you're good to go.
- Many things become super-easy to write (again, because of orthogonality).  Compare [this](https://github.com/philc/vimium/blob/master/content_scripts/vimium_frontend.coffee#L705), which becomes [this](https://github.com/smblott-github/vimium/blob/modes---proof-of-concept---dev/content_scripts/vimium_frontend.coffee#L848).  Or [this](https://github.com/smblott-github/vimium/blob/modes---proof-of-concept---dev/content_scripts/vimium_frontend.coffee#L770), which allows us to collect together all (well, probably most) of the quirks of the interactions between `find` and `insert` modes in a single place (a hideous problem to get right).

I've been working on my [dev](https://github.com/smblott-github/vimium/tree/modes---proof-of-concept---dev/content_scripts) version of this, and the orthogonality comes to the rescue again and again.

> onDeactivate/onReactivate are already provided by...

I got rid of them quite early on.  I agree, they're not useful.

> having separate keyup/down/press handlers is only really useful for insert mode

Not really.  Passkeys uses this, as do some of the utility modes I implemented along the way.  Some modes should bubble events they're not interested in (e.g. passKeys, and some utility modes), others want to capture all keyboard events.  It's trivial to write the handlers -- e.g. [here, again](https://github.com/smblott-github/vimium/blob/modes---proof-of-concept---dev/content_scripts/mode_visual.coffee#L12).

smblott-github:
Consider this problem.  We might have visual mode and edit mode.  Edit mode always works within an editable element.  Visual mode might have access to the whole document.  Or, if called from within edit mode, it should only have access to the element being edited.

Now, what happens when the user clicks outside of the element being edited?  Hmm, things get tricky.  And you have to solve it for every case.

Here's (roughly) how you handle that within this framework.  You create a class as follows....

```coffeescript
# When the user clicks anywhere outside of the given element, the mode is exited.
class ConstrainedMode extends Mode
  constructor: (@element, options) ->
    super options

    @handlers.push handlerStack.push
      "click": (event) =>
        element = event.srcElement
        @exit() if @element and not @isDOMDescendant @element, element
        @continueBubbling
        
  isDOMDescendant: (parent, child) ->
    node = child
    while (node != null)
      return true if (node == parent)
      node = node.parentNode
    false
```

Then, visual mode and edit mode are just subclasses...

```coffeescript
  class VisualMode extends ConstrainedMode
    ...

  class EditMode extends ConstrainedMode
    ...
```

And those new modes don't have to worry at all about users clicking outside of them (well, probably mostly).

That's orthogonality.  That's the big win from this proposed approach.

(To compare, look at the hideous tangled mess of fragile logic we currently have in `vimium_frontend.coffee` for handling normal, find and insert modes.)

mrmr1993:
> Not really. Passkeys uses this, as do some of the utility modes I implemented along the way.

These are all basically the same handler as far as I can see; the logic distills down to:
* does this key match any commands? if so
  * execute a command if its whole mapping has been typed.
  * suppress the key event.
* if not
  * suppress the key event (if we want to do so for this mode)

which I think should be divorced from the modes themselves, in some dedicated key handling module. This is especially true if we're planning to get #1315 in (are we, by the way?) since command matching is just checking whether the current key sequence matches any mappings.

> That's the big win from this proposed approach.

I agree, this is the right way to handle mode transitions, and the code looks very elegant. I hadn't been looking at your branch, so my comments may or may not apply to there.

smblott-github:
> if we're planning to get #1315 in (are we, by the way?)

Initially, I don't want to handle keys mappings.  That's really a separate issue.  However, because of keyboard layouts and internationalisation issues, we *will* be forced to go there eventually, probably sooner rather than later.  When we do, I cannot imagine us implementing new mappings via the background page.  So, #1315 or something like it is definitely on the cards.

smblott-github:
Closing in light of #1413.

Issue #1408, opened 2015-01-02
dom_tests::createGeneralHintTests does not generate tests for filtered link hint mode


maximbaz:
Take a look at the `dom_tests::createGeneralHintTests` function.
It takes `isFilteredMode` argument, but never makes a use of it.

I suppose the argument should dictate the value of `filterLinkHints` setting, which is currently always `false`. But replacing `false` with `isFilteredMode` results into 4 failures when running the tests.

```coffee
# Generate tests that are common to both default and filtered link hinting modes.
createGeneralHintTests = (isFilteredMode) ->
  context "Link hints",
    setup ->
      ...
      stub settings.values, "filterLinkHints", false
```

Checked in `post-1.46`.

PR #1405, opened 2014-12-31
[Post-1.46] Enable link hints for all elements with click listeners, and drop some legacy link hint conditions

mrmr1993:
This PR is #1167 rebased onto `post-1.46`.

This also contains a check for whether our `addEventListener` hook has been applied successfully:
* if so, we drop a couple of legacy link hint conditions (`tabIndex`, `button` in `className`)
* if not, we fall back to the legacy conditions.

This removes a lot of the duplicate/useless link hints we were seeing on a lot of pages (eg. the options page).

Issue #1404, opened 2014-12-31
No link-hints for click handlers, registered with jQuery.live


max-baz:
Consider the following example:

``` html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>Link-hints for jquery.live</title>
    <style>.spoiler_title { text-decoration: underline; cursor: pointer; }</style>
  </head>
  <body>
    <div class="html_format">
      <div class="spoiler">
        <b class="spoiler_title">Spoilers!</b>
        <div class="spoiler_text" style="display: none;">
          <span>Hidden text is here!</span>
        </div>
      </div>
    </div>

    <script src="https://code.jquery.com/jquery-1.8.3.min.js"></script>
    <script>
      $(".html_format .spoiler_title").live("click", function() {
        var spoiler = $(this).closest(".spoiler");
        $("> .spoiler_text", spoiler).slideToggle();
        spoiler.toggleClass("spoiler_open");
      });
    </script>
  </body>
</html>

```

Press `f` to show link-hints.
**Expected**: There is a link-hint to expand/collapse spoiler.
**Actual**: There is none.

Tested on both `master` and `post-1.46`.

The following code was extracted from the popular Russian-speaking collaborative blog [habrahabr.ru](http://habrahabr.ru), so the impacted number of people for this particular example is rather high.

z0rch:
Hey @mrmr1993, I'd like to hear your suggestions on the approach to choose.

First of all, keep in mind that the syntax `$(document).on("click", selector, handlerFn)` is used to setup a handler for elements, that do not necessarily exist at the time of setting up the handler.

This means that we need some way to remember the selector, and find matching elements by ourselves when a user hits `f`, just like jQuery itself would do.

Now, in order to hook the `.on()` method, we need to place the hook **after jQuery is loaded**, but **before the first use** of `.on()` method.
In general case, these two things may exist in one js file (minimized and concatenated), so I'm not even sure how to intercept our hook in-between.

To give you a better insight, here is what roughly happens inside `jQuery.on()`:

* Link between `selector` and `handlerFn` is saved in private variable, let's name it `privateHandlers`
* A general event listener for `click` is assigned to `document`, with no information about `selector` or `handlerFn`
* When a `click` event occurs anywhere on the page and bubbles up to document, jQuery uses `privateHandlers` to call `handlerFn` **only if** the chain of clicked events contains an element, matching the `selector`.

[...]

PR #1409, opened 2015-01-03
[post-1.46] Enable link hints for all elements with click listeners + support jQuery delegated events, and drop some legacy link hint conditions

maximbaz:
This PR is based on @mrmr1993's #1405, with added fix for #1404.

Currently `addEventListener()` and `jQuery.on()` are hooked to provide missing link hints.
`.live()` is based on `.on()` since **jQuery v1.7** (which was released in Nov. 2011), so it is hooked as well.
I decided not to spend time on supporting older versions.

Ideally, it would be nice to hook `removeEventListener()` and `jQuery.off()` as well -- I'll look into this later.

* [x] ~~jQuery.off()~~
* [x] ~~removeEventListener()~~

**UPDATE (done)**: ~~There is a bug with number of link hints appearing when listeners for both "normal" and "delegated" events are assigned on the same element. Corresponding test is currently disabled.~~

~~I'll look more into that, but you are more than welcome to provide an early feedback for this PR.~~

**UPDATE 2**: Agreed not to touch `removeEventListener()` and `jQuery.off()` in this PR since we cannot properly handle all cases.

----------------------------------

#1408 is also fixed as a part of this PR, as I used that functionality to create new tests. It appeared to be simply a missing setup data.


content_scripts/link_hints.coffee
```diff
@@ -125,6 +125,22 @@ LinkHints =
...
 
   #
+  # Find children elements that have delegated onclick event listener assigned,
+  # and mark them with `vimium-has-delegated-onclick-listener` attribute.
+  # This will make sure those elements are discovered during `getVisibleClickableElements` call.
+  #
+  markChildrenThatHaveDelegatedOnClickListener: (element) ->
+    return unless element.hasAttribute "vimium-jquery-delegated-events-selectors"
+
+    selectorsStr = element.getAttribute "vimium-jquery-delegated-events-selectors"
```
mrmr1993: This could be rewritten to avoid 2 DOM calls, remove some redundancy:
```
selectorsStr = element.getAttribute "vimium-jquery-delegated-events-selectors"
return unless selectorsStr?
```

content_scripts/link_hints.coffee
```diff
@@ -125,6 +125,22 @@ LinkHints =
...
+    selectors = selectorsStr.split("|").filter (x) -> !!x
+
+    for selector in selectors
+      for child in element.querySelectorAll(selector)
+        unless child.hasAttribute "vimium-has-delegated-onclick-listener"
+          child.setAttribute "vimium-has-delegated-onclick-listener", ""
```
mrmr1993: You can drop the `unless` condition from here.

content_scripts/link_hints.coffee
```diff
@@ -155,10 +171,14 @@ LinkHints =
...
         element.getAttribute("role")?.toLowerCase() in ["button", "link"] or
-        element.getAttribute("class")?.toLowerCase().indexOf("button") >= 0 or
-        element.getAttribute("contentEditable")?.toLowerCase() in ["", "contentEditable", "true"])
+        element.getAttribute("contentEditable")?.toLowerCase() in ["", "contentEditable", "true"] or
+        element.hasAttribute("vimium-has-onclick-listener") or
+        element.hasAttribute("vimium-has-delegated-onclick-listener"))
       isClickable = true
 
+    # Dispose the attribute so next time it is not included, in case listener has been removed
+    element.removeAttribute "vimium-has-delegated-onclick-listener"
```
mrmr1993: If you put `element.hasAttribute("vimium-has-delegated-onclick-listener")` into an `else if` condition, then we no longer have to call this on every element. IE:
```
        ...
        element.hasAttribute("vimium-has-onclick-listener"))
      isClickable = true
    else if element.hasAttribute("vimium-has-delegated-onclick-listener")
      isClickable = true
      element.removeAttribute "vimium-has-delegated-onclick-listener"
```

content_scripts/link_hints.coffee
```diff
@@ -125,6 +125,22 @@ LinkHints =
     marker
 
   #
+  # Find children elements that have delegated onclick event listener assigned,
+  # and mark them with `vimium-has-delegated-onclick-listener` attribute.
+  # This will make sure those elements are discovered during `getVisibleClickableElements` call.
+  #
+  markChildrenThatHaveDelegatedOnClickListener: (element) ->
```
mrmr1993: Can we change this to `markElementsThatHaveDelegatedOnClickListener` and iterate over `document.querySelectorAll("*[vimium-has-delegated-onclick-listener")` instead?

Making DOM modifications in `getVisibleClickable` means that we always have to iterate over the DOM from start to finish, even if we just want to query a subset of the DOM, or query it in a different order.

mrmr1993:
> Ideally, it would be nice to hook `removeEventListener()` and `jQuery.off()` as well -- I'll look into this later.

I'd purposely left this out since it bloats the PR, and doesn't add much value. I can't think of many situations where an element will remain in the DOM and visible but have its click event listener removed. Maybe work on this in a separate branch, so we can consider the merits of the `removeEventListener` and `jQuery.off` hooks separately.

I've left a few comments on the code for you to take a look over.

z0rch:
Thanks for the comments, they have been addressed. 

In regards of `.off()` and `removeEventListener`... well maybe you are right, not a big win. But omitting them just feels not good, it's like we are not cleaning up after ourselves. 

And that's the reason I also don't want to create a separate PR for it -- if you look at those commits separately, you will not be convinced that it is an important change to merge. While here... well, it is sort of a part of the issue.

And the code amount is small, comparing to other changes: [10 lines of `.off()`](https://github.com/z0rch/vimium/commit/ced15c8a44da9a17fcc7e44b03d11b9f1453ac78), guarded by [2 little tests](https://github.com/z0rch/vimium/commit/24c588b94a09f9a6a4df776a5bf06538b74433c2), and [`removeEventListener()`](https://github.com/z0rch/vimium/commit/712df45c798e175320fb12f84e2e8aaf576aec9e) is even smaller.

But if you believe this brings too much risk, I'm fine to delete these 3 commits.

content_scripts/link_hints.coffee
```diff
@@ -125,6 +125,22 @@ LinkHints =
...
+  #
+  markElementsThatHaveDelegatedOnClickListener: ->
+    for element in document.querySelectorAll("*[vimium-jquery-delegated-events-selectors]")
+      selectorsStr = element.getAttribute "vimium-jquery-delegated-events-selectors"
+      continue unless selectorsStr
+
+      selectors = selectorsStr.split("|").filter (x) -> !!x
+
+      for selector in selectors
+        for child in element.querySelectorAll(selector)
```
z0rch: I have been running this for a while, it is wrong to use `querySelectorAll` here. Selectors were extracted from jQuery's `.on()` function, and jQuery supports more kinds of selectors than `querySelectorAll` (it uses [Sizzle](https://github.com/jquery/sizzle/wiki) in some cases). 

An example I found on trello.com is the following selector: `a[href=#]`.
jQuery returns some links, while `querySelectorAll` throws an exception.

Now, it sounds safe to use jQuery defined on a page &mdash; it is definitely defined as we reach this function, we don't bother if jQuery treats same selectors differently in different versions, etc.

But it seems like we cannot access jQuery, defined on the page, from the content script... Is that right, @mrmr1993?

If that's true, what would be the best way to handle this situation? 
I can easily think of the following approaches:

* Catch `querySelectorAll`'s exceptions and simply ignore those selectors.
* Introduce jQuery (or Sizzle) as a part of vimium distribution.

But I like neither of them :smile:

mrmr1993: > it seems like we cannot access jQuery, defined on the page, from the content script... Is that right, @mrmr1993?

Unfortunately it is. I would send a custom event from the content script and mark the elements we want when we catch it in the injected code. Obviously then we can resume code execution in link hints with another custom event.

Since a lot of the time we don't have to do this, we should also only send the event when we have  a `querySelector` match on the delegated event selector.

mrmr1993:
> if you believe this brings too much risk, I'm fine to delete these 3 commits.

The code here unmarks elements which have
```
$(element).on("click", "selector", func1);
$(element).on("click", "selector", func2);
// later
$(element).off("click", "selector", func2);
```

(or similar) which still have a click listener. I'd rather err on the side of overreporting clickable elements, so maybe separating this out until the implementation is right would be better. For this, querying `jQuery._data(element, "events")` after each update means that we don't have to track event listeners ourselves.

mrmr1993:
Ping @z0rch, can you take a look at my comments here?

z0rch:
Sorry for the silence, I agree with both comments, just didn't have time to look into that last week. However I do have time this week :smile:

Issue #1412, opened 2015-01-06
Find into a text area is broken in many cases


smblott-github:
Reproduce:
- Go [here](http://smblott.org/single-textarea.html).
- `/abc<Enter>`
- `gi`
- `o` -- Oops, you're still in normal mode!

Same thing with various other combinations (also involving `n`).

Edit. BTW: This is (partially) fixed in my follow-on branch to #1406, [here](https://github.com/smblott-github/vimium/tree/modes).

smblott-github:
Weirder...
- Go to the same page.
- `/ooo<Enter>`
- `n` a few times until the selection is *outside of* the text area.
- With the mouse, click in the text area.
- Oops, again we're not in insert mode!

smblott-github:
In `master`, after hitting `n` a couple of times, we have...

![snapshot](https://cloud.githubusercontent.com/assets/2641335/5626079/354e23dc-957e-11e4-960c-fc7a3120b07e.png)

With 929fd37aa67cd54791306d16493c3e6e2a15ee01, we have...

![snapshot](https://cloud.githubusercontent.com/assets/2641335/5626112/83773d00-957e-11e4-906a-d326e97c725a.png)

So, we don't leave behind an inappropriate focus.  This fixes the part of this issue, including the "wierder" case.  But not all cases.  That's trickier.

929fd37aa67cd54791306d16493c3e6e2a15ee01 is not merged and not (yet) in a PR.  If there are no objections, I may just merge it directly.

This is completely fixed in #1413.

smblott-github:
Another way to get into a broken insert-mode/normal-mode state is:
- find into a text area
- `f`
- pick the text area you're already in.

smblott-github:
And another one:
- find into a text area
- `f`
- `Esc`

Issue #1410, opened 2015-01-03
Cannot search/find through content-editable elements


smblott-github:
Reproduce:
- Go [here](http://demo.hongkiat.com/html5-editable-content/index2.html).
- Search for "muffin": `/muffin<Enter>`.
- Type `n` to find the next match.

Oops, you've been dropped into insert mode.

This stems from this [change](https://github.com/philc/vimium/commit/512aa9919ac19335ef86048fb41ffa051b60dd3a), from this [PR](https://github.com/philc/vimium/pull/1247) addressing this [issue](https://github.com/philc/vimium/issues/1245).

philc:
Nice find!

mrmr1993:
This is fixed in #1273

smblott-github:
> This is fixed in #1273.

I tried #1273.  Doesn't fix the problem for me, unfortunately.

mrmr1993:
> ... when using `n`.

Apologies, I missed that.

I designed #1250 with this use case specifically in mind, can you take a look at that?

smblott-github:
BTW: This is fixed in my follow-on branch to #1407, [here](https://github.com/smblott-github/vimium/tree/modes).

Issue #1183, opened 2014-10-24
Bindable command to activate the vimium page popup


smblott-github:
We should have a bindable command to activate the vimium page popup (is that possible?  If not, see below).  Moreover, if there is an existing rule for this page, then we might consider immediately focusing the passkeys box.  I find my most common use of the popup is to add or remove passkeys.

Separately, chrome has a mechanism for extensions to define global key bindings.  We could also have such a binding for the page popup.

PR #1413, opened 2015-01-06
A modal-browsing framework

smblott-github:
Following on from #1406,  this is a more fleshed-out reworking of Vimium's keyboard handling into a system of (more or less orthogonal) modes.  The goal is to create a framework, based on which we can implement visual mode (and others).

What this does:
- Handles modes as a stack of `Mode` handlers on the `handlerStack`.
- Implements modes: `normal`, `passkeys`, `insert`, `find`, `focus-input` and a `visual` mode template (plus a couple of utility modes).
- Uses badges to provide visual feedback as to the current mode.  (This works surprisingly well.  I miss it already when I move back to `master`.) 
- Fixes a number of small bugs along the way (#1410, #1412 and a couple of others).

How does it work:
- `normal` and `passkeys` modes are always an the stack.
- A trigger mode for `insert` mode is also always on the stack.  If the focus drops into an editable element, this pushes `insert` mode onto the stack).
- Other modes (`insert`, `find`, etc) are pushed and popped as needed.
- In addition to `keydown`, `keypress` and `keyup`, `focus` and `blur` events are also bubbled down the mode stack.  This means we can handle modes/events in a single, unified way.  And there's a stack logic as to how they are distributed to active modes.
- A number of pseudo events are also bubbled down the handler stack.  Also, some `extra` information is attached to each event bubbling down the handler stack.  This allows active modes to communicate (by attaching "extra" information to events).

What does this facilitate:
- The immediate priority is visual mode.
  - So, a new command `v`.
- Thereafter, we could consider a vim-like editing mode (which would be super convenient for longer text areas such as this one).
  - So, a new command `gv`, like `gi`, but drops you into a vim-like editing mode.
  - Within that mode, `v` drops you into visual mode, and `Ctrl-O` drops you (briefly) into a new `normal` mode.

Where to start:
- Start by looking in `mode.coffee`, [here](https://github.com/smblott-github/vimium/blob/modes/content_scripts/mode.coffee).
- Edit: No! Start by running it, and observe how surprisingly cool the badge looks.
- Then perhaps take a look at `passkeys` mode [here](https://github.com/smblott-github/vimium/blob/modes/content_scripts/mode_passkeys.coffee), then perhaps `insert` mode [here](https://github.com/smblott-github/vimium/blob/modes/content_scripts/mode_insert.coffee). 
- Then.... well, the diff?  Post here if you need more information.

My experience is that modes are *considerably* easier to think about and work with.

The hideous, flat, tangle of logic we currently have in `vimium_frontend.coffee` remains, but much of that code can be culled.

This is not quite ready for prime time, but it's pretty close.  I'm still considering whether to flatten out the mode class hierarchy. The OO style can make it difficult to track down bugs.  So, while the hierarchy may be neat, it may not be helpful long term.


smblott-github:
Fixes #1414.
Fixes #1412.
Fixes #1410.

I'm sure these can be fixed in `master`.  However they're tricky, particularly #1410.  You end up with logic like:

```coffeescript
if !insertMode and !findMode and
    justBeenInFindMode and wantToBeInInsertMode and
    ...
```

Everything is a special case, so it's tricky to reason about.

background_scripts/main.coffee
```diff
@@ -339,6 +339,17 @@ updateOpenTabs = (tab) ->
 setBrowserActionIcon = (tabId,path) ->
   chrome.browserAction.setIcon({ tabId: tabId, path: path })
 
+# This color should match the blue of the Vimium browser popup (although it looks a little darker, to me?).
+chrome.browserAction.setBadgeBackgroundColor {color: [102, 176, 226, 255]}
```
philc: Elsewhere we have spaces around braces and the keys inside of them. Let's use spaces for consistency.

content_scripts/mode.coffee
```diff
@@ -0,0 +1,210 @@
+# Modes.
+#
+# A mode implements a number of keyboard event handlers which are pushed onto the handler stack when the mode
+# starts, and poped when the mode exits.  The Mode base class takes as single argument options which can
```
philc: Popped, not poped

content_scripts/mode.coffee
```diff
@@ -0,0 +1,210 @@
...
+# define:
+#
+# name:
+#   A name for this mode.
+#
+# badge:
+#   A badge (to appear on the browser popup) for this mode.
+#   Optional.  Define a badge is the badge is constant.  Otherwise, do not define a badge and override the
+#   chooseBadge method instead.  Or, if the mode *never* shows a badge, then do neither.
```
philc: There's a typo which makes this hard to parse. Providing example of which modes use which combo of behaviors will make this design clear.

content_scripts/mode.coffee
```diff
@@ -0,0 +1,210 @@
...
+# keyup:
+#   Key handlers.  Optional: provide these as required.  The default is to continue bubbling all key events.
+#
+# Additional handlers associated with the mode can be added by using the push method.  For example, if a mode
+# responds to "focus" events, then push an additional handler:
+#   @push
+#     "focus": (event) => ....
+# Any such additional handlers are removed when the mode exits.
+#
+# New mode types are created by inheriting from Mode or one of its sub-classes.  Some generic cub-classes are
```
philc: Typo

content_scripts/mode.coffee
```diff
@@ -0,0 +1,210 @@
...
+  # Default values.
+  name: ""
+  badge: ""
+  keydown: null # null will be ignored by handlerStack (so it's a safe default).
+  keypress: null
+  keyup: null
+
+  constructor: (options={}) ->
+    Mode.modes.unshift @
+    extend @, options
```
philc: This is a strange and very loose contracted. I think it will be much more understandable if we keep options around as its own object, and delegate to it when needed (e.g. keyup invokes option.keyup), or we explicitly assign the properties we expect options to have to `this` (e.g. @keyup = options.keyup).

content_scripts/mode.coffee
```diff
@@ -0,0 +1,210 @@
...
+# To install and existing mode, use:
+#   myMode = new MyMode()
+#
+# To remove a mode, use:
+#   myMode.exit() # externally triggered.
+#   @exit()       # internally triggered (more common).
+#
+
+# For debug only; to be stripped out.
+count = 0
```
philc: Not sure if this applies to @count, but if you find some facilities especially helpful for debugging as you develop this, it would be nice to clean them up, leave them in and document their usage for future developers.

content_scripts/mode.coffee
```diff
@@ -0,0 +1,210 @@
...
+      Mode.updateBadge()
+      @modeIsActive = false
+
+  # The badge is chosen by bubbling an "updateBadge" event down the handler stack allowing each mode the
+  # opportunity to choose a badge.  chooseBadge, here, is the default. It is overridden in sub-classes.
+  chooseBadge: (badge) ->
+    badge.badge ||= @badge
+
+  # Shorthand for a long name.
+  alwaysContinueBubbling: (func) -> handlerStack.alwaysContinueBubbling func
```
philc: Can you explain the use case for "always continuing to bubble" so the purpose of this mechanism is clear? Either here or in the general comments at the top of mode.coffee

content_scripts/mode.coffee
```diff
@@ -0,0 +1,210 @@
...
+  # has the focus.
+  @updateBadge: ->
+    if document.hasFocus()
+      handlerStack.bubbleEvent "updateBadge", badge = {badge: ""}
+      chrome.runtime.sendMessage
+        handler: "setBadge"
+        badge: badge.badge
+
+  # Temporarily install a mode to call a function.
+  @runIn: (mode, func) ->
```
philc: What is this used for?

content_scripts/mode.coffee
```diff
@@ -0,0 +1,210 @@
...
+      SingletonMode.kill @singleton
+      SingletonMode.instances[@singleton] = @
+    super options
+
+  # Static method. Return whether the indicated mode (singleton) is currently active or not.
+  @isActive: (singleton) ->
+    @instances[singleton]?
+
+  # Static method. If there's a singleton instance active, then kill it.
+  @kill: (singleton) ->
```
philc: Why not `exit` or `deactivate`?

content_scripts/mode.coffee
```diff
@@ -0,0 +1,210 @@
...
+      "registerStateChange": ({enabled: enabled, passKeys: passKeys}) =>
+        @alwaysContinueBubbling =>
+          if enabled != @enabled or passKeys != @passKeys
+            @enabled = enabled
+            @passKeys = passKeys
+            @registerStateChange()
+
+  # Overridden by sub-classes.
+  registerStateChange: ->
+
```
philc: typo, pseudo

content_scripts/mode_find.coffee
```diff
@@ -0,0 +1,44 @@
+# NOTE(smblott).  Ultimately, all of the FindMode-related code should be moved to this file.
+
+# When we use find mode, the selection/focus can end up in a focusable/editable element.  In this situation,
+# PostFindMode handles two special cases:
+#   1. Be an InsertModeBlocker.  This prevents keyboard events from dropping us unintentionaly into insert
+#   mode.  Here, this is achieved by inheriting from InsertModeBlocker.
+#   2. If the very-next keystroke is Escape, then drop immediately into insert mode.
+#
+class PostFindMode extends InsertModeBlocker
```
philc: Cool. Tangentially related, but could/how would #556 be implemented in this new and wonderful world of modes?

content_scripts/vimium_frontend.coffee
```diff
@@ -110,8 +123,19 @@ initializePreDomReady = ->
   settings.addEventListener("load", LinkHints.init.bind(LinkHints))
   settings.load()
 
+  # Install normal mode. This is at the bottom of both the mode stack and the handler stack, and is never
+  # deactivated.
+  new NormalMode()
```
philc: Seeing this code here, it's kind of strange that instantiating a NormalMode also installs it. Seems to me that handler stack should be an array, and the code which is adding a handler to the handler stack actually does this, e.g.:

@handlerStack.push(new NormalMode())

content_scripts/vimium_frontend.coffee
```diff
@@ -342,28 +376,37 @@ extend window,
...
+        DomUtils.removeElement hintContainingDiv
+        visibleInputs[selectedInputIndex].element.focus()
+        new InsertMode visibleInputs[selectedInputIndex].element
+        super()
 
 # Decide whether this keyChar should be passed to the underlying page.
 # Keystrokes are *never* considered passKeys if the keyQueue is not empty.  So, for example, if 't' is a
 # passKey, then 'gt' and '99t' will neverthless be handled by vimium.
 isPassKey = ( keyChar ) ->
+  return false # Diabled.
```
philc: Perhaps add a TODO so we remember to delete this code.

lib/handler_stack.coffee
```diff
@@ -1,37 +1,65 @@
...
-        break
+    # This is more expense than splicing @stack, but better because splicing can interfere with concurrent
+    # bubbleEvents.
+    @stack = @stack.filter (handler) ->
+      # Mark this handler as removed (to notify any concurrent bubbleEvent call).
+      if handler.id == id then handler.id = null
+      handler?.id?
+
+  # The handler stack handles chrome events (which may need to be suppressed) and internal (fake) events.
+  # This checks whether that the event at hand is a chrome event.
```
philc: Typo

philc:
Incredibly well documented as usual =) Reads like a novel.

This is ambitious. There's a lot that I like: it introduces a clean conceptual framework for key handling by different modes, it clearly separates the implementation for each mode from the frontend, and I assume the understandability and debuggability wins open up the implementation of new, more sophisticated modes (visual, edit). This is a really powerful refactoring of the core, and I can tell the decisions made were based on hard-won lessons working with Vimium's existing handler code. Nice work.

In general it LGTM; with some more cleanup we should merge it. Some considerations:

* Can we migrate the linkhints mode to this framework?
* Can we get rid of the use of inheritance, favoring composition instead? I don't know what the verbosity tradeoff is, but I'm pretty sure it will be a clarity and debuggability win.
* When we implement a visual mode, let's also keep in mind a visual line mode.

Feedback on the mode indicator UX:

1. In Vim/Emacs, I'm used to looking at the cursor to determine what mode I'm in. It would be really, really nice if this UI was somehow possible in Vimium, where appropriate (i.e. when the user drops into edit or visual mode).

2. If we can't use the cursor effectively in enough situations to make it a practical mode indicator, then as a second option, I'd explore using some kind of modeline indicator, as we do today for insert mode. I'm trained by Vim to look at the bottom of the screen for the current mode, and not in the upper right at Chrome's extension icon.

3. If we ultimately decide to use Vimium's toolbar icon as the primary indicator of mode, I think we should make the entire icon dedicated to displaying the mode, rather than just as a badge. The badge is very small on high res screens. I may be blind but I can hardly discern the difference between N and V.

@mrmr1993, great feedback on this as it's evolved. Reading your discussion on #1406 was very helpful.

smblott-github:
Thanks for your input, @philc.  I'll rework this over in my [dev branch](https://github.com/smblott-github/vimium/tree/modes-dev), and merge that back here once it's in a presentable state.

content_scripts/mode_insert.coffee
```diff
@@ -0,0 +1,96 @@
+
+# Input or text elements are considered focusable and able to receieve their own keyboard events, and will
+# enter insert mode if focused. Also note that the "contentEditable" attribute can be set on any element
+# which makes it a rich text editor, like the notes on jjot.com.
+isEditable =(element) ->
```
mrmr1993: This function make more sense in `DomUtils`.

content_scripts/mode_insert.coffee
```diff
@@ -0,0 +1,96 @@
...
+  return true if element.isContentEditable
+  nodeName = element.nodeName?.toLowerCase()
+  # Use a blacklist instead of a whitelist because new form controls are still being implemented for html5.
+  if nodeName == "input" and element.type not in ["radio", "checkbox"]
+    return true
+  nodeName in ["textarea", "select"]
+
+# Embedded elements like Flash and quicktime players can obtain focus.
+isEmbed =(element) ->
```
mrmr1993: Ditto.

content_scripts/mode_insert.coffee
```diff
@@ -0,0 +1,96 @@
...
+  element.nodeName?.toLowerCase() in ["embed", "object"]
+
+isFocusable =(element) ->
```
mrmr1993: Ditto.

content_scripts/mode_find.coffee
mrmr1993: @philc The only way to implement #556 without it clearing the selection when we focus the input is to have it in an iframe. I've done this in #1401 (although it's actually a `contentEditable` element there, so it resizes with its contents).

smblott-github: > how would #556 be implemented

We could implement it, and quite easily.  But #1401 is probably the way to go, as @mrmr1993 says.

content_scripts/vimium_frontend.coffee
smblott-github: I don't agree with this one.  You activate a mode by creating a new instance of the relevant class.  The mode itself knows how to install itself.  And modes deactivate themselves when they're done.  Their logic is self contained.  They're self managing after launch.  This is the right way to handle things.  Code elsewhere does not need to worry about what mode we're in.

content_scripts/vimium_frontend.coffee
smblott-github: > Perhaps add a TODO so we remember to delete this code.

Actually, there's a quite a lot of unused code now in the front end.  Most of it's just sitting there.  In a couple of places (like this), I disabled things to *make sure* they're not being used.

I'll come back and strip out the unused code later, once we get this and #1401 sorted.

smblott-github:
Thanks for the input, @philc, @mrmr1993.  If I haven't responded to a specific comment, then it means the appropriate change has been made.  The biggest changes are:

- Collapse (almost completely) the class hierarchy.  It doesn't help readability or maintainability.
- Lots of simplifications.

> Feedback on the mode indicator UX:

Here's the case for using the browser button (either via badges, or via different icons):
- It feels more integrated with chrome.
- It doesn't take up extra screen real estate.
- We look up there anyway to see if we have passkeys active.
- Unlike the HUD, it doesn't appear in odd places (with frames).
- If we had been using the browser button in this way all along, then I don't imagine anyone would be proposing the HUD as an improvement.
- The HUD takes up a lot of space for about two or three bits of information.
- If users accept Microsoft  completely changing the look-n-feel of their OS on every release, then changing Vimium's mode indicator cannot be too unreasonable.  In fact, UI change is one way of drawing attention to new features.

I propose we stick with badges for a trial period.

Next steps:
- Perhaps another code review of this PR.
- Get #1401 into shape and merge it.  It has some great usability improvements for find mode.  We do that first because #1401 and this have overlapping code footprints.
- Merge this.  I've been using this live, and I'm not seeing issues.  I'm pretty sure there aren't any show stoppers.  However, for sure there will be corner cases which need to be ironed out.  And we won't find those unless we give the code a bit of a shake out.
- Work on developing #1418.
- Separately, I'm going to take a look at handling link hints within this framework.

(The Travis tests are failing.  Looks like #1419.  They pass for me, locally.)

mrmr1993:
I'm still very uncomfortable with what this means for debugging; having lots of separate event listeners for each mode makes it hard to work or where the code is going, or to keep track of the mode hierarchy.

A unified event listener for `keyup`/`keydown`/`keypress` that explicitly checks each mode in turn, runs the appropriate listener if the mode is active, and stops propagation as appropriate would seem like a big improvement over having to check every mode to check for its relative behaviour.

> * Collapse (almost completely) the class hierarchy. It doesn't help readability or maintainability.

With a centralised handler, we could put the class hierarchy back, since they're a really powerful way to express mode transitions. The main problem (at least for me) with all these classes is of key event listener flow, which a centralised handler does away with.

smblott-github:
> having lots of separate event listeners for each mode makes it hard to work or where the code is going

Having all events going through the handler stack is also a big advantage.  You can track them there.  You know exactly what's being called, and in what order.

Here's what happens when I press `j` without a passkey...

    8 "keydown" "insert-trigger" "continue"
    8 "keydown" "passkeys" "continue"
    8 "keydown" "anon" "continue"
    8 "keydown" "normal" "continue"
    9 "keypress" "passkeys" "continue"
    9 "keypress" "normal" "continue"
    10 "keyup" "anon" "continue"
    10 "keyup" "normal" "continue"

And with a passkey...

    9 "keydown" "insert-trigger" "continue"
    9 "keydown" "passkeys" "stop-true"
    10 "keypress" "passkeys" "stop-true"
    11 "keyup" "anon" "continue"
    11 "keyup" "normal" "continue"

(Oh, look.  The `keyup` handling could be tweeked. It's still relying on `HandledKeyDown` from `normal` mode.)

Here's a more complicated example, pressing `n` and landing in a text area...

    9 "keydown" "insert-trigger" "continue"
    9 "keydown" "passkeys" "continue"
    9 "keydown" "anon" "continue"
    9 "keydown" "normal" "continue"
    10 "keypress" "passkeys" "continue"
    10 "keypress" "normal" "continue"
    5 "create:" "insert-blocker"
    11 "focus" "insert-trigger-extra" "continue"
    12 "updateBadge" "insert-blocker" "continue"
    12 "updateBadge" "insert-trigger" "continue"
    12 "updateBadge" "passkeys" "continue"
    12 "updateBadge" "normal" "continue"
    12 "updateBadge" "badge" "continue"
    11 "focus" "badge-extra" "continue"
    5 "exit:" "insert-blocker"
    13 "updateBadge" "insert-trigger" "continue"
    13 "updateBadge" "passkeys" "continue"
    13 "updateBadge" "normal" "continue"
    13 "updateBadge" "badge" "continue"
    6 "create:" "post-find"
    14 "keyup" "anon" "continue"
    14 "keyup" "normal" "continue"

That's a pretty detailed trace of what's going on.  And it looks correct.  (Although, it might be an idea to take a look at how much badge updating is going on.)

mrmr1993:
When modes have been set in the right order, then yes, this is fine. What I'm concerned about it when the modes have been set in the incorrect order.

To have to poke `HandlerStack` at runtime to see if which of eg. find mode or insert mode gets priority for keypresses is a massive loss in our ability to reason about modes when coding. I was never a fan of `HandlerStack` anyway, since it obfuscates the order in which potentially order-sensitive code is run, causing exactly this problem.

smblott-github:
> What I'm concerned about it when the modes have been set in the incorrect order.

That would be revealed in the trace above.

mrmr1993:
>That would be revealed in the trace above.

At runtime, yes. This is useless when coding if eg. the code isn't in a compilable state yet.

Instead of maintaining an n*n map of which modes can go above/below others in the stack *in my head* &mdash; and updating it every time the interaction between modes changes &mdash; I would like (/need, at this stage) the code to explicitly tell me which modes can come in which order. The barrier to entry of mentally creating this map for new coders is also a consideration.

smblott-github:
> Instead of maintaining an n*n map

There's no new n*n problem with this PR.  Here, there are n modes.  In `master`, there are n state variables (`findMode`, and the like) and things that have been pushed onto `handlerStack`.  You have to reason about which can be set/active at the same time.  It's the same thing.

In fact, I'd argue that it's worse in `master`.

Try this...

Go [here](http://jsfiddle.net/smblott/9u7geasd/).  In the result window:
- Type `/Fish` (do not press enter).
- Click in one of the text areas.
- Press `Esc`.
- Type `aaa` - you're in insert mode.
- Type `jk` - hmm, where did they go?
- Type `o` - oops, you're also in normal mode.

This is an example of one of the unexpected n*n combinations that can exist in `master`.  This PR just makes it easier to reason about such cases.

mrmr1993:
> There's no new `n*n` problem with this PR. Here, there are n modes.

The `n*n` is, "if mode x and mode y are active, which mode should handle the key"? This was explicit before. Now we have to compare the interplay of all possible active modes to work it out.

> In fact, I'd argue that it's worse in `master`.

> Try this...

If we think of it as "which mode should handle the key when insert mode and find mode are active", it's clear that there's a longstanding bug that we've hacked around by not setting insert mode when we're in find mode:

```coffee
if isInsertMode()
  # Insert Mode Stuff
else if findMode
  # Find Mode Stuff
```

should be

```coffee
if findMode
  # Find Mode Stuff
else if isInsertMode()
  # Insert Mode Stuff
```

Aside from this bug, what's worse in `master`?

smblott-github:
> Aside from this bug, what's worse in master?

#1414.
#1412.
#1410.
#1415.
#1423.

smblott-github:
More rant on why this is the right way to go.

See 158b369a54c3dc2a3bc9c96cd07649a08207728f.  This demonstrates how modes can be nested, and the benefits of of orthogonality.

It implements a dummy edit mode with one command only: "i".  You can try it out [here](https://github.com/smblott-github/vimium/tree/modes-dev---edit-mode-demo), or just believe the following explanation.

To try this out:
- Go to a page with a text area.
- Type `gv` for "go vim". Unfortunately, `ge` is already used.
- It uses the same input selector as `focusInput`.  So, if there's only one input, you'll be dropped straight into it, if there're more than, you get to pick one.
- You're now in edit mode (note the "E" in the badge).
- The keys all do nothing, except...
- "i" takes you into insert mode; from which `Esc` takes you back to edit mode, and so on...
- Use `Esc` from edit mode to get back to normal mode.

What does this show?

- It shows how modes can be stacked.  We're moving back and forwards between modes on a stack.
- It shows how modes can be re-used.
  - The insert mode you drop into with "i" is *exactly the same* insert mode used elsewhere.
  - Ideally, we would be able to re-use find mode and visual mode within a fully-fleshed out edit mode.
- It shows how code can be re-used.
  - The `focusInput` input selector is *exactly the same* one used for `gi`.
- It shows how modes, once installed, don't interfere with each other.  So we don't have to deal with endless *if this and not that but perhaps something else* logic.
- It shows how the framework of modes, as proposed here, is a flexible way in which we can push Vimium into new areas (initially visual mode, but soon other modes too).

(I also had to tweek the handler stack to get the first `keydown` event to be handled correctly.  `focusInput` and the handler stack now play nicely.)

mrmr1993:
> See 158b369.

Doesn't this make more sense as a sub-mode? This seems to handle de-focusing the current input multiple times in multiple places, instead of making the mode go away because its parent (insert mode on the current element) has been deactivated. `InsertModeBlocker` makes work and adds confusion that a good sub-mode implementation would get rid of entirely.

> * It shows how modes can be stacked.

Can this mode be active while insert mode isn't? What assurance of this (other than checking the whole of our frontend code to ensure this can't happen) do we get?

It should clearly never handle keys when input mode isn't active, but since we don't explicitly specify this, it could happen through a mistake somewhere in the code. *This* is the point I've been trying to make.

> * ... So we don't have to deal with endless *if this and not that but perhaps something else* logic.

With sub-modes, we only need to check the parent mode, and then the parent mode delegates to its children. The "*if this and not that but perhaps something else*" logic then becomes
```coffee
if findMode
  Mode.modes.find.keydown event
else if isInsertMode()
  Mode.modes.insert.keydown event
else if # etcetera...
```
or perhaps, if we want to fall through between modes
```coffee
passThrough = true
if passThrough && findMode
  passThrough = Mode.modes.find.keydown event
if passThrough && isInsertMode()
  passThrough = Mode.modes.insert.keydown event
if passThrough && # etcetera...
```

And then we're certain that we never have insert mode jump in before find mode, regardless of what crazy things we do (unless we explicitly enable it in the find mode handler).

To end on a positive; it's simple and easy, I really like how easy and clean it makes adding a new mode.

smblott-github:
Added [tests](https://github.com/smblott-github/vimium/blob/modes/tests/dom_tests/dom_tests.coffee#L265).

With this, more stuff is easier to test.  For example, we can [test](https://github.com/smblott-github/vimium/blob/modes/tests/dom_tests/dom_tests.coffee#L607) whereby `Escape` immediately after find drops you into insert mode.

mrmr1993:
The interplay of `InsertMode`, `InsertModeBlocker` and `InsertModeTrigger` here are confusing, and seem to be explicitly nonorthogonal. This means that basically all the modes implemented here *appear* orthogonal, but aren't because they inherit from `InsertModeBlocker`. @smblott-github, can you clarify why this is better than explicitly stating the order in which modes can take priority?

Also, it seems we get into a broken state if the page focuses an input while we're in an `InsertModeBlocker` superclass. And maybe also if the user tabs from link hint mode (I suspect at least this is true in windows), but I'm not at a computer to test this out. Also the same for a native find from any `InsertModeBlocker` superclass. These are all symptoms of pretending that the modes can be order-agnostic.

smblott-github:
You're right, @mrmr1993.  The whole insert-mode handling is unnecessarily complicated.  It can be simplified *and* improved -- which is always a good thing.

mrmr1993:
Been thinking about badges a little bit:
* we're going to have problems with users who've hidden the Vimium icon
  - up until now the icon has basically been clutter, only giving useful feedback when something's gone wrong and the extension isn't running on the page.
  - we're asking users to squash their address bar permanently to get feedback about extension state which is almost always obvious.
  - there's (currently) no fallback option if the user refuses to show the icon.
* we don't benefit from the descriptive power of the HUD, so the user has to stop and work out what the badge *could* mean, slowing them down
  - eg. it's hard to check they're right link hint mode unless they've memorised what the badge codes mean. Especially for incognito, where the user might be confused and think they're in insert mode.
* a mode change is shown on all open windows, which draws your eye away from the current window for no benefit
  - the [`tabId` property](https://developer.chrome.com/extensions/browserAction#property-details-tabId) would probably help with this.
* it's hard to see what mode we're in at a glance, since the badge is quite small. (I think Phil may have already mentioned this.)
* having a persistent badge is irritating/bad UX.

Since badges aren't an integral part of this PR and are a significant UX change, I'm going to suggest removing them from here and revisiting the idea later.

z0rch:
Consider the action of hiding Vimium icon as `:set noshowmode` in vim -- some people might use it to save space, others might not.

**+1** for eventually using the entire icon for displaying the mode. Hopefully we could use the power of colors too (entire icon is green for normal mode, orange for visual mode, blue for insert mode, etc), so that users can see the current mode in the corner of their eyes.

mrmr1993:
> Consider the action of hiding Vimium icon as `:set noshowmode` in vim -- some people might use it to save space, others might not.

I don't think we can detect whether the user has hidden the icon or not.

z0rch:
No no, I mean "mentally consider" :smile: 

If we decide to show mode in Vimium icon, users have an option to show this icon and to hide it (natively supported by Chrome).

If they hide the icon, consequences are clear -- they can't see which mode are they in. Just like if they execute `:set noshowmode` in vim.

So what I'm trying to suggest, we would need to teach users to treat Vimium icon as an indicator of the current mode, but as soon as we achieve this, all the section "we're going to have problems with users who've hidden the Vimium icon" does not apply anymore -- if users hid the icon, they knew the consequences!

mrmr1993:
> No no, I mean "mentally consider" :smile:

Sorry, I misunderstood.

> ... we would need to teach users to treat Vimium icon as an indicator of the current mode ...

The issue is, it's hard to achieve this when users have already hidden the icon, since we need them to explicitly re-enable it to show them the new UI. Otherwise we're just penalising them for hiding the icon back when it held little value.

z0rch:
> The issue is, it's hard to achieve this when users have already hidden the icon, since we need them to explicitly re-enable it to show them the new UI.

That is true, but I'm sure we can come up with some notification, and show it once after upgrade. 

Do you know if it is possible to open the extension popup programmatically (the one we use for passkeys)? We could open similar popup once after the upgrade, and put some notes there. 

In the worst case, we could open a new tab with html page (like most extensions do after the installation), or even embed this info in the HUD.

> Since badges aren't an integral part of this PR and are a significant UX change, I'm going to suggest removing them from here and revisiting the idea later.

As this discussion grows, perhaps it is a good idea to move it out of this PR.

mrmr1993:
> Do you know if it is possible to open the extension popup programmatically (the one we use for passkeys)? We could open similar popup once after the upgrade, and put some notes there.

Unless anything's changed since I last investigated this [here](https://github.com/philc/vimium/issues/1183#issuecomment-60380644), then no, it isn't.

> As this discussion grows, perhaps it is a good idea to move it out of this PR.

Specifically, I think this PR should do what the title indicates, and nothing more. So (from the description):

* ~~Handles `modes` as a stack of Mode handlers on the `handlerStack`.~~
* Implements modes: `normal`, `passkeys`, `insert`, `find`, `focus-input` and a `visual` mode template (plus a couple of utility modes).
* ~~Uses badges to provide visual feedback as to the current mode. (This works surprisingly well. I miss it already when I move back to `master`.)~~
* Fixes a number of small bugs along the way (#1410, #1412 and a couple of others).

Alternatively we could look at using #1425, which is already limited to this scope. (I never wanted to work on migrating our code to mode classes, and only created that as an alternative to here, so if anybody wants to take that off me and run with it I'd be grateful!)

philc:
I think this is a huge win and enables lots of new use cases. Let's merge it. I haven't diff'd the code of this vs #1425, and so I don't have a strong opinion about the debugging experience tradeoffs. I have looked at this one carefully and it is a nice win for the code base.@smblott-github, since you wrote the first implementation, please merge the version you think provides a reasonable foundation for future work, and let's get it into master so we can start building on top of it.

Regarding the badge UX, I've been using it for a week and I think it could use more thought (as discussed in this thread), but it's a not a blocker, and if necessary the discussion of how we treat mode notifications can be hashed out later in a separate thread.

smblott-github:
Thanks for all of the input, @mrmr1993, @philc, et al.  A lively debate!  We can follow up on some of the issues you raised, @mrmr1993, separately.

I intend to merge this very soon:
- The only way to shake out any remaining issues is to use it.
- Ideally, we would merge #1401 first.  However, that's blocked at the moment.  I've done the merge (twice, actually), and there are no significant issues.  I'll hold off stripping now-redundant code from the front end until after that merge.

On badges:
- Propose we wait a couple of weeks, then open an issue to discuss how best to handle the matter.
- I have made a couple of changes in the hope of making badges more useful:
  - Do not show a badge in normal mode.  This makes the presence of a badge more striking, and easier to catch out of the corner of your eye.
  - I guess the most common use case is "am I in insert mode or not?".  With this change, in most cases, that's now just the presence/absence of a badge.
  - The badge is slightly darker, making it stand out a little more, and the character easier to read.
- I have *not* reinstated the `Insert mode` HUD when the user uses `i` to enter insert mode.  Reinstating it would be easy.  But I'd like instead to try an experimental period with badges only.

smblott-github:
Merging.

All of the `Blocker`, `Trigger` stuff is gone.  So that aspect is substantially simplified.

I've tried to maintain the existing behaviour (except where that behaviour was clearly broken -- see various modes-related issues that have been posted recently).

Nevertheless, it's possible (likely?) that there are use cases I've overlooked.  Please post anything you find.

Issue #1415, opened 2015-01-06
Find into text area then non-vimium keys broken


smblott-github:
Use find or `n` to get to a situation that looks like this:

![snapshot](https://cloud.githubusercontent.com/assets/2641335/5628734/34accc88-95a3-11e4-8b8b-1f2c92f52937.png)

(the selection is in a text area, or similar).

Type `j`, `k`, `j`, `k`.  Nothing happens.  That's good, you're in normal mode, as you're supposed to be.

Now type any unmapped key (I used `a`).  Oops, we're actually in insert mode too!  It's not useful to be inserting text into a text area, but only have some keys available.

~~This is hard to fix.~~  The source of the problem is this [trick](https://github.com/philc/vimium/blob/master/content_scripts/vimium_frontend.coffee#L724), which prevents us dropping into insert mode.  However, there's no memory of what happened.

(This is fixed in #1413.)

Edit: On reflection, it's not so hard to fix.  It requires (yet) another state variable and a blur listener.

mrmr1993:
Disputing this issue; if the user presses a key that they don't expect us to handle, we shouldn't handle it.

Also, 'fixing' creates some wonkiness:
* We've given the user a selection that they can only look at &mdash; they can't do *anything* to it with the keyboard until they hit `esc`. Specifically,
  - changing the selection with `shift` and a direction doesn't work,
  - copy/paste with keyboard shortcuts doesn't work, and
  - any actions (which may want to use the selection) that other extensions/the page have bound to keys we don't handle don't work.
* We suppress browser keyboard actions using `f1`-`f12` (and possibly some others).

I vote for this issue to be closed as invalid.

smblott-github:
@mrmr1993. You make reasonable points.  Obviously it's a close call.  Try this argument.

Option 1. Pass unmapped keys through in this situation (the status quo).
Option 2. Block them.

Imagine the status quo was Option 2, and someone proposed changing it to Option 1.  How might such a proposal be received?  I guess we'd think it strange.  Pass through `a` and `s` from the middle row, but nothing else?  And a few select keys from the other rows, but not the others?  That's a pretty weird UX.

In fact, we could even do the experiment.  Head on over to the vim forums, and propose that unmapped keys should be inserted into the buffer...

With Option 1, recovering after a goof is pretty ugly:
- delete the unwanted keys, replace with whatever is in the selection, `Escape`

That's a pretty bad UX too.

smblott-github:
There's also an Option 3:
[...]

PR #1425, opened 2015-01-12
[RFC] Implement modes as classes

### LINKED ISSUE #1410: Cannot search/find through content-editable elements

### LINKED ISSUE #1412: Find into a text area is broken in many cases

### LINKED ISSUE #1414: Find then click on focussed input broken
Use find to get to a situation than looks a bit like this:

snapshot

Clicked on the focused text area.  You're still in normal mode (you should be in insert mode, as you would be if you clicked on any other text area on the page).

mrmr1993:
This PR re-implements modes as classes, in a similar style to #1413, but retaining a centralised `keydown`/`keypress`/`keyup` handler. This achieves:

* Modes as classes, which can be invoked by calling `new ModeName()` (eg. `new InsertMode()`, `new InsertMode element`, `new FindMode()`, etc.)
* By default, modes are registered in `Mode.modes` under their name, and are also accessible with a getter `Mode.getMode("MODE_NAME")`
* Creates sub-modes
  * this lets modes be instantiated in the context of other modes with slight modifications
  * the sub-modes are deactivated with the parent mode, so the mode is only active when it's context is relevant
  * sub-modes can be accessed via the `Mode` object using `Mode.modes.PARENT_MODE_NAME.modes.CHILD_MODE_NAME` or `Mode.getMode("PARENT_MODE_NAME.CHILD_MODE_NAME")` (with each child separated from its parent by a `.`).
* Code flow is (largely) the same as in `master`, but wrapped by classes, so it's easy to reason about which modes take priority over others
* Implements insert mode, normal mode, normal mode (for focused `input`/`contentEditable` elements).
* Follows on from #1424, which makes modes behave properly after performing a search. This also means that this
  - fixes #1414
  - fixes #1412 (and all the related issues in the comments there)
  - fixes #1410
* Moves several dom-related functions to `DomUtils`, to reduce the weight of the frontend.

This facilitates the same things as #1413 does.

This is still a work in progress, with several things left to do:
* [x] Finish migrating all find mode code to `find_mode.coffee`
* [x] Finish migrating all pass keys code to `passkey_mode.coffee`
* [x] Implement a template for visual mode
* [x] Migrate link hints to a mode.
* [x] Comments
* [x] Tweaks to the `Mode` class (a few things still don't feel quite right)
* [ ] Tests
* [ ] Remove legacy frontend code.

Happily, this also requires fewer modes than #1413, looks as if it won't need as much code, and deals with the issues I had over there.

Any thoughts @smblott-github?

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -111,6 +111,7 @@ Commands =
       "goUp",
       "goToRoot",
       "enterInsertMode",
+      "enterVisualMode",
       "focusInput",
       "LinkHints.activateMode",
       "LinkHints.activateModeToOpenInNewTab",
@@ -195,6 +196,7 @@ defaultKeyMappings =
   "gs": "toggleViewSource"
 
   "i": "enterInsertMode"
+  "v": "enterVisualMode"
 
   "H": "goBack"
   "L": "goForward"
@@ -283,6 +285,7 @@ commandDescriptions =
   openCopiedUrlInNewTab: ["Open the clipboard's URL in a new tab", { background: true, repeatLimit: 20 }]
 
   enterInsertMode: ["Enter insert mode", { noRepeat: true }]
+  enterVisualMode: ["Enter visual mode", { noRepeat: true }]
 
   focusInput: ["Focus the first text box on the page. Cycle between them using tab",
     { passCountToFunction: true }]
--- a/content_scripts/find_mode.coffee
+++ b/content_scripts/find_mode.coffee
@@ -1,19 +1,24 @@
+findModeQuery = { rawQuery: "", matchCount: 0 }
+findModeQueryHasResults = false
+findModeAnchorNode = null
+
 class FindMode extends Mode
-  constructor: ->
+  constructor: (query = "") ->
     super "FIND"
+    @update query if query?
   keydown: (event) ->
     if KeyboardUtils.isEscape event
-      handleEscapeForFindMode()
+      @handleEscapeForFindMode()
       DomUtils.suppressEvent event
       KeydownEvents.push event
 
     else if event.keyCode == keyCodes.backspace || event.keyCode == keyCodes.deleteKey
-      handleDeleteForFindMode()
+      @handleDeleteForFindMode()
       DomUtils.suppressEvent event
       KeydownEvents.push event
 
     else if event.keyCode == keyCodes.enter
-      handleEnterForFindMode()
+      @handleEnterForFindMode()
       DomUtils.suppressEvent event
       KeydownEvents.push event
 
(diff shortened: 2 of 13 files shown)


smblott-github:
All of the advantages you propose for this are true also in #1413.  In particular...

> Code flow is (largely) the same as in master, but wrapped by classes, so it's easy to reason about which modes take priority over others

... is true of #1413 too.  If you pursue this, you'll end up where #1413 is now.

> Happily, this also requires fewer modes than #1413

And you'll end up with the same number of modes, since in #1413 each mode has a specific function, and those functions are required regardless of how we implement them.  #1413 is just further developed.

> as if it won't need as much code

And, once you start moving the code around and separating out those specific functions, you'll end up with the same amount of code.

> but retaining a centralised keydown/keypress/keyup handler.

#1413 has a centralised handler too.  It's the handler stack.  The unnecessarily hard-wired logic *if-then-else* logic [here](https://github.com/philc/vimium/pull/1425/files#diff-d0bdd7e52a95f0b05be2cd0f116c20edR399) is equivalently present in the initialisation of the handler stack in #1413.  This is an advantage of #1413.  For example, c5f3abacb27afd2c014a18bede8e6281e272a14a brings link hints into the mode framework.  All of the changes are local to link hints. And all of the state transitions and badge updates just work.

There is no n*n problem.  Because the transitions which are possible here, in #1413 and in `master` are basically the same.

I suggest you take another look at #1413.  Debugging is currently on by default for modes.  Look at the console, you see all of the active modes as they are activated and deactivated.  It's easy to track the state you're in.  If you want more detail, enable the debugging flag in the handler stack.  Then you see how every event is handled.

mrmr1993:
> > so it's easy to reason about which modes take priority over others

> ... is true of #1413 too. If you pursue this, you'll end up where #1413 is now.

I'd argue not. Imagine you have no knowledge of the codebase, and then try and answer each of the following from the code/comments in #1413:

* Can we end up in insert mode while we're in find mode?
* When insert mode and find mode are both active (if the answer to the last question was yes), which receives the keys?

It's fine for us, Phil, and possibly some of the other 78 people who've contributed to this project *because they can remember what the code said before*. However, anyone new would have to read a large chunk of the codebase before they can answer these questions, which is an appalling state of affairs.

> ... you'll end up with the same number of modes ...

> ... you'll end up with the same amount of code.

Because of sub-modes, I've avoided the need for quite some helper classes, such as `InsertModeBlocker`. As such, I expect to have less of a code footprint, but I am totally fine with being wrong if that's the way it turns out.

> The unnecessarily hard-wired *if-then-else* logic here is equivalently present in the initialisation of the handler stack in #1413.

I seem to keep saying this over and over again, but the handlers are order sensitive (since if one says it's handled the key, the others will not get a chance). Given that our behaviour is heavily dependant on our modes executing their code in the right order, it's ludicrous to take the laissez-faire "the modes won't ever be in the wrong order, don't worry" attitude. The "hard-wired *if-then-else* logic" just being explicit about what we would expect from `HandlerStack`, for the benefit of new contributors (/me when I'm tired), and possibly to help save our bacon later.

> There is no n*n problem.

But there is. A new coder has to sit down and work out which modes can go above and below each other mode, and then collapse it to a mental list once they've convinced themselves that the modes can only arise in a fixed order (assuming they're able to do so). This PR just gives you the list, so, old or new to the codebase, you can tell what's going to happen.

> I suggest you take another look at #1413. Debugging is currently on by default for modes.

I've had a look. We replace clarity in the code with a runtime-only overview of what our modes are doing, and generate an unnecessarily large amount of logging data in the process. I don't like it at all. The mode classes are nice though, which is why I made this PR.

> Edit. Better example. See #1415. ...  #1413 implements Option 2 by installing a handler underneath normal mode. ... This is easy to achieve with #1413, because all modes are equal (orthogonality).

So nobody knows that the top-level key handling logic has potentially changed without re-checking **all the mode files *and any file that might invoke them*.** That's nightmareish. Why is this something we'd want?this PR

I note that, on this PR, you can write much the same code to describe the mode, but have to go through all the agony of adding 4 lines (6, including leading line breaks):
```
  else if Mode.isActive "THE_NAME_OF_MY_MODE"
    Mode.getMode("THE_NAME_OF_MY_MODE").keydown event
```
to the bottom of `onKeydown`, and
```
  else if Mode.isActive "THE_NAME_OF_MY_MODE"
    Mode.getMode("THE_NAME_OF_MY_MODE").keypress event
```
to the bottom of `onKeypress`. This now means that anybody who's trying to work out what's happening to key events knows to look at the code for this new mode if something isn't right/they're trying to get their bearings in the codebase.

mrmr1993:
Also, finished migrating find mode code to `find_mode.coffee`.

smblott-github:
> Can we end up in insert mode while we're in find mode?

No.

mrmr1993:
> No.

Which we know.. how?

smblott-github:
> Which we know.. how?

```coffeescript
class FindMode extends InsertModeBlocker
```

mrmr1993:
Can we end up in insert mode while we're in visual mode?

smblott-github:
> Can we end up in insert mode while we're in visual mode?

We don't have a visual mode.  Just a placeholder for an implementation.

This is silly.

mrmr1993:
> We don't have a visual mode. Just a placeholder for an implementation.

But when we do, what will behaviour be? If we click an element that normally would trigger insert mode, will we go into insert mode? Do we stay in visual mode? Will we then be in insert mode when we leave visual mode?

> This is silly.

So is obfuscating the mode hierarchy.

smblott-github:
> But when we do, what will behaviour be?

Whatever we want it to be.

mrmr1993:
Sorry, I mistyped. I meant, how will someone who isn't you know the answer to these questions? Will we have to implement that as `InsertModeBlocker` too?

And then, where do we go to find out whether a mode is affected by passkeys? We know from the previous implementation that normal mode is, but find mode isn't. Will we have to have an `InsertModeBlockerWithPasskeys`?

This way of doing things is even more tangled than the *if-then-else* logic, which is why *if-then-else* is a paradigm, while *orthogonal-classes-which-stop-each-other-from-working-if-they're-unhappy-with-the-order-in-which-they've-been-instantiated* isn't.

mrmr1993:
Migrated passkeys logic to `passkey_mode.coffee`

mrmr1993:
Implemented visual mode template.

mrmr1993:
This is now at a stage where I'm fairly happy with it. Not everything has been perfectly ported to modes, but
* we have modes, which are instantiated by calling the constructor for their class, as in #1413 
* most mode-like things use modes (except for `focusInput`, purely because I haven't got around to it)
* we've completely avoided the evil of `handlerStack`, so we have assurances that things are run in a sensible order
* insert mode behaves properly in all situations (except maybe #1415)
* behaviour is the same as in #1424 (so everything is consistent with master except those cases that #1424 fixed).

Everything else should follow fairly easily from what's already here; certainly, I can't think of any more interesting changes to make.

@smblott-github can you take a look at this? Hopefully it achieves all the things you were trying for with #1413, and manages to deal with my concerns there too.

Issue #1361, opened 2014-12-19
Usability issues with vomnibar in iframe (post-1.46 only)


smblott-github:
In `post-1.46`, the vomnibar is in an iframe (see #1214 and #1139).  Using this for everyday browsing, I've found:
- The favicon and refresh button twitch every time the vomnibar is opened.
- The vomnibar is slower to load.  Not much slower, but noticeably slower.
- Several times, by the time the vomnibar opens, I've already typed the first few characters of the query, so the vomnibar misses them.  (Was that possible previously?)

Any ideas as to how this might be improved, @mrmr1993?

Probably not related, but I'll mention anyway:
- Several times, a bookmark I expected to open in a new foreground tab, opened instead in a new background tab.  (However, I could just have fumbled the keyboard somehow.)

(This does not affect `master` or `1.49`.)

PR #1427, opened 2015-01-12
Fix race condition on the options page.

### LINKED ISSUE #1426: DOMContentLoaded race condition in master
I'm seeing a race condition in `master` on the options page.

`DOMContentLoaded` is firing *before* we've registered our listeners in the front end.  So we never call `Vomnibar.init()`, so no vomnibar (which is what tipped me off the the problem).  It happens about half the time, for me.

Two possible sources of the problem could be:

1. On the options page, we're running as a regular, hosted script.  Generally, we're running as an extension.  The injection point may therefore be different.

1. On the options page, we've already fetched all of the necessary HTML before we begin injecting our scripts.  So, it could be that this is a general problem, but we only see it on the options page, because the options page is fast, and injecting our scripts is slow.

smblott-github:
Call onDOMContentLoaded handlers immediately, if the DOM is already loaded.

Fixes #1426.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -1142,7 +1142,7 @@ CursorHider =
 initializePreDomReady()
 window.addEventListener("DOMContentLoaded", registerFrame)
 window.addEventListener("unload", unregisterFrame)
-DomUtils.runWhenDOMLoaded initializeOnDomReady
+DomUtils.documentReady initializeOnDomReady
 
 window.onbeforeunload = ->
   chrome.runtime.sendMessage(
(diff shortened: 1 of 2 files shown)


mrmr1993:
Nice, LGTM.

gdh1995:
Nice. But, why we call `Vomnibar.init()` in initializeOnDomReady? `init` create a iframe of "pages/vomnibar.html", which will slow down loading of pages.
It would be better to call it when `Vomnibar.open` is called the first time, as what the old version does. As for the time taken to load a html file on local, I have done this in my vimium, and feel no difference.

smblott-github:
> It would be better to call it when Vomnibar.open is called the first time

In the general case there's a noticeable delay if you don't preload the vomnibar.  See #1361.

PR #1431, opened 2015-01-14
Don't change the selection if the target element is already selected

### LINKED ISSUE #1430: Sometimes, do not change selection on simulateSelect
In `simulateSelect`, we move the caret to the end.  Probably that is the wrong thing to do if the selection is already in the target element (for example, after find).

mrmr1993:
Fixes #1430.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -750,13 +750,6 @@ focusFoundLink = ->
     link = getLinkFromSelection()
     link.focus() if link
 
-isDOMDescendant = (parent, child) ->
-  node = child
-  while (node != null)
-    return true if (node == parent)
-    node = node.parentNode
-  false
-
 selectFoundInputElement = ->
   # if the found text is in an input element, getSelection().anchorNode will be null, so we use activeElement
   # instead. however, since the last focused element might not be the one currently pointed to by find (e.g.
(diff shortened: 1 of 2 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -764,7 +757,7 @@ selectFoundInputElement = ->
   # heuristic of checking that the last anchor node is an ancestor of our element.
   if (findModeQueryHasResults && document.activeElement &&
       DomUtils.isSelectable(document.activeElement) &&
-      isDOMDescendant(findModeAnchorNode, document.activeElement))
+      DomUtils.isDescendant(findModeAnchorNode, document.activeElement))
```
smblott-github: Suggest revert name to `isDOMDescendant`, unless there's a good reason to change it (?).

lib/dom_utils.coffee
```diff
@@ -141,11 +141,20 @@ DomUtils =
...
+      @isDescendant element, range.commonAncestorContainer
+    else
+      false
+
   simulateSelect: (element) ->
     element.focus()
-    # When focusing a textbox, put the selection caret at the end of the textbox's contents.
-    # For some HTML5 input types (eg. date) we can't position the caret, so we wrap this with a try.
-    try element.setSelectionRange(element.value.length, element.value.length)
+    unless @isSelected
```
smblott-github: Perhaps:

```coffeescript
@isSelected element
```

mrmr1993: Oops!

mrmr1993:
Updated.

smblott-github:
This doesn't seem to work as expected, @mrmr1993.  Could you check, please?

- Go [here](http://static.smblott.org/inputs.html).
- `/sssw<Enter>`
- `f`, and choose the first/selected text area.

I expect the selection to remain selected.  Instead, the caret is nevertheless moved to the end.

smblott-github:
Closing.  Fixed in #1436.

Issue #1375, opened 2014-12-22
Search is all wonky (jumps around all over the place)


smblott-github:
Go [here](https://github.com/philc/vimium/pull/1355/files):
- Make sure you're at the top of the page: `gg`.
- Type: `/show notes`

The search jumps around all over the place, and ends up selecting a match from nearer the bottom of the page.

PR #1434, opened 2015-01-16
Fix slightly wonky perform-find-in-place

smblott-github:
When we use `/` for find, the selection should stay put if it continues to match the query as characters are added.  The is handled currently by first searching backwards, before beginning the real search.  This trick doesn't always work.  Here's an example: 

- Go [here](http://static.smblott.org/inputs.html) (my wonky-find test page).
- Search: `/ooo`

The search jumps around unnecessarily. Bad UX.

This PR uses a different (simpler?) method for sticking to the existing match.  Basically, we store the initial selection, then restore it before each `performFindInPlace`.  So we're always starting from the same place.

(I was hoping to get a fix for #1375; but no joy.)

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -702,18 +702,10 @@ handleEnterForFindMode = ->
   settings.set("findModeRawQuery", findModeQuery.rawQuery)
 
 performFindInPlace = ->
-  cachedScrollX = window.scrollX
-  cachedScrollY = window.scrollY
-
   query = if findModeQuery.isRegex then getNextQueryFromRegexMatches(0) else findModeQuery.parsedQuery
 
-  # Search backwards first to "free up" the current word as eligible for the real forward search. This allows
-  # us to search in place without jumping around between matches as the query grows.
-  executeFind(query, { backwards: true, caseSensitive: !findModeQuery.ignoreCase })
-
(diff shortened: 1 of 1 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -928,7 +918,27 @@ showFindModeHUDForQuery = ->
...
+    range = document.createRange()
+    range.setStart document.body, 0
+    range.setEnd document.body, 0
+    range
+  else
+    selection.collapseToStart() if selection.type == "Range"
+    range = selection.getRangeAt 0
+
+findModeSaveSelection = ->
+  findModeInitialRange = getCurrentRange()
```
philc: Worth a comment to describe why we need to do this?

philc:
Clever =)

smblott-github:
Any more input on this?

Here's an example where this changes the UX:
- Begin find: `/abc`.
- With the mouse, change the selection.
- Continue adding/removing characters to/from the query.

In `master`, the search now proceeds from the selection.  With this, the search nevertheless reverts to the original selection.

In my view, the benefit of this fix outweighs the loss of a corner use case.  Other views?

(This use case was pointed out by @mrmr1993 in #1401.)

smblott-github:
I'm going to merge this:
- It's not too hard to back out if it creates a problem.
- And we'll never find such problems unless we give it a shake out.

PR #1436, opened 2015-01-18
On simulateSelect, do not change selection if selection exists.

### LINKED ISSUE #1430: Sometimes, do not change selection on simulateSelect

smblott-github:
Fixes #1430.

This is the same as #1431, except it uses a different version of `isSelected` ([here](https://github.com/philc/vimium/blob/master/lib/dom_utils.coffee#L171)).

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -177,9 +177,10 @@ DomUtils =
 
   simulateSelect: (element) ->
     element.focus()
-    # When focusing a textbox, put the selection caret at the end of the textbox's contents.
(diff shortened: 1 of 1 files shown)


smblott-github:
Merged.  This is considerably better than the existing UX.

Nevertheless, note the following from 45dd6235a31fecfd3235db9881ddfad8e5a2835a:

```
isSelected makes the wrong decision if the user has placed the caret at
the start of element.  We cannot distinguish that case from the user
having made no selection.  Nevertheless, this is substantially better
than the existing UX, which *always* moves the caret to the end.
```

PR #1438, opened 2015-01-18
Give focusInput (gi) a memory

smblott-github:
When there are multiple inputs available (such as on the new-issue page, here), `focusInput` allows the user to choose which input to focus (via `Tab`).  It's really annoying when the input you want is the third or fourth one.  And it's *particularly* annoying when you move in and out of insert mode several times, but always have to tab back to the same spot.

This PR gives `focusInput` a memory:
- It remembers the most recently-active input.
- When offering a choice, the most recently-active input is selected (not always the first).
- As a side effect, when `focusInput` is run again, the previous choice is immediately selected.

All of this applies only if the count is `1`.  Unfortunately,  this breaks the existing meaning of `1gi`.  In fact, the user has no way of saying *focus the fist input*.

Nevertheless, I propose that this is a big win for a fairly common use case, and losing the ability to force focus the first input is not so bad.

Agree?  Disagree?

Edit:  There's a suitable (junk) test page [here](http://static.smblott.org/inputs.html).

More edit: for correctness.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -338,7 +338,11 @@ getFocusedElementIndexByRecency = do ->
   installListener window, "focus", (event) ->
     focusedElement = event.target if DomUtils.isEditable event.target
 
-  (elements) ->
+  # Only for tests.
+  window.resetFocusInputFocusedElement = ->
+    focusedElement = null
+
+  (elements = null) ->
     Math.max 0, elements.indexOf focusedElement
 
 extend window,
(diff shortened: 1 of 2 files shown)


smblott-github:
The diff looks really big (due to a change in indentation).  In fact, the only changes are:
- These few [lines](https://github.com/smblott-github/vimium/blob/focus-input-with-memory/content_scripts/vimium_frontend.coffee#L336), to track the focus.
- And these two [lines](https://github.com/smblott-github/vimium/blob/focus-input-with-memory/content_scripts/vimium_frontend.coffee#L360), which change the way in which the starting element is selected if `count` is `1`.

smblott-github:
Merged.  This is a big UX improvement (in my view).  @philc... Since this is a UX change, you might want to take a look at it.  It can easily be reverted if you think it could confuse users.

philc:
LGTM @smblott-github !

PR #1439, opened 2015-01-20
Asynchronous DOM tests

smblott-github:
This is a re-working of the DOM test framework for asynchronous tests. It:
- installs the main frontend event handlers (for keyboard events, focus, blur, etc) in the DOM-test test page, and
- uses phantomjs to generate (and, therefore, test) real keyboard events (so, no more need for mock keyboard events).

DOM tests can now traverse the actual code path traversed in live use.  For example, we can verify that insert mode is triggered when an input element is focused and deactivated when that element blurs.  Or, in various modes, we can verify whether keyboard events are handled by Vimium or passed through.

Here's a concrete example from the `PostFindMode` tests.  This tests the feature whereby, when a find/search lands in an input element, an immediate `Escape` enters insert mode (but subsequent `Escape`s do not):

```coffeescript
  should "enter insert mode on immediate escape", ->
    sendKeyboardEvent "escape"
    assert.isFalse @postFindMode.modeIsActive
    
  should "not enter insert mode on subsequent escapes", ->
    sendKeyboardEvent "a"
    sendKeyboardEvent "escape"
    assert.isTrue @postFindMode.modeIsActive
```

Edit:  The diff looks really messy, and is impossible to understand.  The main changes are:
- Export `initializeWhenEnabled`, so we can install the actual frontend event handlers within the tests.
- Implement `reset` methods for the handler stack and modes.  These are for the tests only.
- Move some of the `initializePreDomReady` handling to a function `initializeModes`, which is exported (again, only for the tests).
- Implement a `sendKeyboardEvent` function within the phantomjs framework which uses a phantomjs callback to send "real" keyboard events to the test page.
- Rework the individual tests to work within the new framework (e.g., to use "real" keyboard events rather than "mock" keyboard events).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ node_js: 0.10
 before_install:
   - "npm install -g coffee-script"
   - "npm install path"
+  - "npm install util"
   - "cake build"
 script: "cake test"
 notifications:
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -111,6 +111,7 @@ Commands =
       "goUp",
       "goToRoot",
       "enterInsertMode",
+      "enterVisualMode",
       "focusInput",
       "LinkHints.activateMode",
       "LinkHints.activateModeToOpenInNewTab",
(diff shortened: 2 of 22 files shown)


smblott-github:
Merged.  The tests are clearly more rigorous.

PR #1441, opened 2015-01-22
[WIP] Visual and edit modes

smblott-github:
Heads up.  I started working on visual mode, and it turns out an editing mode comes almost for free.  This is a work in progress.  It's not ready for code review, and far from ready for merging.  Visual mode kinda works, though, and edit mode has movement characters implemented (but no `i`, `a` or `O`, yet).

Use `v` to enter visual mode, and `gv` to enter edit mode. `o` works in visual mode, although I'm not yet convinced of the current implementation.  Use `y` to yank, `Esc` to exit.

### DIFF
--- a/content_scripts/mode_visual_edit.coffee
+++ b/content_scripts/mode_visual_edit.coffee
@@ -221,8 +221,8 @@ class Movement extends CountPrefix
     "(": "backward sentence"
     "}": "forward paragraph"
     "{": "backward paragraph"
-    "$": "forward lineboundary"
     "0": "backward lineboundary"
+    "$": "forward lineboundary"
     "G": "forward documentboundary"
     "gg": "backward documentboundary"
 
@@ -254,6 +254,10 @@ class Movement extends CountPrefix
             if command and (@movements[command] or @commands[command])
               @selection = window.getSelection()
               @keyQueue = ""
+
+              # We need to treat "0" specially.  It can be either a movement, or a continutation of a count
+              # prefix.  Don't treat it as a movement if we already have a count prefix.
+              return @continueBubbling if command == "0" and 0 < @countPrefix
 
               if @commands[command]
                 @commands[command].call @, @getCountPrefix()
--- a/content_scripts/mode_visual_edit.coffee
+++ b/content_scripts/mode_visual_edit.coffee
@@ -112,10 +112,6 @@ class Movement extends CountPrefix
     else
       @selection.modify @alterMethod, movement...
 
-  # # Return a simple camparable value which will be different for different selections.
-  # hashSelection: ->
-  #   [ @element?.selectionStart, @selection.toString().length ].join "/"
-
   # Return a simple camparable value which depends on various aspects of the selection which may change when
   # the selection changes.  This is used to detect, after a movement, whether the selection has changed.
   hashSelection: (debug) ->
@@ -266,7 +262,7 @@ class Movement extends CountPrefix
   # Yank the selection; always exits; either deletes the selection or collapses it; returns the yanked text.
   yank: (args = {}) ->
     @yankedText = @selection.toString()
-    console.log "yank:", @yankedText if @debug  This is used to detect, after a movement, whether the selection has changed.
+    console.log "yank:", @yankedText if @debug
 
     if args.deleteFromDocument or @options.deleteFromDocument
       @selection.deleteFromDocument()
--- a/content_scripts/mode_visual_edit.coffee
+++ b/content_scripts/mode_visual_edit.coffee
@@ -619,25 +619,32 @@ class CaretMode extends Movement
     @selection.modify "extend", forward, character
 
   # When visual mode starts and there's no existing selection, we launch CaretMode and try to establish a
-  # selection.  As a heuristic, we pick the first non-whitespace character of the first visible text node
-  # which seems to be big enough to be interesting.
-  # TODO(smblott).  It might be better to do something similar to Clearly or Readability; that is, try to find
-  # the start of the page's main textual content.
+  # selection.  As a heuristic, we score visible text nodes by their visible area and the number of
+  # non-whitespace characters they contain.  We pick the first non-whitespece character in the highest scoring
+  # node.
   establishInitialSelectionAnchor: ->
     nodes = document.createTreeWalker document.body, NodeFilter.SHOW_TEXT
-    while node = nodes.nextNode()
-      # Don't choose short text nodes; they're likely to be part of a banner.
-      if node.nodeType == 3 and 50 <= node.data.trim().length
+
+    # Find and score candidate text nodes.
+    candidates =
+      for node in (n while n = nodes.nextNode())
+        continue unless node.nodeType == 3 and 0 < node.data.trim().length
         element = node.parentElement
-        if DomUtils.getVisibleClientRect(element) and not DomUtils.isEditable element
-          # Start at the offset of the first non-whitespace character.
-          offset = node.data.length - node.data.replace(/^\s+/, "").length
-          range = document.createRange()
-          range.setStart node, offset
-          range.setEnd node, offset
-          @setSelectionRange range
-          return true
-    false
+        rect = DomUtils.getVisibleClientRect element
+        continue unless rect and not DomUtils.isEditable element
+        area = (rect.bottom - rect.top) * (rect.right - rect.left)
+        chars = node.data.split(/\s+/).join().length
+        { node: node, score: chars * area }
+
+    if 0 < candidates.length
+      node = (candidates.sort (a,b) -> b.score - a.score)[0].node
+      # Start at the offset of the first non-whitespace character.
+      offset = node.data.length - node.data.replace(/^\s+/, "").length
+      range = document.createRange()
+      range.setStart node, offset
+      range.setEnd node, offset
+      @setSelectionRange range
+      return range
 
 class EditMode extends Movement
   constructor: (options = {}) ->
--- a/content_scripts/mode_visual_edit.coffee
+++ b/content_scripts/mode_visual_edit.coffee
@@ -19,7 +19,7 @@ class SuppressPrintable extends Mode
           if event.keyCode in [ 8, 46 ]
             @suppressEvent
           else
-            DomUtils.suppressPropagation
+            DomUtils.suppressPropagation event
             @stopBubblingAndFalse
         else
           @suppressEvent
--- a/content_scripts/mode_visual_edit.coffee
+++ b/content_scripts/mode_visual_edit.coffee
@@ -2,7 +2,7 @@
 # To do:
 # - better implementation of `o`
 # - caret mode
-# - find operations
+# - find operations (needs better implementation?)
 
 # This prevents printable characters from being passed through to underlying page.  It should, however, allow
 # through chrome keyboard shortcuts.  It's a backstop for all of the modes following.
@@ -285,6 +285,29 @@ class VisualMode extends Movement
       extend @commands,
         "d": -> @yank deleteFromDocument: true
         "c": -> @yank(); enterInsertMode()
+
+    # Map "n" and "N" for poor-man's find.
+    unless @options.underEditMode
+      do =>
+        findBackwards = false
+        query = getFindModeQuery()
+        return unless query
+
+        executeFind = => @protectClipboard =>
+          initialRange = @selection.getRangeAt(0).cloneRange()
+          caseSensitive = /[A-Z]/.test query
+          if query
+            window.find query, caseSensitive, findBackwards, true, false, true, false
+            newRange = @selection.getRangeAt(0).cloneRange()
+            range = document.createRange()
+            range.setStart initialRange.startContainer, initialRange.startOffset
+            range.setEnd newRange.endContainer, newRange.endOffset
+            @selection.removeAllRanges()
+            @selection.addRange range
+
+        extend @movements,
+          "n": -> executeFind()
+          "N": -> findBackwards = not findBackwards; executeFind()
 
     @clipboardContents = ""
     @paste (text) => @clipboardContents = text
--- a/content_scripts/mode_visual_edit.coffee
+++ b/content_scripts/mode_visual_edit.coffee
@@ -1,13 +1,32 @@
 
-# Todo:
-# Konami code?
-# Scroll is broken (again).  Seems to be after dd.
-# Paste of whole lines.
-# Arrow keys.
-# J
+#
+# The main modes defined here are:
+# - VisualMode
+# - VisualLineMode
+# - CaretMode
+# - EditMode (experimental)
+#
+# SuppressPrintable and CountPrefix are shared utility base classes.
+# Movement is a shared vim-like movement base class.
+#
+# The inheritance hierarchy is:
+# - Mode, SuppressPrintable, CountPrefix, Movement, [ VisualMode | CaretMode | EditMode ]
+# - Mode, SuppressPrintable, CountPrefix, Movement, VisualMode, VisualLineMode
+#
+# The possible mode states are:
+# - ..., VisualMode
+# - ..., VisualLineMode
+# - ..., CaretMode
+# - ..., VisualMode, FindMode
+# - ..., VisualLineMode, FindMode
+# - ..., CaretMode, FindMode
+# - ..., EditMode
+# - ..., EditMode, InsertMode
+# - ..., EditMode, VisualMode
+# - ..., EditMode, VisualLineMode
+#
 
-# This prevents printable characters from being passed through to the underlying page.  It should, however,
-# allow through Chrome keyboard shortcuts.
+# This prevents printable characters from being passed through to underlying modes or to the underlying page.
 class SuppressPrintable extends Mode
   constructor: (options = {}) ->
     handler = (event) =>
@@ -19,15 +38,13 @@ class SuppressPrintable extends Mode
       DomUtils.suppressPropagation event
       @stopBubblingAndFalse
 
-    super extend options,
-      keydown: handler
-      keypress: handler
-      keyup: handler
+    super extend options, keydown: handler, keypress: handler, keyup: handler
 
-# This watches keypresses and maintains the count prefix as number keys and other keys are pressed.
+# This monitors keypresses and maintains the count prefix.
 class CountPrefix extends SuppressPrintable
   constructor: (options) ->
     @countPrefix = ""
+    # This allows us to implement both "d3w" and "3dw". Also, "3d2w" deletes six words.
     @countPrefixFactor = options.initialCountPrefix || 1
     super options
 
@@ -43,7 +60,6 @@ class CountPrefix extends SuppressPrintable
               else
                 ""
 
-  # This handles both "d3w" and "3dw". Also, "3d2w" deletes six words.
   getCountPrefix: ->
     count = @countPrefixFactor * if 0 < @countPrefix?.length then parseInt @countPrefix else 1
     @countPrefix = ""
@@ -61,7 +77,7 @@ paragraph = "paragraph"
 vimword = "vimword"
 lineboundary= "lineboundary"
 
-# This implements movement commands with count prefixes for both visual mode and edit mode.
+# This implements vim-like movements, and includes quite a number of gereral utility methods.
 class Movement extends CountPrefix
   opposite: forward: backward, backward: forward
 
@@ -71,16 +87,14 @@ class Movement extends CountPrefix
   copy: (text, isFinalUserCopy = false) ->
     chrome.runtime.sendMessage handler: "copyToClipboard", data: text
     # If isFinalUserCopy is set, then we're copying the final text selected by the user (and exiting).
-    # However, we may be called again from within @protectClipboard, which will later try to restore the
-    # original clipboard contents.  Therefore, we disable copy so that subsequent copies will not propagate.
+    # However, @protectClipboard may later try to restore the original clipboard contents.  Therefore, we
+    # disable copy so that subsequent copies do not propagate.
     @copy = (->) if isFinalUserCopy
 
   # This s used whenever manipulating the selection may, as a side effect, change the clipboard's contents.
   # We restore the original clipboard contents when we're done. May be asynchronous.  We use a lock so that
-  # calls can be nested.
-  #
-  # We do this primarily for edit mode, where the user does not expect caret movements to change the clipboard
-  # contents.
+  # calls can be nested.  We do this primarily for edit mode, where the user does not expect caret movements
+  # to change the clipboard contents.
   protectClipboard: do ->
     locked = false
 
@@ -97,9 +111,10 @@ class Movement extends CountPrefix
     if @options.parentMode
       @options.parentMode.launchSubMode mode, options
     else
-      new mode
+      new mode options
 
-  # Return the character following the focus, and leave the selection unchanged.
+  # Return the character following (to the right of) the focus, and leave the selection unchanged.  Returns
+  # undefined if there is no such character.
   getNextForwardCharacter: ->
     beforeText = @selection.toString()
     if beforeText.length == 0 or @getDirection() == forward
@@ -111,7 +126,7 @@ class Movement extends CountPrefix
     else
       beforeText[0]
 
-  # Return the character preceding the focus, and leave the selection unchanged.
+  # As above, but backwards.
   getNextBackwardCharacter: ->
     beforeText = @selection.toString()
     if beforeText.length == 0 or @getDirection() == backward
@@ -125,10 +140,10 @@ class Movement extends CountPrefix
 
   # Test whether the character following the focus is a word character.  Leave the selection unchanged.
   nextCharacterIsWordCharacter: do ->
-    regexp = /[A-Za-z0-9_]/
-    -> regexp.test @getNextForwardCharacter()
+    regexp = /[A-Za-z0-9_]/; -> regexp.test @getNextForwardCharacter()
 
-  # Run a movement.  For convenience, the following three argument forms are available:
+  # Run a movement.  This is the core movement method, all movements happen here.  For convenience, the
+  # following three argument forms are supported:
   #   @runMovement "forward word"
   #   @runMovement [ "forward", "word" ]
   #   @runMovement "forward", "word"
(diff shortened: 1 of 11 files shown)


smblott-github:
Status report.

None of this is ready for code review.  It is, however, ready to be taken for a spin.

#### Visual Mode

Visual mode is fairly functional.  The available movements are [here](https://github.com/smblott-github/vimium/blob/visual-and-edit-modes-snap-1/content_scripts/mode_visual_edit.coffee#L122). Enter visual mode with `v` or `V`, leave with `y` or `Escape`.

Super cool: `o` works.  This is essential for most realistic use cases.

#### Edit Mode

Edit mode is kinda cool, but a bit flaky in places.  There's quite a struggle deciding what a line is.  Does it end with a newline?  That's the vim way, but it breaks a lot of browser idioms.  Here, we follow the Chrome model, a line ends where it wraps.  This has some odd consequences.

The edit mode commands are [here](https://github.com/smblott-github/vimium/blob/visual-and-edit-modes-snap-1/content_scripts/mode_visual_edit.coffee#L401).  It leverages visual mode a lot.  You can say things like `d3w`, or `3dw`, `ce`, or `2ce`.  Or `C`, `c$` or `c0`.

It's not obvious that edit mode is up to snuff, or even can be brought up to snuff. I'd describe it as an experiment, for now. That being said, I'm using it to type this message, and it's pretty cool being able to move around with vim-like movements.

Edit mode works in the compose window of Google's inbox, which is kind of required.  However, the scrolling there if flaky there (actually, it's just not implemented for contentEditable at all).

Next on the list: `dap`, `cas` and so on.

leoj3n:
Yes! This would be awesome @smblott-github — I'll take it for a :cyclone:

smblott-github:
Forgot to say, enter edit mode with `e` or (just added), `gv`.  Still experimenting with (whether!) the default binding.

Also, now added `das`, `cap`, etc.

leoj3n:
This message is being composed using the new vimium visual and edit modes! Already works great and feels natural. Would be nice if there was some indication of which mode is current. Love it!

smblott-github:
@leoj3n... Great!  Be aware, though, that this is a work-in-progress.  There are still plenty of glitches.

> Would be nice if there was some indication of which mode is current.

Check out the Vimium browser icon (make sure it's visible).  It should give you feedback.

leoj3n:
I'll be careful when composing important messages @smblott-github...

Making the icon visible on the toolbar does the trick!

smblott-github:
Update...

Visual mode should be pretty solid now.  Improvements:
- Implements a caret mode if no initial selection exists.
- You can enter and use find mode from within visual mode.
- Visual line mode too, and you can toggle between the two.

Edit mode:
- Is better, but still quirky in places (notably, it does a poor job of melding vim's line-oriented operations (`dd`, `yy` and so on) with Chrome's notion of a line, which is wrapped).
- Still pretty experimental.
- Works on Google's Inbox, probably GMail too.

z0rch:
@smblott-github, great work!

Is it a good time to start reporting some issues/questions? 
I'll put only a couple for now, until you ask for more testing :smile: 

1. `e` correctly jumps to the end of a word, but `w` does something rather unpredictable -- it jumps to the end of the page/textarea, in both visual and edit mode.

2. Entering visual mode scrolls to the last search position.

Perform these actions on the current github page:

* `gg` -- Jump to the top.
* `/whether!<RET>` -- Find your message, it is in the middle of this page. The goal is to find any place, from which you can scroll up and down.
* `G` -- Jump to the bottom.
* `ve` -- Enter visual mode and select a word.

Actual: Page will be scrolled back to the last search occurrence. Sometimes page was scrolled even back to the top, but it seems I cannot reproduce this constantly.

Expected: I'd like page not to scroll anywhere. Since the word I searched previously is out of my viewport, `v` will enter visual mode on the first character, that I can see.

---------------

Windows, `febe308`.

smblott-github:
> Is it a good time to start reporting some issues/questions?

Now is a great time!  Thanks for your input, @z0rch.

>  in both visual and edit mode.

I'm most interested in debugging visual mode.  Edit mode is experimental.  Let's get visual mode right first.

> w does something rather unpredictable

Let me know if you find a page where you can reliably reproduce this (in visual mode). `w` is different.  It's not one of the native Chrome movements.

> Since the word I searched previously is out of my viewport, v will enter visual mode on the first character, that I can see.

That seems reasonable. The UX is weird if it starts jumping around.  I'll try it.

An alternative I've been wondering about is... When there's no (visible) selection, launch visual mode, then immediately launch find mode (so the user can establish a selection).  This caret lark is not a great UX.  In many cases, the caret may be nowhere near what the user is looking for.

> Windows, febe308.

Yup! Please *do* quote the commit.

z0rch:
>  `w` is different. It's not one of the native Chrome movements.

That explains a lot, and is rather unfortunate....

Right now I have a Windows and a Linux machine next to each other, both have Chrome 64bit with https://github.com/smblott-github/vimium/commit/70c7cb66e96305b0f8bbd8687fe8272d989ed3d9.

Right on this github page, performing the following actions:

* `/That seems <RET>` to find the starting point.
* `v` to enter visual mode.
* `w` to select a word.

Linux: `That seems reasonable` is selected.

Windows: 

image

z0rch:
I can see you are making some changes right now, so perhaps you already know the reason, but I want to give you my debugging results as well.

We will inspect the following sentence (`[` and `]` will denote selection): 

> [That seems ]reasonable. The UX is weird.

Running the following command:

```js
window.getSelection().modify("extend", "forward", "word")
```

On Linux gives us:

> [That seems reasonable]. The UX is weird.

On Windows:

> [That seems reasonable. ]The UX is weird.


Since your implementation of `[forward, vimword]` in running `[forward, word]` as a prerequisite (using the snippet above) until the next character is a letter, on Windows the code loops infinitely.

smblott-github:
Thanks, @z0rch.
- 6cff1b0dfccdb137e95e52c9a931fed14cec82d5 now changes visual-mode start up (as discussed).
- 11199de98a2c29601e38e75af561057ab55a0a83 changes the definition of a vim word (for `w`).  This works nicely (for me) on Linux.  I'm going to guess it jumps one word too far on Windows, but not to the end of the document any more.  I don't have a Windows machine to test.  We may have to give up on a vim-like `w` and just use `e` instead. The operation is lines 151-154, [here](https://github.com/smblott-github/vimium/compare/visual-and-edit-modes#diff-0d02c6fa5282032d83254084e0629953R151), if you can get a version that works on Windows. 

> On Windows:

>> [That seems reasonable. ]The UX is weird.

Ah, so that's exactly the definition of `w` we want.  Is `e` broken on Windows?

z0rch:
> Ah, so that's exactly the definition of `w` we want. Is `e` broken on Windows?

Oh, indeed, I didn't notice! 

On Windows, `e` is behaving exactly like `w` on Linux.

z0rch:
With the https://github.com/philc/vimium/commit/11199de98a2c29601e38e75af561057ab55a0a83, Linux still behaves as expected, on Windows `w` and `e` now behave exactly the same, just like `w` is supposed to be. So as for now, `e` is broken.

smblott-github:
Thanks for your help, @z0rch.

Could you try d03ba099fcbe5e07dfdfa454afdb86fd5aad2c89, please?  It resorts to character-by-character movements for the cases we know to be different.  That's pretty bad.  But words can't be that big, right?  So it shouldn't be slow.  And these details are important for the UX.

z0rch:
With https://github.com/philc/vimium/commit/d03ba099fcbe5e07dfdfa454afdb86fd5aad2c89, `e` and `w` on Windows behave as expected. 

It doesn't feel slow while pressing `w` or `e` multiple times. However, if I press and hold `w` or `e` for a couple of seconds, then there is a visible lag before a new selection is displayed. Holding `b` is fast, for example. Just tried https://github.com/philc/vimium/commit/11199de98a2c29601e38e75af561057ab55a0a83 again, it was better there as well, but not "fast".

Since the behavior is correct now, perhaps we should leave it as it is. We can address this particular performance issue when it becomes an issue :smile:

z0rch:
One more finding: if I was in visual/caret mode, then clicked inside a text area with a mouse -- is it possible to detect this, and switch into insert mode?

smblott-github:
> One more finding: if I was in visual/caret mode, then clicked inside a text area with a mouse -- is it possible to detect this, and switch into insert mode?

That makes sense: 847aca4859567aaa88e19d41139d236102bc2024.

z0rch:
Yank copies the selected text and cancels the selection. Can we also cancel the selection when hitting `<ESC>`? I just find it inelegant, if I want to escape, I want to escape from the selection either :smile:

smblott-github:
> Can we also cancel the selection when hitting <ESC>?

I agree that it's pretty ugly leaving the selection behind.  On the other hand, the thinking was that there should be a way out which doesn't trash the selection.  So I'm 50-50 on this one.

z0rch:
That is true, sometimes we want to return to the last selection. What do you think of another idea:

* `Esc` cleans the selection
* Implement Vim's `gv` command for `normal` and `caret` modes, which will start `visual` mode and re-select the last selection.

Then it is clear: if we see a selection, we **always** know that (a) we are in `visual` mode and (b) the selection is  actionable. Plus think it would be more Vimmy :smile:

smblott-github:
> What do you think of another idea....

Or, another idea:
- When entering visual mode, note the selection (which may be `None`).
- When leaving with `Escape`, restore the original selection.

This:
- kinda makes sense (it's easy for the user to understand),
- gets around your original problem vis-a-vis leaving behind the selection on `Escape`, and
- does not require an extra command.

z0rch:
I'm playing with your idea now.

> This kinda makes sense (it's easy for the user to understand),

It is easy to understand **how** it works, but I'm not sure I get **why** it works like this. Why do we want to treat selections made by mouse and with vimium differently? 

Example of what I tried: I'm in normal mode, I make a approximate selection with a mouse, enter visual mode, extend the selection with vimium's commands, then:

* `y` -- entire selection is gone.
* `ESC` -- I'm back to the approximate selection... what am I supposed to do with it, or how to clear it?

Perhaps you have a better example, when a user wants to treat mouse selection and vimium selection separately?

Also, it just occurred to me, in vim `gv` scrolls the view if the last selection is out of viewport now (and that is great, I think). 

Current implementation does not scroll, so if I had my mouse selection somewhere out of viewport, entering visual mode starts a new selection (as we discussed previously), and exiting with `ESC` re-selects mouse selection somewhere out of viewport, and user knows nothing about it. 

Is it expected? Could you please also cover this part in your usage example(s)?
  
  
At any rate, if the user never touches the mouse, then initial selection is always `None`, and so `ESC` works as I expected.

> This does not require an extra command.

In regards of `gv` command, I think we would want it anyway, sooner or later :smile: Certainly many use-cases are in edit mode (when we get to it), but I can also imagine a situation to use `gv` in visual mode. 

Example: enter visual mode, select a text, yank, switch to another tab, get distracted and copy something else, return to the original tab, `gvy` to re-select last selection and yank it again.

smblott-github:
Visual mode works well on Google Inbox *if there's an established, visible selection*.  If there is not, however, it's pretty broken.  The problem seems to be `DomUtils.getVisibleClientRect` returning a positive, visible rectangle for elements which aren't actually visible at all.

smblott-github:
> I'm not sure I get why it works like this...

I'm not completely wedded to the approach described (and implemented), but nevertheless there is some logic to it. `Esc` means "abandon", "give up", "I didn't really want to do that".  So we make it as if the user had never entered visual mode at all.

If you enter visual mode with an existing selection then exit with `Esc`, you have three choices:
1. Leave the partial selection from visual mode active (@z0rch questioned this previously).
2. Clear the partial selection. This seems unnecessarily destructive (you're removing two selections!) but not wholly unreasonable.
3. Restore the previous selection. This is what this PR does currently, and also seems reasonable.

content_scripts/mode_visual_edit.coffee
```diff
@@ -0,0 +1,789 @@
...
+          char = @getNextForwardCharacter()
+        @runMovement forward, character
+
+  # Try to scroll the focus into view.
+  scrollIntoView: ->
+    @protectClipboard =>
+      if @element and DomUtils.isEditable @element
+        if @element.clientHeight < @element.scrollHeight
+          if @element.isContentEditable
+            # Help(smblott)! How do we do this?  This case matters for Gmail and Google's inbox.
```
mrmr1993: [`window.getSelection().getRangeAt(0).getClientRects()[0]`](https://developer.mozilla.org/en-US/docs/Web/API/Range.getClientRects) looks like it does the right thing for this, at least on Chrome for android.

The non-editable element case below should probably also use this somewhere so we scroll to the selection in `overflow: scroll` elements.

mrmr1993:
>The problem seems to be `DomUtils.getVisibleClientRect` returning a positive, visible rectangle for elements which aren't actually visible at all.

That function is designed to give the rect of a descendant if one is visible but the element itself isn't; presumably that's what's causing the issue. Using `element.getClientRects()` makes more sense for this (and everything else apart from link hints).

z0rch:
> If you enter visual mode with an existing selection then exit with Esc, you have three choices:

Let's get more opinions on this... @philc, @mrmr1993 ?

My main reason to vote for no.2 (remove all selections, previous and vimium's) is because I'm missing a real-world example, describing the purpose of preserving any selection while switching into `normal` mode. It is unclear, what further action can I apply for the selected text while being in the `normal mode`, and if there is no action -- why would I want to see such selection at all?

mrmr1993:
> Let's get more opinions on this...

I would suggest keeping the selection, which would enable a few things:

* running a bookmark in the Vomnibar that uses the current selection
* click a link on the page that does something with the selection (eg. make text bold)
* trigger a page shortcut using the selection (eg. `r` here to quote the selection to use in a reply)
* scrolling the page (looks like, from the code, automatic scrolling to the selection point might still be a bit sticky)
* use the selection as the base point for the next search (this may happen anyway, I'm not in a position to test anything at the moment)

I can see the value in being able to clear the selection, but forcing a clear just because the user changes mode seems a bit heavy handed; perhaps a subsequent `esc` could do that. I don't understand the use case for restoring the old selection.

z0rch:
> but forcing a clear just because the user changes mode seems a bit heavy handed; perhaps a subsequent esc could do that.

Perhaps my brain is too used to vim, where visual selection is always removed when switching to the `normal` mode :smile: However, the idea of subsequent `ESC` making complete cleanup also sounds interesting.

Thanks for the examples! I think it is clear that they would benefit from preserving *entire* selection, not only the initial selection, made by the mouse.

------------------

> trigger a page shortcut using the selection (eg. `r` here to quote the selection to use in a reply)

This particular example made me wonder, what if our passkeys could understand the mode? So that `r` in normal mode reloads the page, but `r` in visual mode is passed to the page (which would result in quoting the selection).

smblott-github:
@mrmr1993's examples are good.  I'll revert to option 1.

smblott-github:
I tried @z0rch's idea about a second escape clearing the selection (95e086563b918364d3038f6489cc97c73fcb7180).  It works rather nicely, and feels right.  The only issue would be... *how would a user know this*.  But we have other even more obscure uses of Escape, so I wouldn't worry too much.  It's pretty obvious what happened.

Suggest trying it out like this for a bit.

z0rch:
It really feels nice, and I think people will find this hidden gem.

> I tried @z0rch's idea

Kudos shall go to @mrmr1993, he proposed this first :+1: 

-----------

>  @mrmr1993: I don't understand the use case for restoring the old selection.

2 vs 1, then let's postpone discussing `gv` until we have a clear use-case. 

-----------

I found a bug, it may be related due to the recent changes, but I haven't tried earlier commits. It is reproducible for both `caret` and `visual` modes, but it is more clear what happens in `caret` mode:

* Use [this page](https://github.com/smblott-github/vimium/commit/95e086563b918364d3038f6489cc97c73fcb7180) as a playground.
* Enter `caret` mode approx. at the position, marked as (1)
* Hold `j` -- you will move down up until position (2), then you are stuck.
* Hold `k` -- you will move up until position (3), but there is no way you can get out of the textarea.

image

* Well no, from position (2) by holding `l` you can move down until the footer, but that doesn't help returning into position (1).

smblott-github:
> I found a bug, it may be related due to the recent changes....

It's not related to recent changes, it's related to how the selection works (which is pretty broken).  Here's another example.  On [this page](https://github.com/philc/vimium/pull/1441/files), click at a position just above a file boundary, enter visual mode and tap `j` a few times until you're below the file boundary.  Now, tapping `k` won't bring you back above the boundary.  You end up stuck like this...

snapshot

In this case:
- `selection... "extend", "forward", "line"` succeeds, but an immediate
- `selection... "extend", "backward", "line"` fails.

And here's another ugly example.  Start with the selection at the word "become", and use `j` to go down.  The links on the right are included in the selection.

snapshot

I'm not sure what we should do about these cases.  This last example is just *how-the-selection-works*.  You get the same thing with the mouse, and users are happy to live with it.

Probably we should focus on the important use cases, ensure they work well, and accept that there are corner cases which don't.  We could consider getting rid of caret mode, and requiring the user to establish an initial selection.  If the user establishes the selection (with find, say), they're probably pretty close to the text they want.  If we let caret mode guess the initial selection, it can end up in an odd place.  Or we, could do a better job of guessing the initial caret position, but I'm not sure how.

Any better ideas?

smblott-github:
Let's call these bad use cases resulting from strange selection behaviour *selection traps*.

If our basic tool is the selection, then we may just have to live with selection traps.  What we can do, however, is try to reduce the chance of users encountering them in the common use cases.  If the user has already established a selection, then we just assume that they know what they are doing.  If they have not, then caret mode kicks in.  If caret mode makes a poor guess of the initial selection, then we increase the chance of the user encountering a selection trap.

c1681fea2f2629c6bee1e27c5dfc704d77553d96 implements a different heuristic for guessing the initial selection in caret mode.  Instead of choosing the first visible text node, it scores text nodes based on their visible area and the number of non-whitespace characters they contain.  It seems to do a better job on the BBC News web site and on Wikipedia.

I'll try this out for a few days, and see how it feels in practice.

(Another possibility is to also weight by position within the viewport.)

z0rch:
I agree with you, the initial selection rules the game. We could spend some time improving heuristics, but since the root case cannot be fixed, we have to live with it.

Some structured (*strictly subjective*) thoughts and "just another idea":

* Making initial selection with a mouse eliminates the need of visual mode in 90%. If I've already spent time fetching my hand towards the mouse, in most cases I'll bother making the exact selection I want (unless it is complicated, but that falls into the remaining 10%).
* Caret mode without initial selection is a basic tool. "Better than nothing", but I'm *not* sure than in general case entering caret mode and scrolling with `hjkl` just to find the starting position of my future selection is faster than grabbing the mouse.
* Find mode is much better tool to make initial selection. What I dislike in using find mode for visual selection is the fact that... it jumps! When I want to select something, I'm looking directly at the desired starting position (very focused and almost not blinking :smile:). But if I make a typo, high chances are that page is scrolled somewhere, and I had lost my visual contact. Even though it is as easy as to hit `backspace`, I find this behavior very annoying.

Out of the last thought comes another idea: what if there was a special mode, very similar to our usual find mode, but restricted to the current viewport? If there are no matches in the current viewport, it would not scroll anywhere, but just say "0 Matches" (preferably in bold red, so that I can see it with the corner of my eye :smile:).

smblott-github:
> Making initial selection with a mouse eliminates the need of visual mode in 90%.

Generally, I agree.  However, my laptop has a twitchy mousepad, and then visual mode still wins out.

> but I'm not sure than in general case entering caret mode ... is faster than grabbing the mouse.

> Find mode is much better tool to make initial selection.

Agreed.

I realise now that c1681fea2f2629c6bee1e27c5dfc704d77553d96 (place caret at biggest visible text node) actually makes matters worse in some cases. From that position, `/` is searching forward, and missing half of the visible content.  It's probably better to revert to placing the caret near the top.  That way, `/` starts from there.  Which is almost your idea, @z0rch.  It doesn't restrict `/` to the viewport, but at least we do search all of the visible content.  Or, we could just change `/` under visual/caret mode to start from the top of the viewport and search from there.  Maybe that's the way to go?

(I'm not too keen on yet-another user-facing mode, just to solve a UX issue which we should be able to get right.)

z0rch:
> Or, we could just change `/` under visual/caret mode to start from the top of the viewport and search from there. Maybe that's the way to go?

That solves many problems, except the one -- when you make a typo, in case the only match is now somewhere out of the viewport, the page will be scrolled away. Yes, `backspace` would scroll back (more or less), but the situation itself is still annoying. 

But perhaps, for v1 we can consider it as acceptable.

------------

> I'm not too keen on yet-another user-facing mode, just to solve a UX issue which we should be able to get right.

There was an evil plan behind this idea :smile:. Initially it could just be "usual find mode, restricted to the current viewport", but at the same time it could be a baseline for future improvements, which we could make after v1.

WARNING: Whatever I say below, I'm not proposing to implement it now :smile: 

For example, how I would love to improve this "restricted find mode" (I think I've already described it earlier somewhere):

* When you type `/the`, *all* occurrences of "the" are highlighted in the viewport.
* When you hit `Enter`, for every occurrence a `link hint` appears.
* Type the letter in preferred link hint, and the selection will start from there.

This works perfectly for my "staring, non-blinking" workflow -- I never move my eyes away from the place where I want the selection to start. While typing the search query, I can see that it becomes highlighted (among other matches). I can also see if I made a typo, if highlighting suddenly disappears. I type a few letters, hit `Enter` and in the place where I am looking appears a link hint. I hit the letter in the link hint -- and voila, I've just started a selection!

smblott-github:
> except the one -- when you make a typo, in case the only match is now somewhere out of the viewport

Actually, no.  Any new search (after adding or removing characters, or after typos) begins at the same place as the original search.  This has always been the case, although the implementation was recently changed (#1434).

z0rch:
I'll explain what I mean. 

Suppose on this github page I scrolled to the bottom of the page. In my viewport there are a couple of "will" words. The only occurrence of "WIP" word is the title, somewhere high above my viewport. I type `/wi` and then accidentally hit `p` instead of `l`. Since the only occurrence of `WIP` is in the top, the page was scrolled to that location. This part is what I called confusing, it takes time to understand, what has just happened. But then, if I guess to hit `backspace`, I am more or less returned to the same scroll location as before, maybe slightly shifted.

smblott-github:
I agree.  See #1459.  I think that would fix your concern for this use case.  I'm not sure yet whether it's always the right thing to do.

z0rch:
Thanks, I'll have a look later today.

z0rch:
@smblott-github, I had a look at #1459, it is not exactly what I meant here, but that is an interesting idea. I'm going to play with it for a couple of days to see how it works for me.

Back to this discussion. I'll try to be more specific in my steps:

Suppose this is my current viewport:

image

I want to visually select a phrase `will be scrolled`. To do that, I will press `v/wil<Enter>`. So I invite you to follow my thoughts in slow motion.

- I enter `visual` mode, then enter `find` mode to find the initial selection. I type `w`, `i`....

image

- ... but unfortunately I make a mistake, and instead of `l` I press `p`...

image

- ... wow, what just happened? What am I looking now? Where has the phase `will be scrolled` gone, that I wanted to select?..... aaaah, I made a typo, so let me hit `backspace`.

The annoying part is the **confusion**, that appears after I was thrown completely out of context. 

What I would prefer is **not being scrolled anywhere**.

Obviously the issue described is relevant **only** for find in visual mode, since in 99% of cases the start of a selection will be inside the current viewport. While in normal mode, if I use find I expect the jump to happen, since I am searching across the entire page.

philc:
Yeah this is awesome. (And this is an epic discussion thread).

Notes on the UX:

* When exiting the caret mode with esc, we should clear the selection. This is the behavior you'd expect from your editor.
* When exiting visual mode with esc, I also believe we should clear the selection. Otherwise how do you ever get rid of the selection? Scenario: make a selection with v using caret mode. Hit esc. Now use v again in an attempt to make a different selection. How can you? You're stuck editing/expanding your previous selection.
* "gv" should reselect the most recent selection and put you in visual mode. This is a canonical vim binding.
* Have you given some thought to how to allow for the customization of the keybindings while in visual mode?

Notes on caret mode:

* If you use caret mode to scroll up on this page, eventually your cursor will get sutck in one of "XYZ added some commits") section on this page. That's not ideal.
* How do you feel about C-D and C-U working in caret mode as they normally do? The cursor will move up with the viewport.
* A potentially nice extension to caret mode would be to support easymotion/AceJump style jump-to-hint functionality

mrmr1993:
> When exiting visual mode with esc, I also believe we should clear the selection.

@philc, how should we handle things the user might want to do with a selection (a brief list in [my earlier comment](https://github.com/philc/vimium/pull/1441#issuecomment-72424862), not including uploading the selection to an online clipboard service, sharing to social media, etc.)? If we do clear it, then do we implement all of them ourselves (and leaving the ones we practically can't do, like pasting rich text over the selection in a `contentEditable` element), or leave the user with a really nice way of controlling a crippled-functionality selection?

I would rather see the `esc` key take the following actions, in decreasing order of priority:
* Cancel the current partially-typed command
* Blur the active element if it is editable/an `<embed>`/`<object>`
* Exit the current mode, if it can be exited (ie. isn't normal mode)
* Clear the selection (possibly only if we created it via visual mode)

Keeping these distinct and in a consistent order makes it really easy to reason about what's going to happen, and still gives us maximum functionality.

I think @smblott-github has implemented basically this (although I'm not sure about the interplay of the first and last wrt being back in normal mode).

z0rch:
> how should we handle things the user might want to do with a selection

**Ideologically** I'm on the opposite side in this particular discussion. In Vim it is very clear -- if you see a selection, you are in visual mode. If you want to do something with the selection -- you bind a command for visual mode.

Following this pattern, if you want to run a native action on the selection -- you should introduce a passkey for visual mode. 

**Practically**, this requires much bigger effort to achieve. And "double/triple" `ESC` is an easy and fast trick, that we already have.

smblott-github:
*---SNIP--- Long and complicated draft post deleted ---SNIP---*

Issues:
- Custom key mappings.
- What to do with the selection on exit (as of 449462b2580478b9c6d8b35f05231aad989d01aa, the selection is always removed -- although I know, @mrmr1993, that this may not be the best approach).
- Interaction with find mode (this requires more discussion, @z0rch).
- Is caret mode really helping us at all?
- Possibly others...

None of these issues are insumountable.  However, all of them require further discussion, and perhaps some experimentation. This thread is already too long and the issues too diverse for us to have a productive discussion on each of them simultaneously.

On the positive side, I think the core functionality (selecting short texts via a vim-like visual mode) is in pretty good shape.

Proposal:
- I will bring the code into reasonable shape such that it can be merged.
- Then... Merge this with the issues above unresolved, and address each in its own thread.
- In the merge, I'd like to include the edit-mode code.  This remains highly experimental, so should be hidden from the help menu.  However, it too needs to be discussed, and separately from the details of entering and exiting visual mode.  If edit mode dies, I will remove all trace of it at a later date.

z0rch:
I highly agree, let's merge and later take all the issues one by one.

philc:
SGTM

smblott-github:
I've merged this with edit mode and mode debugging turned off.  Time to move on to discussing the other issues in separate threads.

PR #1448, opened 2015-01-29
Fix parsing of command mappings.

### LINKED ISSUE #1443: Spaces after commands in "Custom key mappings"
Hello! 
If you accidentally put space after command in textarea "Custom key mappings" in options, that command won't be parsed (e.g. "map л scrollUp " instead of "map л scrollUp"). I think that it is the defect which can be fixed trivially.

smblott-github:
This fixes the parsing of command mappings with trailing whitespace, this case:

```
map q scrollDown_
```

(where `_` is whitespace).

It turns out there's actually another problem.  The following doesn't work either:

```
_map q scrollDown
```

I didn't fix this case.  It doesn't seem to be causing problems in the wild, and there may be people relying on it as an alternative way of commenting out rules.

Fixes #1443.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -58,7 +58,7 @@ Commands =
 
     for line in lines
       continue if (line[0] == "\"" || line[0] == "#")
-      splitLine = line.split(/\s+/)
+      splitLine = line.replace(/\s+$/, "").split(/\s+/)
 
       lineCommand = splitLine[0]
 


background_scripts/commands.coffee
```diff
@@ -58,7 +58,7 @@ Commands =
 
     for line in lines
       continue if (line[0] == "\"" || line[0] == "#")
-      splitLine = line.split(/\s+/)
+      splitLine = line.replace(/\s+$/, "").split(/\s+/)
```
z0rch: `line.trimRight()` is easier to read.

smblott-github: > line.trimRight() is easier to read.

Well, you learn something every day!

PR #1454, opened 2015-02-01
Give find mode a history

smblott-github:
This gives find mode a history.

(Am I missing something?  Why have we never had this before?)

Edit:
- Also fix the long-standing issue whereby we report `1 Matches`.
- Also fix the long-standing issue whereby we report a match count for empty queries.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -762,10 +762,7 @@ handleDeleteForFindMode = ->
     exitFindMode()
     performFindInPlace()
   else
-    findModeQuery.rawQuery = findModeQuery.rawQuery.substring(0, findModeQuery.rawQuery.length - 1)
-    updateFindModeQuery()
-    performFindInPlace()
-    showFindModeHUDForQuery()
+    updateQueryForFindMode findModeQuery.rawQuery.substring(0, findModeQuery.rawQuery.length - 1)
 
 # <esc> sends us into insert mode if possible, but <cr> does not.
 # <esc> corresponds approximately to 'nevermind, I have found it already' while <cr> means 'I want to save
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -1014,11 +1014,13 @@ window.goNext = ->
   findAndFollowRel("next") || findAndFollowLink(nextStrings)
 
 showFindModeHUDForQuery = ->
-  if (findModeQueryHasResults || findModeQuery.parsedQuery.length == 0)
+  if findModeQuery.rawQuery and (findModeQueryHasResults || findModeQuery.parsedQuery.length == 0)
     plural = if findModeQuery.matchCount == 1 then "" else "es"
     HUD.show("/" + findModeQuery.rawQuery + " (" + findModeQuery.matchCount + " Match#{plural})")
-  else
+  else if findModeQuery.rawQuery
     HUD.show("/" + findModeQuery.rawQuery + " (No Matches)")
+  else
+    HUD.show("/")
 
 getCurrentRange = ->
   selection = getSelection()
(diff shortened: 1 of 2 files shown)


smblott-github:
I think this is a no-brainer.  However, for the record, it does change the existing behaviour in that, in find mode, the arrow keys no longer scroll the page (as they did previously).  That doesn't seem a big price to pay.

smblott-github:
> it does change the existing behaviour in that, in find mode, the arrow keys no longer scroll the page

@philc, @z0rch... Do you have a view on this?  *Arrow-keys-scroll-page* is how Chrome's native find works.

z0rch:
I find it quite natural to use arrows for navigating through the history. Personally I'm completely fine without scrolling **while I type my search query**, and anyways after I hit `Enter` both `jk` and `up/down` work again. 

However I have a couple of other concerns/questions:

1. I don't like the fact that the history is global for all tabs. I think in most cases users would prefer having separate search history per tab. Would you agree?
2. Search history is preserved after I restart the browser. There is a storage limit of 50 elements, as far as I can tell. Perhaps having the way to cleanup the history would also be nice? Not sure if I would want to use it ever, it's just a thought...
3. This is just a bug, feels weird: 
    * Type `/smth<RET>` just to create a history record.
    * Type `/long`, remember the number of matches (for example 6).
    * Hit `Enter`.
    * Type `/long` again, notice the number of matches is the same (6).
    * Hit `up` and see the number of matches became 7
    * Hit `up` and `down`, now it is back to 6.
    * Hit `down`, now it is 7.

smblott-github:
> I think in most cases users would prefer having separate search history per tab. Would you agree?

I think *global-for-all-tabs* is the way to go.  Here are a couple of reasons:
- The current *repeat-last-search* is global.  It would be odd of that were global but the history was not.
- A common use case is *tweak-last-search*.  Again, since the last search is global, it would be odd if this weren't.
- The history is most likely to be useful if it's global.  The search history for any one tab is likely very small.
- That's how vim does it (at least with my set up).

> This is just a bug, feels weird:

Yes, I noticed that too.  I think it's a symptom of a long-standing, existing bug:
- On this page, `/aaa bbb`. Watch the number of matches as you go from `aaa` to `aaa<Space>`.
- Now start removing characters from the search.  You'll see exactly the same off-by-one error.

I haven't been able to figure out where this is coming from.

mrmr1993:
> long-standing, existing bug:

Doing `HUD.hide()` before the call to `updateFindModeQuery` (in `handleKeyCharForFindMode` and `handleDeleteForFindMode`) will fix that. It also clarifies why we need to call `showFindModeHUDForQuery` later in those functions, and lets us decouple the HUD from `executeFind`.

I don't have a computer, so I can't write or test any code.

smblott-github:
> Doing HUD.hide(true) before the call to updateFindModeQuery

Yes... b082b11942dac88cff4ddb22ae0bc4426cf98bf5.

smblott-github:
Merging this.  It's a simple enough change, and probably the benefits of find history outweigh the loss of being able to scroll the page (via the arrow keys) while in find mode.  We'll soon find out if I'm wrong!

philc:
Great discussion, and +1 on up and down for history entry selection.

How do you feel about clearing the history on browser restart? Otherwise,
having no way to clear it is kind of strange. When would you want the find
history to persist across browser sessions?

smblott-github:
> How do you feel about clearing the history on browser restart?

We could, @philc.  Is that the direction you want to go?

Pre this PR, the only find history we have is `n`.  The search term for `n` has never been cleared on browser restarts.  More than that, it's synced between browser installations.  It seemed natural to handle find history the same way.

On the other hand, the common use case for find history is really just repeating that search I did ten minutes ago.  Not the one from last last week.  Clearing the history (and not syncing it at all) would have no impact on that use case.

PR #1456, opened 2015-02-02
Strip unreachable code post #1413.

smblott-github:
#1413 relocated a fair amount of functionality, but did not strip the consequently unreachable code.  That's what this does.  It also tidies up `sync.coffee`, which still has some debugging capability embedded.


smblott-github:
Merging.  This should be a no-op (unless I've made a mistake...).

PR #1459, opened 2015-02-03
Restore scroll position after failed find with /.

smblott-github:
When using /, if the first few characters match, the scroll position may change.  If further characters then fail to match, we're left at an arbitrary position within the document.  When considering how find mode might work with visual mode, this becomes a problem.  The user can no longer see the text they could originally see.  See @z0rch's comments in #1441.

This restores the scroll position when search using `/` fails.  It's not obvious that this *always* the right thing to do.  However, it does fix @z0rch's concerns in #1441.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -11,6 +11,8 @@ findModeQuery = { rawQuery: "", matchCount: 0 }
 findModeQueryHasResults = false
 findModeAnchorNode = null
 findModeInitialRange = null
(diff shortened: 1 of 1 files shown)


philc:
I don't have any opinion here, other than I believe Chrome's find UX will
be reasonable and familiar to people using Chrome.

smblott-github:
On reflection, I'm just going to drop this.  I doesn't wholly address @z0rch issues in #1441, and it's not at all obvious whether it's better than or worse than the status quo.

PR #1464, opened 2015-02-07
Find history should not survive browser restart (or should it?)

smblott-github:
This follows on from #1454, and resets the find-mode history on browser restart. We keep only the most recent search (so `n` still works, as it always has).

Should we reset the find-mode history in this way?  I'm not sure.  However, I doubt resetting the history has much negative impact on any important use case.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -680,5 +680,4 @@ chrome.windows.getAll { populate: true }, (windows) ->
         (response) -> updateScrollPosition(tab, response.scrollX, response.scrollY) if response?
       chrome.tabs.sendMessage(tab.id, { name: "getScrollPosition" }, createScrollPositionHandler())
 
-# Start pulling changes from synchronized storage.
-Sync.init()
+Settings.init()
--- a/background_scripts/settings.coffee
+++ b/background_scripts/settings.coffee
@@ -4,6 +4,13 @@
 
 root = exports ? window
 root.Settings = Settings =
+  init: ->
+    # Start pulling changes from synchronized storage.
+    Sync.init()
+    # Reset findModeRawQueryList to contain only the most recent query (so "n" still works, but all earlier
+    # history is cleared).
+    @set "findModeRawQueryList", @get("findModeRawQueryList")?[0..0] or []
+
   get: (key) ->
     if (key of localStorage) then JSON.parse(localStorage[key]) else @defaults[key]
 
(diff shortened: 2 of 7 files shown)


mrmr1993:
> I doubt resetting the history has much negative impact on any important use case.

This probably isn't the place to discuss this, but it's nice to keep history

smblott-github:
> This probably isn't the place to discuss this, but it's nice to keep history

This *is* the place to discuss it.  I marginally favour keeping the history (so **not** merging this PR). In #1454, @philc wondered allowed...

> How do you feel about clearing the history on browser restart? Otherwise, having no way to clear it is kind of strange.

This PR is just an opportunity to discuss that point.  I'm guessing the common use case is *repeat the search I did ten minutes ago*, not *repeat the search I did last week*.  So, if we *do* discard the history, then I doubt we lose important use cases. Hence, I'm not too fussed either way.

(For the record, we keep at most 50 searches, so the history doesn't grow indefinitely.)

z0rch:
I'd like to take a scenario approach, i.e. if we could brainstorm an example or two, showing that **keeping** the history is painful sometimes, then we could discuss those reasons. A contrary example is obvious -- user accidentally closed the browser and reopened it immediately to continue his job.

So at this point of time, keeping the history is more useful.

mrmr1993:
> > Otherwise, having no way to clear it is kind of strange.

If this is the major concern, we could add a button to the options page that lets the user clear the find history, or store the last use date and let entries 'expire' after a certain amount of time.

>... if we could brainstorm an example or two, showing that **keeping** the history is painful sometimes ...

A user might want to loose comprimising searches (think incognito-mode-style searches). I'm not sure that we'd necessarily want to give a UI for going through and manually deleting entries, but I think we should consider clearing searches from incognito windows when the user exits incognito. Specifically, from an incognito window, I propose:
* if the search was already in history, move it to the front and
  - if it was previously marked as incognito, keep it marked as incognito
  - if it wasn't, leave it unmarked (so we don't delete a users regularly used searches just because they used it once in incognito)
* if it wasnt in the history, add it to the front and mark it as incognito
* when the user exits incognito mode, clear all the searches marked as incognito.

z0rch:
Incognito mode is a nice catch, I agree we should clear searches made in incognito mode.

smblott-github:
It looks like there's no compelling reason to clear the search history.  So let's close this.  The issue re. incognito mode is raised in #1465.

PR #1465, opened 2015-02-08
Find mode history for incognito mode

smblott-github:
Following on from #1464, this ensures that find-mode queries do not leak out of incognito mode.

The approach is to track the history separately in every tab.  Only new queries from non-incognito mode tabs are pushed to the background page (via `settings.set`), and from there they are pushed out (synchronously) to all tabs (via a new message type, `updateSettings`).  This allows all tabs (including incognito tabs) to track the find history (synchronously).  However, in addition to the global history, incognito tabs also track their own queries (and in the correct order).

Considerations:
1. This changes the long-established behaviour whereby queries in incognito tabs leak (via `n`) to all other tabs.  This might be considered a bug fix.  Or, it might be a bad UX change for some users.
2. ~~Queries in incognito windows are not shared between tabs.  So, search for [booby](http://en.wikipedia.org/wiki/Blue-footed_booby) in one incognito tab, then switch to another incognito tab, and `n` doesn't do what the user might expect.~~

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -382,7 +382,7 @@ root.updateActiveState = updateActiveState = (tabId) ->
           setBrowserActionIcon(tabId,disabledIcon)
         # Propagate the new state only if it has changed.
         if (isCurrentlyEnabled != enabled || currentPasskeys != passKeys)
-          chrome.tabs.sendMessage(tabId, { name: "setState", enabled: enabled, passKeys: passKeys })
+          chrome.tabs.sendMessage(tabId, { name: "setState", enabled: enabled, passKeys: passKeys, incognito: tab.incognito })
       else
         # We didn't get a response from the front end, so Vimium isn't running.
         setBrowserActionIcon(tabId,disabledIcon)
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -188,7 +188,7 @@ handleSettings = (args, port) ->
     value = Settings.get(args.key)
     port.postMessage({ key: args.key, value: value })
   else # operation == "set"
-    Settings.set(args.key, args.value)
+    Settings.set(args.key, args.value) unless args.incognito and args.key in [ "findModeRawQueryList" ]
     # For some settings, we propagate changes to all tabs immediately.
     # In the case of findModeRawQueryList, this allows each tab to accurately track the find-mode history.
     if args.key in [ "findModeRawQueryList" ]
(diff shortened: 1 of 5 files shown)


mrmr1993:
> Queries in incognito windows are not shared between tabs.

Also between successive URLs in the same tab. This behaviour is confusing for e.g. paginated search result pages, where the user is still mentally in the same context even though they've navigated.

smblott-github:
> Also between successive URLs in the same tab.

That's a good point.

My original "consideration 2" is fixed in 7b451aaca8d8c84aab77ac262753c6239b466791.  Your issue remains, though.

mrmr1993:
> My original "consideration 2" is fixed in 7b451aa. Your issue remains, though.

This behaviour is really weird. Any tabs/navigations after a search is performed don't get the search, but old tabs do. The user will probably not work this out themselves either, so basically they'll just have a WTF moment when search history is partially shared.

An easy way to implement this properly would be:
* having a separate incognito list and copying normal history in
* pushing changes from normal tabs to both lists
* pushing changes from incognito tabs to only the incognito list
* purging the list when no incognito tabs are left

smblott-github:
> This behaviour is really weird

Yes, I know.  As I said before, you made a good point.  I'm working on it.

smblott-github:
OK, @mrmr1993.  I think I have it now (hopefully?).  See 11609b918485ee4c3a659bee9dc0c420fb04ddcc.

Basically, 11609b918485ee4c3a659bee9dc0c420fb04ddcc puts the current history for incognito tabs in the `tabInfoMap` (for each incognito tab), and fetches it from there on start up for incognito tabs.  That way, the history goes away by itself when the last incognito tab closes.

smblott-github:
Off topic, but my experience from looking at this is that we'd be *a lot* better of using `chrome.storage`.  In fact, I might migrate this code in that direction.  It'd probably come out a lot more efficient and easier to understand.

mrmr1993:
> ... my experience from looking at this is that we'd be *a lot* better of using `chrome.storage`.

You beat me to it, I was going to comment the same. I'm having real problems following the flow of the code in this PR, and since I have no way of running the code I can't otherwise comment on the behaviour. But that could be me, not the code.

Having a copy of the list for every tab seems also like not a good idea.

smblott-github:
> You beat me to it, I was going to comment the same.

> I'm having real problems following the flow of the code...

Indeed. With `chrome.storage`, things are much better.  It eliminates the need for a lot of plumbing, and allows the implementation to just sit (almost independently) in two places.  The plumbing happens automagically.  This is the way to go as we make changes going forward.

Status:
- Maintain two lists of queries in `chrome.storage`, one for normal tabs, and one for incognito tabs.
- New queries only ever propagate from normal tabs to incognito tabs, never the other way around.
- The query list for incognito-mode tabs is removed when the last incognito tab closes. (And is always removed on startup, if it survived close down.)

smblott-github:
Merged this.  I'm pretty confident the UX is right now, and the code is considerably easier to reason about.

background_scripts/main.coffee
```diff
@@ -655,6 +655,21 @@ sendRequestHandlers =
...
+chrome.storage.local.remove "findModeRawQueryListIncognito"
+
+# Remove chrome.storage.local/findModeRawQueryListIncognito if there are no remaining incognito-mode tabs.
+# Since the common case is that there are none to begin with, we first check whether the key is set at all.
+chrome.tabs.onRemoved.addListener (tabId) ->
+  chrome.storage.local.get "findModeRawQueryListIncognito", (items) ->
+    if items.findModeRawQueryListIncognito
+      chrome.windows.getAll { populate: true }, (windows) ->
+        for window in windows
+          for tab in window.tabs
```
mrmr1993: We can query `window.incognito` directly, so we don't have to loop over all the tabs (or use `populate: true` above).

smblott-github: Excellent!  Thank you: 90afb58f3752d562b5c5683d170317cc5f15b42c.

PR #1474, opened 2015-02-11
Move the HUD to a UIComponent iframe (v2)

smblott-github:
Edit: Ooops. Hit "submit" before text was ready.

This is @mrmr1993's work from #1401 merged into `master`.

Known issues:
- The colour of the selection while find-mode is active grey, rather than orange as in `master`.
- `PostFindMode` isn't triggered correctly when the search lands in an input (this is because the input element isn't active, because the HUD has the focus).

@mrmr1993.  Could you take this for a spin and check for errors, please?


mrmr1993:
> * The colour of the selection while find-mode is active grey, rather than orange as in `master`.

The only fix (read: hack) I can think of uses
* [`Range.getClientRects`](https://developer.mozilla.org/en-US/docs/Web/API/Range.getClientRects),
* our rect library functions (to prevent the rects from overlapping), and
* elements with
  - orange background color
  - high z-index
  - some opacity that keeps the selection beneath visible
  - positioned at each of these rects.

Since this only needs a selection at calculation time, it could be useful for showing multiple matches simultaneously too.

> Could you take this for a spin and check for errors, please?

Sorry, I have no computer at the moment, so I can't really help with this.

PR #1475, opened 2015-02-12
Try to fix a problem about registerFrame, may caused by frame stack is not refreshed with frame freshing

d8660091:
Problem:
http://music.163.com/
What happens in this website is that when click a link in the main frame, "gf" would not work, because Vimium forget the frame on the bottom.

Fixing:
refresh the tab's frame stack in nextFrame function.

This might not the best solution, if you guys have better idea, please let me know.


mrmr1993:
I've opened PR #1476, which will stop us from incorrectly clearing the frames list in this case.

PR #1476, opened 2015-02-12
Retain frame data for a tab unless the top frame tells us its closing

mrmr1993:
When
* a tab's URL changes without a new page loading (eg. `history.pushState`), and
* some resources are being loaded (eg. images/iframes) when we get the `chrome.tabs.onUpdated` in the background page

then we clear our record of frames in the tab. This PR changes our behaviour so that we only clear the frames for a tab when the top-level frame says it's going away.

This fixes the issue mentioned in PR #1475.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -324,7 +324,7 @@ selectTab = (callback, direction) ->
       selectionChangedHandlers.push(callback)
       chrome.tabs.update(toSelect.id, { selected: true })))
 
-updateOpenTabs = (tab) ->
+updateOpenTabs = (tab, deleteFrames = false) ->
   # Chrome might reuse the tab ID of a recently removed tab.
   if tabInfoMap[tab.id]?.deletor
     clearTimeout tabInfoMap[tab.id].deletor
@@ -336,7 +336,7 @@ updateOpenTabs = (tab) ->
     scrollY: null
     deletor: null
   # Frames are recreated on refresh
-  delete frameIdsForTab[tab.id]
+  delete frameIdsForTab[tab.id] if deleteFrames
 
 setBrowserActionIcon = (tabId,path) ->
   chrome.browserAction.setIcon({ tabId: tabId, path: path })
@@ -619,7 +619,7 @@ unregisterFrame = (request, sender) ->
   tabId = sender.tab.id
   if frameIdsForTab[tabId]?
     if request.tab_is_closing
-      updateOpenTabs sender.tab
+      updateOpenTabs sender.tab, true
     else
       frameIdsForTab[tabId] = frameIdsForTab[tabId].filter (id) -> id != request.frameId
 


smblott-github:
@mrmr1993.  Is this the best way to do this?  Perhaps I have this wrong, but would the following work (on the background page, only)?
- Have `unRegisterFrame` **never** call `updateOpenTabs`.
- `unRegisterFrame` **only** maintains `frameIdsForTab`.
- Leave it to `chrome.tabs.onUpdated` to call `updateOpenTabs` to delete `frameIdsForTab[tab.id]`.

Then, `registerFrame` and `unRegisterFrame` are only responsible for maintaining `frameIdsForTab`, and tab removal is a separate matter.  If it's correct (?), then it simplifies the logic.

mrmr1993:
Yes, that's fine.

I think we should eventually store the ports for each frame with the frame description. Then, we could have a control port which does the job of `registerFrame` and the `DomReady` port, and delete the frame description when it closes, to do away with `unregisterFrame` too. (Also could let us make the decision about whether to focus a frame in the background page, which is clearer.) But we don't need to do that right now.

d8660091:
Assume that we have a frame for login, we input the username passeword to that frame, then the frame disappear by itself. Will this PR handle this kind of situation?

mrmr1993:
> Will this PR handle this kind of situation?

The only situations that will have changed are those where the URL has changed for the tab, but the tab itself does not redirect. Does that happen?

mrmr1993:
Sorry @smblott-github, I was eating when I replied your comment, I ended up reading what i wanted to see, not what was actually there.

> * Leave it to `chrome.tabs.onUpdated` to call `updateOpenTabs` to delete `frameIdsForTab[tab.id]`.

We can't do this, since we can't tell whether the change is destroying frames or not at `onUpdated`. This was exactly the cause of the issue (and the old frame issue where we forgot all frames at a redirect). `onUpdate` should update the tab data, but do nothing with frames.

When the `frame` option lands in stable for [`chrome.tabs.sendMessage`](https://developer.chrome.com/extensions/tabs#method-sendMessage), we can stop worrying about all of this and just use [`chrome.webNavigation`](https://developer.chrome.com/extensions/webNavigation) to keep us up to date.

d8660091:
>The only situations that will have changed are those where the URL has changed for the tab, but the tab itself does not redirect. Does that happen?
It doesn't happen I think. From my limited understanding of the code, I thougt the frameId would only be deleted when the tab is closing. But I realized later that I am wrong, the frameId is deleted when "page unload". Sorry, you are right.

mrmr1993:
@smblott-github since this is a small, easy change bugfix, can this be merges? Or is there something else I should do to it first?

smblott-github:
Yes, @mrmr1993.  I'll merge this soon.  Just let me have one more think about it later.

philc:
Nice work guys.

PR #1480, opened 2015-02-12
Grab back focus when page auto-focuses an input (optionally)

### LINKED ISSUE #523: Add a preference to prevent pages from stealing Vimium's focus when they first load

smblott-github:
Some pages (like [Google](https://www.google.ie/?gws_rd=ssl) and [Bing](http://www.bing.com/)) automatically focus an input on load.  This PR adds an option (off by default) to grab back the focus in such cases.  We only grab back the focus if the user has not yet interacted with the page (so, until the first `keydown` or `mousedown`).

This is an alternative to #1282.
Fixes #523.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -96,6 +96,31 @@ settings =
 #
 frameId = Math.floor(Math.random()*999999999)
 
+# If an input grabs the focus before the user has interacted with the page, then grab it back.
+class GrabBackFocus extends Mode
+  constructor: (@insertMode) ->
+    return if @shouldBeDeactivated()
+    super name: "grab-focus", keydown: => @alwaysContinueBubbling => @exit()
+
+    @push
+      _name: "grab-focus-handlers"
+      mousedown: => @alwaysContinueBubbling => @exit()
+      focus: (event) => @grabBackFocus event.target
+
+    # An input may already be focused. If so, grab back the focus.
+    @grabBackFocus document.activeElement if document.activeElement
+
+  grabBackFocus: (element) ->
+    if DomUtils.isEditable(element) and not @shouldBeDeactivated()
+      element.blur()
+      @insertMode.exit null, element
+      return @suppressEvent
+    @exit() if @shouldBeDeactivated()
+    @continueBubbling
+
+  shouldBeDeactivated: ->
+    false and settings.isLoaded and not settings.get "grabBackFocus"
+
 # Only exported for tests.
 window.initializeModes = ->
   class NormalMode extends Mode
@@ -114,6 +139,7 @@ window.initializeModes = ->
   new NormalMode
   new PassKeysMode
   new InsertMode permanent: true
+  new GrabBackFocus InsertMode.permanentInstance
 
 #
 # Complete initialization work that sould be done prior to DOMReady.
@@ -179,7 +205,7 @@ window.initializeWhenEnabled = ->
   unless installedListeners
     # Key event handlers fire on window before they do on document. Prefer window for key events so the page
     # can't set handlers to grab the keys before us.
-    for type in ["keydown", "keypress", "keyup", "click", "focus", "blur"]
+    for type in [ "keydown", "keypress", "keyup", "click", "focus", "blur", "mousedown" ]
       do (type) -> installListener window, type, (event) -> handlerStack.bubbleEvent type, event
     installListener document, "DOMActivate", (event) -> handlerStack.bubbleEvent 'DOMActivate', event
     installedListeners = true
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -98,28 +98,27 @@ frameId = Math.floor(Math.random()*999999999)
 
 # If an input grabs the focus before the user has interacted with the page, then grab it back.
 class GrabBackFocus extends Mode
-  constructor: (@insertMode) ->
-    return if @shouldBeDeactivated()
-    super name: "grab-focus", keydown: => @alwaysContinueBubbling => @exit()
+  constructor: ->
+    super
+      name: "grab-back-focus"
+      keydown: => @alwaysContinueBubbling => @exit()
 
     @push
-      _name: "grab-focus-handlers"
+      _name: "grab-back-focus-mousedown"
       mousedown: => @alwaysContinueBubbling => @exit()
-      focus: (event) => @grabBackFocus event.target
 
-    # An input may already be focused. If so, grab back the focus.
-    @grabBackFocus document.activeElement if document.activeElement
+    chrome.storage.sync.get "grabBackfocus", (items) =>
+      return @exit() unless items.grabBackfocus and not chrome.runtime.lastError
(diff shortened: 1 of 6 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -96,6 +96,30 @@ settings =
...
+  constructor: ->
+    super
+      name: "grab-back-focus"
+      keydown: => @alwaysContinueBubbling => @exit()
+
+    @push
+      _name: "grab-back-focus-mousedown"
+      mousedown: => @alwaysContinueBubbling => @exit()
+
+    chrome.storage.sync.get "grabBackfocus", (items) =>
```
mrmr1993: It seems like bad and confusing practise to set this with our settings library but read it with the underlying mechanism.

smblott-github: Yes.  But it gets around a race condition.

An alternative would be to add extra infrastructure (`postUpdateHook`, `onLoaded`) to `settings` in the content script.  But I'd prefer to see everything moved (over time) to using `chrome.storage`, with caching only where we can't do things asynchronously.

mrmr1993: If those are things we need the settings library to do, we should do them there, rather than introducing implementation details throughout the code that we'll forget to re-integrate later.

I agree that the frontend should use `chrome.storage` directly; I started work on using the backend settings library for the frontend too in #1290, and all those things (except no caching) come for free.

smblott-github: > If those are things we need the settings library to do, we should do them there

I agree.  I was thinking about whether to do it in this PR, or wait.  I'd also like to have a `cachedSetting` object to handle the synchronous case cleanly and consistently.

smblott-github:
@philc.  In #523, you commented that you were worried about the UX (in particular vis-a-vis vimperator).  You might want to check that aspect of this PR.  It's clean on all of the sites I've tried (no flicker, no visible focus/blur).  So I think it's good (modulo the issue @mrmr1993 and I discussed re. settings).  I'll merge this, unless we find a problem.

mrmr1993:
@smblott-github can we not leave this, open an issue for the settings, and mark this blocked on that? That way we won't merge this, deprioritise the settings issue, and leave the bare `chrome.storage` call in the codebase forever.

mrmr1993:
@smblott-github Also, if we want all the settings available at the moment the first content script starts, we can do

```javascript
chrome.webNavigation.onBeforeNavigation.addEventListener(function(details){
  chrome.tabs.executeScript(details.tabId, {
    "code": " settingsPreload = " + jsonEncodeSettings(),
    "allFrames": true,
    " runAt": "document_start"
  });
});
```

from the background page. That way, we can use `settingsPreload` until our normal method of loading settings is ready, and so have no race condition concerns at all.

smblott-github:
After toying with various possible API changes for settings, I think the simplest and cleanest thing to do is to just use `chrome.storage.local`.  It does exactly what we need.

Here's something which works...

1. d478aa5c56acba34b7fb83a2c49adc48baf47889.  Store the actual current value for each setting (default value or otherwise) in `chrome.storage.local`.  Anybody who wants it (and can tolerate being asynchronous) can grab it from there.  Keep all of the existing infrastructure: it works, so there's little value in messing with it.

1. c5ad4790f42cbdfc62f0914fc0dc0ee250ab4033. Modify `settings` in the front end (which provides synchronous access to settings) to pre-load its settings from `chrome.storage.local`.  This simplifies the code and eliminates an astonishing amount of unnecessary messaging.

The effect:

- If you need synchronous access to settings then use (unchanged):
  * `settings.get` in the front end, or
  * `Settings.get` on the background page.

- If you can tolerate (or want) asynchronous access to settings:
  * use `chrome.storage.local.get` (anywhere).

mrmr1993:
Don't want to unnecessarily push my own code, but is there a reason we can't use #1290? It has the same properties, stops us from having to duplicate the settings in `chrome.storage.local` and `chrome.storage.sync`, and gives us one unified settings library.

Obviously it needs finishing and a rebase, but I'm happy to do that when i get home after the Valentines day weekend.

smblott-github:
> ....is there a reason we can't use #1290?

Nothing fundamental.  There are usually several ways to do anything.  The two commits above are arguably simpler and, if the cost of that simplicity is a little extra storage, then it's a cost well worth paying.

We often don't really need synchronous access to settings.  So moving towards the `chrome.storage` approach is a positive for me.

mrmr1993:
> The two commits above are arguably simpler

They are, but #1290 or a similar approach would simplify the *code* more. I think it's worth putting in slightly more effort now and having a simpler codebase for the future. But that's just my $0.02.

smblott-github:
> similar approach would simplify the code more.

In what way?

mrmr1993:
> In what way?

With your changes, the two settings implementations work in a very similar way. Making them into 1 implementation gives us half the code to maintain/test, and consistency everywhere.

smblott-github:
> With your changes, the two settings implementations work in a very similar way. Making them into 1 implementation...

Or, you could just think of them as one implementation as is.  Just with changes propagated through two channels.  The amount of code with the two commits above is less, as is the complexity.  That has to be a win for maintainability.

(Slightly off topic.  But I've also been wondering about doing away with the `postUpdateHooks`.  We only have them because we weren't using `chrome.storage` to start with.  Also, this business about not storing settings with the default value is bogus and is hurting us.  If the default value changes, then we can still migrate.  And, at that point, you don't need `Settings` at all, apart from it being a list of the default values.  If you follow this line of thinking then you: use `chrome.storage.onChanged` in place of update hooks, do away with `Settings` almost entirely (just keep defaults), use `settings` from the frontend in the back end too (for synchronous access), and then just have a list of which settings should be synced and which not.  So, I think it can all be simpler still. Oh.  And you don't need any of this nonsense with JSON.  This is all just engineering debt.)

smblott-github:
...... And you can get rid of `sync.coffee` entirely.  Just have a listener which listens for changes in `storage.sync` and `storage.local` and copies them across, if appropriate.

mrmr1993:
> you could just think of them as one implementation as is.

True, but it is still 2 implementations. Just that one feeds the other. I think that, if there can easily be one implementation, there probably should be.

I agree with your whole bracketed section. If we're changing the settings anyway, I think we should go that way.

smblott-github:
> True, but it is still 2 implementations.

Ah.  I misunderstood, @mrmr1993.  I agree.  One is better than two.

I'm going to move this discussion over to #1290.

smblott-github:
I'm merging this.  It seems to be working nicely, and I'm not seeing any odd behaviour (failures or flickers), or race conditions.

FTR: 05f229201a05101ab4947bd436ec02d8864392f9 removes the unsatisfactory reach into the guts of the settings implementation (and should remove @mrmr1993's concern).

PR #1481, opened 2015-02-15
Added Markup code tags to commands in Readme.md

rkho:
Changed references to vimiim commands that were either in single-quotes or not otherwise distinguished easily to the reader as a command to be wrapped around backtick quotes so they will stand out. Here is a list of all changes made:

Line 92, the reference to ‘5t’ used to be in single quotes, is now `5t`. 
Line 174, updated gU to `gU`. 
Line 175, updated yt to `yt`. 
Line 176, updated W to `W`. 
Line 177, updated mX to `mX` and X to `X`. 
Line 228, updated “o” to `o`. 
Line 240, updated g0 to `g0` and g$ to `g$`. 
Line 241, updated p/P to `p`/`P`. 
Line 249, updated Tab to `Tab` and shift-tab to `shift-tab`. 
Line 318, updated ‘u’ to `u`. 
Line 320, updated ‘unmapAll’ to `unmapAll`. 
Line 327, updated “?” to `?`. 
Line 335, updated Ctrl+D to `Ctrl+D` and Ctrl+U to `Ctrl+U`. 
Line 346, updated J to `J` and K to `K`. 
Line 353, updated ‘gt’ to `gt` and ‘gT’ to `gT`. 
Line 354, updated ‘yy’ to `yy`. 
Line 356, updated Shift+F to `Shift+F`. 
Line 357, updated ESC to `ESC` and ‘g’ to `g`. 

No other changes have been made to this file, this update is intended to strictly improve readability.

PR #1482, opened 2015-02-15
Updated README.md readability -- missed two lines

rkho:
Missed two lines from my original pull request (https://github.com/philc/vimium/pull/1481). Basically:
Changed references to vimiim commands that were either in single-quotes or not otherwise distinguished easily to the reader as a command to be wrapped around backtick quotes so they will stand out. Here is a list of all changes made:

Line 215, 'b' updated to `b`
Line 240, 'g0' updated to `g0` and 'g$' updated to `g$`

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ Release Notes
 
 1.36 (2012-07-07)
 
-- 'b' brings up a bookmark-only Vomnibar.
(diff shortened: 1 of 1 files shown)


smblott-github:
Ta!

PR #1486, opened 2015-02-15
Move "show-advanced-options" to a fixed-position button.

smblott-github:
What about replacing this...

sss

With this...

snapshot

Why? The show-advanced options link/button should be in a fixed place.  On my Chromebook, I have to scroll to see it.

### DIFF
--- a/pages/options.coffee
+++ b/pages/options.coffee
@@ -208,14 +208,13 @@ initOptionsPage = ->
       (event) ->
         if advancedMode
           $("advancedOptions").style.display = "none"
-          $("advancedOptionsLink").innerHTML = "Show advanced options&hellip;"
+          $("advancedOptionsButton").innerHTML = "Show Advanced Options"
         else
           $("advancedOptions").style.display = "table-row-group"
-          $("advancedOptionsLink").innerHTML = "Hide advanced options"
+          $("advancedOptionsButton").innerHTML = "Hide Advanced Options"
         advancedMode = !advancedMode
+        $("advancedOptionsButton").blur()
         event.preventDefault()
-        # Prevent the "advanced options" link from retaining the focus.
-        document.activeElement.blur()
 
   activateHelpDialog = ->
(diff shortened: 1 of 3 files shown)


mrmr1993:
LGTM.

It might be clearer to change the button to a checkbox, in keeping with Chrome UI.

smblott-github:
> It might be clearer to change the button to a checkbox

You could.  But I can't think of a way to lay it out such that the look is consistent with the rest of the footer.  Here's something ugly...

snapshot

(Also, a checkbox is harder to hit with the mouse.)

mrmr1993:
You can make the text clickable too by wrapping both the checkbox and the input in a `<label>` tag, or by wrapping the text in a `<label for="checkbox-id">`.

If the checkbox was aligned with the text, and both were aligned with the top line of text to the left, I think that would look okay. I would prefer the checkbox on the left (á la Chrome's settings) but I'm not that bothered either way.

smblott-github:
@philc.  Do you have a view on this?  I dislike the current "Show advanced options..." toggle.  It moves around, and you can have to scroll to find it.  Another option would be to remove the fold entirely.  I'm guessing pretty much every new user has a look below the fold anyway.

Or, put another way, if we didn't already have a fold, would we wouldn't introduce one in its current form?

smblott-github:
@philc.  I'd like to merge this, but would prefer to have your input.  See the first post for details.

smblott-github:
Ping @philc.  I'd like to merge this.
Do you have a view?

smblott-github:
(Better still, consider getting rid of the fold entirely.  Here are three reasons:
- It's ugly.
- Most new users will take a peak down there anyway.
- You can't expand the *custom-search-engine* input vertically (on my laptop) without opening the fold anyway, which is pretty bad.)

smblott-github:
Merging.  @philc, let me know if you think this needs tweaking.

PR #1490, opened 2015-02-18
Fix link hints for html5 inputs

### LINKED ISSUE #1489: Link hints no longer work correctly with some HTML5 input types
The link hint keydown listener errors out with some input types, and leaves the user stuck with link hints showing but not responding to the keyboard. This makes the keyboard unusable on the page until refresh.

This can be easily reproduced with `master` [here](http://html5doctor.com/demos/forms/forms-example.html), or by focusing the "scroll step size" input on our own options page.

mrmr1993:
This PR
* catches the error thrown by input elements which do not support `selectionStart`/`selectionEnd`,
* removes redundant checks in `DomUtils.isSelected`,
* changes the behaviour of `DomUtils.isSelected` to match its description, and
* adds tests for activating html5 input types with link hints, so this doesn't happen again.

This fixes #1489.

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -167,7 +167,8 @@ DomUtils =
       node = node.parentNode
     false
 
-  # True if element contains the active selection range.
+  # True if element contains the active selection range, or if the element does not support its selection
+  # being accessed.
   isSelected: (element) ->
     if element.isContentEditable
       node = document.getSelection()?.anchorNode
@@ -175,7 +176,13 @@ DomUtils =
     else
       # Note.  This makes the wrong decision if the user has placed the caret at the start of element.  We
       # cannot distinguish that case from the user having made no selection.
-      element.selectionStart? and element.selectionEnd? and element.selectionEnd != 0
+      try
+        element.selectionEnd != 0
+      catch
+        # This input element doesn't support selectionStart/selectionEnd.
+        # NOTE(mrmr1993): We choose true here because it does the right thing everywhere in the code. I am
+        # not certain that this is necessarily what we should do.
+        true
 
   simulateSelect: (element) ->
     # If element is already active, then we don't move the selection.  However, we also won't get a new focus
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -191,6 +191,18 @@ DomUtils =
       handlerStack.bubbleEvent "click", target: element
     else
       element.focus()
+        # If the cursor is at the start of the element's contents, send it to the end. Motivation:
+        # * the end is a more useful place to focus than the start,
+        # * we've been moving the cursor to the end for quite some time now,
+        # * this way preserves the last used position (except when it's at the beginning), so the user can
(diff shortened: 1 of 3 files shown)


lib/dom_utils.coffee
```diff
@@ -167,15 +167,20 @@ DomUtils =
...
     if element.isContentEditable
-      node = document.getSelection()?.anchorNode
+      node = document.getSelection().anchorNode
       node and @isDOMDescendant element, node
     else
-      # Note.  This makes the wrong decision if the user has placed the caret at the start of element.  We
-      # cannot distinguish that case from the user having made no selection.
-      element.selectionStart? and element.selectionEnd? and element.selectionEnd != 0
+      try
+        element.selectionStart != element.selectionEnd
```
smblott-github: @mrmr1993.  This changes the current behaviour if the selection is a caret/cursor.  E.g. go [here](http://html5doctor.com/demos/forms/forms-example.html), put some text in the billing address like this:

snapshot

(i.e. with the cursor in the middle of the text), blur, then return via link hints.  The cursor is now at the end.  It also means the cursor ends up in different places depending upon whether we enter the input with `gi` or link hints, which is odd.

Why not just wrap the existing test with a `try`/`catch`?

smblott-github:
Off topic.  I'm wondering about this business of setting the selection at the end of the input.  There are obvious cases where it's the wrong thing to do:
- In a scrollable text area, the cursor can end up scrolled out of view (haven't tested this).
- In GMail, or the like, the cursor ends up after quoted messages, which is a long way from where the user is likely to want to edit.

Might it be an idea to restrict the *move-cursor-to-end* feature to single-line inputs only?  In other cases, just let Chrome do its thing.

lib/dom_utils.coffee
mrmr1993: I liked that feature too, but that's not what the comment above the function said it was doing. This function needs a new name and description if that's what it should be doing.

mrmr1993:
> Might it be an idea to restrict the *move-cursor-to-end* feature to single-line inputs only?

Agreed, that sounds preferable.

If we mark inputs we've visited before (eg. `input.vimiumVisitedBefore = true`) then we end up moving the cursor only in the cases you intended. Can you think of any reasons not to do it this way?

mrmr1993:
> I'm not a big fan of adding attributes to elements.

I was proposing a property rather than an attribute, so it wouldn't be accessible from outside our javascript scope. We already do the same with link hints.

> Absent a strong use case, I'd suggest dropping the feature.

So we can just drop the `try`, `catch` block and always return `true` if it's not a `contentEditable`? Sounds good to me.

smblott-github:
> So we can just drop the try, catch block and always return true if it's not a contentEditable?

Or just drop the feature entirely.

mrmr1993:
I've updated this so the behaviour is exactly the same as it is in `master`, except it doesn't crash for any input types and the function names/descriptions are accurate.

smblott-github:
Merged.  Thanks, @mrmr1993.

Issue #1453, opened 2015-01-31
Should exclusion rules apply per tab or per frame?


z0rch:
Use [this page](http://jsfiddle.net/v0fznbrf/embedded/result/) as a playground.

* Open playground, make sure no passkeys are defined.
* Hit `t` to make sure that vimium creates a new tab.
* Define `t` as a passkey.
* Hit `t` and ensure that `t` is now passed to the page.
* Reload the playground page.
* Hit `t` again.

EXPECTED: `t` is still a passkey, so it is passed to the page.  
ACTUAL: `t` creates a new tab.

* Modify the passkeys, for example add another character there.
* Now `t` is passed to the playground page again, until the next refresh.

------------------------

I wonder if you can reproduce this, or something is wrong with my setup. I think it is used feature, that's why I'm concerned that nobody reported it yet.

Windows and Linux, `philc/master`.

smblott-github:
Nice.  It's not actually a passkeys issue, you get the same effect in 1.42 setting the page as an excluded site.  The exclusion rule applies to `jsfiddle.net`, but the page contains an iframe from `fiddle.jshell.net`.

I'd say this is actually the correct behaviour, and what's wrong is that we're proposing the wrong pattern in the page popup.

z0rch:
Thanks for looking into it! 

So this actually uncovers a bigger problem, how to treat multiple iframes with our passkey popup. I'm going to rename this issue then.

As a user, I never know how many iframes are in the current page. But if I want to create a passkey, I probably expect it to work on the entire page, no matter if I'm inside an iframe or outside.

Can you think of an example, when we would like to be more granular, and define different sets of passkeys for different iframes inside a single page?

smblott-github:
> Can you think of an example, when we would like to be more granular, and define different sets of passkeys for different iframes inside a single page?

Yes.  The hangouts frame on Google's Inbox.  Within the main frame, `j` and `k` are passkeys (for me).  If I select the hangouts frame, `j` and `k` scroll. This seems to be the right behaviour, in this particular case.

That being said, this "feature" was introduced (unintentionally) in #1413.  I'll fix it.

z0rch:
I'm curious which approach you are going to take. List all possible frames in the popup? Then it must be easy to understand, which line in popup corresponds to which part of the page.

By the way, you may recall that Google Inbox has ~20 frames, and only 2-3 of them are actually visible. We have some extra logic for `gf` to skip invisible frames, it may be interesting to reuse it.

[...]

PR #1495, opened 2015-02-19
Remove the setState function and rework the code for setting the browser icon

### LINKED ISSUE #1494: setState is not called for pages which are enabled and have no passkeys, doesn't respect frames
The background page messages all frames and

* asks them whether they are enabled, and what passkeys they have (`getActiveState`)
* calculates whether the tab's url should be enabled, and whether it should have passkeys
* messages the frame back if it's current enabled state/passkeys don't match what was calculated (`setState`)

Consequences:
* this request tells the tab whether it is incognito (so only tabs where vimium is disabled/has passkeys know if they're incognito)
* the listener for this request triggers the new `FindModeHistory.init()` function, so some (a lot) of the time `FindModeHistory` is uninitialised.
* ~~frames get the enabled/passkeys state of the top level url, rather than for the url of their frame, which makes very little sense, and is confusing for users.~~ The behaviour is more confusing than I thought: if one frame in the tab changes, then passkeys for all other frames are reset to the passkeys for the top frame. There's also a nice race condition here `checkIfEnabledForUrl`.

mrmr1993:
This PR fixes #1494 so that our behaviour is consistent and as the user would expect (ie. an URL matching an exclusion/passkey rule has that rule enforced regardless of whether it is in a frame or not). More importantly, this fixes the issue with `FindModeHistory.init` sometimes not being called, which broke the extension after searches.

This also reverted a couple of commits which were attempting to improve the UX of the browser action icon, since they weren't compatible with these changes, and (I think) this is more important.

@smblott-github can you take a look at this? The `FindModeHistory` bug keeps messing things up while I'm trying to test my newest version of #1490.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -77,11 +77,12 @@ getCurrentTabUrl = (request, sender) -> sender.tab.url
 # Checks the user's preferences in local storage to determine if Vimium is enabled for the given URL, and
 # whether any keys should be passed through to the underlying page.
 #
-root.isEnabledForUrl = isEnabledForUrl = (request) ->
+root.isEnabledForUrl = isEnabledForUrl = (request, sender) ->
   rule = Exclusions.getRule(request.url)
   {
     isEnabledForUrl: not rule or rule.passKeys
     passKeys: rule?.passKeys or ""
+    incognito: sender?.tab.incognito
   }
 
 # Retrieves the help dialog HTML template from a file, and populates it with the latest keybindings.
@@ -371,8 +372,7 @@ root.updateActiveState = updateActiveState = (tabId) ->
     setBadge badge: ""
     chrome.tabs.sendMessage tabId, { name: "getActiveState" }, (response) ->
       if response
-        isCurrentlyEnabled = response.enabled
-        currentPasskeys = response.passKeys
+        # The top frame of the page is responding.
         config = isEnabledForUrl({url: tab.url})
         enabled = config.isEnabledForUrl
         passKeys = config.passKeys
@@ -382,10 +382,6 @@ root.updateActiveState = updateActiveState = (tabId) ->
           setBrowserActionIcon(tabId,enabledIcon)
         else
           setBrowserActionIcon(tabId,disabledIcon)
-        # Propagate the new state only if it has changed.
-        if (isCurrentlyEnabled != enabled || currentPasskeys != passKeys)
-          chrome.tabs.sendMessage(tabId, { name: "setState", enabled: enabled, passKeys: passKeys, incognito: tab.incognito })
-
 
 handleUpdateScrollPosition = (request, sender) ->
   updateScrollPosition(sender.tab, request.scrollX, request.scrollY)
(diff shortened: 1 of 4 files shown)


smblott-github:
Yes, I'll look at it soon.

Did you consider moving *all* of the *check-if-enabled*, *check-what-passkeys-are-active* logic to the front end?

mrmr1993:
I've just pushed an update that makes options and the popup use the new mechanism.

> Did you consider moving *all* of the *check-if-enabled*, *check-what-passkeys-are-active* logic to the front end?

I did, but I'd like to have the new settings library in place before doing so. It's on my list.

smblott-github:
> I'd like to have the new settings library in place before doing so

Yes.  That would make this better.  I had the same thought.  I'll look at this over the next few days.  Quite busy.

smblott-github:
@mrmr1993.  See 055090e25d9865304f200aeb4fad94d8baebb795 as a starting point.  All it does is fix the trivial things (ensure find-mode history and the incognito flag are initialised correctly).  (Also, note, `passKeys` is always initialised to *something* in `master`, but see below.)

At this point, the core problem here is #1453: should exclusion rules apply per tab or per frame?

This PR implements per-tab (and also fixes a second race condition vis-a-vis which frame responds to the `getActiveState` message from the background page).

However, there's also a pretty good case for the per-frame behaviour.  For example, on Google Inbox, per-frame behaviour allows the user to have `j` and `k` as passkeys in the main frame, but use them for scrolling in the hangouts frame.  That makes sense (because `j` and `k` are otherwise useless in the hangouts frame).

In `master`, you sometimes get per-tab behaviour, and sometimes per-frame.  Which is clearly broken.  And it's been like that vis-a-vis `isEnabledForURL` forever.

In the past, you've used the phrase "lying to the user" to describe this type of situation, @mrmr1993.  In this case, *both* choices are lying to the user.  On the one hand, the only URL the user sees is the one in the address bar, so we're lying to the user if we don't implement per-tab exclusion rules.  On the other, the user may have specifically written a rule for the URL of the frame (e.g. `talkgadget.google.com`).  And we should respect that too.

I marginally favour per-frame exclusion rules, and fixing the icon and the population of suggested exclusion rules in the popup to reflect the frame's URL, not the tab's.

Separately, there's another bug in `master`.  Since only one frame's response to a `getActiveState` request is handled, only that frame receives any updated state from the options page.

mrmr1993:
> should exclusion rules apply per tab or per frame?

I think per-frame. Since iframes allow pages to embed whole other websites, it's confusing when passkeys don't work the same on the same page just because of context (frame vs. non-frame). Embedded youtube videos are an easy example of this.

> the core problem here is #1453

I've proposed a solution to that issue over there.

> In this case, *both* choices are lying to the user.

I disagree; we're doing exactly what we said we would in the per-frame version, and we're lying to the user if we don't. We *appear* to be lying in the per-frame version because the user isn't getting the complete picture from the browser. This is partially addressed by my suggestion in #1453, and changing the browser icon to represent the passkey state for the active frame frame, as it's focused, would improve the situation more.

mrmr1993:
> Since only one frame's response to a `getActiveState` request is handled, only that frame receives any updated state from the options page.

I've just looked at the code and I can't see where this is happening. The handler for `getActiveState` doesn't seem to filter responses by frame, so it should be acting on the responses for all of the frames.

smblott-github:
> I've just looked at the code and I can't see where this is happening.

Try 96f4fae76394e37d9c1324aebe58e5af8f60069e in a page with multiple frames.  The "receive" is logged just once for every "send".

smblott-github:
Closing.  See #1534.

Issue #1496, opened 2015-02-19
Add an option for displaying the tab history


tarek2:
It would be nice if vimium has the ability for displaying the history for the tab in a list to be activated by a keyboard shortcut.

smblott-github:
Not sure if we want another command/option, but...


I we were to do this, it might be natural as a vomnibar mode.

mrmr1993:
Chrome doesn't currently have an API for this, but you could open/find an issue on the Chromium bug tracker to propose one. It's a pity because I'd really like to see this feature.

If you're interested in a challenge, it looks like you could make a decent attempt at hacking together such an API by

* associating the relevant `id` from `chrome.history` with each event in `chrome.webNavigation.onCommitted`, and
* keeping the tab history synced with the appropriate tab `id` by watching `chrome.webNavigation.onTabReplaced`.

From that point, it should be simple to hook into the Vomnibar and show the history there.

> Not sure if we want another command/option

This seems like a good, useful feature. If this is really a terrible issue, then we could add it onto the history navigation commands (long hold for history &mdash; but which would require key handling in the frontend), or lose some low-value commands like url editing and pin tab instead.

Personally, I think we should be careful about adding new commands which don't add much value, but shouldn't not add good ones because we've hit some arbitrary boundary of "too many".

1995eaton:
If I'm understanding this issue correctly, the requested feature is equivalent to the list of items shown when you right click the back button in the toolbar? If so, I ended up using `chrome.tabs.onUpdated` when I added the `:tabhistory` command to cVim (https://github.com/1995eaton/chromium-vim/blob/master/background_scripts/main.js#L52).

mrmr1993:
How do you deal with history like

```
Facebook home -> a Facebook profile -> Facebook home -> a Facebook event -> that first Facebook profile
```

and user actions on them? I can't see where/how you distinguish back/forward from the Facebook event, or how an original navigation is considered different from a back/forward.

It seems like your history could diverge from the tab's actual history quite quickly if you use back/forward a lot.

1995eaton:
I only keep track of the history states that aren't duplicates. Rather than using the `history` object to navigate back and forth, I have the tab's history URLs opened in a new tab. I wanted to process the history states in a similar fashion to Chrome, but like you said, there doesn't appear to be an API to do this. When a history state is pushed/replaced, I simply add the URL to the `TabHistory` set if it isn't already in it.

mrmr1993:
[...]

PR #1505, opened 2015-02-23
Ensure incognito-mode flag and find-mode-history are initialised.

smblott-github:
This fixes the most basic race-condition issues around the initialisation of the incognito-mode flag and find-mode history from #1494.  It does the small and obvious part of #1495 (that part which doesn't really require discussion -- I think?).

@mrmr1993.  Should we merge this and take some time to think about what to do with #1496 and #1453?


mrmr1993:
LGTM.

Issue #1506, opened 2015-02-24
Once Vomnibar loses focus, Vimium practically stops working -- is this deliberate or a bug?


saheel1115:
I am currently using the `b49d1bf` version of `master`.

Step 1: Vomnibar.activate (or Vomnibar.activateInNewTab)
Step 2: Click anywhere on the page which takes focus away from Vomnibar

*After this, all Vimium keybindings stop working.*

To make Vimium start functioning again, I have to either:
1. Click on the Vomnibar using a mouse to give back the focus to it, or
2. Switch to some other tab or other application window, and then come back to this tab. This also gives back the focus to Vomnibar and everything resumes working (for the time being anyway).

I feel the expected behavior after Step 2 above should be that the Vomnibar disappears completely. If not that, at least Vimium should not stop working (!) in which case I could do Vomnibar.activate again and get the Vomnibar to work again.

mrmr1993:
This is partially fixed in the `master` branch here; Vimium remains functional after blurring the Vomnibar. It seems like hiding and disabling the Vomnibar when it's blurred is the logical thing to do, and an easy change.

(@smblott-github, this sounds like an easy fix for a part of the problem we discussed over email, too.)

mrmr1993:
Had another look, it's because the Vomnibar is blurred but its `iframe` is still focused.

Adding the following to `VomnibarUI::initDom` (in `pages/vomnibar.coffee`) should do the trick:

```coffee
@input.addEventListener "blur", => @input.focus()
```

saheel1115:
It works, great! Basically, now, even on clicking elsewhere the Vomnibar doesn't lose focus (gets 'blurred'). If there's a better patch in the future, and if you happen to remember this issue, can you please add a comment here?

saheel1115:
I am closing this issue for now.

mrmr1993:
@saheel1115 Can you re-open this? I'll throw together a proper fix when I get home from the pub/tomorrow, and I'd like to be able to associate it with this issue. This is related to something Steve and I were discussing anyway, so I don't think there should be any trouble in getting the fix in the codebase.

smblott-github:
#1511 was intended to fix this.  It hides the vomnibar on blur.  It looks like this approach isn't going anywhere, because we get blur events when we change tab.  We need another approach... (?).

smblott-github:
> If there's a better patch in the future, and if you happen to remember this issue, can you please add a comment here?

[...]

PR #1511, opened 2015-02-28
Hide the vomnibar on blur

smblott-github:
@mrmr1993:  Could you take a look at this, please?  It's intended to implement your suggestion from #1506.

Considerations:

- ~~The tests fail.  @mrmr1993: I don't understand these tests.  Would you be able to take this and fix it?  Or submit an alternative?~~

- This removes the ability to select text with the mouse while the vomnibar is active.  Is that such a loss?  Probably not... The vomnibar already obscures the "focus" of the page.

- ~~Edit.  And, clicking anywhere on the vomnibar closes the vomnibar.  That's bad UX.~~

In #1506, @mrmr1993 wrote:

> (@smblott-github, this sounds like an easy fix for a part of the problem we discussed over email, too.)

Nice observation. I think this fixes the most serious part.

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -29,13 +29,7 @@ class UIComponent
 
   activate: (message) ->
     @postMessage message if message?
-    if @showing
-      # NOTE(smblott) Experimental.  Not sure this is a great idea. If the iframe was already showing, then
-      # the user gets no visual feedback when it is re-focused.  So flash its border.
-      @iframeElement.classList.add "vimiumUIComponentReactivated"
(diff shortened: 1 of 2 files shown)


smblott-github:
342cdc765b76d7581e2cba61e33d228076d76ac2 now:
- Exits vomnibar only when its window loses the focus.  This is a better UX.
- Fixes the tests (although I'm not too happy with how this is done, see 342cdc765b76d7581e2cba61e33d228076d76ac2 log record).

smblott-github:
Closing in favour of #1517.

PR #1514, opened 2015-03-03
Do not include duplicated texts (from descendents) in link hints

smblott-github:
This only affects link hints with *Use the link's name and numbers for link-hint filtering* enabled.

Currently, we match the *entire text content* of each link-hint element.

With two (or more) hints, and with one of the elements a descendent of the other, we use the entire text content of the outer node, which includes the text content of the inner node.  This leads to odd situations where the inner element cannot be selected just by typing its text, because its text is a substring of the outer element's text, so any text matching the inner hint also matches the outer hint.

For example, on Google calendar, the "Today" button shows up as two hints, one inside the other.  Typing "today" never disambiguates the hint.  You always have to hit enter.

There's another nasty example on feedly, where an outer container is clickable, but its text contains all of the (many) texts of the (many) contained links.  So the inner hint (the one you most likely want) always has to be selected manually.  This case is worse (I think), because you end up having to specifically choose the *second* hint (so `Enter` doesn't help).

In this PR, when generating the text for a hint, we exclude the texts from any descendent node which has already been considered.

Note: this also drops the inclusion of `innerHTML` if `textContent` is empty.  It's not clear why you would want to match against the HTML.  But there must have been a reason for including this originally, so...

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -231,6 +231,8 @@ LinkHints =
     # Remove rects from elements where another clickable element lies above it.
     nonOverlappingElements = []
     # Traverse the DOM from first to last, since later elements show above earlier elements.
+    # NOTE(smblott). filterHints.generateLinkText also assumes this order when generating the content text for
+    # each hint.  Specifically, we consider descendents before we consider their ancestors.
     visibleElements = visibleElements.reverse()
     while visibleElement = visibleElements.pop()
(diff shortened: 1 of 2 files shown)


smblott-github:
Merged.  I'm seeing a fair few instances of better UX with this, and haven't yet come across any odd behaviours.  And it's easy to revert if any problem cases emerge.

PR #1517, opened 2015-03-05
Hide vomnibar if host frame regains focus.

### LINKED ISSUE #1506: Once Vomnibar loses focus, Vimium practically stops working -- is this deliberate or a bug?

smblott-github:
#1511 cannot solve #1506 because the vomnibar iframe gets a blur event (and hence hides itself) every time we change tab.  This takes the opposite approach:  we hide the vomnibar when the host frame receives the focus.

Fixes #1506.

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -29,24 +29,25 @@ class UIComponent
 
   activate: (message) ->
     @postMessage message if message?
-    if @showing
-      # NOTE(smblott) Experimental.  Not sure this is a great idea. If the iframe was already showing, then
-      # the user gets no visual feedback when it is re-focused.  So flash its border.
-      @iframeElement.classList.add "vimiumUIComponentReactivated"
-      setTimeout((=> @iframeElement.classList.remove "vimiumUIComponentReactivated"), 200)
-    else
-      @show()
+    @show() unless @showing
     @iframeElement.focus()
 
   show: (message) ->
     @postMessage message if message?
     @iframeElement.classList.remove "vimiumUIComponentHidden"
     @iframeElement.classList.add "vimiumUIComponentShowing"
+    window.addEventListener "focus", @onFocus = (event) =>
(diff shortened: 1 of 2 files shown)


mrmr1993:
I don't think this will fix #1506 &mdash; the issue there is that the iframe has focus but the input doesn't, so we get stuck with no available actions.

Also, it doesn't fix the issue we discussed if there are nestled same-origin iframes manipulating focus.

The easy fix is to
* message the Vomnibar iframes to tell them to disable and hide from `handleFrameFocused` (`background_scripts/main.coffee`)
* focus the parent frame of the Vomnibar iframe when the input is blurred
  - if at some point we decide to implement clickable Vomnibar entries, we can easily switch this to `document.body` being focused instead.

I have a branch with this already at home, but I'm not sure how long it will be until I can get back and push it.

smblott-github:
> the issue there is that the iframe has focus but the input doesn't, so we get stuck with no available actions.

A click anywhere in the vomnibar focuses the input.  Does that not do it?

mrmr1993:
> Does that not do it?

It does, sorry! You got in before I could edit my original comment.

smblott-github:
Merged.  Seems to work well as a fix for #1506.

PR #1519, opened 2015-03-06
Fix vomnibar visibility

### LINKED ISSUE #1485: Vomnibar closing visible/late after changing tab
This is one for you, @mrmr1993.
- Use `O` or `B` to open a new tab.
- Return to the original tab.
- Briefly, the original vomnibar is visible.

Might this be fixed by adding a short delay (a la `nextTick`) before opening the new tab?

smblott-github:
Delay calling any action from the vomnibar until after it has closed.  This prevents flicker/late visibility when we return to a tab from which the vomnibar previously opened a new foreground tab.

Fixes #1485.


smblott-github:
This seems to be working nicely.  Merged.

PR #1521, opened 2015-03-07
Reload content scripts after install/update.

smblott-github:
This is @mrmr1993's work from #1041.

Reload content scripts when vimium is installed or updates.

(@mrmr1993:  The automatic merge was really messy (or, at least, I couldn't figure out what was going on).  Since the bulk of #1041 was actually quite compact, I took the liberty of just copying it in.  Hope you don't mind.  There are also a couple of small fixes here, e.g. checking `chrome.runtime.lastError`.)

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -2,6 +2,25 @@ root = exports ? window
 
 currentVersion = Utils.getCurrentVersion()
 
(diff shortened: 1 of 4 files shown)


smblott-github:
This is working nicely.  Merged.  Thanks, @mrmr1993.

PR #1524, opened 2015-03-09
Pass keyup events after entering link-hint mode.

### LINKED ISSUE #1522: LinkHints.activateModeToOpenInNewForegroundTab opens new tab in the background
When `LinkHints.activateModeToOpenInNewForegroundTab` is remapped to `F`, it opens new tabs in the background and doesn't switch to them. If I remap it to anything else, such as `gf`, for instance, it works as expected, opening new tabs in the foreground.

These screenshots show that my remapping was done and recognized by vimium:

vimium options - google chrome_002

vimium options - google chrome_003

Restarting Google Chrome didn't make any difference.

I'm using Vimium version 1.49, Google Chrome 41.0.2272.74 beta (64-bit), Xubuntu 14.10

PS: What a fabulous extension vimium is ! I can't live without it anymore.

smblott-github:
When we enter link-hints mode with a key such as `F`, the page (or another extension) sees the `keydown` event for `Shift`.  We need to also pass the corresponding `keyup` event, otherwise the page/extension can get confused.

Fixes #1522.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -68,6 +68,7 @@ LinkHints =
     @hintMode = new Mode
       name: "hint/#{mode.name}"
       badge: "#{mode.key}?"
+      passInitialKeyupEvents: true
       keydown: @onKeyDownInMode.bind(this, hintMarkers),
       # trap all key events
       keypress: -> false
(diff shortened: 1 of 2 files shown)


mrmr1993:
@smblott-github would it be preferable to move `KeydownEvents` and normal mode's `onKeyup` listener to the mode prototype, and filter out `keyup`s for every filtered `keydown` automatically? I think we currently end up leaking `keyup` events for other mode transitions too, which isn't ideal.

smblott-github:
> I think we currently end up leaking keyup events for other mode transitions too, which isn't ideal.

Yes.  I think we do too.  It *would* be nice if there was a general way to just *do-the-right-thing* with one piece of logic, somewhere.  Possibly in the handler stack?  All key processing happens there.

PR #1508, opened 2015-02-27
Fix #1507 Open new tab in incognito mode

wadkar:
This PR is a fix for issue #1507  
 - Change openUrlInNewTab to pass selected tab.windowId
 - Change createTab to use openUrlInNewTab

Issue #1507, opened 2015-02-26
Open new tab in incognito mode should open new tab in same window


wadkar:
Vimium version 1.49 (stable downloaded from Chrome Store)

Expected behavior: Pressing `t` while in incognito mode should open a new tab in the same incognito window
Actual behavior: Pressing `t` while in incognito mode opens a new window, i.e. _non-incognito window_

How to reproduce: Open a new incognito window and browse to any page. Press `t`
Comments: It is interesting to note that `Vomnibar.` shortcuts (`o`,`O`) and follow link shortcuts (`f`, `F`) do not open a new _non-incognito window_. They merely open a new tab in the current window.

wadkar:
After a little bit of digging around the source code and looking through Chrome JS APIs, it appears to be an upstream [bug](https://code.google.com/p/chromium/issues/detail?id=308171)
Not sure whats the policy on closing the issue. But keeping it open for the sake of posterity.

mrmr1993:
We can fix this easily by changing the `openUrlInNewTab` function (`background_scripts/main.coffee`) from

```coffee
chrome.tabs.create({ url: Utils.convertToUrl(request.url), index: tab.index + 1, selected: true }))
```

to

```coffee
chrome.tabs.create({ url: Utils.convertToUrl(request.url), index: tab.index + 1, selected: true, windowId: tab.windowId }))
```

wadkar:
Actually, the appropriate place for the change would be [Line 223](https://github.com/philc/vimium/blob/master/background_scripts/main.coffee#L223) where createTab is called.
I might suggest that we could change
`createTab: (callback) -> chrome.tabs.create({url: Settings.get("newTabUrl")}, (tab) -> callback())`
to
`createTab: (callback) -> chrome.tabs.create({url: Settings.get("newTabUrl"), windowId: tab.windowId}, (tab) -> callback())`
Could you please try it out and see if it works?

mrmr1993:
I don't think that would work, since tab is undefined in that context. This bug will also affect the `openUrlInNewTab` function, so we need a fix there too.

`createTab` should be rewritten anyway. This (with my above fix) should work:

```coffee
createTab: openUrlInNewTab Settings.get "newTabUrl"
```

rajeevksahu:
[...]

PR #1532, opened 2015-03-14
Fix #1515 related issues in PR #1508 for new tabs

### LINKED ISSUE #1515: Vimium on new-tab page
When creating a new tab with `Ctrl-T`, Vimium is not active and the focus is locked in the address bar.  The whole Vimium UX is broken.

- Does anyone know how to fix this?

Absent a Vimium fix, I've been using [New Tab Redirect](https://chrome.google.com/webstore/detail/new-tab-redirect/icpgjfneehieebagbmdbhnlpiopdcmna?hl=en) to redirect `chrome://newtab` to a blank page with a very-long TTL (on which Vimium is active).  Unfortunately, we can't do the same trick in Vimium.  First, it completely disables `chrome://newtab` (and it wouldn't be possible to have an option to re-enable it).  Second, redirecting to `pages/blank.html` still leaves the focus in the address bar.

As a super lightweight alternative to New Tab Redirect, I implemented this:
- https://chrome.google.com/webstore/detail/new-tab-url/njigpponciklokfkoddampoienefegcl
- https://github.com/smblott-github/chrome-new-tab-url

(which might be of interest to some Vimium users, absent a Vimium solution).

wadkar:
PR #1508 introduced another issue that setting vimium's `newTabURL` to
the inbuilt `pages/blank.html` resulted in opening a new tab with search
query set to `pages/blank.html`.

This commit solves the issue by conditionally calling `Utils.convertToUrl`.
Also, it changes `openOptionsPageInNewTab` call to use
`openUrlInNewTab` for consistency and reduce usage of
`chrome.tabs.create` API call.

Note that this may not solve #1507 as chrome seems to open
`chrome-extension://...` URLs in _non-incognito_ window *by design*.
You may want to set it to (some other blank page)[1] so that it can be
accessed from incognito mode. You can always open `chrome://newtab` from
incognito window.

[1] http://this-page-intentionally-left-blank.org/

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -164,9 +164,13 @@ openUrlInCurrentTab = (request) ->
 # Opens request.url in new tab and switches to it if request.selected is true.
 #
 openUrlInNewTab = (request, callback) ->
-  chrome.tabs.getSelected(null, (tab) ->
-    chrome.tabs.create({ url: Utils.convertToUrl(request.url), index: tab.index + 1, selected: true, windowId: tab.windowId },
-                         (tab) -> callback()))
+  chrome.tabs.getSelected null, (tab) ->
+    chrome.tabs.create {
+      url: if request.rawUrl then request.url else Utils.convertToUrl request.url
+      index: tab.index + 1
+      selected: true
+      windowId: tab.windowId
+    }, (tab) -> callback?()
 
 openUrlInIncognito = (request) ->
   chrome.windows.create({ url: Utils.convertToUrl(request.url), incognito: true})
@@ -232,7 +236,10 @@ moveTab = (callback, direction) ->
 # These are commands which are bound to keystroke which must be handled by the background page. They are
 # mapped in commands.coffee.
 BackgroundCommands =
-  createTab: (callback) -> openUrlInNewTab({ url: Settings.get("newTabUrl") }, (tab) -> callback())
+  createTab: (callback) -> openUrlInNewTab {
+    url: Settings.get "newTabUrl"
+    rawUrl: true
+  }, (tab) -> callback()
   duplicateTab: (callback) ->
     chrome.tabs.getSelected(null, (tab) ->
       chrome.tabs.duplicate(tab.id)
@@ -598,8 +605,7 @@ shouldShowUpgradeMessage = ->
   Utils.compareVersions(currentVersion, Settings.get("previousVersion")) == 1
 
 openOptionsPageInNewTab = ->
-  chrome.tabs.getSelected(null, (tab) ->
-    chrome.tabs.create({ url: chrome.runtime.getURL("pages/options.html"), index: tab.index + 1 }))
+  openUrlInNewTab { url: "pages/options.html", rawUrl: true }
 
 registerFrame = (request, sender) ->
   (frameIdsForTab[sender.tab.id] ?= []).push request.frameId


wadkar:
I based off this commit from PR #1508. Do you want me to rebase it to some other branch/current master? If you could resolve the conflicts yourself, that will be great though.

smblott-github:
An alternative would be cced8f7e10cef684e0bdd70c36e05ea0a9add145, which just fills in the absolute URL for `pages/blank.html` directly.  cced8f7e10cef684e0bdd70c36e05ea0a9add145 differs from what's here if the new-tab URL is set to `options.html` (cced8f7e10cef684e0bdd70c36e05ea0a9add145 sends it to the default search engine, whereas this PR opens the options page).  It's not at all clear which is correct.  However, `pages/blank.html` is the only URL which we've suggested should work as a new-tab option.

(It looks like there may be no way to use `pages/blank.html` in incognito mode.)

wadkar:
Hmm, alternative cced8f7e10cef684e0bdd70c36e05ea0a9add145 works too. In that case, we might include @mrmr1993's suggestion for [`openOptionsPageInNewTab`](https://github.com/philc/vimium/blob/master/background_scripts/main.coffee#L633). See [my comment on `cced8f7`](https://github.com/philc/vimium/commit/cced8f7e10cef684e0bdd70c36e05ea0a9add14#commitcomment-10201787).
Feel free to close this PR in case cced8f7e10cef684e0bdd70c36e05ea0a9add145 is accepted.
On a side note, how do I keep in sync with `upstream` for build, test, as well as running bleeding edge vimium? It feels silly when I cant see cced8f7e10cef684e0bdd70c36e05ea0a9add145 on my local.

smblott-github:
@wadkar.  Thanks for this.  But I think we should go with cced8f7e10cef684e0bdd70c36e05ea0a9add145.

mrmr1993:
@smblott-github Can we still explore having a unified tab opening function? It seems like calling the same chrome API in slightly different ways with different levels of detail throughout the codebase is generally not the best practise.

Some other thoughts:
* `openUrlInNewTab` should probably only open an url in a new tab (so no conversion to a search url; the frontend should point to a different function called eg. `openUrlOrSearchInNewTab` which does this and then calls `openUrlInNewTab` itself)
* hardcoding `pages/blank.html` as an exception is bad practise and a cheap fix (sorry).

smblott-github:
> It seems like calling the same chrome API in slightly different ways with different levels of detail throughout the codebase is generally not the best practise.

If we create our own centralised function, we'll just end up calling *it* in slightly different ways with different levels of detail throughout the codebase.  Or, put another way, we'd just be creating our own `chrome.tabs.create` API.  It might be worth doing that, at some point.  But probably only if we need a consistent API which is clearly different from `chrome.tabs.create`.

> hardcoding pages/blank.html as an exception is bad practise and a cheap fix (sorry).

Fair, but also a bit harsh.  `createTab` is th *only* place where `pages/blank.html` has a spacial meaning.

Changing topic (and mentioning you, @mrmr1993, to get your attention)...

I'm wondering whether we shouldn't abandon our new-tab page functionality entirely.  We can't make `t` and `Ctrl-T` do the same thing, and we can't get `pages/blank.html` to work in incognito-mode windows.  That's no great.  Personally, I just use my little new-tab [extension](https://chrome.google.com/webstore/detail/new-tab-url/njigpponciklokfkoddampoienefegcl) (which solves the first problem, but not the second).  If Vimium provided a reliable blank page somewhere with a very-long TTL, we could point users towards that.

PR #1534, opened 2015-03-15
Rework exclusion rule popup population, propagation and page icon

smblott-github:
Edit: Actually, it now does considerably more.  See second post.  The following is out of date.

This PR does two related things:

1. 3876a4f06de77d190fdaecc6cf99eb57134d0372 populates the page popup with the active frame's URL (as opposed to the tab's URL).

2. ea05de3f114dcc81a91bf863538f20754641cadd ensures changes to exclusion rules are propagated to all frames, not just the first to respond.  This is a long-standing bug.

With these changes, we can now - for example - set and (immediately) use exclusion rules in the hangouts frame on gmail.

There are a lot of long-standing issues around how frames, exclusion rules and the page icon are handled.  This is a first step towards fixing some of these.  A better way to fix all of this would be to propagate options via `chrome.storage`.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -150,7 +150,7 @@ initializePreDomReady = ->
   settings.load()
 
   initializeModes()
-  checkIfEnabledForUrl true # true means checkIfEnabledForUrl is being called on start up.
+  checkIfEnabledForUrl()
   refreshCompletionKeys()
 
   # Send the key to the key handler in the background page.
@@ -547,7 +547,7 @@ onKeyup = (event) ->
   DomUtils.suppressPropagation(event)
   @stopBubblingAndTrue
 
-checkIfEnabledForUrl = (onStartUp = false) ->
+checkIfEnabledForUrl = ->
   url = window.location.toString()
 
   chrome.runtime.sendMessage { handler: "isEnabledForUrl", url: url }, (response) ->
@@ -556,7 +556,7 @@ checkIfEnabledForUrl = (onStartUp = false) ->
     isIncognitoMode = response.incognito
     if isEnabledForUrl
       initializeWhenEnabled()
-    else if onStartUp and HUD.isReady()
+    else if HUD.isReady()
       # Quickly hide any HUD we might already be showing, e.g. if we entered insert mode on page load.
       HUD.hide()
     handlerStack.bubbleEvent "registerStateChange",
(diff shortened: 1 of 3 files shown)


smblott-github:
This has turned into a major reworking of how exclusion rules are handled.

1. The page popup is pre-populated (when no other rule applies) with a URL rule matching the URL of the current *frame* (not tab, as previously).
2. When exclusion rules change, they are propagated to *all frames* (not just whichever one happens to respond first).
3. The page icon changes to reflect the actual exclusion rules in effect for the current frame (not the rules for the tab's URL, as previously).

For example, on GMail, the user may have one exclusion rule for GMail, and another for the hangouts frame.  With this, the UI elements (popup and icon) are now consistent with the actual rules in play.

This also now fixes a race condition in how rules are propagated to pages.

Replaces #1495.
Fixes #1453.
Fixes #1494.

mrmr1993:
Minor nitpick, but I think frames should disable the icon again in `window.onblur` so we show the right state if the user clicks/focuses a frame where we don't work (eg. another extension's frame on a page).

Otherwise, this looks great, definately approve.

content_scripts/vimium_frontend.coffee
```diff
@@ -183,13 +182,14 @@ initializePreDomReady = ->
...
     return if sender.tab and not sender.tab.url.startsWith 'chrome-extension://'
-    return unless isEnabledForUrl or request.name == 'getActiveState' or request.name == 'setState'
+    return unless isEnabledForUrl
     # These requests are delivered to the options page, but there are no handlers there.
     return if request.handler in [ "registerFrame", "frameFocused", "unregisterFrame" ]
     sendResponse requestHandlers[request.name](request, sender)
     # Ensure the sendResponse callback is freed.
     false
 
+
```
mrmr1993: Why is there an extra newline added here?

smblott-github:
> I think frames should disable the icon again in window.onblur so we show the right state if the user clicks/focuses a frame where we don't work

Like this: e87a8dca875537b2c1e4e39dab1986551281f09a (not on this branch).

Technically, this is the correct thing to do.  However, we get blur and focus events every time the focused frame changes, and there's a race condition.  I'm not sure the benefit (e.g. handling another extension's frame correctly) outweighs the risk (incorrect behaviour on an ordinary web page).

(There's a comment in the insert-mode code somewhere suggesting that we cannot rely on blur and focus events happening in the expected order for inputs.  I'm guessing the same might apply for frames.)

smblott-github:
> I think frames should disable the icon again in window.onblur so we show the right state if the user clicks/focuses a frame where we don't work

Trying out https://github.com/philc/vimium/commit/e87a8dca875537b2c1e4e39dab1986551281f09a, I can't trigger a race condition (which is little comfort!).

However, https://github.com/philc/vimium/commit/e87a8dca875537b2c1e4e39dab1986551281f09a greys out the icon each time the Chrome window loses the focus.  Not good.

mrmr1993:
> I'm not sure the benefit (...) outweighs the risk

Agreed. I hadn't thought of window focus being a factor. LGTM.

smblott-github:
Merging this.  It seems to be working nicely.

PR #1537, opened 2015-03-16
Activate vomnibar in main/top-level window only (WIP)

smblott-github:
This changes vomnibar commands to activate not in the current frame, but in `window.top`.  Consequently, the vomnibar always appears in the same position, and we don't get odd looking vomnibars in small frames.

Apart from the better UX, this seems to be the right thing to do. Vomnibar commands apply to tabs (not frames).

Currently incomplete:
- ~~On exit, the focus is not returned to the frame which originally had the focus.  (It's returned to window.top).~~ Fixed.

@mrmr1993.  What do you think?  Have you any ideas how to get the focus back to the original frame?

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -657,7 +657,6 @@ handleFrameFocused = (request, sender) ->
   if frameIdsForTab[tabId]?
     frameIdsForTab[tabId] =
       [request.frameId, (frameIdsForTab[tabId].filter (id) -> id != request.frameId)...]
-  console.log frameIdsForTab[tabId]
   # Inform all frames that a frame has received the focus.
   chrome.tabs.sendMessage sender.tab.id, 
     name: "frameFocused"
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -660,9 +660,9 @@ handleFrameFocused = (request, sender) ->
   # Inform all frames that a frame has received the focus.
   chrome.tabs.sendMessage sender.tab.id, 
     name: "frameFocused"
-    frameId: request.frameId
+    focusFrameId: request.frameId
 
-# Send a message to a all frames in the current tab.
+# Send a message to all frames in the current tab.
 sendMessageToFrames = (request, sender) ->
   chrome.tabs.sendMessage sender.tab.id, request.message
 
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -666,6 +666,10 @@ handleFrameFocused = (request, sender) ->
 sendMessageToFrames = (request, sender) ->
   chrome.tabs.sendMessage sender.tab.id, request.message
 
+# For debugging only. This allows content scripts to log messages to the background page's console.
+bgLog = (request, sender) ->
+  console.log "#{sender.tab.id}/#{request.frameId}", request.message
+
 # Port handler mapping
 portHandlers =
   keyDown: handleKeyDown,
@@ -694,6 +698,7 @@ sendRequestHandlers =
   gotoMark: Marks.goto.bind(Marks)
   setBadge: setBadge
   sendMessageToFrames: sendMessageToFrames
+  log: bgLog
 
 # We always remove chrome.storage.local/findModeRawQueryListIncognito on startup.
 chrome.storage.local.remove "findModeRawQueryListIncognito"
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -657,6 +657,10 @@ handleFrameFocused = (request, sender) ->
   if frameIdsForTab[tabId]?
     frameIdsForTab[tabId] =
       [request.frameId, (frameIdsForTab[tabId].filter (id) -> id != request.frameId)...]
+  # Inform all frames that a frame has received the focus.
+  chrome.tabs.sendMessage sender.tab.id, 
+    name: "frameFocused"
+    frameId: request.frameId
 
 # Send a message to a all frames in the current tab.
 sendMessageToFrames = (request, sender) ->
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -658,14 +658,9 @@ handleFrameFocused = (request, sender) ->
     frameIdsForTab[tabId] =
       [request.frameId, (frameIdsForTab[tabId].filter (id) -> id != request.frameId)...]
 
-# Send a message to a frame (in the sender's tab).  The sender should set:
-# - request.targetFrameId (the target frame)
-# - request.name (the name of the handler in the target frame, e.g. "focusFrame")
-# In addition, request.senderFrameId will be set to the frameId of the sender.
-sendMessageToFrame = (request, sender) ->
-  request.senderFrameId = request.frameId
-  request.frameId = request.targetFrameId
-  chrome.tabs.sendMessage sender.tab.id, request
+# Send a message to a all frames in the current tab.
+sendMessageToFrames = (request, sender) ->
+  chrome.tabs.sendMessage sender.tab.id, request.message
 
 # Port handler mapping
 portHandlers =
@@ -694,7 +689,7 @@ sendRequestHandlers =
   createMark: Marks.create.bind(Marks)
   gotoMark: Marks.goto.bind(Marks)
   setBadge: setBadge
-  sendMessageToFrame: sendMessageToFrame
+  sendMessageToFrames: sendMessageToFrames
 
 # We always remove chrome.storage.local/findModeRawQueryListIncognito on startup.
 chrome.storage.local.remove "findModeRawQueryListIncognito"
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -658,6 +658,15 @@ handleFrameFocused = (request, sender) ->
     frameIdsForTab[tabId] =
       [request.frameId, (frameIdsForTab[tabId].filter (id) -> id != request.frameId)...]
 
+# Send a message to a frame (in the sender's tab).  The sender should set:
+# - request.targetFrameId (the target frame)
+# - request.name (the name of the handler in the target frame, e.g. "focusFrame")
+# In addition, request.senderFrameId will be set to the frameId of the sender.
+sendMessageToFrame = (request, sender) ->
+  request.senderFrameId = request.frameId
+  request.frameId = request.targetFrameId
+  chrome.tabs.sendMessage sender.tab.id, request
+
 # Port handler mapping
 portHandlers =
   keyDown: handleKeyDown,
@@ -685,6 +694,7 @@ sendRequestHandlers =
   createMark: Marks.create.bind(Marks)
   gotoMark: Marks.goto.bind(Marks)
   setBadge: setBadge
+  sendMessageToFrame: sendMessageToFrame
 
 # We always remove chrome.storage.local/findModeRawQueryListIncognito on startup.
 chrome.storage.local.remove "findModeRawQueryListIncognito"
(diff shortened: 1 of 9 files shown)


mrmr1993:
@smblott-github this causes the Vomnibar to fail when the top level frame has a `<frameset>` rather than a `<body>`.

> Have you any ideas how to get the focus back to the original frame?

I think `frameIdsForTab[tab.id]` should still be intact (in the background page) when exiting the Vomnibar, so we can message the last focused frame to tell it to refocus itself. Unfortunately, this involves a round trip to the background page.

I think (but haven't tested that) it also might be possible to entangle a port with the Vomnibar frame by doing

```coffee
vomnibarUrl = chrome.runtime.getURL "pages/vomnibar.html"
foundVomnibarFrame = false
for frameWindow in window.top
  try # Catch errors from accessing location on cross-origin frames.
    if frameWindow.location == vomnibarUrl
      foundVomnibarFrame = true
      # Set up a port with the window here.
      break
unless foundVomnibarFrame
  # Code to mitigate there being no such frame.
```

which would avoid the round trip to the background page, by having the vomnibar communicate with the opener frame directly.

smblott-github:
> this causes the Vomnibar to fail when the top level frame has a <frameset> rather than a <body>

Actually, not.  I've been using [this](http://www.quackit.com/html/templates/frames/frames_example_2.html) as the main test page, and it works fine there.

> I think frameIdsForTab[tab.id] should still be intact

It turns out to be better/easier than that.  The request to open the vomnibar contains the frameId of the originating frame.  So the top window knows which frame to focus when it's done.

The basics are now functional (https://github.com/philc/vimium/commit/efada88f419933c5bd1478faada3b4eff3082103).  For example, opening the vomnibar in the hangouts sidebar opens a full-window vomnibar, and returns the focus to the sidebar when the vomnibar closes.

The approach I took involves extra massaging when the vomnibar closes, but no more messaging than when the vomnibar opened.  So no big deal.

To do: previously, we hid the vomnibar when the host frame receives the focus.  Now, we probably need to hide the vomnibar when *any* other frame receives the focus.  Not sure.  Requires more thought.

mrmr1993:
> Actually, not.

Sorry! That was a problem that I ran into before, something must have changed (presumably VIF).

> Now, we probably need to hide the vomnibar when *any* other frame receives the focus.

This seems like the most sensible choice. We can easily do this by sending a vomnibar-specific close message in `handleFrameFocused`.

smblott-github:
> We can easily do this by sending a vomnibar-specific close message in handleFrameFocused.

It's actually implemented now.  I could describe the parts individually, but it'd probably be better if you just review the code (which would be appreciated, if you're able @mrmr1993).

Summary:
- Vomnibar activation is only handled in the top frame.  However, the request contains the frameId of the original frame.  This is passed to the vomnibar so that we can refocus that frame later.
- The main/top frame is focused and the vomnibar operates there.
- When the vomnibar closes, we use a new background handler (`sendMessageToFrames`) to send a `focusFrame` request back to the original frame.
- However, there's a subtlety.  When the vomnibar closes, Chrome focuses the containing frame.  There's therefore a race condition as to which frame ends up focused.  To get around this, we wait until the containing frame receives the focus before sending the `focusFrame` message.
- Whenever *any* frame receives the focus, we now broadcast a `frameFocused` message to all frames.  This is the same as @mrmr1993's idea, above.  The vomnibar receives this and, if it's active, then it hides itself.

UX decision:
- Currently, when the vomnibar hides, we flash the original frame, but only if it isn't the main/top frame.  This looks OK to me, but may not be necessary.

background_scripts/main.coffee
```diff
@@ -657,6 +657,19 @@ handleFrameFocused = (request, sender) ->
   if frameIdsForTab[tabId]?
     frameIdsForTab[tabId] =
       [request.frameId, (frameIdsForTab[tabId].filter (id) -> id != request.frameId)...]
+  console.log frameIdsForTab[tabId]
```
mrmr1993: This should be removed

content_scripts/vimium_frontend.coffee
```diff
@@ -94,7 +94,12 @@ settings =
 #
 # Give this frame a unique id.
 #
-frameId = Math.floor(Math.random()*999999999)
+window.frameId = Math.floor(Math.random()*999999999)
```
mrmr1993: The assignment to window is already done at the bottom of the file; this change is unnecessary.

content_scripts/vimium_frontend.coffee
```diff
@@ -229,10 +235,13 @@ getActiveState = ->
...
-registerFocus = ->
-  # settings may have changed since the frame last had focus
-  settings.load()
-  chrome.runtime.sendMessage({ handler: "frameFocused", frameId: frameId })
+registerFocus = (event) ->
+  if event.target == window
+    # Settings may have changed since the frame last had focus.
+    settings.load()
+    # Don't register frameset containers; focusing them is no use.
+    unless document.body?.tagName.toLowerCase() == "frameset"
```
mrmr1993: This mean that focusing a `<frameset>` will leave the Vomnibar visible, no? Is this the right behaviour?

content_scripts/vimium_frontend.coffee
```diff
@@ -94,7 +94,12 @@ settings =
 #
 # Give this frame a unique id.
 #
-frameId = Math.floor(Math.random()*999999999)
+window.frameId = Math.floor(Math.random()*999999999)
+
+# For debugging only. This logs to the console on the background page.
+window.bgLog = (args...) ->
```
mrmr1993: This (and the background code that supports it) don't seem to be used anywhere in this PR. Why is it included?

content_scripts/vimium_frontend.coffee
smblott-github: > This mean that focusing a `<frameset>`

Can that happen?  If it can, then you're right.

(I guess it could happen if the entire document is a frame set, but the frames don't cover the entire viewport.  Clicking outside of the frame set may then focus it.  Do you have a better example?)

(Separately, if we can focus a `<frameset>`, the in `master` we end up registering the frame set on `focus`, and it becomes part of the `gf` cycle.)

content_scripts/vimium_frontend.coffee
smblott-github: It isn't used.

I think we should leave it in though.  You had something similar in a PR some time ago.  It can be useful for debugging to have all messages in a single console.

content_scripts/vimium_frontend.coffee
mrmr1993: I believe that it's focused when you drag the boundary between two frames to resize them, and probably also if you click the frames' borders.

> we end up registering the frame set on `focus`, and it becomes part of the `gf` cycle.

If they *are* focusable, we should probably register them as normal and just mark them as `<frameset>`s, so we can switch out of them/keep track of our position reliably, but can still skip them when switching frames with `gf`.

content_scripts/vimium_frontend.coffee
```diff
@@ -255,9 +264,19 @@ unregisterFrame = ->
...
     frameId: frameId
-    tab_is_closing: window.top == window.self
+    tab_is_closing: DomUtils.isTopFrame()
 
 executePageCommand = (request) ->
+  # Vomnibar commands are handled in the tab's main/top frame.
+  if request.command.split(".")[0] == "Vomnibar"
+    if DomUtils.isTopFrame()
+      # We pass the frameId from request.  That's the frame which originated the request, so that's the frame
+      # which whould receive the focus when the vomnibar closes.
```
mrmr1993: Typo; `whould` should be `should`.

content_scripts/vimium_frontend.coffee
smblott-github: > If they are focusable, we should probably register them as normal...

Seems reasonable.

content_scripts/vimium_frontend.coffee
```diff
@@ -178,6 +183,7 @@ initializePreDomReady = ->
     currentKeyQueue: (request) ->
       keyQueue = request.keyQueue
       handlerStack.bubbleEvent "registerKeyQueue", { keyQueue: keyQueue }
+    frameFocused: -> # A frame has received the focus.  We don't care, here. The Vomnibar/UI-component cares.
```
mrmr1993: Is this ever called?

smblott-github: No.  But the `onMessage` handler (just below) doesn't check to see if a handler is defined.  So I added one here.

(It's arguably better not to check.  That way, we get errors when we make a mistake.)

mrmr1993: I'd say drop it then. It took me quite a while to persuade myself that it wasn't being called, which could have been completely avoided :(

smblott-github: > I'd say drop it then.

OK.  I'll add a test to the `onMessage` handler instead.

smblott-github:
A reason for *not* doing this is...

If Vimium is disabled for the main/top frame, but enabled for some other frame, then the vomnibar doesn't work.

How important is that?  Not sure.  On the one hand, it's not a common case.  On the other, it's pretty broken.

Imagine the status quo was *vomnibar-in-main-frame* and someone proposed getting around this quirk by moving the vomnibar to the host frame (with strange, small vomnibars).  Would we buy that?  Not sure.  Probably not.

mrmr1993:
The most logical fix I can think of is to execute `Vomnibar.init` when one or more frames are active and `Vomnibar.deinit` (doesn't exist, but basically just has to remove the Vomnibar iframe from the document) when none are active. Obviously this wants to be repeated every time exclusion rules are updated. Checking whether any frames are active is easy (in the background) using `chrome.webNavigation.getAllFrames` and checking the urls of all the frames.

smblott-github:
Update.  Having used this for some time, it seems to be robust and work well.  To summarise, when the vomnibar opens within frames, we get something like this:

snapshot

With this PR, the vomnibar is always in the main frame, always in the same place, always the same size and always visible.

The down side, however, is that if Vimium is disabled in the main frame but active in the active frame, then the vomnibar doesn't work.  That's pretty serious.  However, I've yet to come across a live example of this.  One case would be Google's Inbox with its thin and tall Hangouts frame.  However, there,  the vomnibar is useless currently (because it's not visible at all anyway).

Nevertheless, if we were to merge this PR, it would be a good candidate for [Vimium Labs](https://github.com/philc/vimium/issues/1542).  So I've put it together as "Vimium Labs" setting on the options page:

snapshot

(This is at the bottom of the options page, below the fold.)

smblott-github:
Another update.  With d67347461da5a0468547e4c807a54bb40746faf4, the Vomnibar now activates in the top/main frame even if Vimium is disabled in that frame.  As a consequence, there's no longer a need to consider this as a "Vimium Labs" feature.

smblott-github:
Status:
- Seems to work nicely. (I've been using this for quite some weeks.)

Issues...
- If the current and main frames are the same: works fine.
- If the Vimium is enabled in both the current and the main frame: works fine.
- If Vimium is not enabled in the main frame: works fine.
- If Vimium *cannot run* in the main frame (e.g. it's a chrome extension page): broken.  How important is this last case?  I haven't come across it in the wild.  Are there any new-tab extensions, for example, which embed frames on which Vimium can run?

Another issue:  There's sometimes a slight Vomnibar flicker when returning to a tab after previously using the Vomnibar to open a link in a new tab.

philc:
This LGTM. The 4th case you mention sounds incredibly niche to me.

smblott-github:
More update.  The Vomnibar flicker issue has been fixed.

Opinion:  We should merge this.  It's a big UX win.

I'll merge soon, unless there's significant push back.

PR #1538, opened 2015-03-16
Flash a frame's border when it receives the focus.

smblott-github:
When the focus is in a frame other than the main frame (e.g. in a Google Hangouts frame), we get no visual feedback as to which frame has the focus when we return to a tab.

This flashes the (non-main) frame's border when it receives the focus.

However, having tried this out, it seems to be too noisy.  The frame flashes even when we just change window, or when the vomnibar closes.  So not a great idea.  I'm not proposing to merge this.

Are there any better ideas as to how to provide visual feedback as to the active frame?

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -304,13 +304,13 @@ BackgroundCommands =
   moveTabLeft: (count) -> moveTab(null, -count)
   moveTabRight: (count) -> moveTab(null, count)
   nextFrame: (count,frameId) ->
-    chrome.tabs.getSelected(null, (tab) ->
+    chrome.tabs.getSelected null, (tab) ->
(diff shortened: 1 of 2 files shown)


mrmr1993:
How about covering everything in the parent frame that *isn't* the focused frame with semi-transparent gray element when the frame is focused, and then fading it to transparent over a short period (400ms?) after?

We also probably *shouldn't* flash when the user has changed the frame manually themselves - only when we focus a frame, or when returning to a tab.

smblott-github:
Hmm.  Interesting ideas, @mrmr1993.  Still sounds too noisy, though.  This idea probably isn't going anywhere at the moment.  Perhaps it's just something to keep in mind.

PR #1543, opened 2015-03-22
Use chrome.notifications for upgrade notifications.

smblott-github:
Instead of using the HUD for upgrade notifications, this uses `chrome.notifications`, which just seems like a more "chrome-y" way of doing things:

snapshot

Clicking on the notification takes you to the release notes, as currently.

Notes:
- A notification is shown separately on each chrome installation (so, you get the notification on your laptop, and again on your desktop, and so on).  This is the same as the current HUD notification.
- We do not re-show the notification, or expect the user to acknowledge it.  We show it once, and are done.  This is a change from the status quo.

To do:
- ~~If the user has to explicitly accept the "notifications" permission, then we lose the notification until the next time Vimium starts.  (Fix coming.)~~ Done.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -58,7 +58,7 @@ completers =
   bookmarks: new MultiCompleter([completionSources.bookmarks])
   tabs: new MultiCompleter([completionSources.tabs])
 
-chrome.runtime.onConnect.addListener((port, name) ->
+chrome.runtime.onConnect.addListener (port, name) ->
   senderTabId = if port.sender.tab then port.sender.tab.id else null
   # If this is a tab we've been waiting to open, execute any "tab loaded" handlers, e.g. to restore
   # the tab's scroll position. Wait until domReady before doing this; otherwise operations like restoring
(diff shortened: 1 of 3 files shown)


philc:
This is a great idea!

smblott-github:
@philc:  I'm going to go ahead and merge this.  Reviewing it, it seems good to go.

philc:
What happens when you click on the chrome notification? Also, does it auto-dismiss (if so, after how long), or does the user need to hit the X button?

I wonder how many people click on the "release notes" link we have in our current upgrade notification. I'm guessing we lose that flow with the chrome notifications.

If we do, it would be nice to put a "what's new?" link under the "current version" text in the help dialog, so it's somewhat possible to discover what's new.

smblott-github:
> What happens when you click on the chrome notification?

You're taken [here](https://github.com/philc/vimium#release-notes) (which is the same as the HUD notification).

> Does it auto-dismiss (if so, after how long), or does the user need to hit the X button?

It dismisses itself after some (fairly small) number of seconds.  Thereafter, the notification remains available in the "notifications area" (which is platform specific).

> it would be nice to put a "what's new?" link under the "current version" text in the help dialog

Good idea.

philc:
Cool, this SGTM.

PR #1550, opened 2015-04-03
Reinstate "Insert mode" indicator in HUD.

smblott-github:
When entering insert mode with `i`, this reinstates the `Insert mode` indicator in the HUD.  This may appease those who are used to looking to the bottom right to check whether they are in insert mode or not.

Note:  This implementation differs slightly from the legacy indicator.  When, after entering insert mode with `i`, the user clicks on an input, the HUD indicator now disappears immediately.  Previously, the indicator remained visible until the user leaves (blurs) the input.

### DIFF
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -7,6 +7,7 @@ class InsertMode extends Mode
 
     # If truthy, then we were activated by the user (with "i").
     @global = options.global
+    HUD.show "Insert mode" if @global
 
     handleKeyEvent = (event) =>
       return @continueBubbling unless @isActive event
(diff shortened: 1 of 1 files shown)


philc:
I haven't tried this PR, but based on your helpful description, isn't the legacy behavior superior? On some controls (like combo boxes) where the focus/selection isn't obvious, it's comforting to see the insert mode indicator so that I know if I type something, I won't reload or close the tab.

Note that the current badge behavior remains in "insert" until the input box is blurred in the scenario you described.

smblott-github:
@philc.  There are inconsistencies and oddities in the legacy behaviour (I can go into them later, if necessary).

Let's park this discussion until we decide which direction we're going in.

smblott-github:
Replaced by #1570.

Issue #1554, opened 2015-04-08
LABEL aren't listed as a click-able item in f mode


SRachamim:
When pressing f to see all available links, LABEL tags are not listed.

smblott-github:
Thanks @SRachamim.  Could you give more detail, please?  Which site?  What do expect to happen?  What does happen?  And so on.

SRachamim:
Yes, sorry for not giving more details.

When I press f I expect to see not only <a> and <button> tags, but all click-able items on the page. <label> is one of them. You can visit http://www.tipranks.com to see an example to a web site that uses <input> tags inside <label> tags instead of using for attribute (in order to avoid id attributes).

When you pressing  f in order to see all click-able items on the page, I don't see any <label> tag.

smblott-github:
It's the "Trending" labels you're interested in, @SRachamim, right?

On first inspection, it looks like we're not picking these up because we're concluding (incorrectly) that they're not visible (In `DomUtils.getVisibleClientRect`).  They are reported as having zero height.

(Temporarily removing [this test](https://github.com/philc/vimium/blob/master/lib/dom_utils.coffee#L64) fixes the problem.)

mrmr1993:
There isn't currently any code in `LinkHints` that marks `<label>` elements as clickable, unless they satisfy some other condition. I threw together [this JSFiddle](http://jsfiddle.net/mhwa9zep/1/) with all the obvious configurations I could think of.

Some thoughts:
* It's probably too heavy handed and confusing to always show a link hint for `<label>`s which are associated with an `<input>`.
* It's not unreasonable that a webdev might use a label and a hidden native checkbox/radio button to create their own custom checkbox/radio button.
  - This means that visible `<label>`s associated with a hidden  `<input>` should probably have a link hint.
* A `<label>` not associated with an `<input>` (its `for` attribute doesn't correspond to the `id` of any element, or there's no `for` attribute *and* it doesn't wrap an `<input>`) shouldn't have a link hint.

(Sorry if this is over-analysis or I misunderstood the issue.)

PR #1557, opened 2015-04-11
Remove condition from getVisibleClientRect().

smblott-github:
@mrmr1993.  Please see the diff.  Basically, this PR removes the requirement that we only consider children (of zero-height/width elements) which are not floated or not positioned absolutely.

Why do we need this condition?

An example came up in #1554.  On [this site](https://www.tipranks.com/) there are these clickable features:

snapshot

The DOM looks like:

snapshot

We don't currently pick them up as hints because the `<a>` is of zero height while the `<span>` is not (and is not floated or fixed).

That's one example where this PR is the right thing to do.  Are there examples where this PR would be the wrong thing to do?

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -54,7 +54,7 @@ DomUtils =
     clientRects = (Rect.copy clientRect for clientRect in element.getClientRects())
 
     for clientRect in clientRects
-      # If the link has zero dimensions, it may neverthe less be wrapping visible elements.
+      # If the link has zero dimensions, it may nevertheless be wrapping visible elements.
       if (clientRect.width == 0 || clientRect.height == 0)
         for child in element.children
           childClientRect = @getVisibleClientRect(child)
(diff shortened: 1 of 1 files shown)


mrmr1993:
Presumably there's `overflow: visible` set on the `<a>` element, and so its contents aren't bounded by its client rect?

> Are there examples where this PR would be the wrong thing to do?

I expect the existing code was supposed to deal with `overflow-x: hidden; width: 0px;` and `overflow-y: hidden; height: 0px` elements (although it seems odd that there wasn't a check for overflow being `in ["hidden", "scroll", " auto"]` for the relevant direction). It's probably fine to go.

mrmr1993:
I've revisited this briefly and actually looked at the webpage. The offending elements take the form

```coffee
<a href="foo" style="display: inline; font-size: 0px;"><span style="display: inline; font-size: 16px">Text</span></a>
```

Indeed, from a [small JSFiddle](http://jsfiddle.net/ayyfban9/16) I put together, it's obvious that only `display: inline` elements around other `display: inline` elements misreport their height in this way. Instead of removing the condition, we should therefore be adding a check for `display: inline`.

I'll submit a PR doing this with some tests for it.

mrmr1993:
I've pushed #1616, which fixes the original issue without opening us up to showing hints for links hidden in an object's overflow.

PR #1558, opened 2015-04-12
Fix undefined domain completion

musically-ut:
If one visits a site with `undefined` in its domain name (say, `http://undefined.io`), then upon opening the ViOmniBox, immediately that domain name would pop up.

This pull requests fixes that behavior.


smblott-github:
Thanks, @musically-ut.  Nice catch!

PR #1563, opened 2015-04-13
Update README.md

deslee:
Added a backtick literal to the section about section marks. Before there was no backtick, which gave misleading instructions for returning to a mark.


smblott-github:
Thanks, @deslee.

PR #1570, opened 2015-04-18
[WIP] Remove badges and revert to HUD for mode indicator.

### LINKED ISSUE #1569: Backspace to exit Find Mode finds "undefined".
Go [here](http://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html).  Type `/` then `Delete` (Edit: should be `Backspace`).  We find the occurrence of `undefined` on the page.

Bug is in 1.49 and `master`.

smblott-github:
After much discussion and trying-it-out-in-the-wild, it looks like we're going back to using the HUD for mode indications.  This PR begins that process (and fixes #1569 in passing).

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -355,32 +355,12 @@ updateOpenTabs = (tab, deleteFrames = false) ->
   # Frames are recreated on refresh
   delete frameIdsForTab[tab.id] if deleteFrames
 
-chrome.browserAction.setBadgeBackgroundColor
-  # This is Vimium blue (from the icon).
-  # color: [102, 176, 226, 255]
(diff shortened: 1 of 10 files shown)


mrmr1993:
Code LGTM, but haven't tested.

smblott-github:
Thanks for the feedback, @mrmr1993.

For the record, I think we should merge this pretty soon:
- We need this for the next release.
- We therefore also need to shake out any corner cases which have been overlooked.

smblott-github:
Merged this.

It seems to be working well.  And, if we're to find any remaining broken corner cases, then it needs to be used by people other than me.

Please provide feedback if the mode indicator behaviour doesn't seem right.

PR #1575, opened 2015-04-21
Add mainFrame command to select main/top window

### LINKED ISSUE #426: Focusing content page
On many instances main page gets deselected and needs to be clicked once in other for vimio to continue working. 
It would be good to at least have an option to select the main frame.

For instance, on google plus if I use vimio to open notification panel from the google toolbar, if I press ESC toolbar is gone and vimio stops working until I click on the page.

smblott-github:
With hat tip to @mrmr1993, this implements a new `mainFrame` command (default binding `gF`) which focuses the current tab's main frame.

Why is this needed?  Go [here](http://www.thejournal.ie/bus-eireann-transport-unions-days-action-strike-2060286-Apr2015/?utm_source=shortlink), type `gf`.  Now, try to get back to the main frame without using the mouse.

(Notwithstanding @int3's comment in #426, this seems to work well on GMail and Google's Inbox.)

(It would be better if we could achieve this without a new command.)

Fixes #426.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Navigating the current page:
     yy      copy the current url to the clipboard
     yf      copy a link url to the clipboard
     gf      cycle forward to the next frame
+    gF      focus the main/top frame
 
 Navigating to new pages:
 
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -129,6 +129,7 @@ Commands =
       "goPrevious",
       "goNext",
       "nextFrame",
+      "mainFrame",
       "Marks.activateCreateMode",
       "Vomnibar.activateEditUrl",
       "Vomnibar.activateEditUrlInNewTab",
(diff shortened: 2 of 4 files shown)


mrmr1993:
Sorry I couldn't help with the actual "main" frame issue you were asking about; that sounds like it would need some reasonably complex data-based heuristics.

Code LGTM. Though it might be more user-friendly to choose frames visually (eg. via link hints), instead of adding commands for frame movement.

smblott-github:
@mrmr1993.  I was rather wondering whether there might be a way to achieve the same thing with existing commands.  Here are two ideas which don't work:
- `0gf` takes you to the main frame.
- Make the count for `gf` absolute (instead of relative).  Then `gf` or `1gf` would take you to the main frame.  (I'm not proposing doing this.)

smblott-github:
Update: I'm liking this feature a lot.  It eliminates many of the cases where you have to click on the page to get the focus back.

smblott-github:
More update.  This command proving really useful, and I can't think of another (or, rather, a better) way than to introduce a new command.

Unless there's significant pushback (or a better idea emerges), I propose we should merge this.

PR #1577, opened 2015-04-22
Show yanked text in HUD for "yf"

smblott-github:
If `yy` shows the yanked text, then so too should `yf`.

(We need to provide the user with feedback that something happened.)


smblott-github:
Also, now:

> If the element does not have an HREF, then we shouldn't be offering to copy it's link (caca2b6e81ba6fcad9cdc406c8cde15084db7539).

So we don't offer links which do nothing when selected.

smblott-github:
Merging.
This is a no brainer.

philc:
Nice!

Issue #1573, opened 2015-04-20
I can't enter Korean after I enter '/'(find mode)


cypicaso:
Hi.

I can enter only English in find mode.
I want to enter Korean(2-byte characters) in find mode.

Thanks.

smblott-github:
@cypicaso.  Thanks for the report.  Afraid I don't know much about non-English keyboards.  Could you perhaps go to [this site](http://jsfiddle.net/mrcoles/5TJc9/light/), focus the bottom-right frame, press the key you want ot enter and post back the results?

cypicaso:
I entered 8 Korean character: "가너도루"----Click here and start typing to see
how keyboard events are triggered. Hover over a field to see the event
property. (via mrcoles.com <http://mrcoles.com/keys/>)
evt:which:keyCode:charCode:modifier:

2015. 4. 20. 오후 7:30:36 >keyup78780

2015. 4. 20. 오후 7:30:36 >keypress110110110

2015. 4. 20. 오후 7:30:36 >keydown78780

2015. 4. 20. 오후 7:30:36 >keyup70700

2015. 4. 20. 오후 7:30:36 >keypress102102102

2015. 4. 20. 오후 7:30:36 >keydown70700

2015. 4. 20. 오후 7:30:35 >keyup72720

2015. 4. 20. 오후 7:30:35 >keypress104104104

2015. 4. 20. 오후 7:30:35 >keydown72720

2015. 4. 20. 오후 7:30:30 >keyup69690

2015. 4. 20. 오후 7:30:30 >keypress101101101

2015. 4. 20. 오후 7:30:30 >keydown69690

2015. 4. 20. 오후 7:30:28 >keyup74740

2015. 4. 20. 오후 7:30:28 >keypress106106106

2015. 4. 20. 오후 7:30:28 >keydown74740

2015. 4. 20. 오후 7:30:27 >keyup83830

2015. 4. 20. 오후 7:30:27 >keypress115115115

2015. 4. 20. 오후 7:30:27 >keydown83830

2015. 4. 20. 오후 7:30:23 >keyup75750

2015. 4. 20. 오후 7:30:22 >keypress107107107

2015. 4. 20. 오후 7:30:22 >keydown75750

2015. 4. 20. 오후 7:30:18 >keyup82820

2015. 4. 20. 오후 7:30:18 >keypress114114114

2015. 4. 20. 오후 7:30:18 >keydown82820

smblott-github:
[...]

PR #1579, opened 2015-04-24
Update exclusion rules when the URL is changed by history.pushState/popState or its hash changes

### LINKED ISSUE #1578: Broken exclusion rules (possibly related to web navigation)
With the exclusion rule: `https?://www.facebook.com/ jk`, `j` and `k` are excluded on the Facebook home page, but not on any other Facebook pages.

snapshot

Clicking on "Stephen", the URL changes, but our active exclusion rule does not.  Changing tab and coming back fixes things (because we check the active exclusion rule on `focus`).  Similarly, now clicking on "Home", the exclusion rule isn't picked up until there's a focus event on the window.

This might have something to do with Web Navigation.

mrmr1993:
This PR fixes #1578, using `chrome.webRequest` to check exclusion rules for frames that changed URL without a page load.

Side effects:
* we need the `webRequest` permission
* introduced `minimum_chrome_version: 41` so we know we can message frames individually
  - current version is 42
* event listeners are always registered, even when Vimium is disabled
  - justification in 6446cf0, the relevant commit

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2,6 +2,7 @@
   "manifest_version": 2,
   "name": "Vimium",
   "version": "1.49",
+  "minimum_chrome_version": "41",
   "description": "The Hacker's Browser. Vimium provides keyboard shortcuts for navigation and control in the spirit of Vim.",
   "icons": {  "16": "icons/icon16.png",
               "48": "icons/icon48.png",
--- a/manifest.json
+++ b/manifest.json
@@ -28,6 +28,7 @@
     "storage",
     "sessions",
     "notifications",
+    "webNavigation",
     "<all_urls>"
   ],
   "content_scripts": [
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -89,12 +89,11 @@ getCurrentTabUrl = (request, sender) -> sender.tab.url
 # Checks the user's preferences in local storage to determine if Vimium is enabled for the given URL, and
 # whether any keys should be passed through to the underlying page.
 #
-root.isEnabledForUrl = isEnabledForUrl = (request, sender) ->
+root.isEnabledForUrl = isEnabledForUrl = (request) ->
   rule = Exclusions.getRule(request.url)
   {
     isEnabledForUrl: not rule or rule.passKeys
     passKeys: rule?.passKeys or ""
-    incognito: sender.tab.incognito
   }
 
 # Retrieves the help dialog HTML template from a file, and populates it with the latest keybindings.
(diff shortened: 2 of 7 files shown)


manifest.json
```diff
@@ -2,6 +2,7 @@
   "manifest_version": 2,
   "name": "Vimium",
   "version": "1.49",
+  "minimum_chrome_version": "41",
```
smblott-github: @mrmr1993.  Thanks for this PR.  I'll take a proper look soon.

What aspect of the PR requires 41?  Is 40 not a bit too recent to be excluding?  It was just released in January (if I have it right).

mrmr1993: `chrome.tabs.sendMessage` only supports messaging individual frames from Chrome 41. #1575 needs 41+ for the same reason.

> Is 40 not a bit too recent to be excluding?

According to [here](http://gs.statcounter.com/#desktop-browser_version-ww-monthly-201504-201504-bar), Chrome 41/42 have 77.8% of Chrome's market total, Chrome 40 has just 2.8%. Still, I'm happy to remove the minimum version and add a check around the version-sensitive code, if that would make you less uncomfortable.

smblott-github:
@mrmr1993.

> chrome.tabs.sendMessage only supports messaging individual frames from Chrome 41. #1575 needs 41+ for the same reason.

Thanks for pointing this out.

My sense is that requiring 41 is leaving too many users out.  I'll rework #1575.  I'm reviewing #1573 and #1575 today, and hope/expect to merge them.  If it's ok with you, I'll then review, rework and probably merge this.

(Separately, whatever Chrome version which introduced `chrome.sessions` is now about 18 months old.  It's probably safe enough to take out the conditional code there now.)

mrmr1993:
> requiring 41 is leaving too many users out.

Both this and your #1575 will need a rewrite to avoid errorring on Chromes <41, then. I'll try to do this after my exam today, and #1575 should be as simple as dropping `{frameId: 0}` and checking that `window == window.top` at the receiving end instead.

> If it's ok with you

It is.

> (Separately, whatever Chrome version which introduced `chrome.sessions` is now about 18 months old. It's probably safe enough to take out the conditional code there now.)

Since we require `chrome.sessions` &mdash; which Chrome 36 doesn't have, &mdash; I believe Chrome 36 users wouldn't have been able to install Vimium 1.49 (released 5 months after Chrome 36 landed) regardless. This means that the conditional was useless from the moment we added the `sessions` permission.

I held off from mentioning this before, but Chrome 36 actually has *higher* usership than Chrome 40 (1.55% vs 1.47% of total market). This was how I justified to myself that killing all Chromes <41 was probably okay, given how we dropped <37 before.

PR #1580, opened 2015-04-25
Update exclusion rules when the URL is changed by history.pushState/popState or its hash changes (v2)

### LINKED ISSUE #1578: Broken exclusion rules (possibly related to web navigation)

smblott-github:
This is @mrmr1993's work from #1579, reworked to eliminate the requirement for Chrome 41 or later.

Changes vis-a-vis #1579:
- When we detect a URL change in the background page, we now just send a message (`checkEnabledAfterURLChange`) to the content script, which triggers a `checkIfEnabledForUrl`, but only in the currently focused window.
- Other minor changes (such as function/message names).

Fixes #1578. (Confirmed.)
Replaces #1579.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -96,14 +96,12 @@ root.isEnabledForUrl = isEnabledForUrl = (request) ->
     passKeys: rule?.passKeys or ""
   }
 
-isEnabledForUpdatedUrl = (details) ->
-  message = isEnabledForUrl details
-  message.name = "updateEnabledForUrlState"
-  chrome.tabs.sendMessage details.tabId, message, {frameId: details.frameId}
+onURLChange = (details) ->
+  chrome.tabs.sendMessage details.tabId, name: "checkEnabledAfterURLChange"
 
 # Re-check whether Vimium is enabled for a frame when the url changes without a reload.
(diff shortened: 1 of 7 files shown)


smblott-github:
@mrmr1993.  I'm going to merge this.  Let me know if you spot any issues. (And thanks for #1579.)

mrmr1993:
Instead of doing an extra round trip of the background page, couldn't we just send the new URL in the request from the bg page, and all frames that match that URL could pick up the updated information? Seems wasteful to recalculate all frames' rules just because one has changed.

I'll try and push a PR with this when I get home.

smblott-github:
> Seems wasteful to recalculate all frames' rules just because one has changed.

We only recalculate the rules for the frame with the focus.  Not for all frames.  See [here](https://github.com/philc/vimium/blob/master/content_scripts/vimium_frontend.coffee#L611).

PR #1581, opened 2015-04-25
Simplify front-end settings logic.

smblott-github:
A couple of front-end settings-logic changes...

- Simplify the logic.
- Send a single request for all required settings.
- Eliminate the need to count messages (which feels fragile).

(This probably isn't the last word in how settings are handled, but it's better than the status quo.  We replace 24 messages *every time a frame receives the focus* with just two.)


mrmr1993:
LGTM, this is a huge improvement.

PR #1590, opened 2015-04-27
Ensure that the Vomnibar iframe background is transparent

### LINKED ISSUE #1583: Grey box at bookmarks bar with Hacker Vision after 1.50 update
I use the [Hacker Vision](https://chrome.google.com/webstore/detail/hacker-vision/fommidcneendjonelhhhkmoekeicedej) extension together with Vimium. After upgrading to 1.50, a huge grey box appears when bringing up the bookmarks/URL/history bar. It worked fine in 1.49.

With 1.49:

selection_031

With 1.50:

selection_032

mrmr1993:
Some extensions (eg. Hacker Vision) override the background colour of iframes to make the screen dark, even when the frame is already set to transparent. This PR gives our rule higher priority, so that the frame is still rendered as transparent, and doesn't obscure the content beneath.

This fixes #1583.

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -275,7 +275,7 @@ body.vimiumFindMode ::selection {
 /* Vomnibar Frame CSS */
 
 iframe.vomnibarFrame {
-  background-color: transparent;
+  background-color: transparent !important;
   padding: 0px;
   overflow: hidden;
 


smblott-github:
Thanks, @mrmr1993.  Works nicely.

PR #1595, opened 2015-04-27
Don't focus <input type="submit" /> elements from link hints

### LINKED ISSUE #1584: <input type="submit"> buttons activated but steal focus
#1342 fixed the clickability issue (with [f]) of \<input type="submit"\> buttons but these buttons still steal the focus once clicked with [f]. Pressing [Esc] is required to return to normal input flow.

Prior to the 1.47 regression and loss of clickability this was not the case.

mrmr1993:
This fixes #1584.


smblott-github:
Thanks, @mrmr1993.

PR #1599, opened 2015-04-27
Separate Settings from the background page, and use it directly in options.html/popup.html

mrmr1993:
This PR decouples the `Settings` object from the background page, with a view to merging it with `settings` in the front-end, creating a unified `chrome.storage`-backed implementation.

It also uses the decoupled `Settings` to provide settings information directly to the options page and the browser icon popup, instead of via the background page.

The diff here is really quite unfriendly, but I've tried to keep every commit so that it
* only makes one, small & obvious, set of changes, 
* compiles/passes tests, and
* behaves the same (from a black-box perspective) as `master`.


smblott-github:
Thanks @mrmr1993.  I'll look at this soon, once the 1.50 issues calm down.

mrmr1993:
Rebased onto master.

mrmr1993:
Rebased onto master.

mrmr1993:
Rebased onto master.

smblott-github:
@mrmr1993.  I looked at this a while ago (but forgot to comment -- sorry!).  Thoughts...

Does this go far enough?

It would be good to:
- Get rid of `settings` in the front end entirely.
- Use a single `Settings.get "setting"` method uniformly in both the front end and the back end.
- Be able to write `Settings.use "setting", (value) -> ...` anywhere in the front end or back end, without having to consider whether `setting` is cached or not.

mrmr1993:
> Does this go far enough?

I was already planning to do all of those things, this is just a step along the way. The diff is already unweildy, so I decided not to make it worse by touching all the frontend code that uses `settings` too.

If there are no issues, it would be great to see this merged so I have a base to start working on unified settings from.

mrmr1993:
Rebased onto master.

lib/settings.coffee
```diff
@@ -0,0 +1,202 @@
...
+  if Utils.isBackgroundPage()
+    settingsCache = localStorage
+  else
+    settingsCache = extend {}, localStorage # Make a copy of the cached settings from localStorage
+else
+  settingsCache = {}
+
+root.Settings = Settings =
+  cache: settingsCache
+  init: -> Sync.init()
```
smblott-github: Should we only initialise `Sync` on the background page?

lib/settings.coffee
```diff
@@ -0,0 +1,202 @@
...
+      @clear(key)
+    else
+      jsonValue = JSON.stringify value
+      @cache[key] = jsonValue
+      Sync.set key, jsonValue
+
+  clear: (key) ->
+    if @has key
+      delete @cache[key]
+    Sync.clear key
```
smblott-github: Should all of these `Sync` operations be...

```Coffeescript
... if Utils.isBackgroundPage()
```

lib/settings.coffee
```diff
@@ -0,0 +1,202 @@
...
+  has: (key) -> key of @cache
+
+  # For settings which require action when their value changes, add hooks to this object, to be called from
+  # options/options.coffee (when the options page is saved), and by Settings.storeAndPropagate (when an
+  # update propagates from chrome.storage.sync).
+  postUpdateHooks: {}
+
+  # postUpdateHooks convenience wrapper
+  performPostUpdateHook: (key, value) ->
+    @postUpdateHooks[key] value if @postUpdateHooks[key]
```
smblott-github: This is probably inherited from before, but the Coffeescript way would be...

```Coffeescript
@postUpdateHooks[key]? value
```

lib/settings.coffee
mrmr1993: No. Starting to use `chrome.storage` (via `Sync`) everywhere *is the point* of this PR*, so we can stop moving settings data between contexts manually.

tests/unit_tests/commands_test.coffee
```diff
@@ -1,5 +1,6 @@
 require "./test_helper.js"
 extend global, require "./test_chrome_stubs.js"
+global.Settings = {postUpdateHooks: {}}
```
smblott-github: I know there's a lot of legacy usage around, but does the style guide not say...

```Coffeescript
global.Settings = { postUpdateHooks: {} }
```

(That is, with spaces.  I usually try to tidy these up en passant.)

lib/settings.coffee
mrmr1993: Same as my above comment.

tests/unit_tests/commands_test.coffee
mrmr1993: > does the style guide not say

No. The style guide says the opposite.

lib/settings.coffee
mrmr1993: I've pushed a commit to make this change.

tests/unit_tests/commands_test.coffee
smblott-github: Fair enough.  @philc always asks me to change it as suggested.

smblott-github:
I can see the elegance of what you're getting at with this.  It's not too far from the three "should-we-not-go-further" issues I mentioned above.

Idea:
- When using this in the front end, you'll have to deal with `isLoaded` and `load` (or event listeners).  That's easily done with an `AsyncDataFetcher`.

smblott-github:
Thanks, @mrmr1993.  It would be great if you could push on with the next steps.

PR #1604, opened 2015-04-28
Don't reorder frames when they are focused manually (eg. by clicking)

mrmr1993:
This PR ensures the order of `frameId`s per-tab stays consistent, so `gf` always lands on the same frame. Previously, the focused frame was extracted from the list and moved to the front, so clicking frames to focus them would reorder the list.


smblott-github:
Very nice.

Issue #1594, opened 2015-04-27
In v1.50, the vomnibar is always visible and won't close on certain sites


ManiacalAce:
I am on Chrome Version 39.0.2171.71 (64-bit) on Arch Linux.
When I open this site: http://motorbeam.com/, the vomnibar is always visible and won't close or even work.

smblott-github:
Yikes.  Confirmed.
What's more, `Ctrl-r` doesn't event get rid of it...

mrmr1993:
If we create the Vomnibar (and probably all UIComponents, or even the whole Vimium UI) in a shadow DOM and just inject vomnibar.css there, we can stop competing with page css altogether.

smblott-github:
#1613 seems to fix this.
Not sure if it's the best approach.

PR #1605, opened 2015-04-28
Only initialize the vomnibar when enabled.

smblott-github:
We preload the vomnibar.  We probably shouldn't do that on sites where Vimium is disabled.

(This is not a solution for #1594, but was inspired by the fact that even disabling Vimium isn't enough to fix that issue.)


mrmr1993:
Now the Vomnibar for all frames is only loaded in the top frame, this needs to initialise it if *any* frame is enabled.

smblott-github:
> Now the Vomnibar for all frames is only loaded in the top frame, this needs to initialise it if any frame is enabled.

Excellent point.
Closing.

PR #1607, opened 2015-04-28
Make getVisibleClientRect behave as expected by default

mrmr1993:
This PR changes `DomUtils.getVisibleClientRect` to require passing an extra truthy argument in order to access the (generally) unexpected behaviour of sometimes returning the rects of child elements. All locations in the code that *actually* wanted this behaviour have been updated to continue to use it.

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -55,21 +55,25 @@ DomUtils =
   #
   # Returns the first visible clientRect of an element if it exists. Otherwise it returns null.
   #
-  getVisibleClientRect: (element) ->
+  # WARNING: If testChildren = true then the rects of visible (eg. floated) children may be returned instead.
+  # This is used for LinkHints and focusInput, **BUT IS UNSUITABLE FOR MOST OTHER PURPOSES**.
+  #
+  getVisibleClientRect: (element, testChildren = false) ->
     # Note: this call will be expensive if we modify the DOM in between calls.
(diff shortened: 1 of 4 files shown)


smblott-github:
Could you provide more info as to why you're proposing this?  Is it because of (incorrect) use of `DomUtils.getVisibleClientRect` in the scroller and elsewhere?

mrmr1993:
> Is it because of (incorrect) use of `DomUtils.getVisibleClientRect` in the scroller and elsewhere?

I'm afraid so. Before the visual mode ones popped up, I'd been trying to rename it to something less inviting. Since the child-less version is useful too, this is probably an easier and better way to handle it.

lib/dom_utils.coffee
```diff
@@ -80,9 +84,7 @@ DomUtils =
 
         # eliminate invisible elements (see test_harnesses/visibility_test.html)
         computedStyle = window.getComputedStyle(element, null)
-        if (computedStyle.getPropertyValue('visibility') != 'visible' ||
-            computedStyle.getPropertyValue('display') == 'none')
-          continue
+        continue if computedStyle.getPropertyValue('visibility') != 'visible'
```
smblott-github: Re. your earlier explanation: ok.
On this line, though, why are we dropping the test on `display`/`none`?

mrmr1993: `Element.getClientRects()` returns no rects when `display: none;` is set on the element, so that check will never pass.

Issue #1588, opened 2015-04-27
Search text box on github "stealing" focus after "follow next link"


infokiller:
This is a similar issue to the one I opened for [cVim here](https://github.com/1995eaton/chromium-vim/issues/276).

smblott-github:
Thanks @infokiller.  I don't understand how reproduce this.

infokiller:
Sorry for the lack of clarity on my part. To reproduce:
1. Set "Don't let pages steal the focus on load" in Vimium settings.
2. Do any search on Github, like [this one](https://github.com/1995eaton/chromium-vim/issues/276).
3. Execute "follow next link" (by default it is mapped to ]]).

infokiller:
Urgh, the search link example is wrong, try this one:
https://github.com/search?utf8=%E2%9C%93&q=whatever

smblott-github:
OK.  I see it now.

The issue is that the `]]` is not loading a new page (on github).  It's doing a web navigation.  No-grab-focus should probably apply in that case too.

mrmr1993:
> It's doing a web navigation. No-grab-focus should probably apply in that case too.

There's no API for detecting `history.pushState`, and detecting the navigation in the background page is asynchronous, and so will open us up to race conditions. Can I suggest that when a user uses Vimium to click a link which has
* `link.href.split("#")[0] != document.location.toString().split("#")[0]`
* `link.target == "" `
then we call `new GrabBackFocus()`? Most `history.pushState` links take this form to support older browsers (and if it's a normal link then there'll be a redirect anyway) so activating `GrabBackFocus` mode should be safe to do.

smblott-github:
Are you sure there's a race condition, @mrmr1993?  E.g. 061e99f895e5655d351ad1585af028a12abf3ec0 seems to work.

(If there's a race condition with 061e99f895e5655d351ad1585af028a12abf3ec0, then there's also a race condition with #1580, is there not?)

mrmr1993:
> Are you sure there's a race condition, @mrmr1993?

Yes. We're doing an asynchronous round-trip of the background page, so the user may have manually selected a textbox 'while we were away', especially on slower computers or when the page runs `history.pushState` 'late' (eg. after images have finished loading).

> then there's also a race condition with #1580, is there not?

Yes, but there isn't any better way of detecting URL changes that I'm aware of.

Other thoughts:
[...]

PR #1608, opened 2015-04-28
Do not grab back focus if Vimium is disabled.

smblott-github:
Thinking about #1588, I realised that we should only be grabbing back the focus on pages where Vimium is enabled.


mrmr1993:
LGTM

PR #1609, opened 2015-04-28
Fix grab back focus after web navigation

### LINKED ISSUE #1588: Search text box on github "stealing" focus after "follow next link"

smblott-github:
@mrmr1993:  I included `form_submit` (in addition to `link`) in the list.  What do you think?

Fixes #1588.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -100,7 +100,7 @@ root.isEnabledForUrl = isEnabledForUrl = (request, sender) ->
   }
 
 onURLChange = (details) ->
-  chrome.tabs.sendMessage details.tabId, name: "checkEnabledAfterURLChange"
+  chrome.tabs.sendMessage details.tabId, extend details, name: "checkEnabledAfterURLChange"
 
 # Re-check whether Vimium is enabled for a frame when the url changes without a reload.
 chrome.webNavigation.onHistoryStateUpdated.addListener onURLChange # history.pushState.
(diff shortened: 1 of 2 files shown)


mrmr1993:
Code LGTM, I'll try and test it when I have some time later.

smblott-github:
Merging.
Seems to work well, and if there are any issues or gotchas it's not hard to unwind.

smblott-github:
> ... and if there are any issues or gotchas it's not hard to unwind.

And there are gotchas.

This breaks search on Google Inbox.  `/` invokes a navigation, so the search box loses the focus.  The UX resolving things is ugly.

The lesson is that, when it comes to grabbing back the focus, we should tread lightly.

Reverted.

mrmr1993:
Broken record, but: my suggestion in #1588 is more conservative, and will fix the issue. I'll open a PR for it at some point.

smblott-github:
> I'll open a PR for it at some point.

Suggest you hold off.  You'll just run into the same types of issue.

(I'm wondering whether we should can this grab-back-focus lark entirely.)

mrmr1993:
> You'll just run into the same types of issue.

Shouldn't do. I considered the types of breakage (including the one you ran into) when I was devising my solution. You can see me predicting this bug in #1588.

PR #1613, opened 2015-04-28
Fix over-riding of vomnibar display CSS.

### LINKED ISSUE #1594: In v1.50, the vomnibar is always visible and won't close on certain sites

smblott-github:
@mrmr1993.  This seems to fix #1594.  I'm no CSS expert.  Is this safe?

Fixes #1594.

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -279,7 +279,6 @@ iframe.vomnibarFrame {
   padding: 0px;
   overflow: hidden;
 
-  display: block;
   position: fixed;
   width: calc(80% + 20px); /* same adjustment as in pages/vomnibar.coffee */
   min-width: 400px;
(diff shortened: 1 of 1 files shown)


mrmr1993:
This is fine as a quick fix, LGTM.

As I mentioned in #1594, we should consider using a shadow DOM to stop these problems altogether. Otherwise we're always in danger of inheriting some weird page style that is `important!`, and there's not a lot we can do about it.

smblott-github:
> As I mentioned in #1594, we should consider using a shadow DOM...

Yes. (But I'm not confident doing that.)

mrmr1993:
> But I'm not confident doing that.

I'll throw together a PR for you to have a look at shortly.

smblott-github:
OK, thanks.  I'll merge this for now.

Issue #853, opened 2013-05-23
Shadow DOM problem/suggestion


bliker:
Hi!
I am currently finishing porting my extension to use _Shadow DOM_. I was wrestling with the same issue like vimium, where css of page I was injecting into has been modifying css of the extension. Shadow DOM separates extension from other styles, Very neat.

But when I have Input element inside of `#documet-fragment` vimium still reacts to typing like  I am not in input.
[Example on JsBin](http://jsbin.com/asufis/2)

if you want to check my [extension](https://github.com/bliker/magickyslovnik/tree/shadow). It is in Slovak, but code is in English.

mrmr1993:
The example above no longer works now `Element.webkitCreateShadowRoot` has been depreciated. Updated example [here](http://jsfiddle.net/0xL12q5w/).

We can detect that an element has a shadow root by checking that `element.shadowRoot != null`, and further can capture `focus` events after one fires on the shadow root with `element.shadowRoot.addEventListener("focus", listener, true)`. Tying this together, we should be able to activate input mode for shadow DOM elements to kill this bug.

mrmr1993:
This was fixed by #1627.

Issue #599, opened 2012-07-25
Current selection wiped out when a vomnibar is opened


mdegat01:
Opening the vomnibar switches focus to a new element on the page so any selection you had is gone.  This poses a problem with the "b" vomnibar, specifically with bookmarklets, many bookmarklets pick up the selection from the current page and operate on it.  By eliminating the ability to select many bookmarklets won't work with Vimium (which is really a shame because having it autocomplete bookmarklets available based on the function you are looking for and then execute them is excellent).

Since the vomnibar is an input on the page and you have to focus on the vomnibar to submit there is no way the selection can be maintained during entry.  So I think that leaves two options: have a new keystroke for submit that tells the vomnibar to wait for a selection then eval the bookmarklet (alt/cmd+enter? ctrl+enter? something with right-arrow key?), or attempt to save the selection during the b entry and then re-select it before evalling the bookmarklet.

The first one should be relatively easy by having the eval fire in a listener for the onselect event but it is clunkier, users must learn a new button and are probably going to have to reselect text a few times before they get used to it.  The second one would probably surprise people the first time it happened but it seems more intuitive and seemless.  This post http://webcloud.se/log/Selecting-text-with-JavaScript/ and this one http://www.webdeveloper.com/forum/showthread.php?t=73493 lead me to believe it's possible but it definitely looks like a PITA.

Of course there could be other solutions as well but there are a lot of bookmarklets out there that use this so it would be a shame if users had to use their mouse for those ones.

mrmr1993:
This is fixed in master.

PR #1614, opened 2015-04-28
Use a shadowDom for the Vomnibar (and all UIComponents)

mrmr1993:
This wraps `UIComponent.iframeElement` in a shadow DOM, so that page CSS doesn't override our own.

@smblott-github suggest this instead of #1613?

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -5,11 +5,16 @@ class UIComponent
   options: null
 
   constructor: (iframeUrl, className, @handleMessage) ->
-    styleSheet = document.createElement "link"
-    extend styleSheet,
-      rel: "stylesheet"
-      type: "text/css"
-      href: chrome.runtime.getURL "content_scripts/vimium.css"
+    styleSheet = document.createElement "style"
+    styleSheet.type = "text/css"
+    # Default to everything hidden while the stylesheet loads.
+    styleSheet.innerHTML = "* {display: none !important;}"
+    # Load stylesheet.
+    xhr = new XMLHttpRequest()
+    xhr.onload = (e) -> styleSheet.innerHTML = xhr.responseText
+    xhr.open "GET", chrome.runtime.getURL("content_scripts/vimium.css"), true
+    xhr.send()
+
     @iframeElement = document.createElement "iframe"
     extend @iframeElement,
       className: className
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -5,12 +5,24 @@ class UIComponent
   options: null
 
   constructor: (iframeUrl, className, @handleMessage) ->
+    styleSheet = document.createElement "link"
+    extend styleSheet,
+      rel: "stylesheet"
+      type: "text/css"
+      href: chrome.runtime.getURL "content_scripts/vimium.css"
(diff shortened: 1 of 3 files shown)


smblott-github:
Thanks, @mrmr1993.  I'm not sure if you're finished yet (a new commit popped up since I last looked). However, at 1d16e4e996aa5234020a73212d74834ba99ec162, this isn't actually working for http://www.motorbeam.com/.

(Nothing interesting in the console.)

mrmr1993:
Sorry, I think I've found the error (a typo), I'm testing now and then I'll push.

mrmr1993:
@smblott-github I've updated this, it should be working as intended now

content_scripts/ui_component.coffee
```diff
@@ -5,12 +5,29 @@ class UIComponent
...
   constructor: (iframeUrl, className, @handleMessage) ->
+    styleSheet = document.createElement "style"
+    styleSheet.type = "text/css"
+    # Default to everything hidden while the stylesheet loads.
+    styleSheet.innerHTML = "* {display: none !important;}"
+    # Load stylesheet.
+    xhr = new XMLHttpRequest()
+    xhr.onload = (e) -> styleSheet.innerHTML = xhr.responseText
+    xhr.open "GET", chrome.runtime.getURL("content_scripts/vimium.css"), true
+    xhr.send()
```
smblott-github: Why do you need the `XMLHttpRequest` for the style sheet but not for the iframe contents?

(Looks odd.)

mrmr1993: Shadow DOM doesn't support `<link>` elements, but it does support `<iframes>`.

smblott-github:
> Shadow DOM doesn't support <link> elements

Well it does seem to, at least to some extent.  One of [these two lines](https://github.com/philc/vimium/blob/master/pages/vomnibar.html#L11) is needed (and one apparently not).

mrmr1993:
> Well it does seem to, at least to some extent.

You can't have a `<link>` element in the shadow DOM (at least, not in Chrome). The iframe DOM is just a normal DOM and can contain all the usual things, though.

> One of [these two lines](https://github.com/philc/vimium/blob/master/pages/vomnibar.html#L11) is needed (and one apparently not).

Don't we need styles for `.vimiumReset` from `vimium.css` to keep the base style consistent with the rest of the extension? Also, the contents of the iframe should be orthogonal to this PR...?

smblott-github:
@mrmr1993.  This seems to work nicely.  I'm not seeing any issues.
We have three options for 1.51 (which will probably happen quite soon -- bug fixes):

1. Stick with #1613 (it works for the original use case).
1. Go with this.

You think we should go with this *and now*, @mrmr1993?

Also... with the Vomnibar now effectively a web component (yes?), what benefit are we getting from involving an iframe?

mrmr1993:
> You think we should go with this *and now*, @mrmr1993?

Yes.

> what benefit are we getting from involving an iframe?

* It retains the current focused element and selection (#599, which can be closed), so bookmarklets can use them.
* Having an in-page Vomnibar lets the page open arbitrary URLs, extract bookmarks, etc. by manipulating the `<input>` and its contents.
* Dealing with `<input>`s in shadow DOMs is fiddly (#853).

PR #1616, opened 2015-04-28
Fix for #1554.

mrmr1993:
Elements with `font-size: 0px; display: inline;` will declare a height of zero, even if a child with `display: inline;` and non-zero `font-size` contains text. This PR adds a work-around to deal with this case, fixing #1554.


smblott-github:
Thanks, @mrmr1993.

PR #1618, opened 2015-04-29
Ignore keyboard repeats in linkHint key handler

### LINKED ISSUE #1617: Open link in new tab workflow has changed
Hi Folks,

Has the open link in new tab workflow changed?

Prior to 1.50 I performed the following to open a new link:

> press 'f' > press 'shift' > press link hint > tab opens in background

However since 1.50 the workflow has changed to:

> press 'shift+f' > press link hint > tab opens in background

The difference here, at least for me, is that in previous versions the act of opening a new link was decided after pressing 'f'. Specifically, the action of opening a link hint was bound to 'f' and didn't require you to decide which action you're taking before hand. I feel that this is a much better way to work.

If the intention was to change the functionality, is it possible to make it an option of choosing between the new way and the old way. If not, then could you please let me know how to revert to (and keep) 1.49 which allows the previous way of working.

Incidentally, love this extension, been using it for ages and for a vi lover it's great.

Thanks much :)

mrmr1993:
This fixes #1617.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -262,7 +262,7 @@ LinkHints =
   # Handles shift and esc keys. The other keys are passed to getMarkerMatcher().matchHintsByKey.
   #
   onKeyDownInMode: (hintMarkers, event) ->
-    return if @delayMode
+    return if @delayMode or event.repeat
 
     if ((event.keyCode == keyCodes.shiftKey or event.keyCode == keyCodes.ctrlKey) and
         (@mode == OPEN_IN_CURRENT_TAB or


smblott-github:
@mrmr1993.  Thanks for this.  I can't reproduce the problem (no repeats on modifiers).  But the "fix" seems safe enough.

ic80:
Hi, thanks for adding a fix in, it's very much appreciated.

Keep being awesome :D

philc:
Awesome guys. I wonder if this is what was causing people issues with shift
on link hints historically.

mrmr1993:
> I can't reproduce the problem (no repeats on modifiers)

Issue was Windows-only, judging from the "repeats" section [here](http://www.quirksmode.org/dom/events/keys.html) and the description of the issue.

PR #1619, opened 2015-04-29
Fix matching text for text link hints

smblott-github:
This ensures that the texts used for filtering link hints are generated by visiting descendants before their ancestors.  There's an explanation of what's intended in this commit: b7535a604954b5873d825eb66bfecd08f1f2c99b.  However, it turns out we actually need to sort the nodes into the intended order.  That's this PR.

To see what's going on:
- enable `￼ Use the link's name and numbers for link-hint filtering`
- visit [this page](http://www.reddit.com/r/peloton/comments/348rxh/race_thread_2015_tour_of_romandie_stage_2/)
- try to select `Profile Map` by typing the text.

In `master` currently, we can't -- because we're visiting the elements in the wrong order.  We end up instead picking some (pretty useless) ancestor.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -64,8 +64,8 @@ LinkHints =
       # When using text filtering, we sort the elements such that we visit descendants before their ancestors.
       # This allows us to exclude the text used for matching descendants from that used for matching their
       # ancestors.
-      textLength = (el) -> el.element.textContent?.length ? 0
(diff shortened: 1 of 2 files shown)


smblott-github:
Just to note, I had previously thought that [this reverse](https://github.com/philc/vimium/blob/master/content_scripts/link_hints.coffee#L243) was enough to achieve the necessary ordering.  It turns out it is not.

smblott-github:
Merging this.
For the record, it fixes an oversight in #1514.

PR #1621, opened 2015-04-30
URL decode bookmarklets, if necessary.

### LINKED ISSUE #1611: Bookmarklet doesn't run when executed through vomnibar
If I hit `b` and in the vomnibar try to run my "+ Pocket" bookmarklet (by hitting ENTER) nothing happends. If I click the bookmark manually in the bookmarks bar everything works fine. Worked fine a couple of days ago.

smblott-github:
This is one possible approach to #1611.

It appears that, in some workflows, Chrome is storing URL encoded bookmarklets.  `chrome.tabs.update` appears to only handle non-URL encoded bookmarklets.  Therefore, for bookmarklets, we test whether the URL has been encoded, and decode it if necessary.

See the diff.  This uses a number of heuristics to test whether the bookmarklet has been URL encoded or not.  Is there a better way? Or are there better heuristics?  The tricky case is a URL-encoded string within an otherwise non-URL encoded URL.

Fixes #1611 (maybe!).

### DIFF
--- a/lib/utils.coffee
+++ b/lib/utils.coffee
@@ -39,19 +39,6 @@ Utils =
     urlPrefix = new RegExp "^[a-z]{3,}://."
     (url) -> urlPrefix.test url
 
-  # Apply heuristics to test whether a string (such as a "javascript:" URI) has been URL encoded.
-  isUrlEncoded: (string) ->
-    # If it doesn't contain "%", it's not URL encoded.
-    return false if -1 == string.indexOf "%"
-    # If it contains certain non-URL characters, then it's not URL encoded.
-    for char in [ " ", "{", "}", ";", '"', "'" ]
-      return false if 0 <= string.indexOf char
-    # If it contains any of these non-ascii characters, then it's not URL encoded.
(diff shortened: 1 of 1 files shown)


mrmr1993:
It's straightforward to test that, for bookmarks and the omnibox, Chrome *always* decodes the URL. For example,

```js
javascript:alert(25%20);alert("#$%&+,/:;<=>?@[\]^`{|}")
```

will still behave differently via `chrome.tabs.update` (the first alert will show `5`) vs the omnibox/bookmarks (the first alert will show `25`). So the right thing to do, until [Chromium issue 483000](https://code.google.com/p/chromium/issues/detail?id=483000) is fixed, is to always decode URLs.

smblott-github:
Ok, thanks @mrmr1993.

So your point is:
- In the omnibox, Chrome blindly URL decodes `javascript:` URLs.
- When, clicked, Chrome blindly URL decodes `javascript:` URLs.
- `chrome.tabs.update` never URL decodes URLs.

So we should blindly decode them.

smblott-github:
Thanks for your input, @mrmr1993.

PR #1622, opened 2015-04-30
Make GrabBackFocus work for javascript page transitions from links

### LINKED ISSUE #1588: Search text box on github "stealing" focus after "follow next link"

mrmr1993:
Pages can load new content dynamically and change the displayed URL using `history.pushState`. Since this can often be indistinguishable from an actual new page load for the user, we should also re-start `GrabBackFocus` for these as well.

This PR does this, by restarting `GrabBackFocus` mode whenever a link is clicked which points to a URL that could be moved to via a `history.pushState`. The expectation is that such a link will either navigate to the URL as normal, or the navigation will br simulated in Javascript.

This fixes #1588.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -135,8 +135,9 @@ class GrabBackFocus extends Mode
 
 # Pages can load new content dynamically and change the displayed URL using history.pushState. Since this can
 # often be indistinguishable from an actual new page load for the user, we should also re-start GrabBackFocus
-# for these as well.
+# for these as well. This fixes issue #1622.
 handlerStack.push
+  _name: "GrabBackFocus-pushState-monitor"
   click: (event) ->
     target = event.target
     while target
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -139,6 +139,9 @@ class GrabBackFocus extends Mode
 handlerStack.push
   _name: "GrabBackFocus-pushState-monitor"
   click: (event) ->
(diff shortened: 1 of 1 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -133,6 +133,28 @@ class GrabBackFocus extends Mode
...
+# Pages can load new content dynamically and change the displayed URL using history.pushState. Since this can
+# often be indistinguishable from an actual new page load for the user, we should also re-start GrabBackFocus
+# for these as well.
+handlerStack.push
+  click: (event) ->
+    target = event.target
+    while target
+      # Often, a link which triggers a content load and url change with javascript will also have the new
+      # url as it's href attribute.
+      if target.tagName == "A" and
```
smblott-github: I like this approach, @mrmr1993.

Four comments:
1. Could you add a reference to the original issues?
2. Elsewhere, we use `toLowerCase` for comparisons like this.  Do we need that here too?
3. If you add a `_name` property, we get better debugging in the handler stack.
4. Is it possible to have an input embedded within an element which satisfies the condition here?  Something like:

```HTML
<a> .... <input> ... </input> </a>
```

mrmr1993: > Elsewhere, we use `toLowerCase` for comparisons like this.

For the built-in elements, `tagName` is always upper case, so there's no need to do this.

I've pushed commits dealing with all your other comments.

smblott-github:
This is a big win!.  Thanks, @mrmr1993.

PR #1623, opened 2015-05-01
Rename UIComponent visible class to match the one used in the stylesheet

mrmr1993:


smblott-github:
Wow.  Since that style name isn't mentioned in the style sheets...
- How on earth was this working at all?
- If it was working before, then why do we need it at all?

mrmr1993:
The default style on the Vomnibar was doing the job; there's some redundancy in the style sheet. It would make life easier to have this in so things work OOTB for the HUD and the help dialog.

smblott-github:
OK, thanks.

PR #1624, opened 2015-05-01
Stop caching Vomnibar completers so we don't handle stale responses

mrmr1993:
With cached completers, Vomnibar completions from the background page for an old instance of the Vomnibar can still be received by a new instance. For example, typing `bkeyword<enter>b`quickly can bring up completions for `keyword` in the new instance. This is easy to reproduce on slower computers.

This PR removes the cache of completers from the Vomnibar, so every Vomnibar instance opens a new port, fixing the issue.

### DIFF
--- a/pages/vomnibar.coffee
+++ b/pages/vomnibar.coffee
@@ -6,12 +6,8 @@
 Vomnibar =
   vomnibarUI: null # the dialog instance for this window
   getUI: -> @vomnibarUI
-  completers: {}
 
-  getCompleter: (name) ->
-    if (!(name of @completers))
-      @completers[name] = new BackgroundCompleter(name)
-    @completers[name]
+  getCompleter: (name) -> new BackgroundCompleter(name)
 
   #
   # Activate the Vomnibox.


smblott-github:
Good idea.  We have so many of these race conditions.

I *think* I've seen cases where you type `abc`, but only see completions for `ab`. It looks like somebody foresaw the issue, but didn't quite complete the implementation (`id`, [here](https://github.com/philc/vimium/blob/master/pages/vomnibar.coffee#L217), is unused).

An alternative would be [this](https://github.com/philc/vimium/compare/master...smblott-github:alternative-drop-stale-vomnibar-completer-responses).  The diff is bigger than the change because of an indentation change.  Basically, lines referring to `messageId` are new.  This version also fixes race conditions within a single Vomnibar instantiation.

mrmr1993:
> An alternative would be [this](https://github.com/philc/vimium/compare/master...smblott-github:alternative-drop-stale-vomnibar-completer-responses).

That LGTM, I'm more than happy to go in that direction instead. Closing.

PR #1627, opened 2015-05-01
Enter insert mode if an input inside a shadow DOM is focused

mrmr1993:
This fixes issue #853. Testbed [here](http://jsfiddle.net/0xL12q5w/).

### DIFF
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -19,6 +19,9 @@ class InsertMode extends Mode
         # the right thing to do for most common use cases.  However, it could also cripple flash-based sites and
         # games.  See discussion in #1211 and #1194.
         target.blur()
+      else if target.shadowRoot and @insertModeLock
+        # An editable element in a shadow DOM is focused; blur it.
+        @insertModeLock.blur()
       @exit event, event.srcElement
       @suppressEvent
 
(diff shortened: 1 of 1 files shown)


content_scripts/mode_insert.coffee
```diff
@@ -53,6 +56,21 @@ class InsertMode extends Mode
...
+        else if event.target.shadowRoot
+          # A focusable element inside the shadow DOM might have been selected. If so, we can catch the focus
+          # event inside the shadow DOM. This fixes #853.
+          shadowRoot = event.target.shadowRoot
+          eventListeners = {}
+          for type in ["focus", "blur"]
+            eventListeners[type] = do (type) ->
+              (event) ->
+                handlerStack.bubbleEvent type, event
+                if type == "blur"
```
smblott-github: @mrmr1993.  What if there are two or more inputs inside the shadow DOM, and one is first focused, then another (and perhaps then another too)?  Wouldn't the `blur` events cause these listeners to be removed before the shadow DOM itself has lost the focus?  Might b40647c48ae7d8016aa77c418eed42c042000f46 be better?

smblott-github:
Ping.
(Since you seem to be active, @mrmr1993.  And I'll get to the Settings PR soon.)

content_scripts/mode_insert.coffee
mrmr1993: Yes, b40647c48ae7d8016aa77c418eed42c042000f46 looks like a better alternative. I have no computer this week, so I haven't tested it.

smblott-github: OK.  This is a nice improvement.  I'll merge it, then cherry pick b40647c48ae7d8016aa77c418eed42c042000f46.

PR #1629, opened 2015-05-01
Remove background page logging.

smblott-github:
Why do we still have this background page logging?  We should remove it.

snapshot

It rarely serves a purpose.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -506,12 +506,9 @@ splitKeyQueue = (queue) ->
 handleKeyDown = (request, port) ->
   key = request.keyChar
   if (key == "<ESC>")
-    console.log("clearing keyQueue")
     keyQueue = ""
   else
-    console.log("checking keyQueue: [", keyQueue + key, "]")
     keyQueue = checkKeyQueue(keyQueue + key, port.sender.tab.id, request.frameId)
-    console.log("new KeyQueue: " + keyQueue)
   # Tell the content script whether there are keys in the queue.
   # FIXME: There is a race condition here.  The behaviour in the content script depends upon whether this message gets
   # back there before or after the next keystroke.


mrmr1993:
We (or at least I) have used it quite often in past to debug issues with users. There must be a better way to log/place to log to, but the information itself still has value for now.

smblott-github:
If it's so important, then we should log the equivalent info in the content script.  The background page could be running for weeks on end.  And there's a whopping amount of unnecessary logging when you press and hold `j`, say.

mrmr1993:
> we should log the equivalent info in the content script.

This is hostile towards web developers who are trying to use the console for their own separate debugging. Please lets not do this.

I was thinking, we should have a dedicated page for logs which
* is openable (in an inconspicuous way) from the options page and the popup;
* only logs when open (removing the bloat of days/weeks of messages);
* is filterable, so it's possible to only look at eg. tab registrations or key handling.

This would mean we could stop having to get users to be in developer mode while we debug and lets us (and them) have more control over the logs they're handing over.

gdh1995:
What we badly need is to tell users what key mappings are wrong when we open options.html. Any idea?

smblott-github:
I still think we should do this, @mrmr1993.

I don't recall many (any?) cases where logging has proved useful since this this was posted.

What about a variable on the background page that we can ask users to turn on if they're having an issue?

mrmr1993:
Every time we have been debugging an issue with keyboard layout it's been useful. I'm not willing to keep debugging them without, so only kill the logging if you can and will.

I still think my idea above is a good alternative (and more straighforward for users). But none of this is urgent.

smblott-github:
@mrmr1993.  I agree, your idea above (separate page for logging) is a good one.

If I were to merge this PR, would that motivate you to actually implement that idea?

mrmr1993:
I've implemented a very basic logging page in a branch [here](https://github.com/mrmr1993/vimium/tree/logging-page).

smblott-github:
That's not a bad start, @mrmr1993.

- It's a fair amount of work to do for *every* keystroke.  It might be better to track the open logging tabs as they're opened and closed, rather than on every keystroke.
- This is a good gotcha... We should not log keystrokes in the logging tab itself!  Try going to the logging tab and pressing and holding `j`.

PR #1632, opened 2015-05-02
Search engine completion [WIP]

smblott-github:
We should consider adding search-engine completion to the vomnibar.  The issues to consider are below, but for the moment it looks like this...

snapshot

(Note: I have never actually searched for Floyd Mayweather before. `booky` is my own search engine, for most people that would probably be Google.)

There are a number of issues:
- Add search engine completion to the existing "omni" vomnibar (so, *not* a new type of search).
- Scoring search suggestions needs to be done in a way that brings those suggestions to the top when that's what the user is looking for, but otherwise doesn't displace useful vomnibar suggestions.
- Which suggestion engines to support?

This last question seems to be the trickiest.  The approach taken here is that we hard-wire particular completion engines for particular search engines.  It looks like this.  For each completion engine, we define three functions...

```Coffeescript
Google =
  match: (searchUrl) ->
    true # TBD.  Check the search URL, and verify whether this is the correct completer.

  # Given a query, generate the completer's URL
  getUrl: (queryTerms) ->
    "http://suggestqueries.google.com/complete/search?ss_protocol=legace&client=toolbar&q=#{Utils.createSearchQuery queryTerms}"

  # Given a successful HTTP request, extract the completions
  parse: (xhr, callback) ->
    if suggestions = xhr?.responseXML?.getElementsByTagName "suggestion"
      suggestions =
        for suggestion in suggestions
          continue unless suggestion = suggestion.getAttribute "data"
          suggestion
      callback suggestions
    else
      callback []

```

Adding a new completion engine is as simple (?) as adding these three functions. `match` is probably just a RegExp check. `getUrl` is a simple string operation. `parse` can probably be broken down into reusable utilities.  For example, parsing completions from Google's services is likely to be similar.

If it turns out there's an important German or Chinese completion engine, it should be straightforward (?) to add it.  That's the idea, at least.  Ideally, we would get user-contributed completion engines.

Scoring currently works as follows.  First, we score the relevancy as normal.  Then, taking `N` as the number of characters in the query term, we multiply by:

```
Math.min(12, N) / 12
```

This ranges linearly from `1/12` to `1.0`.  The idea is that, for short queries, we don't want these suggestions displacing other suggestions.  After the user has typed a fair bit of text and not found what they're looking for, it seems more likely that the search suggestion is going to be relevant.  This will require experimentation and tweaking.

Other things...
- It should be possible to tab to a completion, and have the completed text populated in the vomnibar input.
- This should work too for custom search engines, when we find a matching completion engine.


cjk:
This looks great, thanks for adding it. Missing search engine completion was my major reason to still use the browser-addressbar (or even the Google page) from time to time.
I think the vomnibar should be the "single point of entrance" for opening new / past pages. Vimperator is pretty good at this (but failed in other aspects).

smblott-github:
Moved to #1635.

PR #1635, opened 2015-05-04
Search engine completion [Version 2]

smblott-github:
This is a more worked out version of #1632.

Here are some things which are true:

- Completion for Google, Bing, Amazon, DuckDuckGo, Youtube, Wikipedia and Amazon.  Here are a couple of example completion-engine specifications:

```Coffeescript
class Google extends GoogleXMLRegexpEngine
  # Example search URL: http://www.google.com/search?q=%s
  constructor: ->
    super [
      # We match the major English-speaking TLDs.
      new RegExp "^https?://[a-z]+\.google\.(com|ie|co.uk|ca|com.au)/"
      ]

  getUrl: (queryTerms) ->
    "http://suggestqueries.google.com/complete/search?ss_protocol=legace&client=toolbar&q=#{Utils.createSearchQuery queryTerms}"

class Wikipedia extends RegexpEngine
  # Example search URL: http://www.wikipedia.org/w/index.php?title=Special:Search&search=%s
  constructor: ->
    super [ new RegExp "^https?://[a-z]+\.wikipedia\.org/" ]

  getUrl: (queryTerms) ->
    "https://en.wikipedia.org/w/api.php?action=opensearch&format=json&search=#{Utils.createSearchQuery queryTerms}"

  parse: (xhr) ->
    JSON.parse(xhr.responseText)[1]
```

- Users do not and cannot specify completion engines.  Hopefully devs will help out by submitting engines via PRs.  There's almost no cost to adding a new completion engine.

- For custom-search engine searches, we use a completion engine associated with the search engine's URL.  For other queries, we use a completion engine associated with the default search engine.

- There's no explicit mapping from search engine to completer.  Each completer just knows what kind of search URL it can complete, and if it sees such a matching URL, it goes ahead and completes it.  Exactly one completion engine is chosen for each query.

- Tabbing into a completion suggestion populates the vomnibar input.  Tabbing out restores the input's previous contents (unless the user has typed more stuff).

- ~~The top suggestion always appears at the top of the list.  So things like `who directed st<TAB>` completes to `who directed star wars`. `bridges of<TAB>` gives you `bridges of Madison County`.~~ Removed.  This does not work well in practice.

- Completion searches are cached.  We also never launch duplicate identical searches.

- Completion lookups can fail for a variety of reasons.  Fingers crossed, we should fail safe.  Code which could fail due to external factors is protected with `try`/`catch`.

- There's a short delay (currently 200ms) before launching a completion search (in case the user is still typing).  We deliver all of the other completions immediately, then post completion suggestions a couple of hundred milliseconds later.

- I'll not describe the scoring here.  It will probably need tweaking, though.

(The tests fail because there is some hard-wired debugging code still in `settings.coffee`.)

### DIFF
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -231,8 +231,8 @@ class DomainCompleter
   #     If `referenceCount` goes to zero, the domain entry can and should be deleted.
   domains: null
 
-  filter: ({ queryTerms }, onComplete) ->
-    return onComplete([]) unless queryTerms.length == 1
+  filter: ({ queryTerms, query }, onComplete) ->
+    return onComplete [] unless queryTerms.length == 1 and not /\s$/.test query
     if @domains
       @performSearch(queryTerms, onComplete)
     else
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -365,6 +365,7 @@ class SearchEngineCompleter
       { queryTerms, query } = request
       keyword = queryTerms[0]
       if keyword and engines[keyword] and (1 < queryTerms.length or /\s$/.test query)
+        request.completers = [ this ]
         extend request,
           queryTerms: queryTerms[1..]
           keyword: keyword
@@ -501,9 +502,12 @@ class MultiCompleter
     return @mostRecentQuery = arguments if @filterInProgress
 
     # Provide each completer with an opportunity to see (and possibly alter) the request before it is
-    # launched.
-    for completer in @completers
-      completer.triageRequest? request
+    # launched.  The completer is provided with a list of the completers we're using (request.completers), and
+    # may change that list to override the default.
+    request.completers = @completers
+    completer.triageRequest? request for completer in @completers
+    completers = request.completers
+    delete request.completers
 
     RegexpCache.clear()
     { queryTerms } = request
@@ -512,7 +516,7 @@ class MultiCompleter
     [ suggestions, continuations, filters ] = [ [], [], [] ]
 
     # Run each of the completers (asynchronously).
-    jobs = new JobRunner @completers.map (completer) ->
+    jobs = new JobRunner completers.map (completer) ->
       (callback) ->
         completer.filter request, (newSuggestions = [], { continuation, filter } = {}) ->
           suggestions.push newSuggestions...
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -360,6 +360,16 @@ class SearchEngineCompleter
   cancel: ->
     CompletionSearch.cancel()
 
+  triageRequest: (request) ->
+    @searchEngines.use (engines) =>
+      { queryTerms, query } = request
+      keyword = queryTerms[0]
+      if keyword and engines[keyword] and (1 < queryTerms.length or /\s$/.test query)
+        extend request,
+          queryTerms: queryTerms[1..]
+          keyword: keyword
+          engine: engines[keyword]
+
   refresh: (port) ->
     # Parse the search-engine configuration.
     @searchEngines = new AsyncDataFetcher (callback) ->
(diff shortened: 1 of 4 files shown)


smblott-github:
Screenshots...

snapshot

snapshot

snapshot

snapshot

background_scripts/completion.coffee
```diff
@@ -23,21 +23,28 @@ class Suggestion
...
+    # If @insertText is a string, then the indicated text is inserted into the vomnibar input when the
+    # suggestion is selected.
+    @insertText = null
 
   computeRelevancy: -> @relevancy = @computeRelevancyFunction(this)
 
   generateHtml: ->
     return @html if @html
     relevancyHtml = if @showRelevancy then "<span class='relevancy'>#{@computeRelevancy()}</span>" else ""
+    highlightTerms =
```
mrmr1993: Can this be renamed to `escapeAndHighlight` or similar, so it's obvious from the point of call that there aren't XSS issues?

background_scripts/completion.coffee
```diff
@@ -321,27 +328,80 @@ class TabCompleter
...
+          score *= 0.9
+
+        # We keep at least three suggestions (if possible) and at most six.  We keep more than three only if
+        # there are enough slots.  The idea is that these suggestions shouldn't wholly displace suggestions
+        # from other completers.  That would potentially be a problem because there is no relationship
+        # between the relevancy scores produced here and those produced by other completers.
+        count = Math.min 6, Math.max 3, MultiCompleter.maxResults - existingSuggestions.length
+        onComplete suggestions[...count]
+
+  mkSuggestion: (insertText, args...) ->
```
mrmr1993: This (and the `Suggestion` constructor) takes quite a lot of arguments. Maybe accepting a key-value pair would make it easier to check that the right arguments have been set, and ensure the order isn't wrong? A list of the arguments in a comment here would also be helpful.

background_scripts/completion.coffee
```diff
@@ -23,21 +23,28 @@ class Suggestion
     @title ||= ""
     # When @autoSelect is truthy, the suggestion is automatically pre-selected in the vomnibar.
     @autoSelect = false
+    # If @noHighlightTerms is falsy, then we don't highlight matched terms in the title and URL.
+    @noHighlightTerms = false
```
mrmr1993: Shouldn't this be settable from the constructor, since it is a lifetime property of the suggestion?

background_scripts/completion.coffee
```diff
@@ -23,21 +23,28 @@ class Suggestion
...
+    # If @insertText is a string, then the indicated text is inserted into the vomnibar input when the
+    # suggestion is selected.
+    @insertText = null
```
mrmr1993: Ditto.

background_scripts/completion.coffee
```diff
@@ -374,9 +435,13 @@ class SearchEngineCompleter
 # A completer which calls filter() on many completers, aggregates the results, ranks them, and returns the top
 # 10. Queries from the vomnibar frontend script come through a multi completer.
 class MultiCompleter
-  constructor: (@completers) -> @maxResults = 10
+  @maxResults: 10
+
+  constructor: (@completers) ->
+    @maxResults = MultiCompleter.maxResults
```
mrmr1993: Might be better to write in the object prototype instead as

```coffee
class MultiCompleter
  @maxResults: 10
  maxResults: MultiCompleter.maxResults
  
  ...
```

background_scripts/search_engines.coffee
```diff
@@ -0,0 +1,215 @@
...
+    for suggestion in xhr.responseXML.getElementsByTagName "suggestion"
+      continue unless suggestion = suggestion.getAttribute "data"
+      suggestion
+
+class Google extends GoogleXMLRegexpEngine
+  # Example search URL: http://www.google.com/search?q=%s
+  constructor: ->
+    super [
+      # We match the major English-speaking TLDs.
+      new RegExp "^https?://[a-z]+\.google\.(com|ie|co.uk|ca|com.au)/"
```
mrmr1993: `.` in TLDs should be escaped.

lib/utils.coffee
```diff
@@ -194,5 +207,43 @@ globalRoot.extend = (hash1, hash2) ->
     hash1[key] = hash2[key]
   hash1
 
+# A simple cache. Entries used within an expiry period are retained (for one more expiry period), otherwise
+# they are discarded.
+class SimpleCache
+  # expiry: expiry time in milliseconds (default, one hour)
+  # entries: maximum number of entries
```
mrmr1993: Doesn't this technically store up to `2 * entries` values (but only allowing access to `entries` of them)? Probably worth mentioning this in the comment, so this isn't a surprise.

smblott-github: Yes.  It is worth clarifying.
However, it allows access to all of them.

(I'm about to fix a couple of other issues with this.)

pages/vomnibar.coffee
```diff
@@ -151,38 +165,40 @@ class VomnibarUI
...
+    @completer.filter @input.value.trim(), (@completions) =>
+      @populateUiWithCompletions @completions
+      callback?()
 
   populateUiWithCompletions: (completions) ->
     # update completion list with the new data
     @completionList.innerHTML = completions.map((completion) -> "<li>#{completion.html}</li>").join("")
     @completionList.style.display = if completions.length > 0 then "block" else ""
-    @selection = Math.min(Math.max(@initialSelectionValue, @selection), @completions.length - 1)
+    @selection = Math.min completions.length - 1, Math.max @initialSelectionValue, @selection
```
mrmr1993: Nice catch.

pages/vomnibar.coffee
```diff
@@ -222,7 +238,7 @@ class BackgroundCompleter
     id = BackgroundCompleter.messageId += 1
     @filterPort.onMessage.addListener handler = (msg) =>
```
mrmr1993: Why are we creating anonymous functions and messing with constantly adding and removing handlers, when we could just have one handler and ignore requests with `id != BackgroundCompleter.messageId`? Also makes this `keepAlive` complication redundant.

smblott-github: Yes.  I'm slowly trying to tidy up some of the legacy issues with this part of the code.  I think some of it may have been written quite some time ago.

lib/utils.coffee
mrmr1993: > However, it allows access to all of them.

The code seems to indicate otherwise... It allows access to *any* `entries` of the `2 * entries`, perhaps?

smblott-github:
Update... We should merge this; not just yet, but soon.  The vomnibar needs a lot of work to make it more useful, and this is just a first step.

(I'm suspicious as to whether the existing approach to the `HistoryCompleter` is actually helping the UX at all.)

smblott-github:
Update...

I intend to merge this soon.  It's not yet perfect...

- Getting the UX right will require a little time, and probably some tweaking.
- The code factoring can still be improved (and there even a few fragments of dead code).  I'll sort these issues out in the coming days (post merge).

There's a big change to how custom search engines with completers work.  After typing "y rage against", it looks like this...

snapshot

There is completion for these search engines...

```Coffeescript
searchEngines: [
      "w: http://www.wikipedia.org/w/index.php?title=Special:Search&search=%s Wikipedia"
      ""
      "# More examples."
      "#"
      "# (Vimium has built-in completion for these.)"
      "#"
      "# g: http://www.google.com/search?q=%s Google"
      "# l: http://www.google.com/search?q=%s&btnI I'm feeling lucky..."
      "# y: http://www.youtube.com/results?search_query=%s Youtube"
      "# b: https://www.bing.com/search?q=%s Bing"
      "# d: https://duckduckgo.com/?q=%s DuckDuckGo"
      "# az: http://www.amazon.com/s/?field-keywords=%s Amazon"
      ].join "\n"
```

(Contributions of other search-engine completers would be very welcome.)

smblott-github:
Heads up...

I'm going to merge this soon.  There are still a few UX issues to tweak, and for that we really need feedback.

However, since this is still somewhat experimental, the UX may not be great for all users.  So I added an option (possibly temporary) which can be used to tune the weighting given to search-engine completions in the vomnibar (omni mode), and even turn them off altogether.

snapshot

(It's right at the bottom of the options page.)

mrmr1993:
This works beautifully, nice one @smblott-github! Code LGTM too.

PR #1636, opened 2015-05-04
Re-fix javascript: bookmarklets.

### LINKED ISSUE #1611: Bookmarklet doesn't run when executed through vomnibar

smblott-github:
It seems `javascript:` bookmarklets aren't fixed after all.  This is one possibility for (re-) fixing the issue.

Fixes #1611.

### DIFF
--- a/lib/utils.coffee
+++ b/lib/utils.coffee
@@ -39,6 +39,19 @@ Utils =
     urlPrefix = new RegExp "^[a-z]{3,}://."
     (url) -> urlPrefix.test url
 
+  # Apply heuristics to test whether a string (such as a "javascript:" URI) has been URL encoded.
+  isUrlEncoded: (string) ->
+    # If it doesn't contain "%", it's not URL encoded.
+    return false if -1 == string.indexOf "%"
+    # If it contains certain non-URL characters, then it's not URL encoded.
+    for char in [ " ", "{", "}", ";", '"', "'" ]
+      return false if 0 <= string.indexOf char
+    # If it contains any of these non-ascii characters, then it's not URL encoded.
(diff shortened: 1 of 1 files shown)


smblott-github:
Alternatively, and perhaps better, we could just do this: 1ceb14308a80954ed05021335e52f7f8aa4728ef.

mrmr1993:
1ceb143 looks like the way to go &mdash; and  I presume that's what chrome does natively anyway. Sorry I missed that `decodeURI` could throw an error.

mrmr1993:
I've just done a test with

```js
javascript:alert(25 % 20);alert(25%20);
```

And it seems like neither of these approaches is right still. We need to decode valid escaped patterns, but not error on the invalid ones. I'm not sure how to do this without it becoming overly complicated.

mrmr1993:
@smblott-github It seems like the right way to replicate the omnibox/bookmarklet behaviour is to decode javascript uris with

```coffee
# Decode all valid escape sequences in a javascript uri.
decodeJavascriptURI = (uri) ->
  uri.split(/(?=%)/).map((uriComponent) ->
    try
      return decodeURIComponent uriComponent
    catch
      return uriComponent
  ).join ""
```

smblott-github:
Thanks, @mrmr1993.

96af254973cc15714894c822a2c8a4a66a34fbef seems to work.  Go with that?

smblott-github:
Merged 96af254973cc15714894c822a2c8a4a66a34fbef.

jmound:
Looks good. That commit fixed my issue.

PR #1643, opened 2015-05-06
Added note on enable regexp search.

LarryBattle:
Added a note on how to enable a RegExp search.


smblott-github:
Thanks, @LarryBattle.  While your suggestion is correct, we usually try not to overwhelm new users with advanced features.  These are usually best handled in the [wiki](https://github.com/philc/vimium/wiki/Find-Mode).

LarryBattle:
ok, but I wish the help dialog box provided a link to the wiki.

philc:
That might be a great idea

Issue #1640, opened 2015-05-06
Vimium corrupts Chrome XML view page


difanz:
Running version 1.51 with Chrome Version 43.0.2357.37 beta (64-bit) on Mac OS X 10.10.3.

Example page: http://qz.com/feed/

Intended behavior: Show Chrome XML viewer for this page since it does not have any XSLT associated.

Actual behavior: Shows XML as well as text representation of Vimium CSS code.

Screenshot: http://tifan.la/7DpI

smblott-github:
Thanks, @difanz.

And the Vomnibar fails to load...

![snapshot](https://cloud.githubusercontent.com/assets/2641335/7488959/e326f4ae-f3c8-11e4-87cf-ab7872b27f74.png)

Here...

![snapshot](https://cloud.githubusercontent.com/assets/2641335/7488969/044cdf54-f3c9-11e4-82fd-8400db66c872.png)

mrmr1993:
Do our key handlers still work? If not, it makes sense to abort initialisation completely when `document.contentType == "application/rss+xml" `.

The only code I can see in the screenshot is from vimium.css. The only solutions I can think of are:
* inject the code when the `domReady` port is opened
  - this won't work with frames; we can't inject into specific frames.
* put all Vimium UI components in shadow DOMs (or all in one shadow DOM together) and only inject css into them (/it), not the document itself.

smblott-github:
Our key handlers do work.  Except the Vomnibar, for which initialisation fails.

mrmr1993:
How about the HUD?

smblott-github:
HUD does not work.

mrmr1993:
Okay, so we should block any attempts to create elements/add them to the DOM when we know they're going to fail. Easiest way should be to
* have a single shadow DOM that all our UI goes in, created at load;
  - (this also fixes the CSS issue)
* fail gracefully when creating the shadow DOM errors; and
* block commands that will try to inject UI from running if the shadow DOM errored/isn't available.
  - this should mean adding a one line check to `LinkHints.activateMode`, `Vomnibar.init`, `Vomnibar.activate`, and one or more HUD functions.

There are probably other ways to achieve this, but it's nice to have a single point of failure that we can check (the shadow DOM), and to avoid having to check mime types altogether.

markstos:
"View Source" on an XML page in Chrome is also broken for me now. Content-Type is `application/xml`. Sample URL is private, sorry. Appears to the same issue.

[...]

PR #1647, opened 2015-05-09
Disable UIComponent for XML documents

mrmr1993:
If the page is an XML document, nothing we do works:
* `<style>` elements show their contents inline,
* `<iframe>` elements don't load any content,
* `document.createElement` generates elements that
  - have `element.style == null`, and
  - ignore CSS.

This PR stops us from injecting anything into the DOM from `UIComponent`, fixing #1640.

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -6,6 +6,14 @@ class UIComponent
 
   constructor: (iframeUrl, className, @handleMessage) ->
     styleSheet = document.createElement "style"
+
+    # If this is an XML document, nothing we do here works:
+    # * <style> elements show their contents inline,
+    # * <iframe> elements don't load any content,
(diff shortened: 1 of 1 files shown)


smblott-github:
@mrmr1993.  Two things:

- Are you *absolutely sure* that this is the right test?  What happens if there's just no stylesheet *yet*?
- Might this bet better: 8e6081d102c4f0143e34b487f96b1d79c1782466?

mrmr1993:
> Are you *absolutely sure* that this is the right test?

I am; `Element.style` is always non-`null` for HTML pages and `null` on XML pages, because of the rules in the corresponding specs.

> Might this bet better: 8e6081d?

That looks good, I'm fine with that being merged over this PR.

smblott-github:
Moved to #1694.

PR #1648, opened 2015-05-10
Use classes for border flashes

### LINKED ISSUE #1634: A race condition about `borderWas` when highlighting body
In vimium_frontend.coffee +341: @@ window.`focusThisFrame`
```
borderWas = document.body.style.border
document.body.style.border = '5px solid yellow'
setTimeout((-> document.body.style.border = borderWas), 200)
```
Then it's easy to make body's border stay yellow:
* quickly click `gF` (`mainFrame`) again and again

After a while, a new `borderWas` will become '5px solid yellow', which is set by `focusThisFrame` before.

Fix:
* record `borderWas` as a global variable, and clean it everytime timeout is triggered

mrmr1993:
This fixes #1634 and ensures that the border flash for `gf` and `gF` is always visible by adding/removing a dedicated element.

Note: since the flash is no longer on the body, the characteristic 'jump' as the page content is redrawn no longer occurs. It's easy to add back in if it's missed.

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -109,6 +109,20 @@ div.internalVimiumSelectedInputHint span {
   color: white !important;
 }
 
+/* Frame Highlight Marker CSS*/
+div.vimiumHighlightedFrame {
+  position: fixed;
+  top: 0px;
+  left: 0px;
+  width: 100%;
+  height: 100%;
+  padding: 0px;
+  margin: 0px;
+  border: 5px solid yellow;
+  box-sizing: border-box;
+  pointer-events: none;
+}
+
 /* Help Dialog CSS */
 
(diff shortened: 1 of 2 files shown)


smblott-github:
After a quick test on a couple of sites with known issues, this looks great.

Might you be able to have a look at the tests, @mrmr1993?

mrmr1993:
Updated, sorry about that @smblott-github.

smblott-github:
Thanks, @mrmr1993.

> Note: since the flash is no longer on the body, the characteristic 'jump' as the page content is redrawn no longer occurs.

That's a plus!  Let's try this, and watch out for issues as we go along.

smblott-github:
@mrmr1993. This is working out beautifully!  Huge improvement.

PR #1649, opened 2015-05-11
Use css @import rather than XMLHttpRequest for shadow DOM external style

mrmr1993:
Since moving to using shadow DOMs for some of the UI, we've been manually loading our stylesheet via `XMLHttpRequest` and injecting it into a style element in the shadow DOM. CSS supports `@import` for doing this automatically, so we should use that instead.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -336,13 +336,8 @@ window.focusThisFrame = do ->
   # Inject stylesheet.
   _styleSheet = document.createElement "style"
   if _styleSheet.style?
-    _styleSheet.innerHTML = ""
+    _styleSheet.innerHTML = "@import url(\"#{chrome.runtime.getURL("content_scripts/vimium.css")}\");"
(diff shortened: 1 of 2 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -336,13 +336,8 @@ window.focusThisFrame = do ->
   # Inject stylesheet.
   _styleSheet = document.createElement "style"
   if _styleSheet.style?
-    _styleSheet.innerHTML = ""
+    _styleSheet.innerHTML = "@import url(\"#{chrome.runtime.getURL("content_scripts/vimium.css")}\");"
```
smblott-github: This looks straightforward, @mrmr1993.

Is there a reason for using `\"` and not just `'` for the outer quotes?

mrmr1993: We need double quotes outside for string interpolation. Changing the inner quotes to `'` shouldn't be a problem, though.

PR #1652, opened 2015-05-11
Separate the hud from frontend code and rewrite tween to be more generic

mrmr1993:
This is the first stage of another attempt at HUD-in-iframe, replacing #1474. This re-attempt will focus on
* having smaller, easier to review commits
* being spread across several PRs, so the whole feature doesn't go stale and need re-writing again
* working with modes properly, not as an after-thought (#1474 was a rebase of #1401, which pre-dated modes).

This PR
* moves the `HUD` class to its own file
* moves styles into the css file
* disentangles `Tween`, rewriting it to be generic enough to support a `UIComponent`.

The diff is large because of moving code; reviewing commit by commit is far easier.

@smblott-github can you take a look over this when you get a chance?

### DIFF
--- a/content_scripts/hud.coffee
+++ b/content_scripts/hud.coffee
@@ -3,12 +3,15 @@
 # Note: you cannot interact with the HUD until document.body is available.
 #
 HUD =
-  _tweenId: -1
+  tween: null
   _displayElement: null
 
   # This HUD is styled to precisely mimick the chrome HUD on Mac. Use the "has_popup_and_link_hud.html"
   # test harness to tweak these styles to match Chrome's. One limitation of our HUD display is that
   # it doesn't sit on top of horizontal scrollbars like Chrome's HUD does.
+
+  init: ->
+    @tween = new Tween ".vimiumHUD.vimiumUIComponentVisible"
 
   showForDuration: (text, duration) ->
     @show(text)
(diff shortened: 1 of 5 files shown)


smblott-github:
@mrmr1993.
Will do.

(And I appreciate your step-by-step approach.)

smblott-github:
Just to add...

I remember from before that there was a problem having the focus in an iframe.  It means the focus is not on the page, which means that inputs never have the focus, which means that selections within inputs aren't actually selected, which means they're not visible.

mrmr1993:
> I remember from before that there was a problem having the focus in an iframe.

I'm aware of that; in the next PR I plan to move the HUD into an iframe as is, so that the subsequent PR can be dedicated solely to accepting input in the HUD and the related focus issues. Hopefully this fragmented approach will make it easier to see what changes need to be made for that, make those changes, and keep the code current wrt master.

smblott-github:
Good plan.

smblott-github:
Thanks, @mrmr1993.  I guess you're good to go on the next stage of your plan.
Bit-sized pieces really helps!

smblott-github:
@mrmr1993.  Could you take a look at 7004420e178416cc680416091a23dd804fb9370c, please?

smblott-github:
@mrmr1993.  I was still seeing errors.  See 26f91fea40f11c116497ce20c1fe53f3956a922f too.

mrmr1993:
@smblott-github good call on 26f91fe, I'd completely missed that.

I've pushed #1658 as a follow-up to this.

Issue #1650, opened 2015-05-11
How to inject scripts to an iframe without `src`


gdh1995:
In this page http://output.jsbin.com/raxorevane/1/edit?html,output :

Vimium can switch between two frames: the top  and the bottom-right. But there's an iframe for output content which has no `src` attribute, and its links are not selectable for LinkHints.

I used F12 and found no extension script was loaded, and so was it even when I delete the frame's `sandbox` attribute.

Is there any trick to inject any scripts - even manually triggered by executing something on window.top ?

mrmr1993:
In general, adding [`"match_about_blank": true`](https://developer.chrome.com/extensions/content_scripts#match_about_blank) to all the places where we inject scripts or styles (including `manifest.json`) should do what you want. However, this might not work for your specific example @gdh1995, since (from the linked page):

> Content scripts cannot be inserted in sandboxed frames.

gdh1995:
Thanks. Now my custom Vimium does work on a normal blank page.

While the sandboxed frames are annoying:
* content scripts may be injected. Then frameId is registered and "DOMContentLoaded" is fired.
* no Vimium's listener can be triggered.
* Vimium's onunload function is not called, which causes the frameId not cleaned
* then `nextFrame` does not work...

Anyway, `match_about_blank: true` works well at most websites. Then this issue has closed.

mrmr1993:
Can you reopen this? This isn't fixed in `master` here.


> Now my custom Vimium does work on a normal blank page.

If you submit a PR with those changes, then we can get them included in `master`, so this issue can resolved for everyone else too.

gdh1995:
I've pushed two commits to /gdh1995/vimium. They are:
* [ignore sandboxed iframes for the command "nextFrame"](https://github.com/gdh1995/vimium/commit/4b94b04124a4b298edc9553b9d9b9f9fa4a9c08c)
* [inject to iframes without "src"](https://github.com/gdh1995/vimium/commit/23716a9b43c38761aa2123e6df6d65880be7b354)

I'm not sure whether they are enough, so haven't created a PR for them.

mrmr1993:
Thinking on this a little more: we need to be careful arbitrarily injecting our content into these kind of frames. Lots of WYSIWYG editors use `src`-less `<iframe>`s, and we don't want to pollute the DOM in those. A `document.body.isContentEditable` check should mostly suffice, but are there other similar edge cases?

[...]

PR #1653, opened 2015-05-11
use shadowDOM for hightlightedFrame only if style is not null

gdh1995:
In a XML document, `_frameEl` should not be in a shadow DOM node, and must have an id.

Otherwise there will be no selector matched to it, and nothing will show.

Please see #1634.

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -110,7 +110,7 @@ div.internalVimiumSelectedInputHint span {
 }
 
 /* Frame Highlight Marker CSS*/
-div.vimiumHighlightedFrame {
+div#vimiumHighlightedFrame {
   position: fixed;
   top: 0px;
   left: 0px;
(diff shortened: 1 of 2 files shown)


mrmr1993:
As I said over in #1634, the rest of our UI is far more broken (Vomnibar, HUD, help dialog and link/input hints do not work).

Is there a specific website where you've been switching frame into an XML document and the border not showing is a problem?

gdh1995:
Um, I think, it should be strong as long as there is it.

I've seen only one: (in #1640) http://qz.com/feed/

mrmr1993:
> Um, I think, it should be strong as long as there is it.

I'm not comfortable setting `id` on elements we create, since `id` is supposed to be unique in the document, and we can't make that guarantee for all pages. But it's ultimately not my decision, and I'm possibly being paranoid.

> I've seen only one

There are no frames on that page. Why would we be concerned about the highlighted border displaying correctly there when it serves no useful purpose?

gdh1995:
> id is supposed to be unique in the document, and we can't make that guarantee for all pages

For Vimium, it's unique; for websites, I think, few people may come up with an id like "vimium***". So, I'm not so worried as you.

This PR just fixs one bug and is obviously not enough for so many `document` types including `XML` and `about:blank`, but this does not prevent us to apply it.

As in #1650, iframes without `src`, which may be from WYSIWYG editors, should keep pure. and then, we should use a switch to control this for `focusFrame` and `UIComponent`.

PR #1654, opened 2015-05-12
Add header for Advanced Options.

smblott-github:
The options page might look better with a header...

snapshot

(This leaves the "show/hide advanced options" link as it is, which isn't great.  But the section headers look better, IMO. Personally, I'd also prefer to get rid of the fold entirely.)

Do you have an opinion on this one, @philc?


smblott-github:
Going, going, ... (@philc).

smblott-github:
Gone.

PR #1656, opened 2015-05-12
Vomnibar filter duplicates

smblott-github:
This does two things.

1. It filters vomnibar suggestions to exclude items with the same URL.  This is pretty common.  Typing  `fac` into the vomnibar, the first three suggestions are identical (Facebook: domain, bookmark, history). Bookmarked items usually generate two entries (history and bookmark).

2. Before filtering, this shortens the URL: remove leading and trailing fluff.  This increases the chances of finding a duplicate.

**Important note**... The URL shortening includes some custom rules for Google search (Google adds unique ids to every query, thereby polluting the vomnibar with, effectively, duplicates).

### DIFF
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -41,6 +41,7 @@ class Suggestion
     # or @relevancyFunction.
     @relevancy ?= @relevancyFunction this
 
+  # Note. This always returns a truthy value.
   generateHtml: ->
     return @html if @html
     relevancyHtml = if @showRelevancy then "<span class='relevancy'>#{@computeRelevancy()}</span>" else ""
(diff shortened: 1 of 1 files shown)


smblott-github:
@mrmr1993 (and possibly @philc).  I left this one sitting because it includes site-specific code (specifically for Google search).  That's no something we've done before.

Any input?

My sense is it's the right thing to do, but only for *very* popular sites, and where it has a substantial positive effect on the UX.  Here, there's no arguing that Google search is a popular site.  And when repeated searches (for the same query) lead to different (but equivalent) URLs in the history, filtering out duplicates seems like a UX win.

mrmr1993:
> it includes site-specific code 

I'm not the biggest fan of this idea, but if it's what we have to do, it's what we have to do.

I've got exams the next couple of days, so I won't be able to look at anything properly until Saturday.

PR #1659, opened 2015-05-13
What should the search-completion UX be? It should be like this.

smblott-github:
This is a major reworking of how search-completion works.

Previously, it seemed like emulating the Chrome omnibox approach was the right thing to do.  Now, I'm convinced that that's not the best approach.  It creates two problems:
- First, you have the text the user has typed, and then the completion text (so, the user types "one flew", and "over the cuckoo's nest" is added as completion text to the query, and highlighted).  This creates the issue of what to do if the user types `Enter`.  Query just what the user has typed?  Or include the completion text?  There are common use cases for both which are wrong (and use cases which are right, too).  Also, it's difficult to get the second approach (the Chrome approach) to integrate well with the way the vomnibar works already.
- Second, if we do the Chrome thing (which is to include the completion text), we create a horrible race condition when the user types `Enter`.  Completions are fetched asynchronously. Therefore, the suggestions can change just before the query is submitted.  That's very bad.

This PR eliminates these *trying-to-be-chrome* issues, and just does things the Vimium way:
- For custom search engines, we keep the existing behaviour whereby the first suggestion is auto selected, and subsequent suggestions (including completion suggestions) appear below.  Therefore, `<Tab><Enter>` gets  the top ranked completion suggestion. `Enter` on its own just sends the query the user has typed to the search engine.
- For normal omni-bar queries, completion suggestions are just relevancy scored and mixed in with the rest of the suggestions.  When the user has typed quite some text and not found what they're looking for, the completion suggestions have usually made their way to the top of the list. So they're just a `<Tab><Enter>` away.  `<Enter>` on it's own always just behaves as it always has done -- send the query to the default search engine.

With this PR, if the user types stuff and hits `Enter`, they always get what they expect.<sup>*</sup>

I'm going to merge this directly, because I'm not happy with the state of `master` in this regard.  This long post was to document my thinking -- though feedback is always welcome, of course.

<sup>*</sup>Except for another pre-existing race condition which exists for custom search engine queries, which I'll get to soon.


philc:
I hate the Chrome completion UX for the reasons you mentioned.

Thanks for addressing this. I'll let you know if I encounter any funny
business in regular usage.

smblott-github:
Note.  I omitted to remove these two lines: d9742b3eddd2cfbd655841c5467607dfeb0a203d.  The effect is to disable the `<left>` and `<Right>` keys within the vomnibar.  Fixed now. (My bad.)

PR #1661, opened 2015-05-15
History for the vomnibar...

smblott-github:
Type `star wars` into the vomnibar.  Then, realise you really meant `star wars director`.  You need to retype the whole query.  Some mechanism for a vomnibar history would be helpful.

Here's one (kind of cute) idea. (A couple of other possibilities are mentioned below).

Start retyping the query...

snapshot

Ah!  There's the history item, in top spot, no less.  Hit `Tab`, and we get this...

snapshot

That is:
- We recognise that this history item has the default search URL as a prefix.
- We strip off the prefix (and any extra suffix) and decode the original query (`star wars`, here).
- And, we insert that text into the vomnibar, so the user can continue typing `director`.

(This comes almost for free as a result of some of the search-completion infrastructure.)

Relevancy:
- Additionally, we rescore matching history suggestions for word relevancy, this time considering only the query terms. Then we choose the maximum of the original word relevancy and this new word relevancy.  This has the side effect (usually) of bumping up the history item's score and pushing it up the list.

Caveat:
- This doesn't play well with the way the vomnibar leaves `@selection` unchanged when the vomnibar updates.  It means, in effect, that the text in the vomnibar input could be replaced with whatever text happens to end up as the selected suggestion.
- This PR changes that behaviour.  Every time the vomnibar updates, `@selection` is reset.
- This is not such a big loss.  The existing behaviour is actually pretty weird.  When a suggestion is selected and the user types (or deletes characters from the query), there is practically no way that the user can predict what suggestion is going to end up selected.

This idea is pretty cheap and easy to understand.

Here are two alternative (or possibly additional) ideas.

1. Add a new *query history* vomnibar mode.  This would require a new command.

2. When the vomnibar is empty, `Tab`, `Up` or `Down` toggle into a *query history* mode, whereby the completion list consists only of URLs the user has selected and the default searches they've launched (possibly with scoring weighted heavily for recency).  Add a similar toggle to get back to the original mode.

(For me, 1. is a non-starter, but 2. might have legs.  However, both are actually orthogonal to the idea here.  We could do this PR on its own, or this PR plus 2., if there's any interest.)

### DIFF
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -45,14 +45,17 @@ class Suggestion
     return @html if @html
     relevancyHtml = if @showRelevancy then "<span class='relevancy'>#{@computeRelevancy()}</span>" else ""
     # NOTE(philc): We're using these vimium-specific class names so we don't collide with the page's CSS.
+    insertTextClass = if @insertText then "vomnibarInsertText" else "vomnibarNoInsertText"
+    insertTextIndicator = "&#xfe62;" # A small plus sign.
+    insertTextIndicator = "&#xfe65;" # A small "greater than" sign.
     @html =
       """
       <div class="vimiumReset vomnibarTopHalf">
-         <span class="vimiumReset vomnibarSource">#{@type}</span>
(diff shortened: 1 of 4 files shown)


smblott-github:
Replaced by #1666.

PR #1662, opened 2015-05-15
Vomnibar query-history mode (a proper one).

smblott-github:
This implements a query history for vomnibar queries (default search engine only).

It looks like this...

snapshot

These are the default-search-engine queries run either:
- directly from the vomnibar input
- from a default search-engine completion, or
- (with #1661) from a history entry which matches the default search URL.

The *big* outstanding issue is how to activate this vomnibar mode.  The way it works here, query history mode is activated whenever the user hits `Tab` in a *completely* empty vomnibar (that is, the vomnibar input is empty, and there are no completions).  This is the start state for "omni" and "bookmarks" modes.  So, `o` then `Tab` then `Tab` brings up the most recent search query, like this...

snapshot

... where the user can then either resubmit the query or edit it.

That's a very nice binding... if you know about it.  It's somewhat hidden away if you do not.  If we go with this, then this would be important enough to get a mention in the main README.

(Apart from adding two new commands, I don't really have any other good ideas as to how such a query history mode could be activated.)

### DIFF
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -526,6 +526,19 @@ class SearchEngineCompleter
             console.log "fetched suggestions:", suggestions.length, query if SearchEngineCompleter.debug
             onComplete suggestions.map mkSuggestion
 
+# A completer which provides completions based on the user's query history (that is, those vomnibar queries
+# for which no suggestion was selected).
+class QueryHistoryCompleter
+  filter: (request, onComplete) ->
+    onComplete [
+      new Suggestion
+        queryTerms: request.queryTerms
+        type: "query"
+        url: "https://www.google.ie"
+        title: "Hello"
+        relevancy: 1
+      ]
+
 # A completer which calls filter() on many completers, aggregates the results, ranks them, and returns the top
 # 10. All queries from the vomnibar come through a multi completer.
 class MultiCompleter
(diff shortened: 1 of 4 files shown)


smblott-github:
Additional note.

It would be easy to include custom search engines too.  The problem with doing so is...
- When the user selects query text that was generated from a custom search-engine query then edits the text, should we submit the resulting text to the default search engine, or to the original custom search engine.

Both approaches can be implemented.  It's just not obvious which one is the best.  And whichever one you choose, it'll be a bad UX for the case where the user wants the other.

smblott-github:
Update:
- Now includes completion for the default search engine, if available.
- Two additional **very experimental** bindings: `<Delete>`  or `Enter` when the vomnibar input is empty (and no completion is selected).

Explanation re. bindings.  `<Tab>` is a natural and excellent (if somewhat undiscoverable) binding for toggling the vomnibar mode.  Unfortunately, while it may be great for entering query-history mode, it can *never trigger* for going back again.  This is because the test is "vomnibar input is empty" AND "the completion list is empty".  These two things are always true when we first fire up the vomnibar (great) but can never be true at the same time in query-history mode.

smblott-github:
Discontinued as too complicated and too difficult to get the UX right.
See #1666 instead.

PR #1667, opened 2015-05-17
Do not ignore recency in history scores.

smblott-github:
See [this line](https://github.com/philc/vimium/commit/51f197952208701542eef60b353f248e5b9c6054#diff-d25503b9a051bcc246a7773e67af13f3R95):

```Coffeescript
# Average out the word score and the recency. Recency has the ability to pull the score up, but not down.
score = (wordRelevancy + Math.max(recencyScore, wordRelevancy)) / 2
```

This is the relevancy calculation for the history completer.

I'm suspicious.  This might actually be making matters worse.  In particular, it has the effect of (unduly?) promoting very-old suggestions which just happen to have short titles (or URLs).  Therefore they have a higher word relevancy and their recency is ignored.  So a page I visited five minutes ago is displaced by one I visited in the dim an distant past.

### DIFF
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -252,8 +252,8 @@ class HistoryCompleter
     historyEntry = suggestion.relevancyData
     recencyScore = RankingUtils.recencyScore(historyEntry.lastVisitTime)
     wordRelevancy = RankingUtils.wordRelevancy(suggestion.queryTerms, suggestion.url, suggestion.title)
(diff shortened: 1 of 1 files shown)


smblott-github:
I think we should try this out.  The inability to find history entries from yesterday (which have been displaced by entries from three weeks ago) is a weakness.

smblott-github:
Closing this.

There **is** a problem with recency scores.  But it's not clear that this is the solution.

PR #1674, opened 2015-05-20
Only replace double-slashes in find mode if they precede our flags

### LINKED ISSUE #1673: \ in regex searches requires double escaping to search for the raw character
* Searching for `\w` works as in normal regexps
* Searching for `\\w` works the same as searching for `\w`
* Searching for `\\\w` and `\\\\w` work how `\\w` works for normal regexps.

This behaviour seems wrong and counter-intuitive, but we also need to support `\I`, `\r` and `\R` as our own special escape sequences, which is how this problem arose. The obvious fix is to only replace `\\` with `\` when it precedes `I`, `r` or `R`.

mrmr1993:
This fixes #1673

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -719,19 +719,15 @@ updateFindModeQuery = ->
   findModeQuery.isRegex = settings.get 'regexFindMode'
   hasNoIgnoreCaseFlag = false
   findModeQuery.parsedQuery = findModeQuery.rawQuery.replace /(\\{1,2})([rRI]?)/g, (match, slashes, flag) ->
-    return match if flag == ""
-    if slashes.length == 2
-      # The escape code is double-escaped (ie. \\r rather than \r). Strip the extra slash.
-      "\\#{flag}"
-    else
-      switch (match)
-        when "\\r"
-          findModeQuery.isRegex = true
-        when "\\R"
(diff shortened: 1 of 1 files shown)


smblott-github:
Is this correct, @mrmr1993?

Go here: https://github.com/philc/vimium/issues/1673.
Search `/`, then these three characters: `\\r`.

I think that should find the `\r` of `\rogue`.

mrmr1993:
> I think that should find the `\r` of `\rogue`.

The behaviour for that case hasn't changed, it's just that `\r` means carriage return in normal regexes. I always thought `\r` was a stupid choice of extra flag, for this reason. We could
1. change the flag,
2. disable carriage return activating on `\\r`, or
3. leave behavioir as is.

smblott-github:
OK.  In that case, go back and try `\\I`.  My point is, this should not trigger our escape sequence, but with this PR, I think it does.

mrmr1993:
> My point is, this should not trigger our escape sequence, but with this PR, I think it does.

@smblott-github I've pushed a commit which stops us from changing the search regexp except when we encounter an unescaped `\r`, `\R` or `\I`. It also means that you need to type `\u000D` or `\x0D` to search for a carriage return, but I'm not sure how useful searching for `\r` was anyway. Could you give it another look?

smblott-github:
Thanks, @mrmr1993.

PR #1675, opened 2015-05-21
Allowing mapping <space>.

### LINKED ISSUE #544: Allow remapping the space key in Vimium's settings

smblott-github:
Fixes #544.
See #665 for previous discussion.


faridcher:
trying 
map <c-space> goBack
map <c-s-space> goForward
don't work!

PR #1680, opened 2015-05-24
Queue UIComponent requests until the frame has loaded

### LINKED ISSUE #1679: Another race condition vis-a-vis the HUD
If you enter insert mode immediately as the page loads, you can get a visible but empty HUD.

This seems to be happening because the HUD iframe (and port) have not yet initialised, so we're not sending the HUD contents to the iframe.

mrmr1993:
This fixes #1679.

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -4,6 +4,8 @@ class UIComponent
   showing: null
   options: null
   shadowDOM: null
+  loaded: false
+  queuedActions: []
 
   constructor: (iframeUrl, className, @handleMessage) ->
     styleSheet = document.createElement "style"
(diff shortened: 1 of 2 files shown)


smblott-github:
@mrmr1993.  I had been thinking about using an `AsyncDataFetcher` [here](https://github.com/philc/vimium/blob/master/lib/utils.coffee#L296). to fix this.  Would that be an alternative?

mrmr1993:
Yes, that would be. Equivalently, we could use [Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) if we're willing to drop support for Chrome <32.

smblott-github:
@mrmr1993.  I think #1685 is the way to go on this.  It avoids the need for yet another queue implementation.

mrmr1993:
Okay, that's fine.

Issue #1651, opened 2015-05-11
Search-engine completion


smblott-github:
As of #1635, we have search-engine completion both for custom search engines, and for the vomnibar in "omni" mode.  Here are some notes...

- This is an experimental feature.  Please be patient, and provide feedback (both positive and negative) here.  The UX still requires some tweaking, but we need experience using the feature to get that right.

- Completion is supported whenever we have a completion engine for the search engine being used.  See [here](https://github.com/philc/vimium/blob/master/background_scripts/completion_engines.coffee#L2).  Currently it's (English-language) Google, Youtube, Wikipedia, DuckDuckGo, Bing (why?) and Amazon.

- Contributions of new completion engines would be greatly received, including in particular for non-English search engines.  They're easy to write (see [here](https://github.com/philc/vimium/blob/master/background_scripts/completion_engines.coffee#L2), again) and add almost no memory or runtime costs.

- There's a considerable amount of caching and re-use of previous results, so we're not generating a massive amount of traffic.

- What are the privacy issues?  For custom search engines, privacy seems moot. However, for general vomnibar queries, there may be an issue.  Currently, there's a new option at the bottom of the options page which allows search completion to be tuned (and even disabled).

- There is a very big change in how custom search engines work if we have a completion engine.  In that case, no suggestions from other (Vimium) completers are displayed.  Try it out and provide feedback.

Here's what it looks like for the vomnibar...

![snapshot](https://cloud.githubusercontent.com/assets/2641335/7566875/50b9c132-f7f1-11e4-91ab-218c887c9443.png)

And for a (Youtube) custom search...

![snapshot](https://cloud.githubusercontent.com/assets/2641335/7566905/85f1c6d8-f7f1-11e4-868b-2313ce178b6f.png)

smblott-github:
This seems to work particularly well for custom search engines.  For the default search engine, the biggest issue I'm finding is getting the balance right between the relevancy scoring of search completions and the relevancy scoring of rest of the suggestions (history, bookmarks, etc).  One group or the other tends to dominate.

smblott-github:
I'm experimenting with an alternative scoring approach which is intended to get around the "clumping" effect described above:
- https://github.com/smblott-github/vimium/tree/search-completion-with-regular-relevancy
- eb0e2964fca5ef2eccc06607944df6b208b2b99f

smblott-github:
[...]

PR #1687, opened 2015-05-27
Offer search completion only on custom searches

smblott-github:
This is a substantial change to the search completion (#1651) UX:
- **Do not** offer search completion for default searches.
- **Only** offer completions for custom search engines (when we have a completer).
- For custom searches, **always** only include history suggestions which are from the current custom search engine (previously, we only filtered like this if we had a completer).
- Change the look of the vomnibar for custom searches (as per @philc's proposal in #1651):

snapshot

Why is this the right thing to do?
- For default searches:
  - Some users might not want completion (e.g. for privacy reasons).
  - We can get rid of the newly-introduced "Search weighting" option.
  - We don't generate network traffic when the user is looking for a history entry.
  - We side step the "crowding" problem (whereby completion suggestions can wholly displace other suggestion, or vice versa).
  - The established, vanilla omni-mode vomnibar UX is unchanged (so, familiar to users).
- We're always either in general omni-mode, or custom search mode. We're never mixing suggestions.  So we can change the look of the vomnibar for the latter case (as illustrated above).

Why is this the wrong thing to do?
- Search completion becomes harder to access (the user has to explicitly enter custom search mode).

The user can still get completions on their default search engine, but to do so they have to add (or have) a matching custom search engine, e.g.:

```
g https://www.google.com/search?q=%s Google
```

(We could consider adding this to the defaults, possibly together with the "I'm feeling lucky..." search.)

(Background.  I've been considering doing things this way for some time.  @philc's comment in #1651 vis-a-vis the look of the vomnibar provided the final concrete reason for why this is the *right thing to do*.)


smblott-github:
Originally, the "WIP" tag on this PR indicated that there was still inactive code related to completion for the default engine in the PR.  That has now been removed.

This is part of an ongoing process of tweaking the search-completion UX.  I expect to merge this soon.

smblott-github:
Screenshots posted just for the wiki...

150531-01-snapshot

150531-02-snapshot

PR #1689, opened 2015-05-27
Go directly to next/previous tab.

smblott-github:
This makes `nextTab` and `previousTab` go directly to the relevant tab, without visiting intervening tabs -- all of which avoids a nasty flicker for 5J or 5K.


mrmr1993:
Nice, I like it.

PR #1658, opened 2015-05-12
Move hud to an iframe

mrmr1993:
This PR moves the HUD to an `<iframe>` managed by a `UIComponent`.

PR #1693, opened 2015-05-29
Accept input directly in the HUD iframe for find mode

### LINKED ISSUE #556: Use an text box for key entry in Vimium's find mode

### LINKED ISSUE #1672: Ability to paste into "find on the page"
`/` allows you to enter find on the page which is very helpful. sometimes I've wanted to find something that is quite long or is a symbol that is not easy to type (e.g. searching for a phrase or the character lambda). In these cases it would be good to be able to paste something in to search for it. So far in these cases I've had to resort to chromes default "find on page".

I am unable to paste using C-v or S-Insert. I think allowing the ability to do that with one of those or something similar would be useful.

mrmr1993:
This PR does the bare minimum to get find mode working using a `contentEditable` element in the HUD iframe, instead of using `keydown`/`keypress`. This is a follow-on from #1658.

Issues I'm aware of (and plan to fix in futute PRs):
* no focus checks, so the user can 'escape' the iframe to the parent window and trigger the old code directly
* the current find mode code is nearly unchanged, so there is some code we won't need to keep, and some that will want re-writing
* when the search term gets too long, it overflows the HUD and the caret can be in the overflow

This resolves #556, fixes #1672, and possibly some others too.

@smblott-github sorry I've taken so long to push this, can you take a look over it?

### DIFF
--- a/pages/hud.coffee
+++ b/pages/hud.coffee
@@ -58,7 +58,7 @@ handlers =
 
   showFindMode: (data) ->
     hud = document.getElementById "hud"
-    hud.innerText = "/"
+    hud.innerText = "/\u200A" # \u200A is a "hair space", to leave enough space before the caret/first char.
 
     inputElement = document.createElement "span"
     inputElement.contentEditable = "plaintext-only"
--- a/content_scripts/mode_find.coffee
+++ b/content_scripts/mode_find.coffee
@@ -61,7 +61,6 @@ class FindMode extends Mode
     hasResults: false
 
   constructor: (options = {}) ->
-    console.log "constructor", @options
     # Save the selection, so findInPlace can restore it.
     @initialRange = getCurrentRange()
     FindMode.query = rawQuery: ""
--- a/content_scripts/mode_visual_edit.coffee
+++ b/content_scripts/mode_visual_edit.coffee
@@ -353,7 +353,7 @@ class Movement extends CountPrefix
       do =>
         doFind = (count, backwards) =>
           initialRange = @selection.getRangeAt(0).cloneRange()
-          for [0...count]
+          for [0...count] by 1
             unless FindMode.execute null, {colorSelection: false, backwards}
               @setSelectionRange initialRange
               HUD.showForDuration("No matches for '#{FindMode.query.rawQuery}'", 1000)


smblott-github:
Thanks, @mrmr1993.  This works (and looks) great.  Could you explain the relationship between the find-mode HUD input and `FindMode`?  I kind of expected `FindMode` to disappear, and I don't get why you're bubbling various events down the handler stack.  A bit of an explanation would be appreciated.

(Also, we seem to have lost `PostFindMode` when the selection lands in an input.)

Off topic... I think we're heading towards a release where the feature items are:
- Much improved vomnibar.
- Much improved find.

mrmr1993:
> I kind of expected `FindMode` to disappear

I plan for it to eventually, but I figured it better to just use the existing code for everything except what *had to* change. Since that's just a matter of moving and deleting code, it should work identically to this PR, up to bugfixes.

> (Also, we seem to have lost `PostFindMode` when the selection lands in an input.)

I'll look into this soon.

mrmr1993:
@smblott-github I've just pushed a commit to fix `PostFindMode`.

content_scripts/hud.coffee
```diff
@@ -42,6 +56,28 @@ HUD =
...
+      null
+    else if sel.focusNode == sel.anchorNode and sel.focusOffset == sel.anchorOffset
+      # The selection either *is* an element, or is inside an opaque element (eg. <input>).
+      sel.focusNode.childNodes[sel.focusOffset]
+    else if sel.focusNode.nodeType != sel.focusNode.ELEMENT_NODE
+      sel.focusNode.parentElement
+    else
+      sel.focusNode
+    document.activeElement?.blur()
+    focusNode?.focus()
```
smblott-github: Rather than working out whether we need to focus an input here, might we not instead remember where the selection landed in `execureFind` and restore whatever state we need here?

content_scripts/hud.coffee
```diff
@@ -42,6 +56,28 @@ HUD =
...
+
+    findModeQuery.rawQuery = data.query
+    handlerStack.bubbleEvent "keydown", data.event
```
smblott-github: Why is this needed?

content_scripts/vimium_frontend.coffee
```diff
@@ -5,7 +5,7 @@
 # "domReady".
 #
 
-findModeQuery = { rawQuery: "", matchCount: 0 }
+window.findModeQuery = { rawQuery: "", matchCount: 0 }
 findModeQueryHasResults = false
 findModeAnchorNode = null
 findModeInitialRange = null
```
smblott-github: As this has been extended and extended, it might be better soon to collect these together into an object, something like `findMode.query`, `findMode.hasResults`, etc.

(Perhaps leave as is, for now, to keep the diff clean.)

pages/hud.coffee
```diff
@@ -1,3 +1,21 @@
+document.addEventListener "keydown", (event) ->
```
smblott-github: Why isn't this part of the `keydown` handler for FindMode?

Also, why are we only handling `keydown`, here?  Will that not cause problems for non-Latin keyboards?

pages/hud.coffee
```diff
@@ -10,6 +28,38 @@ handlers =
...
 
+  showFindMode: (data) ->
+    hud = document.getElementById "hud"
+    hud.innerText = "/"
+
+    inputElement = document.createElement "span"
+    inputElement.contentEditable = "plaintext-only"
+    inputElement.textContent = data.text
+    inputElement.id = "hud-find-input"
+    hud.appendChild inputElement
```
smblott-github: Could all of this not be a template within `hud.html`?

pages/hud.coffee
```diff
@@ -10,6 +28,38 @@ handlers =
...
+
+    inputElement.addEventListener "input", (event) ->
+      # Replace \u00A0 (&nbsp;) with a normal space.
+      UIComponentServer.postMessage {name: "search", query: inputElement.textContent.replace "\u00A0", " "}
```
smblott-github: Ah, this is where non-Latin keyboards are handled.

This should be a handler within `FindMode`.

content_scripts/hud.coffee
mrmr1993: The extra code here works out which element would have receieved a `focus` event if the focus wasn't in the HUD frame, which is more complicated than what we calculate and store in `executeFind`. By testing without it, it's quite easy to find examples of where this is necessary.

content_scripts/hud.coffee
mrmr1993: This was the easiest, small-change way to pass keys that are detected in the iframe to the `FindMode` `keydown` handler. I plan to turn it into a simple function call when refactoring after (/if) this PR is merged.

smblott-github:
Some explanation (comments?) regarding how this is working would be helpful.  Where is the focus while find-mode is active?

content_scripts/vimium_frontend.coffee
mrmr1993: Agreed, I was tempted to do that in this PR, but I held off for the sake of the diff. This is one of the things I want to do when I refactor `FindMode` later.

pages/hud.coffee
mrmr1993: This is in a different frame to `FindMode`. Instead, they interact via `UIComponent`, which forwards the events to `FindMode` via `handlerStack.bubbleEvent` (which you commented on above). Most of the `FindMode` `keydown` code will probably end up in here, but not yet, for the sake of a small change footprint.

> why are we only handling `keydown`, here?

We only handle backspace, delete, up, down, enter and escape here. There's an `input` handler a little way down, which handles all the actual text, and is locale-agnostic because it uses `input`.

pages/hud.coffee
mrmr1993: Again, this can't be in `FindMode` (at least yet) because they're in different frames.

mrmr1993:
> Some explanation (comments?)

I'll get on that in a few hours when I'm next at a computer. Thanks for looking over this @smblott-github.

smblott-github:
Here's what I understand so far (correctly or incorrectly).
- The focus is in the HUD, and all keyboard events happen there.
- Duplicate or "fake" or "copies of" `keydown` events are sent back to `FindMode` for `Enter`, `Escape`, etc.  This is also where the find-mode history is handled.
- As things stand, `FindMode` could be removed entirely, keeping only:
  - its `keydown` handler (which would be called in place of `handlerStack.bubbleEvent`, here, and
  - probably its `exit` handler.
- Almost everything (all key handling, `FindModeHistory`, etc.) except (something like) `executeFind` could be moved to `HUD.coffee`.

mrmr1993:
That's exactly right, yes.

smblott-github:
You should probably go whole-hog on this:
- Move everything possible to either `pages/hud.coffee`, or a new `pages/hud_find.coffee`.
- Just leave `executeFind` (or similar), `PostFindMode` and some initialisation/exit code in `frontend.coffee`.
- Move as many state variables as possible to `hud_find.coffee`.
- Use either `chrome.storage.local` or `Settings` to save and share state, where necessary (not global variables).  Probably `chrome.storage`; there's no need for any of that state to be cached.

smblott-github:
(And you may want to leave a shell `FindMode` in place, because I suspect there will be issues related to `focus` events when we think through how things should work with mulitple frames.)

smblott-github:
Ping?

mrmr1993:
@smblott-github I've pushed changes which move a lot of the code around, so the majority of find mode code is now in `mode_find.coffee`. There is still some code that could be moved out of `vimium_frontend.coffee`, but it's fiddly and I'm not really interested enough to do it right now.

The final commit here also fixes a bug that 73f66f25e6b8e5b5b8456074ad4fa79ba1d3ca4d introduced, where `<esc>` doesn't drop back into insert mode. Apart from that (and a few other slight tweaks), everything should work the same as the original PR.

smblott-github:
Thanks, @rmrm1993.

I took a look.  Unsurprisingly, the diff is becoming quite something!

I spotted a couple of [issues](https://github.com/smblott-github/vimium/tree/hud-iframe-input):

- c21aa9ab03d5537d575d3c46d3affce7f3ece5e5
- f34f842417bdac363fac5a3e47366b3ee94c2454
- 4f426affde34c895cf342dac256895ffe98ee4f3
- 14076fe862efc99970d160bd76b5434c89ab6387
- 8151d592b3ec2b37d4c41123cf94ed23d119d3c9

On *very-big* pages (e.g.  http://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html), there's quite some flicker in the HUD when you use the find-mode history.  The query and the match-count are updated asynchronously.  Might it be better to do them together? (Not sure if this is the right thing to do.)

mrmr1993:
Updated with your changes @smblott-github.

I still need to do some follow-up PRs to sort out flicker, make loss of HUD focus close the HUD, and possibly to move the HUD to the top frame only.

> The query and the match-count are updated asynchronously. Might it be better to do them together?

Can you clarify what you mean here? I've been considering the query to have come from the HUD frame and the match count from the main frame as a result of the query, so as far as I can tell, they *need* to be asynchronous.

smblott-github:
> Can you clarify what you mean here?

It's asynchronous.  Previously it wasn't.  And it doesn't look great.  We could make it synchronous.  Or perhaps moving the "Matches" to the right-hand side would make the flicker less obvious.

Is it possible to stop the cursor blinking?

Also, a fraction more space between the `/` and the cursor (when the query is empty) might look better.

mrmr1993:
> We could make it synchronous.

No, we can't:
* we've given up control of when the user's typed text shows, instead letting the browser handle it (which gives us all the advantages of this PR). So we can't just delay it until we have the count.
* we calculate the match count in a different frame (the top frame) to the HUD, so we have to communicate back and fourth, which is necessarily asynchronous.

We can (and should) discard counts that aren't for the current query, but that's about all we can do. I'll put this on my todo list.

> Is it possible to stop the cursor blinking?

Not without weird hacks that also stop selections working properly in the HUD.

> a fraction more space between the / and the cursor (when the query is empty) might look better.

I'll add this to my todo.

smblott-github:
> No, we can't:

Yes, we can! (Or we could.)  I'm was originally talking about using UP and DOWN arrow in the find-mode history.  There, *we* control what appears in the HUD and when.

But in the general case, we could extract `document.body.innerText` when find mode is launched, and pass it to the HUD itself.  Then update the matched count from there.  So, on "input", we:

- update the match count, then
- send the query to the page for `window.find`.

However, that's not the immediate priority.

I suggest you fix the spacing issue (a small thing) and we'll merge, then work forward from there.

mrmr1993:
> I suggest you fix the spacing issue (a small thing) and we'll merge, then work forward from there.

Done.

mrmr1993:
> I'm was originally talking about using UP and DOWN arrow in the find-mode history.

Fair point. There we're in complete control, so it's easily done.

Nonetheless, it would be much nicer to push the match counting/regexp searches out into a web worker:
* we'd be doing a (potentially) slow thing asynchronously; good for UX (but the opposite of what you were after),
* if the user tries to change the search terms while we were part-way through the last search, we could just kill that worker instead of waiting for the search to finish *and then* doing another one.

smblott-github:
OK, @mrmr1993.  Merged this (together with #1739, so I could test regexp mode).

Let's work on cleaning up the remaining issues.

PR #1694, opened 2015-05-29
Disable UIComponent for XML (merge).

smblott-github:
Follow on from #1647.

@mrmr1993. Looks like we probably have to disable the `Tween` too (02aaa6d7d3d852154dc9b0904a17575d857fc9df).

### DIFF
--- a/content_scripts/hud.coffee
+++ b/content_scripts/hud.coffee
@@ -56,6 +56,11 @@ class Tween
   styleElement: null
 
   constructor: (@cssSelector, insertionPoint = document.documentElement) ->
+    # Disable on XML pages.  See comment in UIComponent constructor.
+    unless styleSheet.style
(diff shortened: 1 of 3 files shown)


mrmr1993:
@smblott-github I think you misunderstood the test; `styleSheet` is a local variable containing a new `<style>` element for each `UIComponent`, and we check its `.style` to see if `Element.prototype.style` is defined ((X)HTML) or not (XML).

The test in `Tween` should be moved down to below the definition of `@styleElement`, and then check against that instead of `styleSheet`.

smblott-github:
@mrmr1993.  Thanks for picking up on that.  Do you want to just fix it yourself, directly?  My only point was that, if we're not initialising the HUD, then we shouldn't be initialising the Tween either.

smblott-github:
Closing in favour of #1698.

PR #1695, opened 2015-05-29
Custom search command (experimental, probably not for merge)

smblott-github:
This is an experiment with adding a new command, `Vomnibar.activateCustomSearch` (default binding `s`).  The idea is to provide direct access to custom search engines:
- So, for me, `sg` launches Google search, with completion.
- `sl` launches Google search, with completion and "I'm feeling lucky...".
- `sw` launches Wikipedia search, with completion.

After typing `s`, the HUD looks like this...

snapshot

with the obvious bindings.  In this case, there are two completion engines starting with `d`, so typing `d` gives...

snapshot

Thereafter, `<Enter>` launches the `d` search engine, and `w` launches the `dw` search engine.

Should we do this?  I'm not sure:
- It's yet another command.
- It's only a marginal improvement. "`sg`" with this PR has the same effect as "`og<Space>`" without.  So a saving of just a single keystroke.
- And, of course, we don't just get one new command, we get its *in-new-tab* counterpart too.

An alternative idea (and a hideous hack) would be to pass any trailing characters on vomnibar key mappings as the initial input in the vomnibar, so we'd achieve the same thing with:

```
map sg Vomnibar.activate g
```


smblott-github:
Closing in favour of #1697.

Issue #1269, opened 2014-11-24
Taming the Command Zoo


smblott-github:
We get many (perfectly reasonable) requests for new commands (here's one: #1264).  However, in my view, we already have _too many commands_.  We need a way of taming the command zoo.

Here's one idea:
- Currently, `count` prefixes have no meaning for `o`, and `b`.
- `o` open in current tab.
- `1o` open in current tab (for symmetry).
- `2o` open in a _new tab_.
- `3o` open in a _new window_.
- And similarly for bookmarks.

And, for mappings:

```
map o Vomnibar.activate        # open in current tab
map o cur Vomnibar.activate    # open in current tab (variant of above, for symmetry)
map O tab Vomnibar.activate    # open in a new tab
map w win Vomnibar.activate    # open in a new window
```

Existing bindings would be grandfathered or migrated. No need for three versions of each command.  Probably, a similar thing can be done for link hints.

If we were able to cut the number of vomnibar and link-hint commands by two thirds, we'd have a lot more room for maneuver.

dragon788:
I like where this is going. You can never have too many flags, unless you make it impossible to dump a raw list of them out somewhere. In my experience across a ton of command line driven systems, the only bad option is one that isn't easily accessible or is completely undocumented.

I've joked with my friend that I really need to write an infinite recursion tool in Python that calls every tool on a given system with a dictionary of commands and then read the STDOUT of the commands to find more options to pass until I have a complete map of every possible command and option. It really SHOULDN'T be that hard.

philc:
Closing this now that #4667 is merged. Also related to this discussion: #4634.

Issue #1280, opened 2014-12-02
Make the help dialog more "helpful"


mrmr1993:
To quote myself from [my comment here](https://github.com/philc/vimium/issues/1269#issuecomment-65218294), I think we should

> ... push all but the basic commands out of the help dialog and onto a separate page:

> * we stop needing a scrollable modal window (which is bad UX);
* there's no pressure to reduce the number of commands;
* there's space to discuss quirks (eg. we can't activate hover styles for `LinkHints.activateModeToHover` in #1032, we don't detect elements with click handlers for `LinkHints.activate`, etc.);
* users are more likely to search (searching in a modal dialog is weird);
* users searches for commands don't interact with the underlying page (eg. try opening help and searching for "scroll" on this page);
* the help dialog would be a place where (mainly new) users go to "get help" with using the extension (instead of being our sole user-facing documentation for commands).

In short: we should compile some proper documentation for commands, and then use the help dialog to reveal the most basic commands or direct users to the full documentation.

dragon788:
I have to say I like the idea of more help pages (tabs in the modal experience?) and I'm with mrmr1993 in wanting a keyboard exclusive interface for Chrome. As soon as I get some time to play with tiling window managers on Windows I'll probably stop touching my mouse almost entirely throughout the day.

PR #1697, opened 2015-05-29
Direct keyboard access to custom-search engines via keyword flag

smblott-github:
This extends the `map` command for *Custom key mappings* such that mappings (only for `vomnibar.activate`, here) can optionally take extra flags.  This idea was first mooted in #1269.

Here, the extra flag is a custom search-engine keyword.  Something like:

    map s vomnibar.activate keyword=g

Assuming a custom search engine...

    g: http://www.google.com/search?q=%s Google

the user can then launch Google search (with completion) with a single keystroke. So, with this example, the user types `s` and immediately gets...

snapshot

This is an alternative to #1695 (a better alternative, I think, since 1) it does not require a new command, and 2) it just feels slicker.

The idea is that this is an advanced feature, and would not be listed on the help page (`?`).  It would be described in the wiki.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -24,21 +24,13 @@ Commands =
       noRepeat: options.noRepeat
       repeatLimit: options.repeatLimit
 
-  mapKeyToCommand: (key, command) ->
+  mapKeyToCommand: ({ key, command, extras }) ->
     unless @availableCommands[command]
-      console.log(command, "doesn't exist!")
+      console.log command, "doesn't exist!"
       return
 
-    commandDetails = @availableCommands[command]
-
-    @keyToCommandRegistry[key] =
-      command: command
-      isBackgroundCommand: commandDetails.isBackgroundCommand
-      passCountToFunction: commandDetails.passCountToFunction
-      noRepeat: commandDetails.noRepeat
-      repeatLimit: commandDetails.repeatLimit
-
-  unmapKey: (key) -> delete @keyToCommandRegistry[key]
+    extras ?= []
+    @keyToCommandRegistry[key] = extend { command, extras }, @availableCommands[command]
 
   # Lower-case the appropriate portions of named keys.
   #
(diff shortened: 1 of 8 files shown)


smblott-github:
@mrmr1993 (and @philc).  What do you think of this?  In particular, the use of flags is a big departure from the established Vimium way of things.

mrmr1993:
I like it. Code LGTM.

smblott-github:
This seems to give a very natural search interface.  Here's one configuration (which works for me)...

Custom search engines (a subset):

```
# Wikipedia.
w: http://en.wikipedia.org/w/index.php?title=Special:Search&search=%s Wikipedia

# Google.
g: https://www.google.ie/search?q=%s Google
l: https://www.google.ie/search?btnI&q=%s I'm feeling lucky...
```

Custom search-engine mappings:

```
# Google custom searches.
map ss Vomnibar.activate keyword=g
map sS Vomnibar.activateInNewTab keyword=g

# Wikipedia custom searches.
map sw Vomnibar.activate keyword=w
map sW Vomnibar.activateInNewTab keyword=w

# Google "feeling lucky" custom searches.
map sl Vomnibar.activate keyword=l
map sL Vomnibar.activateInNewTab keyword=l
```

smblott-github:
Two things:
- I expect to merge this soon.  It's the way to go.
- There's a new wiki page describing vomnibar usage and tips [here](https://github.com/philc/vimium/wiki/Using-the-Vomnibar).

FYI ping @philc.

mrmr1993:
@smblott-github Based on this, can we prioritise a review of the help dialog? This opens the door to it becoming either hugely bloated or gradually becoming disconnected from what we offer (in addition to visual mode, which it already ignores).

For me, #1280 proposes the best solution to the help dialog issue, but conversation there seems to have stopped.

philc:
This looks okay to me and it's a cool idea. Before we share this with the
world, let's think about other use cases we may want to support and ensure
this syntax naturally extends to those use cases.

Another approach to parameterizing commands that I've seen is to use
something a little closer to actual javascript syntax, with positional
arguments, e.g.

`map sw Vomnibar.activate("w")`

Some possible future keybinding use cases to think through, off the top of
my head:

Mapping a key to:

* Go N pages back in history
* Close N tabs
* Open a bookmark with a given name, or a specific URL

I think this intersects with a proposal we've had on the issue tracker
awhile back for letting users define custom commands.

With this PR we certainly don't need to think through all of the
implications of doing that, or decide whether we want to, but it's worth
sanity checking whether this PR is a reasonable start down that path, or
whether it should wait until we've further baked our plans for/stance on
more sophisticated keymappings.

smblott-github:
(Unnecessarily long post deleted, and replaced with...)

I'm confident we're not unnecessarily closing any doors here.

smblott-github:
@philc.  The *options-like-Javascript-parameters* approach doesn't seem like the way to go.  With parameters, order matters.  So, if in the future we had multiple options, then the user would have to set the first even if they only cared about the second, and they'd have a to set a default value, and then we couldn't change the default value.  All of which is unnecessarily complicated.

The approach here seems both easy to understand and extensible.  For example, we could extend it to include on/off flags (`something`, `nosomething`), options whose values contain spaces (`key="some value"`), arbitrary JSON (`key="[1,2,3]"`).  I'm not proposing doing any of that now, just noting that those doors are not closed. (We'd possibly have to write a parser for the more complicated cases.)

Important point.  Keep the simple cases simple.  And the approach here achieves that.

@mrmr1993...

> or gradually becoming disconnected from what we offer

My sense is to keep the help page simple for noobs, and put the advanced stuff on the wiki.  Although, separately, we could do a better job of keeping the wiki up to date.

Will merge this.  Having key bindings for custom search engines is a nice feature.  I'll also make sure this is well described in the wiki.

smblott-github:
It turns out there's a problem with this.  It mucks up the layout of the help page...

snapshot

PR #1698, opened 2015-05-29
Disable UIComponent for XML documents

mrmr1993:
This is #1647, with improvements. From there:

> If the page is an XML document, nothing we do works:
* `<style>` elements show their contents inline,
* `<iframe>` elements don't load any content,
* `document.createElement` generates elements that
  - have `element.style == null`, and
  - ignore CSS.

> This PR stops us from injecting anything into the DOM from `UIComponent`, fixing #1640.


smblott-github:
Thanks, @mrmr1993.

PR #1699, opened 2015-05-30
Add a link to the wiki on the help page.

smblott-github:
Looks like this...

snapshot

Wiki opens in a new tab.

As suggested in by @LarryBattle in #1643.


LarryBattle:
+1

PR #1700, opened 2015-05-30
Move Vomnibar commands to own category on help page.

smblott-github:
The help page gets pretty lopsided when advanced commands are shown.  This balances things out a bit by creating a new category for Vomnibar commands in the right-hand column.  Like this...

snapshot

We could break up that left-hand column a bit, too.  Perhaps separate out the scrolling commands.


smblott-github:
Merging this.  It looks better this way.

(Also, note 95f0df630783274b686ee530816a842565413cf2, which didn't seem to warrant a PR of its own.)

PR #1702, opened 2015-05-30
Completion for google maps

smblott-github:
This adds search completion for Google Maps.  For example:

    gm: https://www.google.ie/maps?q=%s Google maps
    gm: https://www.google.com/maps?q=%s Google maps

(It also does some refactoring to simplify the specification of completion engines.)

### DIFF
--- a/background_scripts/completion_engines.coffee
+++ b/background_scripts/completion_engines.coffee
@@ -26,7 +26,6 @@ class RegexpEngine
 
 # Several Google completion engines package XML responses in this way.
 class GoogleXMLRegexpEngine extends RegexpEngine
-  doNotCache: false # true (disbaled, experimental)
   parse: (xhr) ->
     for suggestion in xhr.responseXML.getElementsByTagName "suggestion"
       continue unless suggestion = suggestion.getAttribute "data"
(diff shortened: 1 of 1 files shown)


smblott-github:
Here's a screenshot...

snapshot

(Notice, we're only getting places -- just one, here -- as suggestions.)

smblott-github:
Another screenshot...

snapshot

Most of those are pretty good.  "dark matter" is not so good.  It's being returned because we're searching for "map of matter".

smblott-github:
I'm not at all sure what to do with this; possibly revert it.

This is great for towns, cities, regions, countries, geographical features and the like.  However, it can't do address search, or queries like "chinese restaurant" or "bike shop".

A possible alternative is e62a8b4a9e3111d9d0df9a1e2cdd8ec3e8cc52fa, which does address search, and most of the above.  However, it's unacceptably brittle.  Check out [this line](https://github.com/smblott-github/vimium/commit/e62a8b4a9e3111d9d0df9a1e2cdd8ec3e8cc52fa#diff-fddc9b5a37a2e472642b6703c6004cd8R74)!

PR #1703, opened 2015-05-31
Document completion engines

smblott-github:
This adds automatic generation of documentation of completion engines for custom search engines.  The generated documentation looks like [this](https://github.com/smblott-github/vimium/blob/document-completion-engines/misc/completion_engines/completion_engines.md#completion-engines).

The idea is that, by tagging releases, we can automatically generated the related completion documentation.

(More thought is required as to how to integrate this into the build/release process.)

(This also includes a couple of tweaks to the completion engine search URLs.)


smblott-github:
Doh!

The right way to do this is to embed a page in the extension itself.

Closing.

PR #1704, opened 2015-05-31
Completion engine page

smblott-github:
#1703, but done better.

Add *built-in* documentation of the available completion engines.  The documentation is generated on-the-fly from the actual completion engines themselves.

The documentation page looks like this...

150531-03-snapshot

It is accessed from the link at the bottom right, here...

snapshot

Edit: eabf8f9df81a511c971c2b3cb7ab1cedf476aff9 tidies up the regular expressions so that we don't have all of those unnecessary backslashes.

### DIFF
--- a/background_scripts/completion_engines.coffee
+++ b/background_scripts/completion_engines.coffee
@@ -43,15 +43,15 @@ class Google extends GoogleXMLRegexpEngine
 # A wrapper class for Google completions.  This adds prefix terms to the query, and strips those terms from
 # the resulting suggestions.  For example, for Google Maps, we add "map of" as a prefix, then strip "map of"
 # from the resulting suggestions.
-class GoogleWithPrefix
+class GoogleWithPrefix extends Google
   constructor: (prefix, args...) ->
-    @engine = new Google args...
-    @prefix = "#{prefix.trim()} "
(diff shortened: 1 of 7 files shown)


smblott-github:
I need some input on how best to do this (@philc, @mrmr1993) -- and also on the look, etc.

If we're providing completions for *selected* custom search engines, then we need to tell users which ones.  How?
- They could read the source code (!!!).
- We could say nothing, and just *hope* that our regexps happen to match users' custom search engines.
- We could have an external web page somewhere (e.g. #1703, or on the wiki).  But then the list of completion engines may become out of date wrt the version the user is actually running.
- Or this, which integrates the page into Vimium itself.

Any input?  My sense is to go with this.

smblott-github:
Here's an update to what the page looks like.  Differences to above: better regexp formatting and we now include an example search URL...

snapshot

smblott-github:
(Sorry for all the noise...)

Or, better still, give the actual declaration...

snapshot

smblott-github:
Closing in favour of #1717.

smblott-github:
@philc.  This is one for you to take a look it.  It was committed as part of #1717. Let me know if you have an opinion.

Issue #1386, opened 2014-12-28
Migrate to a testing library that supports asynchronous tests


mrmr1993:
Much of our code relies on asynchronous chrome APIs, javascript events, etc. when run natively. Currently, with [shoulda.js](https://github.com/philc/shoulda.js/), we have to hack code into running synchronously with intrusive stubs. (This is especially obvious when trying to test the Vomnibar in an iframe, as in e33df51c6d1f870e19461ed9941c48dd07b4c68e.)

I propose we migrate to a framework with async capabilities like [CasperJS](http://casperjs.org/), with a view to implementing chrome APIs properly(-ish) and essentially simulating the extension in PhantomJS.

Being able to move beyond our current 'dom-aware unit test'-style tests towards a simulated extension has a few major wins:
- testing that we gracefully handle race conditions.
- testing async actions such as scrolling (and behaviour on key repeats, etc.).
- separate test cases for issues such as #1245, to automatically test our attempted fixes and to ensure we don't regress.
- testing user-facing output rather than implementation details, so we can write higher-level tests.

@philc shoulda.js was your brainchild, what do you think?

philc:
Async testing was introduced in shoulda.js awhile back, so now we're good!

PR #1705, opened 2015-05-31
Unified settings implementation

mrmr1993:
This is a follow-on to #1599 which replaces the frontend-only settings implementation `settings` with the version used everywhere else (`Settings`, in `lib/settings.coffee`). Benefits:
* no more duplicated settings code
* updated settings are passed to everywhere they're needed as soon as they are updated
* less code in `vimium_frontend.coffee` and `main.coffee`

Potential issues:
* the setting for toggling advanced commands in the help dialog is now updated from `chrome.storage.sync`
  - the value was already *stored* in `chrome.storage.sync`
  - updates from `chrome.storage.sync` were ignored by `Settings`, since it had no default value
  - the value in use was the one from `localStorage` in the background page (which should be the same as the one in `chrome.storage.sync` for nearly all users, especially those on a single machine).
* the settings loaded in content scripts are no longer limited to the reduced list from `settings.values`
  - I've left this list in the code, in case this is something we want/need to add back in at some point for eg. memory use
* `Utils.nextTick` had to be overridden for tests, since shoulda.js' lack of async support was causing tests to fail (relevant: #1386).

@smblott-github this is working nicely for me, and pretty much finishes what I'd been wanting to do towards unified settings; could you give it a look?

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -68,10 +68,6 @@ settings =
     @port = true
     Settings.init()
 
-  set: (key, value) ->
-    @init() unless @port
-    Settings.set key, value
-
   load: -> @init() unless @port
 
   addEventListener: Settings.addEventListener.bind Settings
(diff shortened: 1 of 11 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -41,14 +41,10 @@ textInputXPath = (->
...
 )()
 
-#
-# settings provides a browser-global localStorage-backed dict. get() and set() are synchronous, but load()
-# must be called beforehand to ensure get() will return up-to-date values.
-#
+# NOTE(mrmr1993): we use Settings everywhere instead of the dedicated implementation that was here.
+# Previously, only the values listed below would be loaded. If the space used by settings across all of our
+# content scripts is becoming an issue, then we can restrict the values we load to the list below.
 settings =
```
smblott-github: It's probably safe enough to remove these (and the comment above).  If you want to keep them, then they could at least be commented out, could they not?

philc:
Nice!

PR #1706, opened 2015-06-01
Unified settings implementation

smblott-github:
@mrmr1993. #1705 looks great.  What about this tweaked version?

(See ~~eabf8f9df81a511c971c2b3cb7ab1cedf476aff9~~ c62ffa33ad5230b89f44cb8f3268e6a4e48afd52.)

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -728,6 +728,4 @@ chrome.windows.getAll { populate: true }, (windows) ->
         (response) -> updateScrollPosition(tab, response.scrollX, response.scrollY) if response?
       chrome.tabs.sendMessage(tab.id, { name: "getScrollPosition" }, createScrollPositionHandler())
 
-# Start pulling changes from synchronized storage.
-Settings.init()
 showUpgradeMessage()
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -217,20 +217,6 @@ selectSpecificTab = (request) ->
     chrome.windows.update(tab.windowId, { focused: true })
     chrome.tabs.update(request.id, { selected: true }))
 
-#
-# Used by the content scripts to get settings from the local storage.
-#
-handleSettings = (request, port) ->
-  switch request.operation
-    when "get" # Get a single settings value.
-      port.postMessage key: request.key, value: Settings.get request.key
-    when "set" # Set a single settings value.
-      Settings.set request.key, request.value
-    when "fetch" # Fetch multiple settings values.
-      values = request.values
-      values[key] = Settings.get key for own key of values
-      port.postMessage { values }
-
 chrome.tabs.onSelectionChanged.addListener (tabId, selectionInfo) ->
   if (selectionChangedHandlers.length > 0)
     selectionChangedHandlers.pop().call()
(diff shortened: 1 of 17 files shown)


smblott-github:
@mrmr1993.  c62ffa33ad5230b89f44cb8f3268e6a4e48afd52 was my original idea to tidy things up a bit.  Then I just kept going, and got rid of Sync entirely, bringing all of the functionality into Settings, and simplifying it along the way.

Other significant changes:
- I think we have to do things this way (0f90453d269c5f8ab700123aa356a3d41026d925), unfortunately, otherwise we will get a race condition.  See comment in that diff.
- I found a long-standing (but not too worrying) bug: f50d3add7e6c95a5fc2945e067a52634a19c8a65. Not fixed yet.

mrmr1993:
@smblott-github this is quite hard to review in its current state, now it's reached over 2.5 times the size of #1705. Could we break it down into smaller steps, starting with merging #1705?

Beyond that,
* everything up to 83fefcae893f9ba57f291681f7b0328e6ee41db0 LGTM,
  - `Settings.getAsync` would be clearer than `Settings.use`.
* 5f0400ebac5867df74225b987ea1238bdaeb40b2 needs to be broken up if I'm to be able to review it effectively. I've already found a place where the new code and old code have different behaviour, but it's a struggle to check with so much going on.
  - I am happy to do this if you'd rather not.
* you might want to cherry pick 0f90453d269c5f8ab700123aa356a3d41026d925 on top of 83fefcae893f9ba57f291681f7b0328e6ee41db0 or even just #1705 so the bug doesn't land in master.

> I found a long-standing (but not too worrying) bug

I'll open a PR including the first batch of these changes succeeding #1705 shortly.

smblott-github:
> I've already found a place where the new code and old code have different behaviour

What's the different behaviour?

>  I'm to be able to review it effectively

Actually, Settings has become so short it's probably easier to read it from top to bottom.  Ignore the diff.

mrmr1993:
> What's the different behaviour?

Update hooks aren't run on `set`/`clear`.

smblott-github:
> Settings.getAsync would be clearer than Settings.use

Actually, it's just copying the same naming for HistoryCache, searchEngines, iframePort, etc...  I'd leave it the way it is.

smblott-github:
> Update hooks aren't run on set/clear.

The update hooks are, I think, being called in exactly the same places here as in 1.51 (and previous).

(Whether that's correct or not is another matter.)

mrmr1993:
> Actually, it's just copying the same naming for HistoryCache, searchEngines, iframePort, etc...

I'd argue that they're all the wrong name; names are supposed to be informative, but `use` gives no indication of why it's there. For `HistoryCache`, `searchEngines`, etc. `useAsync` might be a better name. I stand by `getAsync` for `Settings`.

mrmr1993:
5f0400ebac5867df74225b987ea1238bdaeb40b2 LGTM

smblott-github:
Thanks for that, @mrmr1993.

smblott-github:
Merged this, @mrmr1993.
With cherry-picked 0506491bc06b7b333042a8d393cd48d810b9ef3a.

Suggest you fix the JSON comprison as suggested and submit a regular PR over here.

PR #1708, opened 2015-06-01
Use JSON.stringify to compare setting values everywhere

mrmr1993:
Fixes the FIXME recently added in `settings.coffee`.


smblott-github:
Thanks,. @mrmr1993.
Merged this.
Not sure why it didn't close.
Tweak: 5812b01c5a3cfedb16d73127a5aec4e7468fcc02.

(Looks like Settings is in pretty good shape, now.)

PR #1714, opened 2015-06-03
Fix marks and add jump back to previous position

### LINKED ISSUE #1712: Global marks don't work (and may never have worked?)
(Am I going mad?)

Going right back to this commit aeb455a577c0a20684f357ff737203b5cb419cd4, it looks like global marks don't work, and may never have worked. (Seriously, am I getting this wrong?)

[This](https://github.com/philc/vimium/commit/aeb455a577c0a20684f357ff737203b5cb419cd4#diff-d1acbe0799862381d234f7aa6f4cffacR8) test never succeeds, because the preceding `keydown` event for `<Shift>` causes the handler to exit.

(So, if I'm right -- and I have tested this -- then we can conclude that nobody is using global marks.)

smblott-github:
This has become a major reworking of the `Marks` code.

###### First some small things...

- Fix global marks.
- Implement as "modes", and add an indicator (so there is visual feedback that the mode is active).
- Use `keypress` events instead of `keydown`.  This should help with internationalisation issues.
- Previously, `A-Z` were global marks and `a-z` were local marks.  Now, if the `Shift` key is depressed, then it's a global mark, otherwise it's a local mark.  This should also help with internationalisation issues.
- Various other refactoring.

###### Then the big change...

This implements "\`\`" (two backticks) to jump back to the previous location following `gg`, `G`, `n`, `N`, `/` or any (local) mark movement.  However, the way it does so is a bit unusual (cunning, perhaps).

Here's how it works. `Marks.activateGotoMode` is bound to some key sequence, usually just "\`".  We take the last character in that sequence (so, usually just "\`" itself) and, if that is the next character entered, then we jump to the previous position.  So, with the default binding, "\`\`" takes you to the previous position.  And a subsequent "\`\`" takes you back to where you started.

This works well for the default binding, and probably for many other natural bindings.  But there's a gotcha, of course.  If the user binds `Marks.activateGotoMode` to (say) "a", then the local mark "a" becomes unusable.

How bad is that?  Not too bad, I think.  The vast majority of users probably leave the default binding in place.  And it's easily discoverable there.

Fixes #1712.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,8 @@ Using marks:
 
     ma, mA  set local mark "a" (global mark "A")
     `a, `A  jump to local mark "a" (global mark "A")
-    ``      jump back to the position before the previous jump -- before gg, G, n, N, / or `a
+    ``      jump back to the position before the previous jump
+               -- that is, before the previous gg, G, n, N, / or `a
 
 Additional advanced browsing commands:
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Navigating the current page:
     F       open a link in a new tab
     r       reload
     gs      view source
-    i       enter insert mode -- all commands will be ignored until you hit esc to exit
+    i       enter insert mode -- all commands will be ignored until you hit Esc to exit
     yy      copy the current url to the clipboard
     yf      copy a link url to the clipboard
     gf      cycle forward to the next frame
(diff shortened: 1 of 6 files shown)


smblott-github:
Two more small points:
- This changes `m3j` from `m` (ignored) then `3j` to `m3` (create a mark named `3`) then `1j`.
- This trick works if (as I do) you re-map backtick to `'` (single quote).  In this case, `''` takes you to the previous position.

mrmr1993:
> Use `keypress` events instead of `keydown`. This should help with internationalisation issues.

What internationalisation issues are there? Surely being unable to access marks created in a locale with a different character set (after switching keyboards) is worse?

smblott-github:
Closed in favour of #1716.

PR #1715, opened 2015-06-04
On the help page, use separate row for very-long bindings.

smblott-github:
If there are many bindings for a command, then the column layout on the help page can become out of whack.  This puts such bindings on a separate table row which spans all three columns.

(Many bindings for a single command are more likely in light of #1697.)

Before:

150604-01-snapshot

After:

150604-02-snapshot


smblott-github:
Merging.  Should be a no-op for most users, and fixes an unfortunate formatting issue.

PR #1716, opened 2015-06-04
Improved global marks

smblott-github:
This makes global marks:
- persistent across tab closures and browser sessions, and
- synchronised between browser instances.

Effectively, this makes global marks another kind of bookmark.

This PR includes and is an extension of #1714.  The differences are difficult to see in the diff.  They are:
- [this](https://github.com/smblott-github/vimium/blob/global-marks/background_scripts/main.coffee#L38) line, so that we can set the scroll position after a tab has been opened, and
- all of [this](https://github.com/smblott-github/vimium/blob/global-marks/background_scripts/marks.coffee) file.

Here's the approach:
- If the tab still exists, use it.
- Otherwise, if another tab exists with a matching URL, then use that instead.
- Otherwise, create a new tab and set the scroll position within that.

We use `vimiumSecret` to detect whether we can re-use a tab Id.  Global marks are stored not in memory, as previously, by in `chrome.storage.sync`.

### DIFF
--- a/background_scripts/marks.coffee
+++ b/background_scripts/marks.coffee
@@ -11,15 +11,27 @@ Marks =
   # tabId can be considered valid.
   create: (req, sender) ->
     chrome.storage.local.get "vimiumSecret", (items) =>
-      item = {}
-      item[@getLocationKey req.markName] =
+      markInfo =
         vimiumSecret: items.vimiumSecret
         markName: req.markName
         url: @getBaseUrl sender.tab.url
         tabId: sender.tab.id
         scrollX: req.scrollX
         scrollY: req.scrollY
-      chrome.storage.sync.set item
+
+      if markInfo.scrollX? and markInfo.scrollY?
+        @saveMark markInfo
+      else
+        # The front-end frame hasn't provided the scroll position (because it's not the top frame within its
+        # tab).  We need to ask the top frame what its scroll position is. (With the frame Id set to 0, below,
+        # the request will only be handled by the top frame within the tab.)
+        chrome.tabs.sendMessage sender.tab.id, name: "getScrollPosition", frameId: 0, (response) =>
+          @saveMark extend markInfo, scrollX: response.scrollX, scrollY: response.scrollY
+
+  saveMark: (markInfo) ->
+    item = {}
+    item[@getLocationKey markInfo.markName] = markInfo
+    chrome.storage.sync.set item
 
   # Goto a global mark.  We try to find the original tab.  If we can't find that, then we try to find another
   # tab with the original URL, and use that.  And if we can't find such an existing tab, then we create a new
--- a/background_scripts/marks.coffee
+++ b/background_scripts/marks.coffee
@@ -65,14 +65,22 @@ Marks =
   focusOrLaunch: (markInfo) ->
     chrome.tabs.query { url: markInfo.url }, (tabs) =>
       if 0 < tabs.length
-        # We have a matching tab: use it.
-        @gotoPositionInTab extend markInfo, tabId: tabs[0].id
+        # We have a matching tab: use it (prefering, if there are more than one, one in the current window).
+        @pickTabInWindow tabs, (tab) =>
+          @gotoPositionInTab extend markInfo, tabId: tab.id
       else
         # There is no existing matching tab, we'll have to create one.
         chrome.tabs.create { url: @getBaseUrl markInfo.url }, (tab) =>
           # Note. tabLoadedHandlers is defined in "main.coffee".  The handler below will be called when the tab
           # is loaded, its DOM is ready and it registers with the background page.
           tabLoadedHandlers[tab.id] = => @gotoPositionInTab extend markInfo, tabId: tab.id
+
+  # Given a list of tabs, pick one in the current window, if possible, otherwise just pick any.
+  pickTabInWindow: (tabs, continuation) ->
+    chrome.windows.getCurrent ({ id }) ->
+      tabsInWindow = tabs.filter (tab) -> tab.windowId == id
+      continuation tabsInWindow[0] ? tabs[0]
+
 
 root = exports ? window
 root.Marks = Marks
--- a/background_scripts/marks.coffee
+++ b/background_scripts/marks.coffee
@@ -81,6 +81,5 @@ Marks =
       tabsInWindow = tabs.filter (tab) -> tab.windowId == id
       continuation tabsInWindow[0] ? tabs[0]
 
-
 root = exports ? window
 root.Marks = Marks
--- a/background_scripts/marks.coffee
+++ b/background_scripts/marks.coffee
@@ -59,6 +59,11 @@ Marks =
           if baseUrl == @getBaseUrl tab.url
             # We have a matching tab.  We'll use it.
             return @gotoPositionInTab extend markInfo, tabId: tab.id
+      # There is no existing matching tab, we'll have to create one.
+      chrome.tabs.create { url: @getBaseUrl(markInfo.url) }, (tab) =>
+        # Note. tabLoadedHandlers is defined in "main.coffee".  This handler will be called when the tab has
+        # is loaded, its DOM is ready and it registers with the background page.
+        tabLoadedHandlers[tab.id] = => @gotoPositionInTab extend markInfo, tabId: tab.id
 
   getBaseUrl: (url) -> url.split("#")[0]
 
--- a/background_scripts/marks.coffee
+++ b/background_scripts/marks.coffee
@@ -63,18 +63,16 @@ Marks =
   # The tab we're trying to find no longer exists.  We either find another tab with a matching URL and use it,
   # or we create a new tab.
   focusOrLaunch: (markInfo) ->
-    chrome.windows.getAll { populate: true }, (windows) =>
-      for window in windows
-        for tab in window.tabs
-          if markInfo.url == @getBaseUrl tab.url
-            # We have a matching tab: use it.
-            @gotoPositionInTab extend markInfo, tabId: tab.id
-            return
-      # There is no existing matching tab, we'll have to create one.
-      chrome.tabs.create { url: @getBaseUrl markInfo.url }, (tab) =>
-        # Note. tabLoadedHandlers is defined in "main.coffee".  The handler below will be called when the tab
-        # is loaded, its DOM is ready and it registers with the background page.
-        tabLoadedHandlers[tab.id] = => @gotoPositionInTab extend markInfo, tabId: tab.id
+    chrome.tabs.query { url: markInfo.url }, (tabs) =>
+      if 0 < tabs.length
+        # We have a matching tab: use it.
+        @gotoPositionInTab extend markInfo, tabId: tabs[0].id
+      else
+        # There is no existing matching tab, we'll have to create one.
+        chrome.tabs.create { url: @getBaseUrl markInfo.url }, (tab) =>
+          # Note. tabLoadedHandlers is defined in "main.coffee".  The handler below will be called when the tab
+          # is loaded, its DOM is ready and it registers with the background page.
+          tabLoadedHandlers[tab.id] = => @gotoPositionInTab extend markInfo, tabId: tab.id
 
 root = exports ? window
 root.Marks = Marks
--- a/background_scripts/marks.coffee
+++ b/background_scripts/marks.coffee
@@ -1,6 +1,6 @@
 
 Marks =
-  # This returns the key which is used for storing mark locations in chrome.storage.local.
+  # This returns the key which is used for storing mark locations in chrome.storage.sync.
   getLocationKey: (markName) -> "vimiumGlobalMark|#{markName}"
 
   # Get the part of a URL we use for matching here (that is, everything up to the first anchor).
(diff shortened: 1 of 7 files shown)


background_scripts/marks.coffee
```diff
@@ -1,34 +1,80 @@
...
-    scrollX: mark.scrollX
-    scrollY: mark.scrollY
-  chrome.tabs.sendMessage mark.tabId,
-    name: "showHUDforDuration",
-    text: "Jumped to global mark '#{req.markName}'"
-    duration: 1000
+  # The tab we're trying to find no longer exists.  We either find another tab with a matching URL and use it,
+  # or we create a new tab.
+  focusOrLaunch: (markInfo) ->
+    chrome.windows.getAll { populate: true }, (windows) =>
```
mrmr1993: This would be equivalent to

```coffee
chrome.tabs.query {url: [markInfo.url, markInfo.url + "#*"]}, (tabs) ->
  if tabs.length > 0
     # We have a matching tab: use it.
    @gotoPositionInTab extend markInfo, tabId: tabs[0].id
  else
    # There is no existing matching tab, we'll have to create one.
    chrome.tabs.create { url: @getBaseUrl markInfo.url }, (tab) =>
      # Note. tabLoadedHandlers is defined in "main.coffee".  The handler below will be called when the tab
      # is loaded, its DOM is ready and it registers with the background page.
      tabLoadedHandlers[tab.id] = => @gotoPositionInTab extend markInfo, tabId: tab.id
```

which skips us checking each window and tab manually.

content_scripts/marks.coffee
```diff
@@ -1,45 +1,75 @@
...
+  activateGotoMode: (registryEntry) ->
+    # We pick off the last character of the key sequence used to launch this command. Usually this is just "`".
+    # We then use that character, so together usually the sequence "``", to jump back to the previous
+    # position.  The "previous position" is recorded below, and is registered via @setPreviousPosition()
+    # elsewhere for various other jump-like commands.
+    previousPositionKey = registryEntry.key[registryEntry.key.length-1..]
+    @mode = new Mode
+      name: "goto-mark"
+      indicator: "Go to mark..."
+      suppressAllKeyboardEvents: true
```
mrmr1993: This mode should have `exitOnEscape` so you can leave it without being forced to set a mark.

content_scripts/marks.coffee
```diff
@@ -1,45 +1,75 @@
...
+      keypress: (event) =>
```
mrmr1993: Why are we now using `keypress`? What issues does this resolve over `keydown`? Surely being unable to access marks created in a locale with a different character set (after switching keyboards) is worse?

smblott-github: With `keydown`, we basically haven't a clue what character the user is typing.  We get lucky for most Latin keyboards, but that's it.

Since this PR drops the requirement for marks to be `/[a-zA-Z]/` what do you propose using for the mark name if we use `keydown`?

> What issues does this resolve over keydown? Surely being unable to access marks created in a locale with a different character set (after switching keyboards) is worse?

I'd say supporting users who change keyboards and locale on the fly is a pretty low priority.

content_scripts/marks.coffee
smblott-github: Thanks: b356e50677e1ae7ba68e27b1b70e59744e129627.

content_scripts/marks.coffee
mrmr1993: > what do you propose using for the mark name if we use `keydown`?

Why not the return value of `KeyboardUtils.getKeyChar`? It will need a fix for the shift issue first, of course.

> I'd say supporting users who change keyboards and locale on the fly is a pretty low priority.

These are the only users we have to give any thought to supporting; everybody else can just press the same key again and it will refer to the same mark.

background_scripts/marks.coffee
smblott-github: Thanks: 5901968be865588c2444055f8e63916a22c01156.

content_scripts/marks.coffee
smblott-github: Not quite, because we show the mark which was set or used in the HUD.  We'd want users to see a message that corresponds to what they pressed.

content_scripts/marks.coffee
```diff
@@ -1,45 +1,77 @@
...
+          else
+              localStorage[@getLocationKey keyChar] = @getMarkString()
+              @showMessage "Created local mark", keyChar
+
+  activateGotoMode: (registryEntry) ->
+    # We pick off the last character of the key sequence used to launch this command. Usually this is just "`".
+    # We then use that character, so together usually the sequence "``", to jump back to the previous
+    # position.  The "previous position" is recorded below, and is registered via @setPreviousPosition()
+    # elsewhere for various other jump-like commands.
+    previousPositionKey = registryEntry.key[registryEntry.key.length-1..]
```
mrmr1993: This is a really confusing thing to do. Vim has marks for \` as the previous location, and `'` as the line of the previous location (which for us should mean only restore y scroll position?), and we should probably stick with those.

content_scripts/marks.coffee
mrmr1993: * For any Latin keyboard layout, it will be the same as the key they pressed for any ASCII key, and users without Latin keyboards will be able to see which mark they could use in a Latin layout.
* If non-Latin keyboards are such a concern, we could show the message that the mark has been saved/loaded on `keypress`, quoting the actual mark name (from `keydown`) and the key pressed (from `keypress` in brackets).
* Restricting to `[a-zA-Z0-9]` would probably be wise (for now) too.

Is this at all reasonable?

content_scripts/marks.coffee
smblott-github: Vimium is not vim, so we should use vim-like concepts where appropriate, but do what's right in the browser at the end of the day.  I'm not sure the distinction you're making is particularly important for us.

I'll accept that this is a bit of a cunning trick.  I wouldn't even consider it if it wasn't easily discoverable for many users (which it is, because with the default bindings it just works, and does pretty much what any vim user would expect it to do).

In fact, it was only when I thought of this trick that I entertained the idea of doing this at all.  This is not important enough for a new option (or even a new command, for that matter).  However, imagine for a moment that we did add a new command.  Because <tt>\`</tt> is already a binding, it would be impossible to map the new command to <tt>\`\`</tt>.  That's why I think this is the way to go.

So, for most users, it just works as they'd expect.  And we can add a wiki entry for those curious enough.

mrmr1993: > Because \` is already a binding, it would be impossible to map the new command to \`\`.

Sorry, I meant keep the same approach (treat it as a mark), but don't compute it from the mapping. That the 'register' containing the last location moves around depending on the user's mapping is weird and confusing.

content_scripts/marks.coffee
smblott-github: We *could* do all of that.  But what is the advantage of `keydown` over `keypress`?

(The change-keyboard, change-locale use case hasn't convinced me.)

With `keypress`, I get the same mark name on my desktop with some whacky gaming keyboard as I do on my Chromebook.  That has to be a plus.

mrmr1993: > whacky gaming keyboard

If the keyboard is a Latin one, then it should be indistinguishable, no?

content_scripts/marks.coffee
smblott-github: So you're proposing hardwiring <tt>\`</tt> as the name of the register?  That would definitely be possible, and natural for most users.

I use `'` for `Marks.activateGotoMode`, and I *really* want to be able to just hit it twice.  Even venerable old vi uses `''` for this.

This trick allows us to support established vim-like movement bindings without having to hardwire anything or add any new options/commands.  And the cost is marginal.

(The cost is that it becomes `aa`, if the user binds `Marks.activateGotoMode` to `a`.  Well, even in that case, I'd say I'd want it to be `aa`.  This is a binding that should be fast to type and require no thought.  That's why it's <tt>\`\`</tt> and `''` in vim.)

mrmr1993: > So you're proposing hardwiring \` as the name of the register?

And `'`, if you'd like.

> even in that case, I'd say I'd want it to be `aa`.

I'd like to think that this isn't going to be a common use case, and it would be nice to think of marks as registers (as in Vim), so we could also use them for macros and copy/paste in the future. To have the 'last location' register jumping around is not useful for this.

There's also the issue of multi-key bindings (which is even weirder), and any non-printing-character mapping (surrounded by `<`, `>`) will map it to `>` without any reasonable explanation for the user beyond 'implementation'.

smblott-github: OK.  I accept that.

(I particular dislike your `>` example.)

We bind to those two special register names, and we catch the vast majority of users with no special explanation or documentation at all.

content_scripts/vimium_frontend.coffee
```diff
@@ -254,16 +255,27 @@ executePageCommand = (request) ->
   # All other commands are handled in their frame (but only if Vimium is enabled).
   return unless frameId == request.frameId and isEnabledForUrl
 
-  if request.registryEntry.passCountToFunction
+  if commandType == "Marks"
+    Utils.invokeCommandString request.command, [request.registryEntry]
```
mrmr1993: This is no longer necessary after 2473475367411fdeb051a6752d0059e337ad09e3

smblott-github: Thanks.  Yes, I'll unwind this (and some of the other related changes).

content_scripts/marks.coffee
mrmr1993: > But what is the advantage of `keydown` over `keypress`?

* For `[a-zA-Z0-9]`, what we calculate from `keydown` is the same as what we get from `keypress` on Latin keyboards.
* For a non-Latin keyboard, we will set the Latin equivalent from the `en_US` keyboard, and show them that we've set the mark under that equivalent.
* For users with multiple non-Latin keyboards, or one or more non-Latin keyboards and `en_US` (potentially across computers, since we now sync marks), the same key in the same position will *always* result in the same mark. With `keypress` the user has to switch to the correct layout first (if it's even on the computer they're using) before they can activate the mark.
* Old non-Latin local marks will no longer work (because they were set in this style originally).

@smblott-github it feels wrong to me to do this just so that single-keyboard non-Latin users can have a prettier message in the HUD. Am I just going to have to live with it?

smblott-github: Sorry, @mrmr1993, but I still don't get this.

Let me try a different angle.

In normal mode, all printable characters (which is what we're talking about here) are handled on `keypress`, not on `keydown`.  If there's a difficulty using `keypress` in this little corner of Vimium (that is, handling marks), then there's a whoppingly more serious problem in normal mode.

If my reasoning above is correct, then the discussion here seems moot.

mrmr1993: The difference here is that there's no config a user can tweak to make multiple layouts consistent, whereas normal mode can be made consistent just by adding the right mappings. If we do something weird here, then we cause pain that users can't mitigate.

Maybe there can be a compromise. Looking at `keypress` it seems as though we should be able to extract `keyCode` and `charCode`. I think it deals with both of our concerns if we
* associate the character (derived from `charCode`) *and* the key (`keyCode`) with each mark
* try to go to the mark for the character when asked
* if that fails, try to go to the mark for the `keyCode`
* if that also fails, show the error message.

This way, everybody should be able to access the mark they meant to from any keyboard layout. Is this plausible?

smblott-github: Still not with you, @mrmr1993, I'm afraid.  The name of the mark should be the character you get when you hit the key in an input, which is usually the character printed on the key.  We shouldn't be in the business of second guessing the user or the user's configuration (e.g. `setxkbmap`) after that.

(Users of multiple keyboard types or locales just have to do several things twice: their vim/tmux/screen configuration, their window-manager key bindings, their Vimium configuration, and here, their marks. C'est la vie.)

smblott-github:
@mrmr1993.  Merging this.  It's a huge functionality win, notwithstanding any discussion around details of the implementation.  Thanks for you input.

smblott-github:
Since this puts more data in `chrome.storage.sync` (and there are quotas for that) it seemed like an idea to check how much space we're actually using (b85ac49b0bbfc244815c944ddcfefd69a9c23254)...

    exclusionRules 14 1129 1143
    filterLinkHints 15 6 21
    findModeRawQuery 16 10 26
    findModeRawQueryList 20 272 292
    grabBackFocus 13 6 19
    helpDialog_showAdvancedCommands 31 6 37
    key 3 348 351
    keyMappings 11 1641 1652
    newTabUrl 9 42 51
    omniSearchWeight 16 5 21
    regexFindMode 13 6 19
    scrollStepSize 14 4 18
    searchEngines 13 2203 2216
    searchUrl 9 37 46
    vimiumGlobalMark|D 18 117 135
    vimiumGlobalMark|F 18 113 131
    vimiumGlobalMark|I 18 114 132
    vimiumGlobalMark|J 18 113 131
    vimiumGlobalMark|K 18 118 136
    vimiumGlobalMark|M 18 117 135
    vimiumGlobalMark|P 18 115 133
    vimiumGlobalMark|R 18 131 149
    vimiumGlobalMark|S 18 117 135
    vimiumGlobalMark|T 18 108 126
    vimiumGlobalMark|£ 18 132 150

    total 7405

The quota is 102400 bytes. (The columns are count for key, count for value and the sum of those two.)

The big-ticket items are `keyMappings`, `exclusionRules` and `searchEngines`, probably not surprisingly.

(Note: I recently trashed my exclusion rules while experimenting with something.  But, even if you multiply by ten, we're still well within quota.)

PR #1717, opened 2015-06-06
Rework custom search-engine completions

smblott-github:
This includes several things:
- Up to 97e5da6cbdcf1b5cf4f80cf550fc26c1ce194a3e, this is just #1704.
- f5db9ea1dc8d23dff732dd3345bf85798d64f1e9 and 2d5a01c9791a81aa87eaa935a1183f10950bdc84 are a refactoring of how completion engines are specified.  Having had a little experience working with completions, it became apparent that it would be helpful to simplify the approach.  Here, completion-engine specifications look like this...

```Coffeescript
class Wikipedia extends BaseEngine
  constructor: ->
    super
      engineUrl: "https://en.wikipedia.org/w/api.php?action=opensearch&format=json&search=%s"
      regexps: "^https?://[a-z]+\\.wikipedia\\.org/"
      example:
        searchUrl: "http://www.wikipedia.org/w/index.php?title=Special:Search&search=%s"
        keyword: "w"

  parse: (xhr) -> JSON.parse(xhr.responseText)[1]
```

- ... So, (usually) everything is data, except the method for parsing the response, which can be quite simple (as here).

Then, 4eda19de339212f86a9b008a4f3142a61d62829e is a significant change.  Say we have a custom search engine:

    j: https://www.google.ie/search?q=javascript+%s Javascript

and a query `object model`.  Previously, we only passed "object model" to the completion engine, so we got suggestions which are unrelated to Javascript.  With 4eda19de339212f86a9b008a4f3142a61d62829e, we pass "javascript object model" to the completion engine, which gives much better quality suggestions.

(4eda19de339212f86a9b008a4f3142a61d62829e also re-works the Google Maps completion engine, in light of the other changes.)

Finally, cb900a255113b8304d8931f7c6294e20f7f9f36d improves the filtering of completions from other completers.  There's an explanation in the commit itself.


smblott-github:
Merging.
This is part of a sequence of changes, as we get experience with completions, which improve the overall UX.

PR #1721, opened 2015-06-07
Use "Tab" to select active link hint (in filtered mode only)

smblott-github:
(This applies to link hints in filtered mode, that is when *Use the link's name and numbers for link-hint filtering* is selected.)

1. Make the active link hint (the one that will be activated if the user types `Enter`) visually distinct.
2. Make `Tab` (and `<Shift-Tab>`) select the activated link hint.

The visual indicator is quite subtle (perhaps too subtle).  It currently looks like this...

150607-01-snapshot


smblott-github:
Closing.  The same functionality is now in #1722.

Issue #1676, opened 2015-05-21
Unify key decoding for keydown


mrmr1993:
At the moment, we decode keys seperately for each `keydown` handler. With a view to reducing bugs and handling international keyboard layouts correctly, I think we should centralise conversion from `event` to the key we want to handle.

Pros:
- duplicated/similar code is moved to a library function.
- we will have to make anomalies in `keydown` handling explicit, which will help with debugging
- when (or before) [Chromium issue 227231](https://code.google.com/p/chromium/issues/detail?id=227231) is resolved, we have a centralised place to implement on the new feature. From this point, we can stop relying on the depreciated `keypress` event, by using `event.key` when it's available:
  - this will solve all keyboard internationalisation issues immediately (passing keys we are going to handle, and blocking the events for those we aren't)
  - we can't be pre-empted by `keydown` handlers for keys we want to handle, where `event.keyCode` disagrees with what we get from a `keypress` event (eg. a page handles `a`, but we handle an internationalised key that represents as `a` during `keydown`)
  - we move to a single key handler, since our `keypress` handlers will never get events for any key we handle
- friendlier key-handling code

Cons:
- code movement/refactoring
- bugs from missed edge cases due to platfrom inconsistency
- no visible benefit for the user
  - the Chromium issue will change this but will also require extra work
- redundancy in the code until old Chrome versions are abandoned (for `event.key` considerations only)

lydell:
Switching to `event.key` won't fix everything. You need `event.code` as well. See akhodakivskiy/VimFx#438 (be sure to read the last comment) and [vim-like-key-notation](github.com/lydell/vim-like-key-notation).

mrmr1993:
> (be sure to read the last comment)

@lydell I can't seem to find the comment you're referring to, could you quote it here?

To my understanding, `event.key` looks to provide the OS's interpretation of the key, which is currently what the codebase expects (often incorrectly) to be handling everywhere we deal with keys. Is there some nuance I've missed?

lydell:
Sorry I linked to the wrong PR. I've updated my comment above.

mrmr1993:
> Switching to `event.key` won't fix everything.

My suggestion in the issue was about replacing `keypress` handlers with `keydown` handlers that use `event.key`. This gives us the right behaviour under all internationalisations, fixing eg. #1411. Any mapping that currently works will work identically under this change, and those that don't will start to.

> You need `event.code` as well.

[...]

PR #1722, opened 2015-06-07
Use "keypress" for link hints (instead of "keydown")

### LINKED ISSUE #1387: Open links won't work with different language input method
I have multiple layouts: english and russian.  I can use almost all functionality of vimium extension with russian layout and mapping. But i can't choose links by letters from another layout (no mappings). 
Please bind shortcuts to actual physical keys or create mappings for letters

smblott-github:
Currently, if the user sets non-ASCII link-hint characters on the options page, they can't select links because we use the `keydown` event.  This uses `keypress` instead.  It should be the case that whatever characters the user can type into options-page option (from which we generate link hints), they can later type those same characters to activate link hints.

(Also, exit link-hints mode on `click` and `scroll` events.)

~~Fixes~~ Related to #1387 (I think).

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,7 @@ Release Notes
 - Bug fixes, including:
     - Bookmarklets accessed from the vomnibar.
     - Global marks on non-Windows platforms.
+    - Link-hints for non-Latin keyboards.
 
 1.51 (2015-05-02)
 
--- a/README.md
+++ b/README.md
@@ -161,6 +161,7 @@ Release Notes
 - A much improved interface for custom search engines.
 - Added <tt>\`\`</tt> to jump back to the previous position after selected jump-like movements.
 - Global marks are now persistent (across tab closes and browser sessions) and synced.
+- For filtered link hints (not the default), you can now use `Tab` to select hints.
 - Bug fixes, including:
     - Bookmarklets accessed from the vomnibar.
     - Global marks on non-Windows platforms.
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -456,6 +456,7 @@ class AlphabetHints
 # Use numbers (usually) for hints, and also filter links by their text.
 class FilterHints
   constructor: ->
+    @linkHintNumbers = Settings.get "linkHintNumbers"
     @activateOnEnter = true
     @hintKeystrokeQueue = []
     @linkTextKeystrokeQueue = []
@@ -476,7 +477,7 @@ class FilterHints
         @labelMap[forElement] = labelText
 
   generateHintString: (linkHintNumber) ->
-    (numberToHintString linkHintNumber + 1, Settings.get "linkHintNumbers").toUpperCase()
+    numberToHintString linkHintNumber + 1, @linkHintNumbers.toUpperCase()
 
   generateLinkText: (element) ->
     linkText = ""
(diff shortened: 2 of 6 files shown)


mrmr1993:
I have the same concerns about this that I did about the related change in #1716: a Russian user who also uses `en_US` (eg. with [this keyboard](http://ep.yimg.com/ay/yhst-16016120845095/russian-and-english-bilingual-black-keyboard-2.png)) could always select the correct link hints with the default settings by pressing the key that shows the character that's next in the link hint.

After this change, they will have to choose a locale for link hints, and switch keyboard layout to it every time they want to click a link. This is arguably worse than what #1716 did for such a user, since they have no way of mitigating it.

@smblott-github can we have an option allowing users to return to the old behaviour?

smblott-github:
I hear you, @mrmr1993.  And I agree that this case is different (the user has just one set of hint characters, so they have to be able to use them).

@vans239, @grey0ne.  You commented over in #1387.  Could you provide some input here?  The idea is that you can set and use non-Latin characters for link hints (e.g. Russian).  @mrmr1993's concern is that, by doing this, the Russian (say) characters wouldn't be available *at all* if the user changed locale.

smblott-github:
(As implemented here, this is broken.  It doesn't handle `Delete` correctly.)

mrmr1993:
> It doesn't handle `Delete` correctly.

Nor the `shift`/`ctrl` modifiers.

smblott-github:
@mrmr1993.  Is the following correct?

###### If we stick with `keydown`, then...

- **A** - A Russian user who sets their alphabet hint keys to `фывапролд` cannot ever use alphabetical hints.
- **B** - Filtered hints can never be used on websites which use non-ASCII characters.

Both of these are very serious.

###### If we move to `keypress`, then ...

- **C** - A user with non-ASCII alphabet hint keys may have to change their locale to use alphabetical hints.
- **D** - Similarly, a user may have to change their locale (to match that of the website) to use filtered hints.

###### So...

If the above is correct, the **A** and **B** are really bad, **C** and **D** aren't great, but they're not show stoppers.

mrmr1993:
That's right.

There's also the case where the user is on a page with one locale but their filter hints characters use another; there the user has to choose before they start (since they probably can't switch given how readily we exit link hint mode).

All in all, I'll conceed that `keypress` is probably the way to go for now, but it would be good to have an option to fall back to/only use `en_US` (via `keydown`). #1676 would be a nice place to start with this, if that's something we would consider doing.

smblott-github:
@mrmr1993....

See 6b29f4bb9fec6c61054c2037289d7ca6c0249c10 (currently on a different branch).  For alphabet hints, we can preserve the existing behaviour for existing users who are used to it, and support non-Latin characters too.  Might that be too confusing for users?

Doing the same thing for number hints isn't quite so obvious (since we match against both the hint characters and the hint texts).  We *could* look for non-ASCII characters in all of the hint texts, and choose `keydownKeyChar` if there are none.  However, then a single non-ASCII character (e.g. `é`) in the hint texts would change the behaviour.  Which seems too confusing for users.

mrmr1993:
@smblott-github that looks great, it should cover the cases I was concerned about nicely. (NB: The `useKeydown` regexp could include `0-9` as well without causing any problems.)

> Doing the same thing for number hints isn't quite so obvious... seems too confusing for users.

Agreed. Being able to filter links with non-ASCII characters is a big positive here, so it seems fair to let that and simplicity take priority.

smblott-github:
@mrmr1993.  This is merged now.  Could you watch out for gotchas vis-a-vis alphabet hints, please?  I use filter hints.

mrmr1993:
@smblott-github I missed it when this was merged, but this breaks the `<ctrl>` modifier for link hints, since chrome shortcuts are fired for e.g. `<c-a>`, `<c-s>`, etc.

(Clearly nobody uses that modifier in link hints mode, but still something worth considering.)

PR #1726, opened 2015-06-10
Rank filtered hints by score.

smblott-github:
Score and rank filtered link hints by relevancy. Better matches are likely to be either first (so just hitting `<Enter>` activates them) or just a `<Tab>` or two away.

Scoring:
- Requires that every search term be matched.
- Assigns higher scores to matches at the start of a word, and higher scores still for whole-word matches.

Note:
- This changes the meaning of a filter like `Hell Worl`. This would previously have required an exact match (modulo case).  Now, this matches `Hello World` and, indeed, `World Hello`.  This is similar to the way the vomnibar does its matching.


smblott-github:
This is a small corner or Vimium.  Nevertheless, this PR is "Wow" in terms of usability.  If you type the obvious filter, then on pretty much any page we find the right link after just a character or two, or the right link is just one or two Tabs away.

PR #1728, opened 2015-06-10
Fix tab move.

### LINKED ISSUE #1727: Feature request: moveTabStart, moveTabEnd
VimiumPlus has these commands. Since there is `moveTabLeft/moveTabRight` corresponding to `nextTab/previousTab`, it makes sense to have `moveTabStart/moveTabEnd` to correspond to `firstTab/lastTab` :)

[for reference](https://github.com/jb55/vimium-plus/blob/master/background_scripts/commands.coffee#L112)

(love vimium btw, thanks for the work!)

smblott-github:
There are two problems with `tabMoveLeft/Right`:
- For negative `count`, if there are pinned tabs, then the index we calculate is invalid.
- For positive `count`, if count is too large, then the index we calculate is invalid.

Fixes #1727 (kind of).

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -239,9 +239,12 @@ repeatFunction = (func, totalCount, currentCount, frameId) ->
 
 moveTab = (count) ->
   chrome.tabs.getAllInWindow null, (tabs) ->
+    pinnedCount = 0
+    for tab in tabs
+      pinnedCount += 1 if tab.pinned
     chrome.tabs.getSelected null, (tab) ->
       chrome.tabs.move tab.id,
-        index: Math.max 0, Math.min tabs.length - 1, tab.index + count
+        index: Math.max pinnedCount, Math.min tabs.length - 1, tab.index + count
 
 # Start action functions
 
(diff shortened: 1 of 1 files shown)


background_scripts/main.coffee
```diff
@@ -237,11 +237,14 @@ repeatFunction = (func, totalCount, currentCount, frameId) ->
...
 
-moveTab = (callback, direction) ->
-  chrome.tabs.getSelected(null, (tab) ->
-    # Use Math.max to prevent -1 as the new index, otherwise the tab of index n will wrap to the far RHS when
-    # moved left by exactly (n+1) places.
-    chrome.tabs.move(tab.id, {index: Math.max(0, tab.index + direction) }, callback))
+moveTab = (count) ->
+  chrome.tabs.getAllInWindow null, (tabs) ->
+    pinnedCount = 0
+    for tab in tabs
```
mrmr1993: Might this be more readable as `(tabs.filter (tab) -> tab.pinned).length`?

mrmr1993:
LGTM.

background_scripts/main.coffee
smblott-github: Thanks, @mrmr1993: abc24123a9bec8f07d019d31c08a8740e695401d.

PR #1729, opened 2015-06-10
Initialise options-page link-hint mode correctly (and an hilarious? bug)

smblott-github:
Trivial:
- We haven't been initialising the options page correctly (see diff).

Possibly hilarious?
- The `linkHintNumbers` option has been (pretty much) ignored for as far back as I've checked.  We ignore it when generating hints.  So setting it to anything other than its default value has (almost) no effect. Well, its effect is to make it impossible to select links by their hint.

(This is not quite true.  It does affect which characters are ignored when filtering link texts.  However, that's not at all its billing.)

What should we do? Probably just remove it as an option.

### DIFF
--- a/pages/options.coffee
+++ b/pages/options.coffee
@@ -229,8 +229,6 @@ initOptionsPage = ->
     element.className = element.className + " example info"
     element.innerHTML = "Leave empty to reset this option."
 
-  $("filterLinkHints").checked = bgSettings.get "filterLinkHints"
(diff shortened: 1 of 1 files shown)


philc:
Nice find. Agree!

smblott-github:
Humble pie.

The trivial bug is indeed a bug and is now fixed.  The "hilarious bug" isn't a bug at all.  I misunderstood part of the code.

Issue #1731, opened 2015-06-12
Regular expressions stopped working on master


mohitleo9:
Hi, regular expressions search stopped working on 532fa0779fd4733fc8a4bec963cd6c73a91b35f2

mrmr1993:
Regexp search seems to work fine for me on 532fa07 on my phone's Arch install (`Mozilla/5.0 (X11; Linux armv7l) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/43.0.2357.81 Safari/537.36`). Can you give more details of what's not working?

mohitleo9:
hmm, I am using Chrome Version 43.0.2357.124 (64-bit), and the above mentioned commit.
(it has the new interface for settings page). On my settings page the option to use regex in find is enabled but as soon as I try to search using regex it does says no results found. The regular text search works though. It's almost the regex search is not enabled. If i revert back a few commits it starts working again. I will try to do a git bisect and narrow down the commit.

mohitleo9:
I did a git bisect and found this is the culprit 0de6b076271b673d0e1dcc2b74b2ddd1646bf08e
. The commit does somethings to settings so I guess that might be breaking the regex search.

mrmr1993:
Thanks @mohitleo9, great find. Looks like your settings aren't all being carried over once we switch to `chrome.storage` in the frontend. As a workaround, you should be able to re-enable regexp searching by disabling it in the options, saving, enabling it, and then saving again.

I'll push a PR in the morning to push settings from `localStorage` to `chrome.storage.sync` as necessary, which should fix the general problem.

mohitleo9:
I tried the mentioned solution it didn't work for me. But I guess I can make the changes locally to get it working, I am really excited to use the new features!

smblott-github:
@mohitleo9.  Is it only the regexp setting you're having an issue with, or is it other settings which are used in the front end too?

E.g. Try changing the "Scroll step size" option, and see if that is effective.

smblott-github:
OK, @mohitleo9.  Got it: https://github.com/philc/vimium/commit/aa00e29dc2533b6701c65935223599671c5833b1 (which is so trivial, I just merged it directly).

@mrmr1993.  I wonder if we could write a test to verify that all of the settings on the options page have default values?

mrmr1993:
> Any user who has migrated to the new settings system (and uses regex search) will have to untick that box on the options page, save, and tick it again

[...]

PR #1732, opened 2015-06-14
Quick-fix: check that all settings have default values.

smblott-github:
On the options page, this outputs a console error message if there is an options-page settings which does not have a default settings value.

The idea is to prevent issues like #1731.

It would be better to have a test, so this is really just a quick fix.  However, it does verify that there are no more missing default values.

### DIFF
--- a/pages/options.coffee
+++ b/pages/options.coffee
@@ -22,6 +22,10 @@ class Option
     @element.addEventListener "change", @onUpdated
     @fetch()
     Option.all.push @
+    # All options-page settings must have default values.
+    # Note(smblott) It would be better to have a test for this.
+    unless bgSettings.defaults[@field]?
+      console.error "error: option #{@field} has no default value in settings."
 
   # Fetch a setting from localStorage, remember the @previous value and populate the DOM element.
   # Return the fetched value.


mrmr1993:
It's easy to check with a bash one-liner that, now, all the settings we use (except for testing and migration) are in the defaults:

```bash
vimdiff <(sort -u <(pattern="(?<=Settings\\.(set|get|has|use)[ \\(])('[^']+'|\"[^\"]+\")"; grep -ohrP --include="*.coffee" "$pattern" * | sed s/[\'\"]//g)) <(sort <(cat lib/settings.coffee | head -n 152 | tail -n 68 | grep -oP "(?<=^    )[^: ]+(?=:)"))
```

It might be more helpful from a development perspective to yell at the developer directly during manual testing, by adding the following to `Settings.set`:

```coffee
  alert "Vimium: called Settings.set on '#{key}', but '#{key}' has no default value. IT WILL NOT SYNC!" unless @shouldSyncKey key
```

I think this PR is more likely to give a false sense of security, I'd much prefer to see a runtime check like the above that is obvious enough for us to notice when testing.

smblott-github:
@mrmr1993.  Then this (83189319490ce02229709b2e8058566942c7a707) looks like a better approach.

smblott-github:
Merged 83189319490ce02229709b2e8058566942c7a707, closing this.

mrmr1993:
That commit LGTM. I'm slightly worried that it will give us a false sense of security that all settings have defaults, but it's a nice improvement from having nothing.

PR #1733, opened 2015-06-14
Sync localStorage to chrome.storage and bump version to 1.52

mrmr1993:
This ensures that settings set in old versions of Vimium are not 'forgotten' when we switch to using `chrome.storage.sync` for all settings implementations.

@smblott-github this is what I meant over in #1731.

### DIFF
--- a/lib/settings.coffee
+++ b/lib/settings.coffee
@@ -158,6 +158,20 @@ if Utils.isBackgroundPage()
   # We use settingsVersion to coordinate any necessary schema changes.
   if Utils.compareVersions("1.42", Settings.get("settingsVersion")) != -1
     Settings.set("scrollStepSize", parseFloat Settings.get("scrollStepSize"))
+
+  # Any settings set before we started using chrome.storage.sync were only stored in localStorage. If they've
+  # remained unchanged, then they won't be in chrome.storage.sync, and hence unusable with chrome.storage-
+  # backed Settings in the frontend. Here, we push settings in localStorage to chrome.storage.
+  # NOTE(mrmr1993): We only push settings for which there is no corresponding value in chrome.storage.sync.
+  # If there is a corresponding value, our value is either the most recent, or will be updated to it soon.
+  if Utils.compareVersions("1.52", Settings.get("settingsVersion")) == 1
+    localStorageClone = {}
+    for key, value of localStorage
+      localStorageClone[key] = value if Settings.shouldSyncKey key
+    chrome.storage.sync.get Object.keys(localStorageClone), (items) ->
+      delete localStorageClone[key] for key of items
+      chrome.storage.sync.set localStorageClone
+
   Settings.set("settingsVersion", Utils.getCurrentVersion())
 
   # Migration (after 1.49, 2015/2/1).
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 2,
   "name": "Vimium",
-  "version": "1.51",
+  "version": "1.52",
   "description": "The Hacker's Browser. Vimium provides keyboard shortcuts for navigation and control in the spirit of Vim.",
   "icons": {  "16": "icons/icon16.png",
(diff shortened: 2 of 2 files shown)


smblott-github:
Is bumping the version number appropriate for this?  I'm not sure.

I had a more direct approach in mind...  https://github.com/smblott-github/vimium/commit/e481644dc0d2f825c13e47901c08b4daa749c085.

mrmr1993:
@smblott-github that's not the (only) bug I'm trying to solve. When you implemented `Sync`, you didn't migrate settings from `localStorage` to `chrome.storage`, so any settings saved in 1.44 or earlier and unchanged until now won't be in `chrome.storage`. This PR fixes that.

smblott-github:
> ... you didn't migrate settings from localStorage to chrome.storage ...

Yes.  And I think that's the right thing to do, and we shouldn't "fix" it now.  In going from non-synced settings to synced settings, it means that the behaviour in a particular chrome instance doesn't change unless the user actively does something.  Thereafter, the settings reflect the user's most-recently expressed preference.

mrmr1993:
I presume that users who
* have been using filter hints, or
* removed `f` from their link hint characters, or
* tweaked their scroll step size, or
* changed their previous/next page patterns, etc.

before 1.45 landed (20th July 2014) are going to be upset if we suddenly start ignoring settings they've been using for over a year. Especially since the setting they're accustomed to will still show as set in the options page.

The way I've implemented this, if they've changed their settings on any instance since, then we won't push the local setting. If they haven't, but they previously changed a setting to a non-default value, we should respect the old settings.

> the behaviour in a particular chrome instance doesn't change unless the user actively does something.

If you want this to be true, we're going to have to back out the unified settings implementation.

smblott-github:
Moving conversation here, @mrmr1993.

Comments:

1. Don't bump the version number.

1. We should record the fact of the migration having been performed in `chrome.storage.sync`.  That is, we should do this once and once only, on whichever installation happens to be updated first.

1. We could restrict this to the settings which were used by content scripts in 1.51.  There's no need to mess with settings which are only used on the background page.  For example, we don't need to make changes key mappings.

mrmr1993:
> We should record the fact of the migration having been performed in `chrome.storage.sync`. That is, we should do this once and once only, on whichever installation happens to be updated first.

I disagree. If the user has gone to the effort of customising a setting, we shouldn't throw that away if another, less configured/unconfigured instance has been loaded first.

Honestly, starting to sync settings without user input was a mistake. This was always going to be a problem that will mess people up just because eg. they've left their chrome logged into someone else's computer. We should disable it and use `chrome.storage.local` by default, and add an option to sync settings.

Is there any reason not to do this?

smblott-github:
> I disagree. If the user has gone to the effort of customising a setting, we shouldn't throw that away if another, less configured/unconfigured instance has been loaded first.

Well, we could try to get the settings from the first instance to upgrade, or from the last.  The latter doesn't seem great, because then the setting may change several times before we find the last installation.

> Is there any reason not to do this?

I don't recall seeing any complaints.  If there have been, there aren't many.  So probably users are happy enough with the status quo.

mrmr1993:
> I don't recall seeing any complaints.

Let me be the first: I use a separate profile directory for testing Vimium and normal web browsing. I'm logged into Chrome on both to sync my *Chrome* settings. Because of this design decision, every time I need to test a setting it messes up my actual settings, and I have to go and change them back. It's stupid.

mrmr1993:
@smblott-github I have a new suggestion, which (hopefully) will keep everyone happy:
* add an option to disable sync and the supporting settings architecture
* default to sync
* if there are settings values in `localStorage` which aren't in `chrome.storage.sync`, disable sync and populate `chrome.storage.local` directly from `localStorage`
  - we should display a brief message next to the option to tell the user that and why it was automatically disabled when we do this.

@smblott-github Would there be any problems with this, or can I start work on it?

mrmr1993:
This was superseded by #1739. Closing.

PR #1734, opened 2015-06-14
Sync localStorage to chrome.storage.local and bump version to 1.52

mrmr1993:
This PR ensures that any setting available from `localStorage` in the background page is also available to content scripts, via `chrome.storage.local`. This fixes the issue:
* any setting set before Vimium 1.45 (and unchanged since) hadn't been propagated to `chrome.storage`, and so is unavailable to content scripts.

@smblott-github this also makes your statement from PR #1733 true, namely:

> the behaviour in a particular chrome instance doesn't change unless the user actively does something.

### DIFF
--- a/lib/settings.coffee
+++ b/lib/settings.coffee
@@ -206,12 +206,18 @@ if Utils.isBackgroundPage()
   # backed Settings in the frontend. Here, we push settings in localStorage to chrome.storage.
   # NOTE(mrmr1993): We only push settings for which there is no corresponding value in chrome.storage.sync.
   # If there is a corresponding value, our value is either the most recent, or will be updated to it soon.
-  if Utils.compareVersions("1.52", Settings.get("settingsVersion")) == 1
+  lessThan152 = Utils.compareVersions("1.52", Settings.get("settingsVersion")) == 1
+  if lessThan152 or not Settings.get "localStorageMigrated"
+    if lessThan152
+      delete localStorage["localStorageMigrated"]
+    else
+      localStorage["localStorageMigrated"] = true
+
     localStorageClone = {}
     for key, value of localStorage
       localStorageClone[key] = value if Settings.shouldSyncKey key
     chrome.storage.sync.get Object.keys(localStorageClone), (items) ->
-      delete localStorageClone[key] for key of items
+      delete localStorageClone[key] for own key of items
       chrome.storage.local.set localStorageClone
 
   Settings.set("settingsVersion", Utils.getCurrentVersion())
--- a/lib/settings.coffee
+++ b/lib/settings.coffee
@@ -171,7 +171,7 @@ if Utils.isBackgroundPage()
       localStorageClone[key] = value if Settings.shouldSyncKey key
     chrome.storage.sync.get Object.keys(localStorageClone), (items) ->
       delete localStorageClone[key] for key of items
-      chrome.storage.sync.set localStorageClone
+      chrome.storage.local.set localStorageClone
 
   Settings.set("settingsVersion", Utils.getCurrentVersion())
 
--- a/lib/settings.coffee
+++ b/lib/settings.coffee
@@ -159,6 +159,20 @@ if Utils.isBackgroundPage()
   # We use settingsVersion to coordinate any necessary schema changes.
   if Utils.compareVersions("1.42", Settings.get("settingsVersion")) != -1
     Settings.set("scrollStepSize", parseFloat Settings.get("scrollStepSize"))
+
+  # Any settings set before we started using chrome.storage.sync were only stored in localStorage. If they've
+  # remained unchanged, then they won't be in chrome.storage.sync, and hence unusable with chrome.storage-
+  # backed Settings in the frontend. Here, we push settings in localStorage to chrome.storage.
+  # NOTE(mrmr1993): We only push settings for which there is no corresponding value in chrome.storage.sync.
+  # If there is a corresponding value, our value is either the most recent, or will be updated to it soon.
+  if Utils.compareVersions("1.52", Settings.get("settingsVersion")) == 1
+    localStorageClone = {}
+    for key, value of localStorage
+      localStorageClone[key] = value if Settings.shouldSyncKey key
+    chrome.storage.sync.get Object.keys(localStorageClone), (items) ->
+      delete localStorageClone[key] for key of items
+      chrome.storage.sync.set localStorageClone
+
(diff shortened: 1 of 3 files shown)


smblott-github:
@mrmr1993. I think this is *very* difficult to get right (and therefore potentially error prone).  And it leaves us with an unnecessarily complex storage model.

There are a couple of issues with this implementation (I'll comment on those separately in a moment).  However, I'm not at all convinced that the cost of the complexity here is worth it.  We may be better off just forcing an update in `chrome.storage.sync`, where needed.

(And don't really need to bump the version number.)

lib/settings.coffee
```diff
@@ -15,9 +15,34 @@ Settings =
...
+
+      chrome.storage.onChanged.addListener (changes, area) =>
+        return unless area == "local"
+        for key, change of changes
+          # If change.newValue is non-null, a value has been added to chrome.storage.sync. For settings,
+          # the only values we care about here, this should only happen on the first run after the 1.52
+          # version bump. Everything else is filtered out in the following function call.
+          @handleUpdateFromChromeStorage key, change.newValue if change?.newValue?
+
       @onLoaded()
```
smblott-github: `@onLoaded()` will not run *at all* (and it must run) if `chrome.runtime.lastError` is set in `Settings.init()`.

lib/settings.coffee
```diff
@@ -15,9 +15,34 @@ Settings =
       unless chrome.runtime.lastError
         @handleUpdateFromChromeStorage key, value for own key, value of items
 
+        @loadSettingsFromLocalStorage()
+
       chrome.storage.onChanged.addListener (changes, area) =>
         @propagateChangesFromChromeStorage changes if area == "sync"
 
+  loadSettingsFromLocalStorage: ->
+    chrome.storage.local.get null, (items) =>
```
smblott-github: When the migration code (at the bottom, here) is run, we could see the values here from either *before* or *after*.  In other words, we may or may not see the values from `localStorageClone` on the first run.

lib/settings.coffee
```diff
@@ -15,9 +15,34 @@ Settings =
...
+            else
+              @handleUpdateFromChromeStorage key, value
+
+        # All of the values in overriddenBySync have equivalents in chrome.storage.sync which take
+        # priority, so we remove them.
+        chrome.storage.local.remove overriddenBySync if Utils.isBackgroundPage()
+
+      chrome.storage.onChanged.addListener (changes, area) =>
+        return unless area == "local"
+        for key, change of changes
```
smblott-github: I don't see how this is ever reached.  (There would have to be a `chrome.storage.local.set` somewhere, and I don't see one.  Or is this where you pick up the `localStorageClone` updates?

However, more importantly:

- I don't see `chrome.storage.local.remove()` anywhere.  And we *have to* remove these because, if the setting is reset to its default value, then an old (non-default) value from `chrome.storage.local` will take precedence, which is incorrect.

- The only mechanism for pushing options-page changes to content pages is `chrome.storage.sync`. When a setting which is not stored in `chrome.storage.sync` is changed back to its default value, there will be no `chrome.storage.sync` update, so such changes won't propagate to content pages.

lib/settings.coffee
mrmr1993: Sorry, that was an indentation typo. I'll fix it and push an update shortly.

lib/settings.coffee
mrmr1993: > I don't see how this is ever reached.

From the migration code. So if the migration happens after we've tried to get the values out of `chrome.storage.local` above, the values are still pushed to `@cache`.

I'll tweak the code to deal with your other concerns here.

smblott-github: > I'll tweak the code to deal with your other concerns here.

I'm not sure we should be doing this at all, @mrmr1993.  This is introducing too much additional complexity for very little benefit.  We should probably just force an update in synced storage for the offending settings and be done with it.

mrmr1993: Okay, can we merge #1733 then?

(I can tweak it to not use a version bump, if you'd prefer.)

smblott-github:
@mrmr1993.  Here's my version of how to make this work: https://github.com/smblott-github/vimium/blob/maintain-settings-in-chrome.storage.local/lib/settings.coffee#L164.

It's basically the same as your idea (we merge settings from `chrome.storage.local` and `chrome.storage.sync`) with a couple of extra ideas:

- Use `postUpdateHooks` to catch updates, and remove the key from `chrome.storage.local`.

- Also push the change to active tabs via `chrome.storage.sync`.

    This is a [hideous hack](https://github.com/smblott-github/vimium/blob/maintain-settings-in-chrome.storage.local/lib/settings.coffee#L193).  It works locally (I've checked, content scripts see both updates).  It's not clear whether it would work remotely.  Chrome might see the `.remove()`, drop the preceding `.set()` and then do nothing (because the key isn't set in `chrome.storage.sync`).

All of this is pretty complicated, and therefore risky.  It's making assumptions about chrome's behaviour, and it's assuming certain invariants as to what's stored in which storage area and how they're related.  What's more, if we go with this, we're basically stuck with all of those assumptions forever.

So it may be better to go with the blunt instrument of simply forcing updated settings to `chrome.storage.sync` (as we're discussing in #1733).

mrmr1993:
@smblott-github I've pushed an update to this that fixes the issues that you pointed out.

Having a look at your alternative
* I like the change to `postUpdateHook`.
* it uses `chrome.storage.sync.get` repeatedly on individual keys when we've already got the full contents.
* it keeps trying to migrate at every startup, forever.
* 11 levels of indentation at one point?!?!
* The ugly hack is ugly.

I prefer how this PR does it to your e0b50c6, but I really only made this PR to illustrate that this approach will make a mess of the settings code.

ron813c:
On Jun 16, 2015 10:06 AM, "Matthew Ryan" <notifications@github.com> wrote:
>
> @smblott-github I've pushed an update to this that fixes the issues that
you pointed out.
>
> Having a look at your alternative
>
> I like the change to postUpdateHook.
> it uses chrome.storage.sync.get repeatedly on individual keys when we've
already got the full contents.
> it keeps trying to migrate at every startup, forever.
> 11 levels of indentation at one point?!?!
> The ugly hack is ugly.
>
> I prefer how this PR does it to your e0b50c6, but I really only made this
PR to illustrate that this approach will make a mess of the settings code.

mrmr1993:
This was superseded by #1739. Closing.

PR #1736, opened 2015-06-16
Make "Show Advanced Options" state persistent.

smblott-github:
The "Show advanced commands" toggle on the help page is persistent (and synced).  This makes the "Show Advanced Options" state on the options page similarly persistent.

This helps with the use case where a user is twiddling with an advanced option, and constantly has to open the options-page fold.

(Separately, I dislike the fold.  This has the side effect of hiding the fold permanently for me.)


mrmr1993:
> I dislike the fold. This has the side effect of hiding the fold permanently for me.

Seconded, this is a great idea @smblott-github. Code LGTM.

Issue #1719, opened 2015-06-07
Remove "CSS for link hints" option


smblott-github:
(@philc has proposed this before, but met some fairly mild resistance.)

We should remove the *CSS for link hints* option.

I've just started looking at tweaking the link hints UX for filtered hints:
- `Enter` activates the "first" hint, but there's currently no visual indication as to which one that is.
- Add `Tab` to make the next hint the active one for `Enter`.

(Why? When you're down to two or three hints, hitting `Tab`-`Enter` requires less thought than actually working out what to type to activate the second or third hint -- or even to figure out which is the first hint.)

So I went looking for the CSS to tweak this, and it's in ~~`options.coffee`~~ `settings.coffee`.  That's crazy.  We can't tweak the appearance of link hints for users who aren't using the default value.

(We don't allow users to customise the appearance of vomnibar, or the HUD.  Why on earth do we allow customisation of the appearance of hints?)

gdh1995:
Well, in fact custom CSS is very useful for me: Vimium will filter out those nodes whose width or height is less than 3px, but I have turned off showing-image on most websites, which makes `<a><img></img></a>` often with a size of 0x0...

Then, I've set:
``` CSS
img {
  min-width: 3px;
  min-height: 3px;
}
```

Cool, isn't it?

smblott-github:
@gdh1995. But that has nothing to do with link hints, and nothing really to do with Vimium at all.  You'd probably be better off doing that kind of thing with TamperMonkey or your own custom extension.

mrmr1993:
@smblott-github it's part of a workaround for a decision we've made. Having a place that people can put css to make the extension work better for them is a good thing.

The real issue is that the default value isn't a no-op, which means that we have to take it into account for any changes we make. Simple fix: move the actual css content into `vimium.css`, and just keep the default as the selectors without any rules. I can see no benefits from removing it entirely.

(Full disclosure: I have a few custom styles on mine too, and so I'd be a bit peeved if this went away.)

gdh1995:
> doing that kind of thing with TamperMonkey

I know, but it's too heavy than just what I need.

We may rename `userDefinedLinkHintCss` to  `userDefinedCss`.

PR #1739, opened 2015-06-17
Store default settings value and fix settings/sync problem

smblott-github:
(@mrmr1993: This is yet another approach to the Settings problem.)

With the new Settings implemetation, settings which have a non-default value and which are not in synced storage (that is, they have not been changed since synced storage was introduced) are not currently accessible to content scripts.  This commit makes such settings accessible via chrome.storage.local.

Important:

- There's a change to the established settings data model here.  Previously, settings with default values were not stored; here, they are.  This eliminates a considerable amount logic from Settings, but means that migrations will be required if default values are changed in future.  (Other than type changes, have we ever changed a default value?)

- There's also a change (bug fix?) to the behaviour when an affected setting is reset to its default value.  Previously, the change would *not* have been synced (whereas all other changes are).  Here, such changes *are* synced.  The previous behaviour was inconsistent with the syncing behaviour of all other options changes.

Note:

- This isn't particularly well tested.  It's being committed mainly just for consideration of the approach, initially.

### DIFF
--- a/lib/settings.coffee
+++ b/lib/settings.coffee
@@ -6,6 +6,11 @@ Settings =
   onLoadedCallbacks: []
 
   init: ->
+    if Utils.isExtensionPage()
+      # On extension pages, we use localStorage (or a copy of it) as the cache.
+      @cache = if Utils.isBackgroundPage() then localStorage else extend {}, localStorage
+      @onLoaded()
+
     chrome.storage.local.get null, (localItems) =>
       localItems = {} if chrome.runtime.lastError
       @storage.get null, (syncedItems) =>
--- a/lib/settings.coffee
+++ b/lib/settings.coffee
@@ -6,19 +6,16 @@ Settings =
   onLoadedCallbacks: []
 
   init: ->
-    if Utils.isExtensionPage()
-      # On extension pages, we use localStorage (or a copy of it) as the cache.
-      @cache = if Utils.isBackgroundPage() then localStorage else extend {}, localStorage
-      @onLoaded()
+    chrome.storage.local.get null, (localItems) =>
+      localItems = {} if chrome.runtime.lastError
+      @storage.get null, (syncedItems) =>
+        unless chrome.runtime.lastError
+          @handleUpdateFromChromeStorage key, value for own key, value of extend localItems, syncedItems
 
-    @storage.get null, (items) =>
-      unless chrome.runtime.lastError
-        @handleUpdateFromChromeStorage key, value for own key, value of items
+        chrome.storage.onChanged.addListener (changes, area) =>
+          @propagateChangesFromChromeStorage changes if area == "sync"
 
-      chrome.storage.onChanged.addListener (changes, area) =>
-        @propagateChangesFromChromeStorage changes if area == "sync"
-
-      @onLoaded()
+        @onLoaded()
 
   # Called after @cache has been initialized.  On extension pages, this will be called twice, but that does
   # not matter because it's idempotent.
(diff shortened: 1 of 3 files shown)


mrmr1993:
> Other than type changes, have we ever changed a default value?

I don't think we have, but I'm [advocating for doing it](https://github.com/philc/vimium/issues/1719#issuecomment-110032836) as a way of solving the issues in #1719 without killing the custom CSS feature.

> Previously, settings with default values were not stored; here, they are.

I'd prefer the alternative of only 'trusting' a single storage area (eg. by using my suggestion [here](https://github.com/philc/vimium/pull/1733#issuecomment-112446595), or PR #1733 itself), being careful to clear settings out of `localStorage` when they are not present in the storage area.

smblott-github:
> but I'm advocating for doing it as a way of solving the issues in #1719

I don't think there's anything here which precludes that.

> I'd prefer the alternative of only 'trusting' a single storage area

Here's how I would think about it.  Each settings value is either in `chrome.storage.sync` or in `chrome.storage.local` (**never both**).  Let's call that the primary storage area for the setting.  Additionally, values are cached in `localStorage`.  Cached values are **always identical** to their primary storage area counterparts.

That's not so complicated, and it allows us to get rid of a knot of ugly logic in `Settings.set()` and friends. (`handleUpdateFromChromeStorage` becomes trivial.)

(We've both tried several ideas to solve this problem.  Getting it just right is really ugly  -- lots of logic and shaky assumptions.  If we relax the constraint about storing default values (which we never really needed), then suddenly all of that complexity goes away.  And the existing `Settings` logic becomes simpler too.  Win-win.  I've wanted to do this for a while, just never had a strong reason for the change until now.)

I think we're down to this (or something like it) or #1733 without the version bump (and with an unresolved question as to whether we record the migration in `localStorage` or `chrome.storage.sync`).

smblott-github:
Update:
- Added comment explaining data model/assumptions.
- Added debug option with logging.
- Tested with various settings states (it works).

This looks like the best option for solving this problem.  The core settings logic is simpler, which is a win in itself.  As a complexity/space trade off (in some cases we're storing more stuff in synced storage), it's a win too.

Will merge.

mrmr1993:
Simpler still (and I think preferable) is

* have disablable sync
* either only use `local` or only use `sync`
  - default to sync
  - if there are unsynced settings of old, force local

I don't like the default values decision, but I guess I could live with it.

There is also an issue with syncing which wasn't considered at its creation:
* chrome staggers extension updates for its server load, so one instance may migrate to new settings but another may still want and use the old settings for a while.
  - my suggestion can handle that easily by dropping back to local settings until the version catches up
  - this PR seems to make that much harder to handle

I'm going away for the weekend, so I probably won't be able to give any more input until Monday or later.

smblott-github:
That would be a change to the established experience.  Apart from your own development environment, @mrmr1993, what would be the reason for that change?

This PR fixes an existing bug.  It's not attempting to change anything.  Just fix a problem with #1706 which we didn't spot at the time.

smblott-github:
> There is also an issue with syncing which wasn't considered at its creation: .... chrome staggers extension updates for its server load

No.  This was considered.  There are two cases: propagating old-style to new-style, and new-style to old-style.  And for each of those, there's change-to-default and change-to-non-default.  If you work out the details, all four cases work as expected.  The only case which requires special consideration is a change-to-default from an old-style remote instance.  For this case, [this line](https://github.com/philc/vimium/pull/1739/files#diff-e3d64c4d1b80a956c1e86eeb7b27dc74R56) grabs the default value (that line wouldn't otherwise be needed).

mrmr1993:
With this merged, any user with
* 2 systems where one is more customised (eg. home) and the other is less customised (eg. work/uni/school)
* the customisations preceded sync, and where
* the new Vimium version is run on the less customised system first,

the user's more detailed settings will be discarded irretrievably. This is directly hurting some of the users that sync was supposed to help.

@smblott-github Can you back this out? Of all of the suggestions we considered, this is the most destructive.

I'm still advocating for my solution:
* it makes that we are syncing settings , through the option to disable them
* it doesn't destroy any legacy settings
* it works as sync was intended on all new installs
* users can disable it if they want to, eg.
  - me and anybody else testing settings
  - people who use a different, but overlapping (eg. QWERTY at work and AZERTY at home), keyboard layout on two different machines, who've tweaked their bindings to be key-location-consistent
  - anybody else who, for whatever reason, doesn't want their settings synced.

smblott-github:
> the user's more detailed settings will be discarded irretrievably. This is directly hurting some of the users that sync was supposed to help.

This PR takes the behaviour to exactly what it was in 1.51 (and 1.49, etc).  It changes nothing in terms of the user experience.  However, master is broken in light of #1705, and this is the cleanest fix for that.

There are no new features in this PR.  It's a bug fix.

If you want to discuss new features, open an issue.

mrmr1993:
>> There is also an issue with syncing which wasn't considered at its creation: .... chrome staggers extension updates for its server load

> No. This was considered.

Consider the change in `scrollStepSize` from being a string to being an integer. An update making this change would break scrolling on any not-yet-updated machine until it receives the update too. This clearly wasn't considered.

We can (and should) version the synced settings, so, from here on we can check to avoid this problem. My suggestion deals with this cleanly, by allowing us to push `localStorage` to `chrome.storage.local` and fall back to that easily. This PR does not.

mrmr1993:
> This PR takes the behaviour to exactly what it was in 1.51

No it doesn't; on multiple instances with unsynced settings, we change user settings without any user interaction.

smblott-github:
> No it doesn't; on multiple instances with unsynced settings, we change user settings without any user interaction.

I think you're mistaken.  There are no "changes".  Just data copied from one local storage area to another local storage area.  Nothing propagates outside of the local instance until the user hits "Save".  That's exactly the established 1.51, 1.49, etc. behaviour.  If you think that that established behaviour is incorrect, then that's a separate matter (one for a new issue).

> ... this change would break scrolling on any not-yet-updated machine until it receives the update too. This clearly wasn't considered.

That *cannot* happen.  The `scrollStepSize` migration is *always* been applied before the new migration, and *nothing* leaves the local instance as a result of this new migration.  So the case you describe *cannot* happen.

All this PR does is push unsynced, non-default settings to `chrome.storage.local`.  It's a local change only.  It does not affect remote instances.  This PR is correct and considerably simpler than the other options, and it fixes an extant bug in master (one which I've encountered several times, including just this morning, when looking at the find-mode stuff).

lib/settings.coffee
```diff
@@ -169,5 +182,19 @@ if Utils.isBackgroundPage()
...
+  # thereby making these settings accessible within content scripts.
+  do (migrationKey = "copyNonDefaultsToChromeStorage-20150717") ->
+    unless localStorage[migrationKey]
+      chrome.storage.sync.get null, (items) ->
+        unless chrome.runtime.lastError
+          updates = {}
+          for own key of localStorage
+            if Settings.shouldSyncKey(key) and not items[key]
+              updates[key] = localStorage[key]
+          chrome.storage.local.set updates, ->
```
mrmr1993: > Nothing propagates outside of the local instance until the user hits "Save".

This code propagates settings outside of the local instance without the user hitting "Save".

PR #1740, opened 2015-06-17
Initialise modes and install listeners at the same time.

### LINKED ISSUE #1738: Typing in textarea that has focus set by javascript still fires function keys
If I open a file that has a textarea that gets focus automatically, I think the expected behavior should be that you can type in that textarea, since the text entry cursor shows, and the field is indicated as having focus.

Keys that aren't mapped DO type into the textarea, but keys that are mapped to function will perform the function instead.

Some example HTML to demonstrate:
```html
<textarea id="focusme"></textarea>
<script>document.getElementById('focusme').focus();</script>
```

If you open a page with that code, then type, e.g., 'x', it will close the tab, rather than putting the 'x' where the text entry cursor is.

smblott-github:
Previously, we initialised modes early, but then installed our listeners only after DOM ready, and then only after we'd heard back from the background page in `checkIfEnabledForUrl`.  This creates a gap during which modes are installed, but they're not receiving events.

If an input is focused during that gap, then we don't pick it up when the modes are installed, and we don't pick it up when the input is focussed (because we're not listening).  This commit moves these two initialisation steps together, so they happen at the same time and there is no gap.

Fixes #1738.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -126,7 +126,6 @@ window.initializeModes = ->
 # Complete initialization work that sould be done prior to DOMReady.
 #
 initializePreDomReady = ->
-  initializeModes()
   checkIfEnabledForUrl()
   refreshCompletionKeys()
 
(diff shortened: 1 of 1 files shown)


mrmr1993:
LGTM.

smblott-github:
Thank you, @mrmr1993.

Issue #1742, opened 2015-06-18
Add support for ngClick to link hints


poacher2k:
There's quite a lot of Angular sites around these days, that use the ngClick directive instead of the native JS onclick.

Supporting ngClick to link_hints would allow Vimium to work on Angular sites, and is quite easy to implement.

I've added a suggestion that I've tested to work properly here: https://github.com/poacher2k/vimium/blob/patch-1/content_scripts/link_hints.coffee

smblott-github:
Thanks, @poacher2k.  @mrmr1993... What do you think?  This your realm.

mrmr1993:
Sounds reasonable and looks like an easy/clean way to do it.

@poacher2k could you submit a PR with that code? I have a few minor tweaks to talk over, and then it'd be great to get this in the codebase.

poacher2k:
Closed as of #1745.

PR #1745, opened 2015-06-19
Add support for ngClick to link hints 

poacher2k:
Elements using AngularJS ngClick directive added to link hints, ref #1742

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -169,7 +169,7 @@ class LinkHintsMode
 
     # Check for AngularJS listeners on the element.
     @checkForAngularJs ?= do ->
-      angularElements = document.body.getElementsByClassName "ng-scope"
+      angularElements = document.getElementsByClassName "ng-scope"
       if angularElements.length == 0
         -> false
       else
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -167,8 +167,7 @@ class LinkHintsMode
         element.getAttribute("aria-disabled")?.toLowerCase() in ["", "true"])
       return [] # This element should never have a link hint.
 
-    # The quite popular (http://w3techs.com/technologies/market/javascript_library/20) JavaScript framework AngularJS
-    # uses the proprietary click attribute "ng-click". This checks for every valid way it may occur.
+    # Check for AngularJS listeners on the element.
     ngPrefixes = ['', 'data-', 'x-']
     ngSeparators = ['-', ':', '_']
     hasNgClick = () ->
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -185,7 +185,7 @@ class LinkHintsMode
         element.getAttribute("role")?.toLowerCase() in ["button", "link"] or
         element.getAttribute("class")?.toLowerCase().indexOf("button") >= 0 or
         element.getAttribute("contentEditable")?.toLowerCase() in ["", "contentEditable", "true"]) or
-        hasNgClick 
+        hasNgClick()
       isClickable = true
 
     # Check for jsaction event listeners on the element.
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -167,25 +167,26 @@ class LinkHintsMode
         element.getAttribute("aria-disabled")?.toLowerCase() in ["", "true"])
       return [] # This element should never have a link hint.
 
-    # checks for every valid version of ng-click
+    # The quite popular (http://w3techs.com/technologies/market/javascript_library/20) JavaScript framework AngularJS
+    # uses the proprietary click attribute "ng-click". This checks for every valid way it may occur.
     ngPrefixes = ['', 'data-', 'x-']
     ngSeparators = ['-', ':', '_']
-    ng = 'ng'
-    click = 'click'
     hasNgClick = () ->
       for prefix in ngPrefixes
         for separator in ngSeparators
-          attr = prefix + ng + separator + click
-          if element.attributes.hasOwnProperty(attr)
+          attr = "#{prefix}ng#{separator}click"
+          if element.hasAttribute(attr)
             return true
       return false
+      
+    if hasNgClick()
+      isClickable = true
 
     # Check for attributes that make an element clickable regardless of its tagName.
     if (element.hasAttribute("onclick") or
         element.getAttribute("role")?.toLowerCase() in ["button", "link"] or
         element.getAttribute("class")?.toLowerCase().indexOf("button") >= 0 or
-        element.getAttribute("contentEditable")?.toLowerCase() in ["", "contentEditable", "true"]) or
-        hasNgClick()
+        element.getAttribute("contentEditable")?.toLowerCase() in ["", "contentEditable", "true"])
       isClickable = true
 
     # Check for jsaction event listeners on the element.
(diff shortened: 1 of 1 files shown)


content_scripts/link_hints.coffee
```diff
@@ -167,11 +167,25 @@ class LinkHintsMode
         element.getAttribute("aria-disabled")?.toLowerCase() in ["", "true"])
       return [] # This element should never have a link hint.
 
+    # checks for every valid version of ng-click
```
mrmr1993: This comment should mention something about Angular.js so we know *why* we're doing it, as well as what we're doing.

content_scripts/link_hints.coffee
```diff
@@ -167,11 +167,25 @@ class LinkHintsMode
...
+    ngPrefixes = ['', 'data-', 'x-']
+    ngSeparators = ['-', ':', '_']
+    ng = 'ng'
+    click = 'click'
+    hasNgClick = () ->
+      for prefix in ngPrefixes
+        for separator in ngSeparators
+          attr = prefix + ng + separator + click
```
mrmr1993: This might be clearer at a glance as

```coffee
attr = "#{prefix}ng#{separator}click"
```

content_scripts/link_hints.coffee
```diff
@@ -167,11 +167,25 @@ class LinkHintsMode
...
+          if element.attributes.hasOwnProperty(attr)
```
mrmr1993: Is there a reason for using this over `element.hasAttribute attr`?

content_scripts/link_hints.coffee
```diff
@@ -167,11 +167,25 @@ class LinkHintsMode
...
+            return true
+      return false
+
     # Check for attributes that make an element clickable regardless of its tagName.
     if (element.hasAttribute("onclick") or
         element.getAttribute("role")?.toLowerCase() in ["button", "link"] or
         element.getAttribute("class")?.toLowerCase().indexOf("button") >= 0 or
-        element.getAttribute("contentEditable")?.toLowerCase() in ["", "contentEditable", "true"])
+        element.getAttribute("contentEditable")?.toLowerCase() in ["", "contentEditable", "true"]) or
+        hasNgClick()
```
mrmr1993: This function call would be better up by its definition, so it's clear what this check is doing, and that it isn't a criterion for natively clickable elements. Something like `isClickable = true if hasNgClick()` above would do.

mrmr1993:
Thanks @poacher2k, this is great. Do you have an example page I can test this on?

content_scripts/link_hints.coffee
poacher2k: No, element.hasAttribute is way clearer.

poacher2k:
Glad to be of help! You can check this out at http://dutiz.com/ (the elements in the "job wall", Jeanette, Trym, etc.)

content_scripts/link_hints.coffee
poacher2k: Looks like proper CoffeeScript magic to me. I agree this looks clearer.

content_scripts/link_hints.coffee
poacher2k: Agreed!

mrmr1993:
Tested on the example page, it works beautifully!

poacher2k:
I should've checked CONTRIBUTING.md better! Still a bit new to contributing.

smblott-github:
Looks great, thanks @poacher2k.  Is it possible (without great complexity) to fix these two issues?

- On the example site (link immediately above) some of the elements move around immediately after load, leaving the hints in the wrong place.

- Similarly (without great complexity), can we detect elements which aren't active?  E.g., in the screenshot below, most of the hints on the left are not active...

snapshot

smblott-github:
Thinking about this a bit more.

- We're doing attribute checks against nine constant attribute names, but we're building the attribute names dynamically for *every* DOM element.  We should use a `do ->` block to build the attribute names just once.

- (The answer is probably "no", but...)  This test is pretty expensive.  Is there a way to detect whether AngularJS is active on the page in the first place, and only do these this test if it is?

- (Marginal...)  Would it better to move this further down (so that we don't do the test for clickable items we've already identified)?  Marginal.

- See here for minor tweaks: 9475c51932fc3331e515886b0495c5b86a1a9e65.

poacher2k:
> On the example site (link immediately above) some of the elements move around immediately after load, leaving the hints in the wrong place.

I noticed this as well, and I assume this is will happen whenever elements are animated after/while link hints are activated. Maybe we could add event listeners and do some smart checks, maybe even just add event listeners to elements that are likely to change.

>Similarly (without great complexity), can we detect elements which aren't active?

Maybe we can use `document.elementFromPoint` for this (see [SO question](http://stackoverflow.com/a/905341/2908761)). I'll try it out.

> This test is pretty expensive. Is there a way to detect whether AngularJS is active on the page in the first place, and only do these this test if it is?

Good idea! We can check for `'angular' in window` or `window.hasOwnProperty 'angular'`.

> Would it better to move this further down (so that we don't do the test for clickable items we've already identified)?

For sure!

poacher2k:
Sorry to ask, but how do I merge your tweaks (https://github.com/philc/vimium/commit/9475c51932fc3331e515886b0495c5b86a1a9e65) into the pull request?

smblott-github:
They're tiny.  Just copy them.

(More generally, add my repo as a remote then `git cherry-pick`.)

- https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes#Adding-Remote-Repositories
- http://git-scm.com/docs/git-cherry-pick

mrmr1993:
> Good idea! We can check for `'angular' in window` or `window.hasOwnProperty 'angular'`.

@poacher2k unfortunately this won't work, since we're in an isolated world, so we can't query `window`.

> Is there a way to detect whether AngularJS is active on the page in the first place, and only do these this test if it is?

@smblott-github Any efficient way is going to use a trick like #1405 (alternatively using `chrome.tabs.update` with a bookmarklet to run the code, or a custom event to commuicate with the content script, or both). If you didn't like #1405, you're probably not going to like those solutions either.

smblott-github:
Thanks @poacher2k, @mrmr1993.  We should keep it simple, then.

poacher2k:
> @poacher2k unfortunately this won't work, since we're in an isolated world, so we can't query window.
Ah! Was not aware of this. In that case, we could check for the existence of any element with the class `ng-scope`. AngularJS adds this to _at least_ one element.

smblott-github:
> In that case, we could check for the existence of any element with the class ng-scope...

That sounds like a good idea.

gdh1995:
Or check the `<script[src~=angular]>`?

poacher2k:
Good idea, but this wouldn't catch cases where scripts are concatenated and
renamed though
24. juni 2015 19:09 skrev "Dahan Gong" <notifications@github.com>:

smblott-github:
Perhaps re-worked like this... 2cc7dc1d2164b5dbb27bcc1d4bc0517402dd7581?

poacher2k:
Should be ready to merge now!

> On the example site (link immediately above) some of the elements move around immediately after load, leaving the hints in the wrong place.

and

> Similarly (without great complexity), can we detect elements which aren't active? E.g., in the screenshot below, most of the hints on the left are not active...

Are challenges not specific to this pull request, but definitely needs a solution.

smblott-github:
Thanks, @poacher2k.

poacher2k:
@smblott-github is there an issue created for `Similarly (without great complexity), can we detect elements which aren't active? E.g., in the screenshot below, most of the hints on the left are not active...` ? Couldn't find one when searching issues for "active" or "visible".

smblott-github:
@poacher2k. I'm not aware of an existing issue.

PR #1747, opened 2015-06-21
fix a bug that `pasteFromClipboard` does not return text from clipboard

gdh1995:
In `main.coffee`, the `pasteFromClipboard` handler calls `Clipboard.paste` but returns `null`, which will be received by a `callback` in `mode_visual_edit.coffee:86`


smblott-github:
Thanks, @gdh1995.

PR #1751, opened 2015-06-25
Set the parent tab when creating tabs.

smblott-github:
Trivial change, but UX win.

When creating new tabs, set the parent tab.  If the tab is subsequently deleted, then we return to the original tab (as opposed to the one on its right).


smblott-github:
(See also d9e8da958d55de5c8880b336255cf8987986be93, which makes the behaviour with marks consistent with the behaviour here.)

PR #1755, opened 2015-06-29
Node 0.12.x now has built in execSync

lantrix:
Remove need for no longer supported exec-sync
Now a built in feature of node.js 0.12.x `child_process`

### DIFF
--- a/Cakefile
+++ b/Cakefile
@@ -56,15 +56,12 @@ task "autobuild", "continually rebuild coffeescript files using coffee --watch",
   coffee = spawn "coffee", ["-cw", __dirname]
 
 task "package", "Builds a zip file for submission to the Chrome store. The output is in dist/", ->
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @lantrix.  Let's do this in August, six months after 0.12.0 was released.

(Is that the right date?)

We would probably also need a note [here](https://github.com/philc/vimium/blob/master/CONTRIBUTING.md#installing-from-source).

lantrix:
Good idea. Yes 7th August is 6 months after 0.12.0 was released
https://github.com/joyent/node/releases/tag/v0.12.0

mrmr1993:
`exec-sync` was removed by #1764 and replaced with `spawnSync` (the double of `spawn`, which we were already using). May want to consider closing this PR.

Issue #494, opened 2012-03-02
Use a link-hints-like UI to select a frame to focus


icx-jonv:
The biggest problem I have with Vimium is Chrome's insistence on picking the incorrect frame to place focus.  I really like the tags used for links, would it be possible to add tags for frames?  For instance, I press gf (or something else) and tags show up on all of the frames, making it easy to select one that will gain focus.

boblehest:
Example website where this is an issue: http://doc.rust-lang.org/nightly/book/
Would be great if I could focus a frame like this: http://i.imgur.com/x4lenuH.png

smblott-github:
Closing.
Old, and discussed elsewhere,

Issue #602, opened 2012-07-27
Show linkhints for content in iframes


philc:
The keypoint website is a good example -- the login form is an iframe, and linkhints do not show for the login boxes.

[https://www.keypointcu.com](https://www.keypointcu.com/)

jaekyeom:
Is there any update on this?

mrmr1993:
I'm working on it (albeit slowly, I'm quite busy at work as of late). I hope to get properly moving with this when Steve is back.

If it's of any interest, my plan is:
* make `LinkHints` code more friendly and orthogonal (so we can call parts of it in frames without needing the whole mode)
  - #1761 is the start of this
* add the ability to update hints when presented with new links (so hints can be shown incrementally as frames report them)
  - we need to ensure that, when we update the hints, to add links, existing hints will only change by having a suffix added (ie. `a` -> `ab` or `aab`, not `a` -> `bb`).
* move the mode to the top frame only, and have other frames enter a 'helper' mode
  - communicating between frames using a `CustomEvent` seems like the easiest, least-brittle way to get only the hints we want (ie. only query visible frames).
  - rendering the hints in the top frame keeps the UI in step and helps with consistent positioning, so I plan to do this too.

All of these are relatively simple, so long as everyone agrees on this implementation. If I get some free time at the weekend I'll try to write and push a full working version.

tsim0:
Any update on this? Google reCAPTCHA doesn't work because of this too.

smblott-github:
It should be fixed in the next release (see #2048), @tsim0.  If anybody'd be willing to take that for a spin (it's in `master`) and provide feedback, then that'd be appreciated.

Closing.

Issue #425, opened 2011-12-21
Make it possible to focus a scrollable div without using the mouse


vbaruh:
Take a look, my expectation is that the scrolling should work on the main frame but it does not affect anything.
Opening links functionality work fine.

marlun:
I often need to click no the part of the page that I want to scroll before being able to use the keyboard. Is it still planned to make it possible using the keyboard?

smblott-github:
Ah.  The "Vimium click", @marlun.

Often, the `gF` (focus the main frame) command helps.

marlun:
Unfortunately that almost never works for me because it's never multiple frames. Take for example http://redux.js.org/docs/introduction/ThreePrinciples.html if I load it, move to another section using vimium F I can no longer scroll the main text div.

smblott-github:
[This branch](https://github.com/smblott-github/vimium/tree/global-link-hints-extended) is an extension of #2048 which (amongst other things) makes scrollable divs selectable.  For example, with filtered hints, it looks like this...

![snapshot](https://cloud.githubusercontent.com/assets/2641335/14063964/dbf6a612-f3e4-11e5-9c28-45d9e98e5853.png)

tamvm:
> [This branch](https://github.com/smblott-github/vimium/tree/global-link-hints-extended) is an extension of #2048 which (amongst other things) makes scrollable divs selectable. For example, with filtered hints, it looks like this...
> 
> ![snapshot](https://cloud.githubusercontent.com/assets/2641335/14063964/dbf6a612-f3e4-11e5-9c28-45d9e98e5853.png)

Hi @smblott-github, thank you for your contribution in this project for the community. 

The idea about choosing scrollable div is quite nice, I just want to ask about the implementation of that, how to config as your image?

Thank you,

PR #1761, opened 2015-07-01
Switch frame with link hints and refactor link hint modes

mrmr1993:
This PR
* refactors the code for modes in `LinkHints` to
  - allow diffferent `linkActivators` and `getVisibleClickable` implementations for each mode
  - make it clearer which mode does what
  - make new mode definitions orthoganal to the definition of `LinkHintsMode`
* fixes bugs where `COPY_LINK_URL` and `OPEN_INCOGNITO` focus the `contentEditable` element containing a link instead of working on the link itself
* adds a new command (`activateModeToChangeFrame`) that selects frames using link hints
  - this is not yet exposed as a mappable command
  - when cross-frame link hints are implemented, this can be exposed to fix issue #494

This is the first part of an effort towards cross-frame link hints (#602), and also allows for useful possibilities such as #425. Each commit here is atomic, and so the commits may be easier to check than the PR diff.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -29,9 +29,18 @@ OPEN_WITH_QUEUE =
 COPY_LINK_URL =
   name: "link"
   indicator: "Copy link URL to Clipboard."
+  linkActivator: (link) -> # `this` will be an instance of LinkHintsMode at call-time.
+    if link.href?
+      chrome.runtime.sendMessage handler: "copyToClipboard", data: link.href
+      url = link.href
+      url = url[0..25] + "...." if 28 < url.length
+      @onExit = -> HUD.showForDuration "Yanked #{url}", 2000
+    else
+      @onExit = -> HUD.showForDuration "No link to yank.", 2000
 OPEN_INCOGNITO =
   name: "incognito"
   indicator: "Open link in incognito window."
+  linkActivator: (link) -> chrome.runtime.sendMessage handler: 'openUrlInIncognito', url: link.href
 DOWNLOAD_LINK_URL =
   name: "download"
   indicator: "Download link URL."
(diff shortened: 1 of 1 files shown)


gdh1995:
Should we mark `<iframe>`s  when `f` is pressed, instead of inventing a new command to find frames specially? Just say I open an unfamiliar page and press `f`, then I might see nothing but a isolated HUD, but there's another kind of behavior:
1. Vimium shows a mark on the top left - like `[a]`
2. I press `a` and the iframe window gets focus
3. I press `f` again, now many marks will be shown properly.

ekucks21:
+1 for getting this in

smblott-github:
Switch frame with link hints sounds good, @mrmr1993 (if we can do it without a new command).

I'm not so keen on the refactoring.

mrmr1993:
> I'm not so keen on the refactoring.

The main idea was to make each link hint mode's version of `getVisibleClickable` easily accessible (ie. without having to instantiate the full mode in every frame), so that we can call the correct one easily when doing cross-frame link hints.

Does this make the refactor any more acceptable/is there an alternative way that you'd like to see this done? Or should I just bin this?

smblott-github:
Do we really need/want a new command for this?

I tried out a poor-man's version of the main idea here: 013965007858f1f80477a422c149b2b1695d2a2c.

It turns out it's not as useful as I'd expected, the problem being that - once you're in a frame - you cannot use link hints to get out of it.  That problem would be solved if we had global hints, but *that* feature would be extremely complex.

A simpler alternative might be:

- as currently, include all clickable things in the current frame,
- however, also include a single hint for each other frame which is in the current viewport.

Bottom line... I'm not sure we should do this in its current form.

(Entirely separately, it would be good if `gf` didn't select frames which are not in the viewport.)

PR #1763, opened 2015-07-08
now you can setup list of css classes which will be treated as clickable elements 

ygun:
useful for Vaadin apps

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -43,5 +43,6 @@ Contributors:
   Timo Sand <timo.j.sand@gmail.com> (github: deiga)
   Shiyong Chen <billbill290@gmail.com> (github: UncleBill)
   Utkarsh Upadhyay <musically.ut@gmail.com) (github: musically-ut)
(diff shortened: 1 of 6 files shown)


gdh1995:
It's useful, but I think, usually one class name is just for one website, or even one page - this feature seems too heavy and slow than what we needed.

mrmr1993:
It doesn't feel reasonable to me to expect users to identify the classes that individual pages use to denote a clickable element when we're not matching them. If I can find a way to re-implement #1405, then these cases should nearly all be handled anyway.

Since we support the ARIA `role` attribute (which sites should be using anyway for accessibility), you could create a tampermonkey script (or an extension, etc.) to mark the clickable elements with `role="link"`, and then they'll work as intended. This avoids the need to add a confusing option *and* would let you distribute your fixes easily if you want to..

ygun:
Thanks for tampermonkey solution. Anyway this patch was a really quick fix. Your explanation sounds reasonable for me.

PR #1764, opened 2015-07-13
Use spawn instead of exec-sync for cake package

mrmr1993:
This PR changes `cake package` to use `spawn` rather than `exec-sync`, reducing its dependencies to be the same as `cake build`. To execute a sequence of commands, the `exit` event of each process is used to call the next, so they run asynchronously but in order.

Full disclosure:

`exec-sync` fails to build on one of my systems, so I wrote this to make `cake package` work for me. PR'd because it's nice to be able to package everywhere that we can build.


philc:
Very nice, thank you @mrmr1993

PR #1772, opened 2015-07-20
Move the caret to end in `focusInput`

### LINKED ISSUE #1765: Text box focusing
The <gi> function works actually really well, and I appreciate how much time it saves. However, whenever focusing action happens, the cursor always appears in the very front of text box, and it quite often requires pressing END additionally for normal editing. So, it there any possibility to change this behavior?

gdh1995:
This fixes #1765 , using `DomUtils.simulateSelect`.

We have been using it in `LinkHints.activateLink`.


mrmr1993:
LGTM

PR #1775, opened 2015-07-26
Normal mode key handling in frontend as a generic inheritable class

### LINKED ISSUE #1545: main.coffee: generateCompletionKeys treats keysToCheck mistakenly
main.coffee:
507: `generateCompletionKeys = (keysToCheck) ->`
508: >>`splitHash = splitKeyQueue(keysToCheck || keyQueue)`

Then, if `keysToCheck` is `""`, `keysToCheck || keyQueue` will be `keyQueue`, which makes

578: `completionKeys: generateCompletionKeys("")`

build `completionKeys` with `keyQueue`.

However, when sending "executePageCommand" messages, `keyQueue` can be a non-empty string, if the current is a two-key command. So I think `generateCompletionKeys` works mistakenly.

mrmr1993:
This PR
* creates a new, generic class `MappingMode` for handling mappable commands and numeric prefixes
* moves the bulk of (easily re-usable) normal mode key handling mode out of `vimium_frontend.coffee` and into this class
* moves the key checking code from the background page into the the class
  - this resolves some race conditions
  - fixes an passkey issue we were testing for, but where Shoulda.js' lack of asynchronous testing made the test pass incorrectly (da419de and 98e7095 for info)
  - we're not messaging the background as much
* gives each `NormalMode` its own key queue
  - switching tab/window with the mouse/a window manager command could cause weirdness before (eg. `g` *switch tab* `y` *switch back* `y` would run `yy` and pass the second `y` to the page)
* uses a single implementation to do all normal mode key handling logic (ie. instead of a frontend one and a more complex background one)
  - because of this, `0` will no longer be leaked to the page in number prefixes
* gives the handler which hides the help dialog its own function, since it's not a part of the mode's main logic.

Consequences of this:
* gives us support for mappings of more than 2 keys
* fixes #1545 
* modularisation

This is also a good starting point for #1188, #1469 (and in general removing a lot of duplicate key handling code from visual etc. modes) and several other issues.

Every commit in this PR should compile, pass all tests (apart from da419de), and work correctly. Each commit is a single simple step, so it should be possible to review this commit by commit or in blocks of commits too.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -499,13 +499,6 @@ root.refreshCompletionKeysAfterMappingSave = ->
 
   sendRequestToAllTabs(getCompletionKeysRequest())
 
-splitKeyQueue = (queue) ->
-  match = /([1-9][0-9]*)?(.*)/.exec(queue)
-  count = parseInt(match[1], 10)
-  command = match[2]
-
-  { count: count, command: command }
-
 handleKeyDown = (request, port) ->
   {keyChar: key, keyQueue: queue} = request
   if queue?
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -25,7 +25,6 @@ tabInfoMap = {} # tabId -> object with various tab properties
 # Queue of keys typed. If keyQueueArray.numericPrefix is true, its 0th entry is the current command's numeric
 # prefix.
 keyQueueArray = []
-validFirstKeys = {}
 commandKeys = []
 focusedFrame = null
 frameIdsForTab = {}
@@ -468,19 +467,12 @@ splitByKeys = (key) ->
       key = key[1..]
   returnArray
 
-populateValidFirstKeys = ->
-  validFirstKeys = {}
-  for keys in commandKeys
-    if (keys.length > 1)
-      validFirstKeys[keys[0]] = true
-
 populateCommandKeys = ->
   commandKeys = (splitByKeys key for key of Commands.keyToCommandRegistry)
 
 # Invoked by options.coffee.
 root.refreshCompletionKeysAfterMappingSave = ->
   populateCommandKeys()
-  populateValidFirstKeys()
 
   sendRequestToAllTabs(getCompletionKeysRequest())
 
@@ -695,7 +687,6 @@ if Settings.has("keyMappings")
   Commands.parseCustomKeyMappings(Settings.get("keyMappings"))
 
 populateCommandKeys()
-populateValidFirstKeys()
 
 # Show notification on upgrade.
 showUpgradeMessage = ->
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -470,16 +470,19 @@ getActualKeyStrokeLength = (key) ->
     key.length
 
 populateValidFirstKeys = ->
+  validFirstKeys = {}
   for key of Commands.keyToCommandRegistry
     if (getActualKeyStrokeLength(key) == 2)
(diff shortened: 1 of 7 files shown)


background_scripts/main.coffee
```diff
@@ -455,138 +452,58 @@ updatePositionsAndWindowsForAllTabsInWindow = (windowId) ->
...
+splitByKeys = (key) ->
+  returnArray = []
+  while key
+    if (key.search(namedKeyRegex) == 0)
+      returnArray.push RegExp.$1
+      key = RegExp.$2
+    else
+      returnArray.push key[0]
+      key = key[1..]
+  returnArray
```
gdh1995: We may rework `namedKeyRegex` into:
``` Regex
/<(?:(?:a-(?:c-)?(?:m-)?|c-(?:m-)?|m-)([A-Z][0-9A-Z]+|[a-z][0-9a-z]+|[^\s])|[A-Z][0-9A-Z]+|[a-z][0-9a-z]+)>|[^\s]/g
```

Then we can use `key.match(namedKeyRegex)` to get the splitted array easily.

PS: If we have to considering that `"<space>"` is changed into `" "`, then its end `|[^\s]/g` should be `|./g`.

mrmr1993: That looks like a valid change, but it's not something I want to include in this PR. You might want to open an issue or create a PR implementing this.

mrmr1993:
I've pushed a [branch](https://github.com/mrmr1993/vimium/tree/normal-mode-mappings-in-frontend+visual-mode-mapping-rework) which reuses the `MappingMode` class methods to provide visual mode mappings too. (At the moment the `Movement` class doesn't inherit from `MappingMode`, since I didn't want to touch `SuppressPrintable` for now.)

@smblott-github could you take a look at this? I'd love to get it in so I can start building upon it.

mrmr1993:
Rebased onto master.

mrmr1993:
@smblott-github @philc is this something that you would consider merging? I'd be really pleased to see this get into the codebase, but I'd rather close the PR and stop trying to keep it up to date with master if it's going to be dead in the water.

mrmr1993:
Ping @smblott-github. Anything on this?

smblott-github:
@mrmr1993.  The status quo isn't what we'd choose if we were starting from scratch -- and what you describe is probably a better approach.  However, this is a big-footprint PR (so quite a bit of time/effort to review, and it may well bring with it its own bugs). And, while it probably does fix some bugs (e.g. the race conditions), those aren't biting us too badly at the moment.

philc:
Epic. Will review soon.

smblott-github:
This is a massive PR, @mrmr1993.

It would take a week to review properly!

mrmr1993:
Every commit is completely atomic and non-breaking, so these can be reviewed in batches of eg. 1, 2, 5, 10 commits to make it easier. This PR is more a destination than a change to go all at once.

I'll rebase in a bit.

smblott-github:
> I'll rebase in a bit.

That doesn't fill me with enthusiasm....

(Big or complicated PR's are less likely to receive my attention. Sorry, but that's just how it is.)

mrmr1993:
> Big or complicated PR's are less likely to receive my attention.

Anything I could do to help this along? I can submit PRs of smaller chunks of changes that logically go together, if that helps.

mrmr1993:
I've rebased this onto master in [this branch](https://github.com/mrmr1993/vimium/tree/rebase-normal-mode-mappings-in-frontend).

smblott-github:
Closing.  #2022 does this.  And #2022 loses more than 200 LoC, whereas this adds code.

PR #1779, opened 2015-07-28
use keyCode only to distinguish a key event in KeydownEvents

gdh1995:
There's no need to distinguish a key using `keyIdentifier`, `ctrlKey` and so on, since:
* Whenever we press a key and release it, the `keyCode` stays the same.
* Most different keys on one keyboard has different `keyCode`.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -461,26 +461,20 @@ extend window,
 # Track which keydown events we have handled, so that we can subsequently suppress the corresponding keyup
 # event.
 KeydownEvents =
-  handledEvents: {}
-
-  stringify: (event) ->
-    JSON.stringify
-      metaKey: event.metaKey
-      altKey: event.altKey
-      ctrlKey: event.ctrlKey
-      keyIdentifier: event.keyIdentifier
-      keyCode: event.keyCode
+  handledEvents: new Uint8Array(256)
 
   push: (event) ->
-    @handledEvents[@stringify event] = true
+    @handledEvents[event.keyCode] = 1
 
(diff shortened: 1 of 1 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -461,26 +461,20 @@ extend window,
...
-  handledEvents: {}
-
-  stringify: (event) ->
-    JSON.stringify
-      metaKey: event.metaKey
-      altKey: event.altKey
-      ctrlKey: event.ctrlKey
-      keyIdentifier: event.keyIdentifier
-      keyCode: event.keyCode
+  handledEvents: new Uint8Array(256)
```
mrmr1993: Why use this rather than a normal object? I don't think we're massively concerned about the speed of this function, and this makes it brittle to future changes.

gdh1995: Yes, we may keep using `{}`

mrmr1993:
This should also fix the bug where the following sequence leaks an `a` keyup (if `<c-a>` is mapped):

`<ctrl>` keydown, `<c-a>` keydown, `<ctrl>` keyup, `a` keyup

The idea of this looks good to me, but I can't see why this PR couldn't be just changing `stringify` to

```coffee
stringify: (event) -> event.keyCode
```

gdh1995:
> `(event) -> event.keyCode`
That's just what I want. Other code is just for simplifying more.

Besides, I find another bug about KeydownEvents: If we press a key mapped `nextTab`, then `push` is called when key is down, but this page may not receive any keyup event, and `pop` won't be called.

To solve this, we need to clean `handledEvents` on blurring window.

So, if I need to start a new PR?

gdh1995:
Well, since there's #1780, this pr is not so useful. Then, closed.

PR #1780, opened 2015-07-28
fix bugs in KeydownEvents so that correct keyup events will be handled

gdh1995:
In this PR, `KeydownEvents` uses only `keyCode` to match events, and will clean left keydown records when `window` loses its focus (e.g. another tab becomes selected).

 Then,
> the following sequence: `<ctrl>` keydown, `<c-a>` keydown, `<ctrl>` keyup, `a` keyup

will be handled correctly, and no irrelevant keyup event will be matched.

This is better than PR #1779 , since `handledEvents` is still `{}`.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -463,13 +463,7 @@ extend window,
 KeydownEvents =
   handledEvents: {}
 
-  stringify: (event) ->
-    JSON.stringify
-      metaKey: event.metaKey
-      altKey: event.altKey
-      ctrlKey: event.ctrlKey
-      keyIdentifier: event.keyIdentifier
-      keyCode: event.keyCode
+  stringify: (event) -> event.keyCode
 
   push: (event) ->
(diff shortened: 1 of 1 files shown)


mrmr1993:
Great, code LGTM.

content_scripts/vimium_frontend.coffee
```diff
@@ -463,13 +463,7 @@ extend window,
...
   handledEvents: {}
 
-  stringify: (event) ->
-    JSON.stringify
-      metaKey: event.metaKey
-      altKey: event.altKey
-      ctrlKey: event.ctrlKey
-      keyIdentifier: event.keyIdentifier
-      keyCode: event.keyCode
+  stringify: (event) -> event.keyCode
```
smblott-github: @gdh1995.
Looks like the name `stringify` is no longer appropriate here, right?

smblott-github: Actually, @gdh1995.  Don't worry about it.
I'll just merge then change it.

Issue #1796, opened 2015-08-26
Vimium may work well on XML view page


gdh1995:
We know that on a XML page, `document.createElement("div")` gets a `<div>` with `null` style, which makes Vomnibar and LinkHints do not work. But, today I find a simple way to create normal DOM nodes and make them work well successfully.

What we should do is just to use `document.createElementNS("http://www.w3.org/1999/xhtml", tagName)`. A node created by `createElementNS` has a normal "style", and its children created by setting "innerHTML" also have good "style"s, and shadowDOMs do no harm to this.

![2](https://cloud.githubusercontent.com/assets/5547703/9486653/a586d86c-4bfc-11e5-9712-27f05a4f01a2.png)

gdh1995:
Besides, there's another way to get a node with normal "style".
``` javascript
var html = document.importNode(document.documentElement); // get a "<html></html>"
html.innerHTML = '<div></div>';
// now, `html` and `html.firstElementChild` both have normal "style"s
```

mrmr1993:
@gdh1995 Good stuff, this works very nicely! I've pushed a PR implementing your idea from here; hope that's okay.

If you'd rather implement it yourself, let me know and I'll close the PR :)

gdh1995:
Besides, there's another problem (or BUG ?) about DOM nodes having `null` style:
On XML view page and XHTML page, whose content types both contain "xml", a normal `div` with a normal `.style` attribute will create children with `null` styles if:
* it belongs to a shadow DOM
* the shadow DOM has been appended to `document` DOM tree.
* we make new nodes by setting `div`'s `innerHTML`.

This problem seems to have no impact on the master version, so this is just a warning.

gdh1995:
Closed, because #1798 has been merged.

PR #1797, opened 2015-08-26
Fix Vimium on XML pages

mrmr1993:
This implements @gdh1995's idea from #1796, using `document.createElementNS` for XML documents and remove the now redundant XML-specific codepaths.

Tested as working correctly [here](http://qz.com/feed/), the example from #1640.

### DIFF
--- a/lib/clipboard.coffee
+++ b/lib/clipboard.coffee
@@ -1,6 +1,6 @@
 Clipboard =
   _createTextArea: ->
-    textArea = document.createElement("textarea")
+    textArea = DomUtils.createElement("textarea")
     textArea.style.position = "absolute"
     textArea.style.left = "-100%"
     textArea
--- a/content_scripts/hud.coffee
+++ b/content_scripts/hud.coffee
@@ -96,7 +96,7 @@ class Tween
   styleElement: null
 
   constructor: (@cssSelector, insertionPoint = document.documentElement) ->
-    @styleElement = document.createElement "style"
+    @styleElement = DomUtils.createElement "style"
 
     unless @styleElement.style
       # We're in an XML document, so we shouldn't inject any elements. See the comment in UIComponent.
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -127,7 +127,7 @@ class LinkHintsMode
   # Creates a link marker for the given link.
   #
   createMarkerFor: (link) ->
-    marker = document.createElement("div")
+    marker = DomUtils.createElement("div")
     marker.className = "vimiumReset internalVimiumHintMarker vimiumHintMarker"
     marker.clickableItem = link.element
 
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -6,28 +6,17 @@ class UIComponent
   shadowDOM: null
 
   constructor: (iframeUrl, className, @handleMessage) ->
-    styleSheet = document.createElement "style"
-
-    unless styleSheet.style
-      # If this is an XML document, nothing we do here works:
-      # * <style> elements show their contents inline,
-      # * <iframe> elements don't load any content,
-      # * document.createElement generates elements that have style == null and ignore CSS.
-      # If this is the case we don't want to pollute the DOM to no or negative effect.  So we bail
-      # immediately, and disable all externally-called methods.
-      @postMessage = @activate = @show = @hide = ->
-        console.log "This vimium feature is disabled because it is incompatible with this page."
-      return
+    styleSheet = DomUtils.createElement "style"
 
     styleSheet.type = "text/css"
     # Default to everything hidden while the stylesheet loads.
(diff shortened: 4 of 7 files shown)


gdh1995:
Yes, this is just what I want. By the way, for speed, I wrote like this:
``` javascript
DomUtils.createElement = function(tagName) {
  var node = document.createElement(tagName);
  this.createElement = node.style
    ? document.createElement.bind(document)
    : document.createElementNS.bind(document, "http://www.w3.org/1999/xhtml");
  return node;
};
```

smblott-github:
@mrmr1993.  Could you go back and check, please, whether the original issues from #1640 still exists?

I can't seem to reproduce it now (44.0.2403.155).

smblott-github:
... In particular, on the test page, ....

snapshot

gdh1995:
@smblott-github Now Vimium checks `.style == null` and does not inject DOM nodes to XML pages.

On Chrome 43, a `<style>` with `null` style will show its text content just like a `<div>`, but one with normal ".style" works well.
I'm going to test Chrome 44 now.

smblott-github:
> Now Vimium checks .style == null and does not inject DOM nodes to XML pages.

No, @gdh1995.  I can't reproduce the problem using Vimium 1.51.

lib/clipboard.coffee
```diff
@@ -1,6 +1,6 @@
 Clipboard =
   _createTextArea: ->
-    textArea = document.createElement("textarea")
+    textArea = DomUtils.createElement("textarea")
```
smblott-github: Moving all of the `createElement` calls to DomUtils seems like a good idea.

However, perhaps not this one (which is only used from the background page).  It seems to break yank from visual mode:

- `v`, select some text, the `y` seems not to be working.

gdh1995:
@smblott-github I've tested Chrome 44 (32bits), nothing is better.

The master version of Vimium will not show any UI components for a XML page, because:
``` coffeescript
# content_scripts/ui_component.coffee:  class UIComponent: constructor: line 11
unless styleSheet.style
  @postMessage = @activate = @show = @hide = ->
    console.log "This vimium feature is disabled because it is incompatible with this page."
  return
```

This part prevents Vimium from injecting CSS code;

Reproduction:
* open the developer tools and create a new `<style`> using `document.createElement`
* set its `innerHTML` and call `document.documentElement.appendChild(node)`
* now the `<style`>'s textContent is shown at the bottom of the page
* NOTE: if wrapping a `<style>` in a `<div>`, and append the latter to `documentElement`, no text will show since height of `<div>` is 0.

smblott-github:
@gdh1995. Nope.

> now the <style>'s textContent is shown at the bottom of the page

That's not what I see.  It works fine (44.0.2403.157).  The text is *not* shown in the page.

In fact, using Vimium 1.51, I have no problem using the Vomnibar or find mode on XML pages.  It seems something in Chrome has changed.

(I think we should probably fix the issue regarding yank, above, and merge this anyway.)

gdh1995:
@smblott-github Are you testing http://qz.com/feed/ ? I find the page's response header has no "Content Type" and Chrome treats it with a "text/plain" type:

When I test a xml file on my computer (Windows 8.1 x64), it breaks again. The content type is:
``` ini
[HKEY_CLASSES_ROOT\.xml]
"Content Type"="text/xml"
```

smblott-github:
OK.  Thanks @gdh1995.  That sorted it.  I see the correct error now -- using http://smblott.org/test.xml.

Now there's a new problem.

With `master`, I get ...

150827-01-snapshot

(Which is as expected.)

With this PR, I get ...

150827-02-snapshot

By a process of binary search, it seems the difference is (first) triggered by this line... https://github.com/philc/vimium/pull/1797/files#diff-d0bdd7e52a95f0b05be2cd0f116c20edR279.

gdh1995:
Oh, I also meet the problem that a XML view with rss content becomes a `<rss>` view if and only if this PR is enabled. But I've applied this PR to my custom Vimium project (its code is very eariler and it has no iframe), and the page works well as before. I have no idea.

* I'm soory I don't understand the github's diff page, so which line do you mean?
* How to debug initing of a page?
* I've tested appending `<div>` with a iframe and style in its shadowDOM to a loaded page, but nothing happened.

smblott-github:
> I'm soory I don't understand the github's diff page, so which line do you mean?

It appears to be the first call of `DomUtils.createElement "div" (line 279)`.  I just did a binary search to find it.

gdh1995:
@smblott-github fixed: a XML-format rss file will show using a RSS-format view, if we call `document.createElementNS("http://www.w3.org/1999/xhtml", tagName)` before document is ready (`document.head` and `document.body` are both `null`). So we need to change `focusThisFrame`, I'll give a pull request later.

lib/clipboard.coffee
gdh1995: Yes, and it's because the background page doesn't contain `lib/dom_utils.js` in manifest.json.

gdh1995:
@smblott-github Now PR #1798 has fixed bugs found. Thanks for your patience, and please test more.

PR #1798, opened 2015-08-28
Run on XML pages (version 2)

gdh1995:
This is based on mrmr1993 version, and  fixs some bugs found in PR #1797.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -276,17 +276,16 @@ setScrollPosition = ({ scrollX, scrollY }) ->
 #
 window.focusThisFrame = do ->
   # Create a shadow DOM wrapping the frame so the page's styles don't interfere with ours.
-  highlightedFrameElement = document.createElement "div"
+  highlightedFrameElement = DomUtils.createElement "div"
   # PhantomJS doesn't support createShadowRoot, so guard against its non-existance.
   _shadowDOM = highlightedFrameElement.createShadowRoot?() ? highlightedFrameElement
 
   # Inject stylesheet.
-  _styleSheet = document.createElement "style"
-  if _styleSheet.style?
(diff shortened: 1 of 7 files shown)


gdh1995:
Sorry. I'll do `cake test` on my computer. Just a while.

gdh1995:
If better, I can rebase my repo and create a new PR ,since these commit logs are awful.

mrmr1993:
Code LGTM (but haven't tested since I'm on holiday). I'll close #1797.

content_scripts/vimium_frontend.coffee
```diff
@@ -274,23 +274,22 @@ setScrollPosition = ({ scrollX, scrollY }) ->
 #
 # Called from the backend in order to change frame focus.
 #
-window.focusThisFrame = do ->
+setTimeout (->
```
smblott-github: @gdh1995.  Thanks for this.

Effectively, you're running this on "nextTick".  How do we know that `createElement` will work then, if it won't work now?  I kinda expected to see `DomUtils.documentReady` here.

(Separately -- and for future reference -- we have `Utils.setTimeout()` and `Utils.nextTick()`.)

gdh1995:
I did such a test: modify `manifest.json` and set `.content_scripts[0].run_at = "document_end"`, and then the old `vimium_frontend.coffee` would not make a XML view page reset into a RSS one. This means that we just need to call every `createElementNS` after `document.readyState` is `"interactive"` (more on [developer.../.../content_scripts](https://developer.chrome.com/extensions/content_scripts)).

Since the main thread parses HTML content and executes scripts serially, it seems that `setTimeout` will work only after the page is ready, but if a server blocks the XML body (e.g. sleeps some seconds before send a next chunk), I do know whether the main thread will be blocked or not.

So, let's use `documentReady` firstly.

smblott-github:
Thanks, @gdh1995.

Issue #1804, opened 2015-09-08
Allow <A-Enter>  to open in a new tab in Vominibar opened by `o`(lowercase)


BohrShaw:
Similar to Chrome's Omnibar, in which pressing `<A-Enter>` could open in a new tab, I like to see this available in Vomnibar.

From time to time, I pressed `o`, then realized it should be `O`, so I would `<Esc>`, then `O`. This is not ideal and the reason I open up this issue.

smblott-github:
Thanks @BohrShaw. `<Ctrl>-<Enter>` does that.
(I guess we could add your suggestion, if that's the Chrome way.)

BohrShaw:
Oh, I didn't realize `<Ctrl>-<Enter>` can make it.
`<A-Enter>` is actually cross-browser and I would love to see it's available.

smblott-github:
Hmm. `<Alt>-<Enter>` doesn't seem to open in a new tab (from the omnibox) for me. ???

BohrShaw:
@smblott-github Strange! Then what does it do?
I usually press `<A-Enter>` when there are at least _some text_ in the omnibar.

smblott-github:
> Then what does it do?

Nothing.  Like no key was pressed. (And I don't think it's a window-manager binding.)

BohrShaw:
I'm pretty sure the behavior of `<A-Enter>` is available in vanilla browsers like Chrome/Chromium, Firefox, IE/Edge. (I don't use OS X though.)

mrmr1993:
The [official keyboard bindings list](https://support.google.com/chrome/answer/157179?hl=en-GB) (under address bar shortcuts) seems to indicate that Chrome for Windows has the shortcut, but Chrome for Linux doesn't

BohrShaw:
@mrmr1993 I'm using Chromium on Ubuntu, which has this shortcut.

smblott-github:
Doh!
I took a closer look and (apparently) I do have a window manager binding. `<Alt-Enter>` does indeed open in a new tab from the omnibox, as @BohrShaw stated.  Sorry for the confusion.

If that's what Chrome does -- and since we don't anything else with the binding -- we should probably do the same, as originally suggested.

philc:
Agree.

smblott-github:
Closed in light of #1805 (which is merged).

mrmr1993:
Late to the party, but I did too. Sorry all!

PR #1805, opened 2015-09-09
Alt modifier on Enter in vomnibar opens in new tab

smblott-github:
`<Alt-Enter>` in Vomnibar opens in a new tab, as suggested in #1804.

What do you think, @mrmr1993?  On the one hand, there's no reason to be non-chrome-like here.  On the other hand, we're wasting a lot of bindings.

(Interestingly, it seems Chrome treats `Ctrl-Enter` as "I'm feeling lucky...".)


smblott-github:
Merged in light @philc's support in #1804.

Issue #1802, opened 2015-09-05
Tweaking vomnibar custom search engine completion UX?


smblott-github:
Here are a couple of pre-release UX issues we could think about for the vomnibar.

1. There is currently no proper way to explicitly disable completion.  Do we need one?

    (A hokey workaround is for the user to use custom search engine URLs which don't match our built-in regexps -- e.g. by adding bogus parameters.)

2. There is currently a delay of up to 200ms before we launch a completion request (up to 100ms from the vomnibar front end, then a guaranteed 100ms in the back end).  The idea is to avoid sending completion requests if the user is still typing.  However, the delay can make completion feel sluggish compared to the omnibox.

(Also on the pre-release to-do list: check that the relevant section of the wiki is up to date and well written.)

mrmr1993:
Sorry for the slow reply, missed the email for this.

> no proper way to explicitly disable completion. Do we need one?

I don't feel strongly about this.

> There is currently a delay of up to 200ms before we launch a completion request (up to 100ms from the vomnibar front end, then a guaranteed 100ms in the back end).

* Looks to me like there's a strict 150ms delay in the front end for all omni mode queries
  - this covers all completions
  - exception: when the entire query is a keyword, there is no delay. (Or, at least, I *think* there's no delay; the code is hard to read.)
* There's then another strict 100ms delay in `CompetionSearch`
  - this would seem to be redundant if we're already delaying by 150ms in the front end.
  - we're wasting 100ms of the front end's 150ms in a delay, so we only have 50ms to make the request and get the information to the front end before it's sent another request and invalidated the ID of the last one.

I think we should
* drop the delay in `CompletionSearch` (perhaps just change it to 0 to keep the change small and revertable).
  - I've implemented this, with a brief commit message to explain, in ad93e1cb2e74a40af8149c501d7a8026eb18efdd.
* use the delay in the front end to block the sending of rapid multiple requests, but send the first request immediately
  - this means that the 150ms delay only applies when we have a request being processed, and not for every update indiscriminately.
  - I've implemented this, and renamed a variable for clarity, in a5a978ed3da9b0bda19f86efce196e916b69310d.

I've also just discovered that, if you stop typing during the front end 150ms delay, you won't ever get a match for your full query, since that request is never sent. This might be more of an issue now we're using web completions, and should probably be fixed. I'll write a fix for it tomorrow if nobody beats me to it.

smblott-github:
[...]

PR #1806, opened 2015-09-09
Adjust delays to make the Vomnibar more responsive

mrmr1993:
This implements the ideas from [this comment](https://github.com/philc/vimium/issues/1802#issuecomment-138775436) for issue #1802.

Ping @smblott-github.

### DIFF
--- a/background_scripts/completion_search.coffee
+++ b/background_scripts/completion_search.coffee
@@ -42,7 +42,7 @@ CompletionSearch =
 
   # The amount of time to wait for new requests before launching the current request (for example, if the user
   # is still typing).
-  delay: 100
+  delay: 0
 
   get: (searchUrl, url, callback) ->
     xhr = new XMLHttpRequest()
--- a/pages/vomnibar.coffee
+++ b/pages/vomnibar.coffee
@@ -238,9 +238,14 @@ class VomnibarUI
     else if not @updateTimer?
       # Throttle requests for a better user experience, and to take some load off the CPU (not every
       # keystroke will cause a dedicated update).
+      @shouldUpdateCompletionsOnTimeout = false
       @updateTimer = Utils.setTimeout @refreshInterval, =>
         @updateTimer = null
+        @updateCompletions() if @shouldUpdateCompletionsOnTimeout
       @updateCompletions callback
+    else
+      # Don't execute this request now, but we need to send a request for it eventually.
+      @shouldUpdateCompletionsOnTimeout = true
 
     @input.focus()
 
--- a/pages/vomnibar.coffee
+++ b/pages/vomnibar.coffee
@@ -212,12 +212,12 @@ class VomnibarUI
     @completer.cancel()
     if 0 <= @selection and @completions[@selection].customSearchMode and not @customSearchMode
       @customSearchMode = @completions[@selection].customSearchMode
-      updateSynchronously = true
+      mustRun = true
     # If the user types, then don't reset any previous text, and reset the selection.
     if @previousInputValue?
       @previousInputValue = null
       @selection = -1
-    @update updateSynchronously
+    @update mustRun
 
   clearUpdateTimer: ->
     if @updateTimer?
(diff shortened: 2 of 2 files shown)


background_scripts/completion_search.coffee
```diff
@@ -42,7 +42,7 @@ CompletionSearch =
 
   # The amount of time to wait for new requests before launching the current request (for example, if the user
   # is still typing).
-  delay: 100
+  delay: 0
```
smblott-github: This may not be a great idea.

I don't have a precise figure, but ...

If you turn on debugging in this file then you can watch what happens to completion requests.  A pretty significant portion (30%-50%, perhaps) of them are cancelled before being launched because the user is still typing.  As good netizens, there's no need to send that unneeded load to the servers.

mrmr1993: I know the intention was good, but a 100ms delay does nothing to reduce the number of requests, it just leaves us with 50ms to do all our processing and requests rather than 150ms.

If we're going to rate limit, let's do it in one place (the frontend) and do it properly (longer than 150ms).

smblott-github: > but a 100ms delay does nothing to reduce the number of requests

Yes it does.  Run with the debug turned on and you'll see them being cancelled.

There are two types of rate limiting going on.  They're different.  On the front end we throttle updates, but allow them to proceed once every 150ms.  On the back end we cancel completion requests completely if the user is still typing (to be good net citizens).

mrmr1993: So that's the 50ms not being long enough for our processing sometimes. Processor speed and OS load seems like a bad way to decide throttling. Lets change the delay in the frontend to reflect what we think reasonable throttling is when we're using completers.

smblott-github:
This is basically two separate ideas, @mrmr1993.  Unfortunately, I'm not sure we should pursue either.

For the front-end delay... It's not obvious -- to me at least -- that the proposed approach is better than the status quo, and there may be cases where it's worse.  Since the existing logic achieves what it aims to (keep load off the CPU while at the same time providing a responsive UX), and nobody has raised any usability issues, we're probably best just sticking with what we've got.

For the back-end delay... We don't *need* the delay, but it does eliminate a fair number of unnecessary network requests.

We could tweak the constants.  The current values seem borderline sluggish to me, but nevertheless quite usable.  Let's just wait and see what the feedback from 1.52 is.

> So that's the 50ms not being long enough for our processing sometimes. Processor speed and OS load seems like a bad way to decide throttling.

The 50ms is 150-100, yes?  This number is not particularly relevant here.  The current approach makes no particular assumption about processor speeds.  We post an advisory "cancel" message to the back end whenever the user types, and -- in the case of completion -- we do indeed cancel any pending HTTP request which has not yet been launched.

mrmr1993:
> For the front-end delay... It's not obvious -- to me at least -- that the proposed approach is better than the status quo,...

I thought the proposal was quite simple: run the code and delay running it again, rather than delay running the code and don't run it again. So all of the requests to the background are 150ms sooner, which is more responsive, and so better.

> and there may be cases where it's worse.

What will doing the same thing 150ms sooner make worse?

> The current approach makes no particular assumption about processor speeds.

I misspoke there -- I thought latency was cancelling some requests, not newer user input.

> We post an advisory "cancel" message to the back end whenever the user types, and -- in the case of completion -- we do indeed cancel any pending HTTP request which has not yet been launched.

We're trying to solve the problem that "we don't want to make a request more than x times per second". Why not just not send a request more than x times a second? Have I missed something fundamental? I can't fathom why we would want to rate limit twice, in two different places, for a single request.

smblott-github:
> I thought the proposal was quite simple: run the code and delay running it again, rather than delay running the code and don't run it again while delayed. So all of the requests to the background are 150ms sooner, which is more responsive, and so better.

Yes.  But also consider the case where the user types at exactly one character every 51ms.  The proposed approach is worse in that case.  Since the UX seems to be fine with the status quo, I'd stick with that.

> I can't fathom why we would want to rate limit twice, in two different places

In the back end we're not rate limiting requests, we're **cancelling them** once we *know* that the results aren't needed.  We do this because otherwise we'd be consuming external resources, and we're good net citizens.

If the user keeps typing, then the internally-generated suggestions update every 150ms.  That's rate limiting.  However, the externally-generated completions don't update at all until the user stops typing.  The two mechanisms are doing different things.

mrmr1993:
> But also consider the case where the user types at exactly one character every 51ms.

More generally, we can consider behaviour across typing speed measured in charactrers per minute (CPM):
* < 125 CPM &mdash; all requests are cancelled
* > 125 CPM & < 150 CPM &mdash; a request is sent for every second-ish key, 100-250ms after the key was pressed. A lot of the responses are then discarded by the front end when they arrive.
* > 150 CPM & < 250 CPM &mdash; all requests are cancelled.
* > 250 CPM &mdash; a request is sent for every key. Requests will be ignored if < 200 CPM + round trip request time.

So, we could rate limit to 1 request per 250ms and we have essentially the same result, less some complication in the 125-150 CPM range.

> In the back end we're not rate limiting requests, we're **cancelling** them once we *know* that the results aren't needed.

I'd argue that, just switching to rate limiting,
* the consequences are much the same,
* rate limiting alone makes fewer requests (the same number except for in the 125-150 CPM range), and
* I'm not sure that the results "aren't needed"; they'll nearly always be better than the suggestions we're showing as we reject the response, so it may be better to just show them.

smblott-github:
Sorry, @mrmr1993.
I do not think that this is the correct approach.  `<opinion>` The status quo is better. `</opinion>`

PR #1808, opened 2015-09-09
Search completion for amazon.de

fnkr:


smblott-github:
Thanks, @fnkr.

Interestingly, this seems to provide English completions initially.  But when the word is clearly not English, it provides German completions.  E.g. "regenb" comlpletes to "regenbogen".  The same happens for French: "parapl" completes to "parapluie".

fnkr:
@smblott-github There are a lot of english words in the article titles on amazon.de, so it's not surprising that their dictionary for amazon.de contains english words too.

Thanks for merging.

PR #1809, opened 2015-09-09
Add "older" to previousPatterns and "newer" to nextPatterns

fnkr:
Useful for blogs where buttons are named "older posts" and "newer posts".


smblott-github:
Thanks, @fnkr.
This will only affect users who haven't changed the original default value.

eggbean:
Newer [[ and Older ]] isn't working for me on github commit history?

gdh1995:
@eggbean It's because of another bug that `goNext` may crash if a page uses SVG buttons.

PR #1811, opened 2015-09-10
Search completion; add support for Qwant completion

PrestanceDesign:


smblott-github:
Is this working for you, @PrestanceDesign?

I'm seeing a parse error in the response.  The data I get back is...

snapshot

Should be easy enough to fix.

PrestanceDesign:
Hi,

You are right, my JSON parsing is wrong.
It's fix on my last commit, now it's working perfect.

smblott-github:
That works nicely, thanks @PrestanceDesign.

PR #1813, opened 2015-09-10
Don't decode javascript URIs on Chrome 46.0.2467.2 and later

mrmr1993:
The fix for the [Chromium issue 483000](https://code.google.com/p/chromium/issues/detail?id=483000) landed in version 46.0.2467.2, so it is no longer necessary for #1611.

Currently, in that version and later, javascript URIs will be decoded twice &mdash; once by us and once by Chrome when opening the URL &mdash; which will break some bookmarklets.

### DIFF
--- a/lib/utils.coffee
+++ b/lib/utils.coffee
@@ -144,7 +144,8 @@ Utils =
     if Utils.hasChromePrefix string
       string
     else if Utils.hasJavascriptPrefix string
-      Utils.decodeURIByParts string
+      # In Chrome versions older than 46.0.2467.2, encoded javascript URIs weren't handled correctly by.
+      if Utils.haveChromeVersion "46.0.2467.2" then string else Utils.decodeURIByParts string
     else if Utils.isUrl string
       Utils.createFullUrl string
     else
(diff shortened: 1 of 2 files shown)


mrmr1993:
Note that this will break some bookmarklets that contain URLs with encoded data in them. So, for example,

```
javascript:var%20inject%3Ddocument.createElement(%22script%22)%3Binject.src%3D%22%2F%2Fmy.website.com%2FdoSomething.js%3Fpadding%3D%2526nbsp%253B%26url%3D%22%2BencodeURIComponent(location.href)%3Bdocument.head.appendChild(inject)%3B
```

will be interpreted as

```
javascript:var inject=document.createElement("script");inject.src="//my.website.com/doSomething.js?padding=&nbsp;&url="+encodeURIComponent(location.href);document.head.appendChild(inject);
```

instead of the correct

```
javascript:var inject=document.createElement("script");inject.src="//my.website.com/doSomething.js?padding=%26nbsp%3B&url="+encodeURIComponent(location.href);document.head.appendChild(inject);
```

This issue is already present in both the dev and beta versions of Chrome, and the affected version will hit stable in around one and a half months. @smblott-github @philc Is there the possibility of merging this and pushing a new version containing it, so the issue doesn't present in the wild?

smblott-github:
We should hold of pushing any new version for a week or two (just to clean up issues emerging from 1.52).  Otherwise, agreed.

PR #1816, opened 2015-09-11
Make the sort used for filetered link hints stable.

smblott-github:
JavaScript's sort function is not stable; this PR makes the sort used for filtered link hints stable.

There are two reasons for doing this:

- High-scoring hints are more likely to keep the same hint string as the user continues typing. (Currently, the hints assigned change based on the vaguaries of the non-stable sort.)

- For equal-scoring hints, we retain the visit-child-before-parent ordering (which is used to NOT match a parent's text if we have already matched that text in a child).

And, as a result of all of that, the UX is more predictable and hence better.

Edit:

A specific use case where this improves things is on Reddit and using filtered link hints to select one of the "comments" links.  In this case, after typing "com", there are still lots of matches.  With this PR, the hints assigned don't change (so much) as the user continues typing "ment".

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -126,18 +126,22 @@ class LinkHintsMode
   #
   # Creates a link marker for the given link.
   #
-  createMarkerFor: (link) ->
-    marker = DomUtils.createElement "div"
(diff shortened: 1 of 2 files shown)


mrmr1993:
Nice, LGTM.

PR #1819, opened 2015-09-11
Emit mouseout for the last link when clicking a new link with link hints

### LINKED ISSUE #1189: Links "hovered" by clicking are never "unhovered"
Links which are clicked using `LinkHints.activateModeToOpenInNewTab`, `LinkHints.activateModeWithQueue`, etc. trigger the hover effect of the link by sending a `mouseover` event, but the corresponding `mouseout` event is never sent, so the hover effect persists even after clicking other links.

An example of this behaviour is the movie posters on [this IMDb list](http://www.imdb.com/list/ls059872711?start=1&view=grid&sort=listorian:asc&defaults=1&scb=0.034486552234739065).

PR #1032 fixes this.

mrmr1993:
This is the unobjectionable part of #1032, intended to fix #1189.

Whenever a new link is clicked with `DomUtils.simulateClick` in `LinkHints`, the previous element to have been clicked is sent a `mouseout` event. This greatly improves usability of some websites with hover effects (eg. [an IMDb grid view page](http://www.imdb.com/list/ls000004180/?start=1&view=grid&sort=listorian:asc&defaults=1&scb=0.23077548970468342), where opening one of the films in a new tab shows a hover pop-up).

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -93,6 +93,7 @@ class LinkHintsMode
       else
         @hintMode.setIndicator "Open multiple links in new tabs."
       @linkActivator = (link) ->
+        @unhoverLast link
         # When "clicking" on a link, dispatch the event with the appropriate meta key (CMD on Mac, CTRL on
         # windows) to open it in a new tab if necessary.
         DomUtils.simulateClick link,
@@ -117,10 +118,13 @@ class LinkHintsMode
     else if @mode is DOWNLOAD_LINK_URL
       @hintMode.setIndicator "Download link URL."
       @linkActivator = (link) ->
+        @unhoverLast link
         DomUtils.simulateClick link, altKey: true, ctrlKey: false, metaKey: false
     else # OPEN_IN_CURRENT_TAB
       @hintMode.setIndicator "Open link in current tab."
-      @linkActivator = (link) -> DomUtils.simulateClick.bind(DomUtils, link)()
+      @linkActivator = (link) ->
+        @unhoverLast link
+        DomUtils.simulateClick link
 
   #
   # Creates a link marker for the given link.
@@ -376,6 +380,12 @@ class LinkHintsMode
         @deactivateMode delay, -> LinkHints.activateModeWithQueue()
       else
         @deactivateMode delay
+
+  unhoverLast: do ->
+    lastHoveredElement = null
+    (element) ->
+      DomUtils.simulateUnhover lastHoveredElement if lastHoveredElement?
+      lastHoveredElement = element
 
   #
   # Shows the marker, highlighting matchingCharCount characters.
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -93,7 +93,6 @@ class LinkHintsMode
       else
         @hintMode.setIndicator "Open multiple links in new tabs."
       @linkActivator = (link) ->
-        @unhoverLast link
         # When "clicking" on a link, dispatch the event with the appropriate meta key (CMD on Mac, CTRL on
         # windows) to open it in a new tab if necessary.
         DomUtils.simulateClick link,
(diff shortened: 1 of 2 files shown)


gdh1995:
Here's a thing I do many times every day: A old-fashion web page use `mousein` and `mouseout` to show a multi-level menu, and I need to click several submenus which are included in one same top menu item.

Usually I use `LinkHints.activateModeWithQueue` to do this, but if this PR is merged, the top menu will receive a `mouseout` and hide the whole second level menus, which makes `activateModeWithQueue` has no use.

This top menu's DOM tree is like:
``` html
<li class="menu-item-top">
  <a class="menu-item-top-title">***</a>
  <ol class="menu-second" style="display: none;">
    <li class="menu-item-second"></li>
    <!-- multi <li>***</li> -->
  </ol>
</li>
```

mrmr1993:
@gdh1995 Can you give an example of a page where this happens? A native mouse movement will also send a `mouseout` and `mouseover` event, so this should behave similarly.

Perhaps we need to set `relatedTarget` (the last argument to `initMouseEvent`) for this to work.

gdh1995:
I'm sorry for that my above argument turns to be wrong - I made a error test.

The DOM tree above is abstract, and this PR works well for a real tree:
``` html
<div class="menu2" id="apDiv1">
    <table width="100" border="0" cellspacing="1" cellpadding="0">
      <tr><td role="link" id="apMenu1" onmouseover="VisitLayer('apDiv2','show')" 
             onmouseout="VisitLayerOut('apDiv2','hide')" class="td1"></td></tr>
    </table>
</div>
<div class="menu2" id="apDiv2" onmouseover="VisitLayer('apDiv2','show')" onmouseout="VisitLayer('apDiv2','hide')">
<table width="100" border="1" cellspacing="1" cellpadding="1">
  <tr><td>second 1</td></tr>
  <tr><td>second item 2</td></tr>
</table>
```

`VisitLayer` is called twice when we move mouse from `#apMenu1` to `#apDiv2`, and `#apDiv2` is hidden and then shown again.
```
function VisitLayer() { //v9.0
  var p,v,obj,args=VisitLayer.arguments;
  if((obj=document.getElementById(args[0]))!=null) {
    v=args[1];
    if (obj.style) {
      obj=obj.style; v=(v=='show')?'visible':(v=='hide')?'hidden':v;
    }
    obj.visibility=v;
  }
}
```

smblott-github:
Is this still a good thing to do, @mrmr1993.  Seems promising to me.

(Some of the code here is in #1948.)

mrmr1993:
Very much yes (try background opening a few of the films from the grid view [here](http://www.imdb.com/list/ls000004180/?start=1&view=grid&sort=listorian:asc&defaults=1&scb=0.23077548970468342), notice how we collect hovered menus).

smblott-github:
> try background opening a few of the films from the grid view...

Yikes.
That's horrible.

gdh1995:
Well, I think Vimium's behavior on the grid view page of films is acceptable enough. What else do you think is "horrible"?

mrmr1993:
screenshot from 2016-01-30 12 25 47_crop

Doesn't look all that acceptable to me.

gdh1995:
Oops, I misunderstood the conversation between you two. Sorry for my poor English.

smblott-github:
Thinking about this...

Might the problem be more general (and the solution simpler)?

The idea is that **any** time we `simulateClick` **anything** we should generate a `mouseout` event for the last thing we clicked.

In which case, the code goes in `simulateClick` (not in the link hints).

mrmr1993:
I've updated this with a new commit that makes every click (actually, simulated hover) trigger a `mouseout` event on the last hovered (by us) element. All of the relevant code now lives in dom_utils.

smblott-github:
That's nice, @mrmr1993.  Thanks.

PR #1821, opened 2015-09-13
Don't show UIComponent <iframe>s while vimium.css loads

### LINKED ISSUE #1817: vomnibarFrame and vimiumHUDFrame visible on new tab
OSX 10.10.5 / Chrome 46.0.2490.22 beta (64-bit)

When a new empty tab is opened, the two frames mentioned in the title are visible as in the attached screenshot.

These don't appear anywhere except the initial empty tab and don't seem to have any adverse effect otherwise
Started within the last week - no idea which specific Chrome update triggered it, sorry.

developer_tools_-_https___www_google_co_uk___chrome_newtab-serviceworker_js_and_new_tab_and_xara_viewer_for_osx_-_google_search_and_react_native_debugger

mrmr1993:
This fixes #1817 by hiding the Vomnibar and HUD when our stylesheet isn't loaded.

This has the effect that `b`/`B`/`o`/`O`/etc. appear to do nothing, but require a click on the new tab page to continue using Vimium normally.

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -9,7 +9,10 @@ class UIComponent
     styleSheet = DomUtils.createElement "style"
     styleSheet.type = "text/css"
     # Default to everything hidden while the stylesheet loads.
-    styleSheet.innerHTML = "@import url(\"#{chrome.runtime.getURL("content_scripts/vimium.css")}\");"
+    styleSheet.innerHTML = """
+      @import url("#{chrome.runtime.getURL("content_scripts/vimium.css")}");
+      iframe {display: none;}
+    """
 
     @iframeElement = DomUtils.createElement "iframe"
     extend @iframeElement,


smblott-github:
Hmm.  Thanks, @mrmr1993.  When I try this out, it worked beautifully on the first new tab I tried it on, then wouldn't work at all thereafter.  Which is weird. (The vomnibar was successfully hidden, but couldn't be used.)

I wonder whether we shouldn't take a different approach.  An alternative would be to explicitly disable Vimium on the NTB for the Chromium versions known to have this bug.

mrmr1993:
> (The vomnibar was successfully hidden, but couldn't be used.)

That was the point of this PR. I figured a purely aesthetic fix should be enough, since
* most people will (reasonably) assume that nothing has changed, and Vimium still doesn't work on the new tab page, so they won't notice at all, and
* they are only broken for a single major version anyway.

Even if you want to disable Vimium completely on the NTP, I still think it's worth having this so that the `<iframe>`s don't show while the stylesheet loads (eg. when the page has hit the concurrent connection limit and loading our stylesheet has to wait).

smblott-github:
> I still think it's worth having this so that the <iframe>s don't show while the stylesheet loads

Yes.  That's worth it in itself.

Thanks.

PR #1822, opened 2015-09-13
Make <label>s clickable with LinkHints

### LINKED ISSUE #1783: Press F or f shows links but cannot navigate
Hi Vimium crew,

As shown in the picture below, when pressed PA or GA, Vimium does not navigate to the link; however, other link in the page seem to work just fine though. Not sure whether this is the problem for Vimium or not.

The problem link: http://www.timeout.com/newyork/movies/the-100-best-movies-of-all-time
Chrome version: 44.0.2403.107 m
Vimium version: 1.51

Thank you very much and love your product!

capture

mrmr1993:
This fixes #1783.

Edit: Updated to only show hints for `<label>`s whose corresponding element doesn't have a hint itself, to avoid lots of duplicated hints where `<label>` wraps the element.

Edit 2: [JSFiddle](http://jsfiddle.net/mhwa9zep/1/) for testing all the cases.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -214,6 +214,8 @@ class LinkHintsMode
                              (element.readOnly and DomUtils.isSelectable element))
       when "button", "select"
         isClickable ||= not element.disabled
+      when "label"
+        isClickable ||= element.control?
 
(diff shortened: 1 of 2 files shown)


smblott-github:
Thanks, @mrmr1993.  Very nice!

PR #1828, opened 2015-09-17
Add <count> prefix for g0 and g$.

### LINKED ISSUE #1298: <n>gt is inconsistent with vim
Hi, in vim, `<n>gt` will switch to the nth tab, but in vimium, `<n>gt` moves 3 tabs to the right.

Sorry for filing an issue if this feature is intentional.
I think it should be mentioned near these lines in the README:

```
J, gT      go one tab left
K, gt      go one tab right
```

smblott-github:
`2g0` takes you to the second tab.
`2g$` takes you to the second last tab.

This gives us `gt`/`gT` for relative tab selection and `g0`/`g$` for absolute tab selection.

`<count>` previously had no special meaning for `g0` and `g$`, and it's not clear it could have any other meaning than that implemented here.

Fixes #1298 (kind of).

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,7 @@ Release Notes
 -------------
 Updates since 1.52 (not yet released)
 
+- `g0` and `g$` now accept count prefixes; so `2g0` selects the second tab, and so on.
 - Bug fixes:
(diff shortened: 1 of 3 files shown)


mrmr1993:
Seems reasonable, and code LGTM. Nice!

PR #1829, opened 2015-09-17
Bound count for H and L by history length.

### LINKED ISSUE #1691: Better count prefix for history (H and L).
`1000H` should take me as far back as I can go in the history.  Currently, it (usually) does nothing.

Similarly for `1000L`.

(Assuming there are fewer than 1000 history entries.)

smblott-github:
Bound the count for H and L by history.length - 1.  Going further back or forward than this is always silently discarded.

This means that, in certain circumstances, you can say 1000H to go back to the first history page and 1000L to go forward to the end of the history.

(However, this doesn't work unless you're starting at the beginning or end of the history.  The history API doesn't expose where in the history we currently are.)

Fixes #1691, kind of.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -325,8 +325,8 @@ extend window,
 
 extend window,
   reload: -> window.location.reload()
-  goBack: (count) -> history.go(-count)
-  goForward: (count) -> history.go(count)
+  goBack: (count) -> history.go Math.max -count, -(history.length - 1)
+  goForward: (count) -> history.go Math.min count, history.length - 1
 
   goUp: (count) ->
     url = window.location.href


smblott-github:
@mrmr1993... Is there another API (not `history`) which would allow us to fix this properly?

mrmr1993:
> Is there another API (not `history`) which would allow us to fix this properly?

Not that I know of, no. Although, looping through indexes seems to work, even if it isn't ideal:

```coffee
  goBack: (count) -> history.go i for i in [-(Math.min count, history.length - 1)..0] by 1
  goForward: (count) -> history.go i for i in [(Math.min count, history.length - 1)...0] by -1
```

gdh1995:
> it will take you back nearly the square of the number of pages you wanted

In fact, it go back `N * (N+1) / 2` times, which means, `history.go i for i in [-(Math.min count, history.length - 1)..0] by 1` has finished its loop before the current page closed.

mrmr1993:
> `N * (N+1) / 2`

Thanks @gdh1995, that's why I said nearly the square.

gdh1995:
> unless there's no navigation

Can we listen `hashchange` event, and stop if only catch one? Just an idea without any test now.

mrmr1993:
> Can we listen `hashchange` event, and stop if only catch one?

In the spec, [history traversal](https://html.spec.whatwg.org/#history-traversal) emits events asynchronously when the *`non-blocking events`* flag is set, and hash (referred to as *fragment identifier*) changes [set the *`non-blocking events`* flag](https://html.spec.whatwg.org/#scroll-to-fragid). So, if Chrome is to spec, `hashchange` should be fired asynchronously.

We should be able to use `setTimeout` with time 0 to queue the next attempt after the `hashchange` event (maybe) fires. So, **in theory, yes.**

I'll note that we would also need to detect the `popstate` event (which is synchronous) to cover all the bases.

Under these assumptions, the following is roughly what I'd expect that we'd need:

```coffee
historyGo: (count, stepSize) ->
    navigationOccurred = false
    hasNavigated = -> navigationOccurred = true
    window.addEventListener "hashchange", hasNavigated, true
    window.addEventListener "popstate", hasNavigated, true
    history.go count
    setTimeout ( =>
      window.removeEventListener "hashchange", hasNavigated, true
      window.removeEventListener "popstate", hasNavigated, true
      @historyGo count + stepSize, stepSize unless navigationOccurred or count == 0
    ), 0
goBack: (count) -> @historyGo -count, 1
goForward: (count) -> @historyGo count, -1
```

mrmr1993:
Sorry for the spam, but I just remembered that tab history spans frames. So any attempt similar to mine above will fail in a content script with frames that have been navigated.

[`chrome.webNavigation`](https://developer.chrome.com/extensions/webNavigation) in the background page would probably be the only reliable way to do this.

smblott-github:
Closing.
Not particularly useful.

PR #1831, opened 2015-09-18
Disable Vimium on NTP for versions with known bug

smblott-github:
In light of #1817, we have a problem coming up with Chrome 47 (this is due to a Chromium bug, see #1817).

- Vimium has been functional on the NTP for quite some time, so some users expect that behaviour.

- With #1821 (merged), Vimium partially runs. Unfortunately, `o` and `b` (arguably the most likely things to do on the NTP), don't work.  In fact, there aren't many useful Vimium features which work.

This gives us a tricky problem.  Reading reviews on the Chrome Store, at least one user has abandoned Vimium in light of the #1817 problem.

What should we do?  This PR is one option.  It disables Vimium completely on the NTP for the offending Chrome versions.  Arguably, this is easier for users to understand than the partially-enabled state we face at present.

@mrmr1993, @philc: what do you think?

(An alternative might be to find another way to inject the Vimium CSS into the page.)

### DIFF
--- a/background_scripts/exclusions.coffee
+++ b/background_scripts/exclusions.coffee
@@ -28,6 +28,7 @@ root.Exclusions = Exclusions =
   # is the default.  However, when called from the page popup, we are testing what effect candidate new rules
   # would have on the current tab.  In this case, the candidate rules are provided by the caller.
   getRule: (url, rules=@rules) ->
+    return forceDisableRule if forceDisableRule = @checkForChromiumNewTabPageBug url
     matches = (rule for rule in rules when rule.pattern and 0 <= url.search(RegexpCache.get(rule.pattern)))
     # An absolute exclusion rule (with no passKeys) takes priority.
     for rule in matches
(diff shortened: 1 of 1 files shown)


smblott-github:
Options...

If we leave 1.52 in the wild as is:

- many users will experience #1817, which is a pretty UX.

If we publish 1.53 with #1821 (merged):

- the most important Vimium features (`o`, `b`, etc.) won't work on the NTP.

smblott-github:
If we leave 1.52 in the wild:

- many users will experience #1817, which is a pretty bad UX.

If we publish 1.53 with #1821 (merged):

- users get a partial Vimium experience on the NTP.  Arguably the most useful features (`o`, `b`, etc.) won't work -- and some users are used to those features working there.  Also a bad UX.

If we publish 1.53 with this PR (Vimium disabled on NTP) or similar:

- users who are used to Vimium working on the NTP will be confused as to why it doesn't work.  We'll get many "Vimium is broken on NTP" issues posted.

All of these options are bad.

Do we have any other options?

smblott-github:
Closing in favour of #1833.

PR #1086, opened 2014-06-30
Rework to make match counting code for searches more DRY and easier to read

mrmr1993:
PR to (hopefully) address @philc 's comments on #1026, and to improve code readability.

PR #1833, opened 2015-09-18
Load UIComponent stylesheet via XMLHttpRequest, use background as needed

### LINKED ISSUE #1817: vomnibarFrame and vimiumHUDFrame visible on new tab

mrmr1993:
This fixes #1817 (properly this time), making the Vomnibar and HUD work on NTP for Chrome 47.

We try to load styles via `XMLHttpRequest`, and if that fails we use `fetchFileContents` in the background page to do the `XMLHttpRequest` for us.

@smblott-github I think this is the silver bullet you wanted over in #1831.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -668,6 +668,7 @@ sendRequestHandlers =
   setIcon: setIcon
   sendMessageToFrames: sendMessageToFrames
   log: bgLog
+  fetchFileContents: (request, sender) -> fetchFileContents request.fileName
 
 # We always remove chrome.storage.local/findModeRawQueryListIncognito on startup.
 chrome.storage.local.remove "findModeRawQueryListIncognito"
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -4,15 +4,16 @@ class UIComponent
   showing: null
   options: null
   shadowDOM: null
+  styleSheetGetter: null
 
   constructor: (iframeUrl, className, @handleMessage) ->
     styleSheet = DomUtils.createElement "style"
     styleSheet.type = "text/css"
     # Default to everything hidden while the stylesheet loads.
-    styleSheet.innerHTML = """
-      @import url("#{chrome.runtime.getURL("content_scripts/vimium.css")}");
-      iframe {display: none;}
-    """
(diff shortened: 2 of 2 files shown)


smblott-github:
Thanks @mrmr1993.

Yes, this *is* the silver bullet.  A couple of small things (I could fix these myself, but I'll give you the opportunity first).

- `fetchStyles` is misnamed (as are its variables).  It has nothing specifically to do with CSS.  Perhaps `fetchFileContents`?

- Can we remove the CSS inclusion from `manifest.json`?

@philc: We should merge this, give it a couple of days to catch any gotchas, then push 1.53 before Chrome 47 starts moving through the pipeline.

mrmr1993:
> * `fetchStyles` is misnamed (as are its variables). It has nothing specifically to do with CSS. Perhaps `fetchFileContents`?

While that url is hardcoded in it, and so the function specifically loads the Vimium stylesheet from a file, I feel like `fetchFileContents` is too vague. Is `fetchStylesheetContents` (or `fetchCssFileContents`, etc.) acceptable, or would it be better to modify the function's contract?

> * Can we remove the CSS inclusion from `manifest.json`?

Not for now, link hints use it.

philc:
Very nice solution @mrmr1993. This could use a comment explaining why the stylesheet request can fail, since the next developer who sees this won't have the NTP case in mind.

mrmr1993:
@philc I'm loathe to make a PR with your suggested small change since #1086 was so readily ignored. However,  in such a PR, I would have comment above the code in the constructor:

```coffee
# Use an XMLHttpRequest, potentially via the background page, to fetch the stylesheet, to circumvent @include fails ( eg. #1817).
```

Hopefully this &mdash; or similar &mdash; will fare better in getting into the codebase.

PR #1835, opened 2015-09-21
Add a comment explaining the choice of XMLHttpRequest for UIComponent

mrmr1993:
@philc thanks for merging #1086. This PR adds the comment you asked for in #1833.

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -12,6 +12,8 @@ class UIComponent
     # Default to everything hidden while the stylesheet loads.
     styleSheet.innerHTML = "iframe {display: none;}"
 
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @mrmr1993.
I'd actually already added the required comment (before Phil mentioned it).

2eb6d9a7d3412e4b46e081caa7054f925ffe7e92

mrmr1993:
I noticed your comment, but I felt we should probably clarify that CSS `@include` is not a viable alternative because of #1817, in case anybody is tempted to move back to it. So IMO this comment serves a different purpose to yours.

smblott-github:
Thanks for the clarification, @mrmr1993.

PR #1845, opened 2015-10-02
Block keyboard events when a filtered hint matches.

### LINKED ISSUE #1842: When opening a link with link-hint mode, add a delay before listening to new actions
The link-hint mode is a good feature which lets the user save time. However, as this feature is implemented for now it steals time instead. The user has to wait for every letter that is pressed to see if the link has started loading. Otherwise, the user may press a key that represents unwanted behavior at this point, such as: closing the tab; opening a new one; going to the clipboard URL, and the list continues.

By adding a small delay for say 0.5 seconds, or even letting the user set this delay itself, this feature may become much better.

smblott-github:
With filtered link hints, there is a risk that a link is matched but the user is still typing the link's text -- in which case, Vimium commands are launched inadvertently.

Previously, we blocked keyboard events for a fixed 200ms to prevent this.

With this PR, we continue blocking keyboard events until 150ms after the last `keydown` or `keypress` event.  So, we wait until we think the user has stopped typing.

Fixes #1842.


mrmr1993:
Nice, code LGTM.

PR #1849, opened 2015-10-04
Suppress trailing key events (after link hints).

### LINKED ISSUE #1844: Excluded keys should be excluded from `f` Open links hotkeys
If a URL has certain keys disabled, they should not be options to select links when using the `f` "Open a link" command. The current behavior causes hitting these keys to both open the link and pass through to the original page, which then causes undesired results.

smblott-github:
This ensures that -- on leaving link hints mode -- we consume any trailing keyup events (and don't let the underlying page see them).

Fixes #1844; possibly.

Additional notes:

- There are other places where we seem to be leaking keyup events.

- A separate bug... It looks like we're calling `exit()` on link-hints
  mode twice.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -69,6 +69,7 @@ class LinkHintsMode
       indicator: false
       passInitialKeyupEvents: true
       suppressAllKeyboardEvents: true
+      suppressTrailingKeyEvents: true
       exitOnEscape: true
       exitOnClick: true
       exitOnScroll: true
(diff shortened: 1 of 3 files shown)


smblott-github:
> A separate bug... It looks like we're calling exit() on link-hints mode twice.

I tracked this down.  Our simulated click is triggering `hintMode.exit()` (because `exitOnClick` is set).  Later, we're explicitly making the same call.  All of this is harmless currently.  Nevertheless, 7158cd670ae5824ebeb134fe3ec35472e4dc692f (in this PR) makes `mode.exit()` explicitly idempotent.

PR #1167, opened 2014-09-25
Enable link hints for all elements with `click` listeners

mrmr1993:
This PR implements a hook on `addEventListener`, which marks elements registering a `click` event listener so that Vimium recognises them.

This enables link hint support for clickable elements such as eg. the quiz choices on [this Buzzfeed quiz](http://www.buzzfeed.com/adamellis/this-palm-reading-quiz-will-reveal-your-future#2s6on3d) or the slideshow navigation buttons on the [Mercedes website](http://www5.mercedes-benz.com/en/).

This PR

* fixes #1091 
* fixes #993 
* fixes #861 
* improves the situation for #1157 (some clickable elements still don't receive link hints).

Issue #1157, opened 2014-09-13
Custom html what would be treated as href/link


mbollemeijer:
I am a Java web dev one of the frameworks we have in every project is http://www.zkoss.org/
ZK basically generates the html for you.

The problem is that I cant use vimium on most of the mouse clickble links because vimium doesn't recognise the generated links by ZK.
I would love the option to that I can specify which html structures must be treated as a link.

Greets

smblott-github:
Closing.
No response from OP.

PR #1859, opened 2015-10-17
Enable link hints for all elements with `click` listeners

### LINKED ISSUE #861: Follow elements with an "onclick" listener
It would be awesome if you could "click" on elements that have "onclick" listener.
Many sites have these elements and sometimes I must just use mouse to click it.

I guess script should iterate through every element on page. Or is there a list of acitve listeners? Maybe there could be a wrapper for addEventListener to catch these elements?

### LINKED ISSUE #993: activate javascript onclick events
Would be nice to have this.  Thank you.  Great tool BTW.  Love it.

### LINKED ISSUE #1091: Add an option to trigger click event on any element that has a bound listener
While "Open link in current tab" (shortcut - "f") works wonderfully for the most part, there are a lot of sites out there that create their menus by using div or li elements and attaching a click handler to them. Sadly, this makes such sites nigh on impossible to navigate.

mrmr1993:
This PR is a rebase of #1167 onto current master, and fixes the method used so that
- there are no longer any problems with CSP
- `addEventListener` is always hooked before any page scripts are run.

This supersedes #1405, but without dropping any legacy hint conditions as done there.

From #1167:

> This enables link hint support for clickable elements such as eg. the quiz choices on [this Buzzfeed quiz](http://www.buzzfeed.com/adamellis/this-palm-reading-quiz-will-reveal-your-future#2s6on3d) or the slideshow navigation buttons on the [Mercedes website](http://www5.mercedes-benz.com/en/).
> 
> This PR
> - fixes #1091 
> - fixes #993 
> - fixes #861 
> - improves the situation for #1157 (some clickable elements still don't receive link hints).

Edit: [This JSFiddle](http://jsfiddle.net/p7w9yvsn/2/) is useful for testing.


pages/addEventListener_hook.coffee
```diff
@@ -0,0 +1,44 @@
...
+  _addEventListener.apply this, arguments
+
+onLoaded = ->
+  loaded = true
+
+  # Create an element detatched from the DOM, register it with the content scripts.
+  registrationElement = document.createElementNS "http://www.w3.org/1999/xhtml", "div"
+  registrationEvent = new CustomEvent "VimiumRegistrationElementEvent"
+  document.documentElement.appendChild registrationElement
+  registrationElement.dispatchEvent registrationEvent
```
gdh1995: Why not dispatch this event to `window` or `document`?

mrmr1993: The only way to pass an element between Chrome's isolated worlds seems to be as an event's target, so we have to dispatch on the element where `addEventListener` was called.

The basic logic of this code is:
* if the element is in the DOM, dispatch an event on it and catch it in the content script as it bubbles to `document`, or
* if the element isn't in the DOM, we need to add it to somewhere that has an event listener that the event will bubble up to. We don't want to add it to `document` since we don't want to render elements prematurely or trigger lots of `MutationObserver` calls. Thus, the solution I came up with is:
  * communicate `registrationElement` to the content script (which the line you commented on is doing), so it can add a event listener
  * add elements not in the DOM to `registrationElement` so that the event listener catches the events, then remove them again so the page can still add them to the DOM at will.

content_scripts/inject_script.coffee
```diff
@@ -0,0 +1,16 @@
...
+  req.responseText
+
+for script in injectScripts
+  # Inject the script, which will be executed before the page scripts, as long as it is injected directly as
+  # text.
+  # TODO(mrmr1993): Find a reasonable way to inline the scripts to inject here at build time, since doing a
+  # synchronous XMLHttpRequest will slow down every page load.
+  scriptEl = document.createElement "script"
+  scriptEl.innerHTML = fetchFileContents script
+  document.documentElement.insertBefore scriptEl, document.documentElement.firstElementChild
```
gdh1995: `req.open(,,false)` will cause a warning, which is a little ugly.

You said in a comment in #1405 :
> CSP stops us from injecting our hook before page scripts are done loading.

What prevent a `<script>` with `[src]` from executing is its file's origin (host). Once I found an explanation:
* a `<script src="/...">` will run immediately, and `$0.async` is `false`
* a `<script src="http://other-site/...">` will be delayed, and two such scripts will run without any order after page is loaded.

So, I think here we can add `scriptEl.async = false` before `insertBefore` to make this script run at once.
``` coffeescript
for script in injectScripts
  ...
  scriptEl.src = chrome.runtime.getURL(script)
  scriptEl.async = false
  document....
```

And a simple test passed:

mrmr1993: In my tests, I couldn't get a `<script src="...">` to run before page scripts in `<head>`, hence why #1405 had 2 injection methods. Feel free to change the line above to `scriptEl.src = script`; if you find it detects the clickable element for a page with the following script in `<head>` then I'm happy to change it:

```javascript
var clickable = document.createElement("div");
clickable.innerHTML = "test";
clickable.addEventListener("click", function(){alert()});
document.addEventListener("DOMContentLoaded", function(){
  document.body.appendChild(clickable);
});
```

gdh1995:
In `addEventListener_hook.coffee`, I'm confused about:
* There's `document.documentElement.removeChild registrationElement`, so `if wrapInRegistrationElement`, next `element.dispatchEvent registrationEvent` seems not to work.
* In my test, it did not call `markTargetClickable` in `link_hints.coffee`.

Confirmed *Bug*: the action `appendChild and then removeChild` has an side effect:
* `appendChild` will remove element from its parent
* but the next `removeChild` will just clean its parent (set to `null`).

We need to restore its parent after `registrationElement.removeChild element if wrapInRegistrationElement`.

mrmr1993:
> I'm confused about:
>
> * There's `document.documentElement.removeChild registrationElement`, so `if wrapInRegistrationElement`, next `element.dispatchEvent registrationEvent` seems not to work.

There is (or at least should be) an event listener on `registrationElement`, so that should capture the event. Is this not happening?

> * In my test, it did not call `markTargetClickable` in link_hints.coffee.

Can you clarify what your test involved? When I test on [this JSFiddle](http://jsfiddle.net/p7w9yvsn/), all of the link hints show as expected for me, despite there being no evidence that they are clickable in the DOM itself.

> Confirmed **Bug**: the action `appendChild and then removeChild` has an side effect:

I'll fix this now.

gdh1995:
> `<script src="">` `.async = false`

You are right. When I tested local "file:///" files, `async = false` worked, but it failed for iframes running on fiddle.

I captured and read the timeline, and found it took so long to load a new script file that Chrome finished parsing DOM tree. In your implementation, a **synced** XHR request blocks the whole thread, which makes the PR have a chance to work. So I shouldn't change it.

> In my test, it did not call markTargetClickable in link_hints.coffee.

I'm sorry I didn't explain my test. What I did is just to call `addEventListener` after the page is loaded. I've retried several times and it did not work.

There's a race condition between:
* addEventListener_hook.coffee:20 `registrationElement.dispatchEvent registrationEvent`
* vimium_frontend.coffee:192 `installListener "VimiumRegistrationElementEvent"` in `window.installListeners`

Fix: add `"VimiumRegistrationElementEvent"` handler to `window` when vimium_frontend is initing.

mrmr1993:
> There's a race condition

Fixed

mrmr1993:
> What I did is just to call `addEventListener` after the page is loaded.

@gdh1995 I can't seem to be able to repro this, at least since the latest commit. Is it a specific page that isn't working?

gdh1995:
> Is it a specific page that isn't working?

The race condition is just the reason why my test page didn't work, so you've fixed it. Chrome on my laptop is clearly slowly - what a sad story.

gdh1995:
To tell you the truth, I don't like the way to inject a script text by synced XHR. Too expensive.

Maybe we can write a simpler injector: Wrap code of `addEventLIstener_hook` with an anonymous function and put the function into `inject_addEventListener.coffee`. Then we just need to call this function's `toString` to get the script text we want to run.

This's because Chrome has a special cache for extension's content scripts, and they are executed  in order and without interruption.

gdh1995:
Oops, please clean elementsToRegister at the end of  `addEventListener_hook.coffee::onLoaded`.

mrmr1993:
> Oops, please clean elementsToRegister at the end of `addEventListener_hook.coffee::onLoaded`.

Thanks, done.

> To tell you the truth, I don't like the way to inject a script text by synced XHR. Too expensive.

I agree, it's a horrible solution. I threw it together purely to make the code work; I was hoping for some better ideas, since I couldn't think of anything especially compelling.

> Wrap code of `addEventLIstener_hook` with an anonymous function and ... call this function's `toString` to get the script text we want to run.

That's exactly the kind of idea I was looking for! I'll implement this when I get a bit of time tomorrow.

gdh1995:
I want the injected script to be idempotent, and that is, not hook twice:
* I write a custom [Vimium++](/gdh1995/vimium-plus), and sometime I'll enable both Vimium and it
* Then it's not needed to hook twice, and one `VimiumRegistrationElementEvent` is OK enough.

By the way, your method works well, so I think it's time to merge this PR into my project ^_^

content_scripts/inject_script.coffee
```diff
@@ -0,0 +1,16 @@
+injectScripts = ["pages/addEventListener_hook.js"]
+
+fetchFileContents = (extensionFileName) ->
+  req = new XMLHttpRequest()
+  req.open("GET", chrome.runtime.getURL(extensionFileName), false) # false => synchronous
+  req.send()
+  req.responseText
```
smblott-github: Should we not re-use [this](https://github.com/philc/vimium/blob/master/content_scripts/ui_component.coffee#L124)?

(Probably moving it to `dom_utils.coffee`.)

content_scripts/inject_script.coffee
smblott-github: We inject this script into *every frame* of *every page* the user visits.  That's a fair amount of overhead.

Might we instead only inject the scripts when the frame first get's the focus (or is it too late then)?

pages/addEventListener_hook.coffee
```diff
@@ -0,0 +1,49 @@
+_addEventListener = Element::addEventListener
+
+elementsToRegister = []
+registrationElement = null
+
+EventTarget::addEventListener = (type, listener, useCapture) ->
+  eventTarget = if this in [document, window] then document.documentElement else this
+  if type == "click" and eventTarget instanceof Element
+    setTimeout (-> registerElementWithContentScripts eventTarget, "onclick"), 0
```
smblott-github: Use `Utils.nextTick`?

pages/addEventListener_hook.coffee
```diff
@@ -0,0 +1,49 @@
...
+  _addEventListener.apply this, arguments
+
+onLoaded = ->
+
```
smblott-github: Why the extra space?

content_scripts/inject_script.coffee
mrmr1993: The plan was for it to be temporary, so this will disappear shortly when I do what I've said [here](https://github.com/philc/vimium/pull/1859#issuecomment-149070551)

pages/addEventListener_hook.coffee
mrmr1993: This is injected into the page, so we have nothing except what's in this file available.

pages/addEventListener_hook.coffee
```diff
@@ -0,0 +1,49 @@
...
+    # it to our registrationElement, which will capture the events without interfering with the DOM.
+    elementToWrap = element
+    elementToWrap = elementToWrap.parentElement while elementToWrap.parentElement?
+    registrationElement.appendChild elementToWrap
+
+  # Dispatch an event to the content scripts, where the event listener will mark the element.
+  registrationEvent = new CustomEvent "VimiumRegistrationElementEvent-#{type}"
+  element.dispatchEvent registrationEvent
+
+  registrationElement.removeChild elementToWrap if wrapInRegistrationElement
```
smblott-github: This is pretty complicated stuff.  Some explanation would be helpful.

Also, it seems to be diddling with the DOM in a pretty complicated way.  How confident are you that it will be robust going forward?

content_scripts/inject_script.coffee
mrmr1993: > (or is it too late then)?

It's too late then; we're aiming to capture every call to `addEventListener` the page makes, and the page won't wait to be focused to do that.

pages/addEventListener_hook.coffee
mrmr1993: > it seems to be diddling with the DOM in a pretty complicated way.

There's intentionally no effect on the DOM, which is why this is a bit more complicated. I'll add a comment clarifying shortly.

mrmr1993:
Pushed some updates:

* Fixed a bug relating to shadow DOMs
* Inlined the injected script (so it no longer uses `XMLHttpRequest`)
* Added a comment to clarify what `registerElementWithContentScripts` is doing.

gdh1995:
I found a bug: this PR crashes on Vimium Options page, because it's not allowed to inject a `<script>` according to Vimium's CSP. Suggest: add
* `if (window.LinkHints && LinkHints.activateMode) { script(); continue; }`

before `inject_script.coffee`:69.

By the way, can we remove the `<script>`s (may got by set and pass random `id`s)? When `F12`, the script is too ugly.

gdh1995:
I've got ideas about fixing some potential race conditions, so now I want the injected script to be idempotent again:

To be updated if we want to achieve this goal:
* after `_addEventListener = Element::addEventListener`, add `return if _addEventListener.vimiumHooked is true` (in `inject_script.coffee`)
* set the new `EventTarget::addEventListener`: `***.vimiumHooked = true`
* at the end of `link_hints.coffee`, `markTargetClickable` and `VimiumRegistrationElementEvent` event handler call `event.stopPropagation()` manually and then return `true`

smblott-github:
@mrmr1993.  I tried this on various sites (w/ and wo/ click listeners) and it seems to work nicely -- and doesn't seem to cause a plethora of unwanted hints.  So that's good.

I find the injected code very difficult to follow.  It would probably be helpful if you could provide more commentary as to how this works

And how confident are you as to its robustness (to web site peculiarities, or to future Chrome changes)?

mrmr1993:
@gdh1995 

> this PR crashes on Vimium Options page

I've written a fix but I forgot to push it before going to the pub. I'll update this either later or tomorrow.

> another bug is that: we should check `document.readyState === "loading"` before adding "DOMContentLoaded"

> ...

> now I want the injected script to be idempotent again:

This can only happen if the extension is reloaded/updated/installed, and so the script needs to be necessarily non-idempotent to interact with the new Vimium instance. Since late loading can be fixed by refreshing and will occur rarely, I think it's overkill to make the changed that it would need to do it properly (unless I've missed something that would simplify it).

> potential race conditions

`CustomEvent` dispatches synchronously (even across isolated worlds), so there shouldn't be any scope for race conditions here.

mrmr1993:
@smblott-github 

> doesn't seem to cause a plethora of unwanted hints.

Because of how we currently decide which elements are clickable, there's still scope for pages to have double hints for some clickable regions. But it is reassuring that they're not abundant. As a complement to this, I have an idea which should allow us to eliminate nearly all duplicate (and unclickable, like links in `overflow: scroll` parent elements) hints, which I will explore at some point.

> It would probably be helpful if you could provide more commentary as to how this works

I'll try to attack this tomorrow, both giving an overview and moving some code about to clarify what it is doing.

> how confident are you as to its robustness (to web site peculiarities, or to future Chrome changes)?

In order for *the idea* to work correctly, the code needs to be run under the following constraints:
* `<script>` tags with content inlined are always run synchronously (ie. as if by a function call, but in the page context) when injected by a content script
  - [This part of the spec](http://www.w3.org/TR/html5/scripting-1.html#attr-script-crossorigin) seems to indicate that "the user agent must synchronously prepare the script element" and "The user agent must immediately execute the script block" in this case, so this should be reliable.
* (For elements not in the DOM) events can be dispatched and bubble up to ancestors in element trees which aren't attached to the DOM.
  - This seems to be well supported by [this part of the spec](https://dom.spec.whatwg.org/#concept-event-dispatch).

There may be bugs in the implementation that make it behave badly on individual web sites, but it seems as though &mdash; in principle &mdash; this should always work correctly when Chrome is spec-compliant on these. So I would say I'm confident up to implementation (both Chrome's and my own).

gdh1995:
A little obsessive: the new two listeners doesn't belong to LinkHints, but should be in inject_script.coffee.

content_scripts/inject_script.coffee
```diff
@@ -0,0 +1,75 @@
+# We can't inject inline scripts into our chrome-extension pages (eg. options), so don't try.
+return if window.location.protocol == "chrome-extension:"
+
+injectScripts = [
+  ( -> # Hook addEventListener to tell link hints when a click listener is added.
+    _addEventListener = Element::addEventListener
```
gdh1995: Why here is `Element::` and why `EventTarget::` is not chosen?

content_scripts/inject_script.coffee
```diff
@@ -0,0 +1,75 @@
...
+
+    onLoaded = ->
+      # Create an element detatched from the DOM, register it with the content scripts.
+      registrationElement = document.createElementNS "http://www.w3.org/1999/xhtml", "div"
+      registrationEvent = new CustomEvent "VimiumRegistrationElementEvent"
+      document.documentElement.appendChild registrationElement
+      registrationElement.dispatchEvent registrationEvent
+      document.documentElement.removeChild registrationElement
+
+      elementsToRegister.map (args) ->
```
gdh1995: Warning: `Array::map` may be rewritten by other websites and not work. Example: In http://jsfiddle.net/p7w9yvsn/2/, the top frame rewrites `Function::bind`, which causes my custom Chrome extension fail again and again.

content_scripts/inject_script.coffee
mrmr1993: Fixed in 4004793

content_scripts/inject_script.coffee
mrmr1993: Fixed in 4004793

mrmr1993:
> A little obsessive: the new two listeners doesn't belong to LinkHints, but should be in inject_script.coffee.

@gdh1995 I disagree; the injected code and the receiving code are run in separate contexts and should be in separate locations as a consequence, akin to content scripts and background scripts. Accordingly, I've moved the injection (content script) into `lib/dom_utils.coffee`, the injected script into a new folder as `injected_scripts/hook_addEventListener.coffee`, and the handlers live in `content_scripts/link_hints.coffee`.

@smblott-github I've pushed an update which moves quite a lot of the code about and adds some more comments, hopefully making it clearer what's going on. Is the code any more readable now?

gdh1995:
Your new work is OKay. What I want is:
* in the same file, one function adds registration event listeners on Vimium context, and the other dispatches those events
* then if I remove this file from `manifest.json`, no *waste* work will be done on the foreground.

Indeed, it's not a perfect way to keep two functions on different contexts in one file. Just a point of view.

gdh1995:
**A serious bug**? Have you tested on a page having several `<iframe>` nodes under the same origin (src have the same host)?

As far as I found on a mailbox webpage, `window.top` could catch `addEventListener` calls to nodes in an `<iframe>` - An odd actions but it did. For example, when adding a "click" listener,
* the target was `<a>`
* `target.parentElement.parentElement.....parentElement` was `<html>` of the `<iframe>`
* `console.log(document)` showed `document` was `window.top.document`.
* the above happened when `<iframe>` was loading and before it triggered `DOMContentLoaded`.
* `iframe.contentWindow` had different `EventTarget::addEventListener` function object from `window.top`'s.

mrmr1993:
> I'm sorry. Some of your last commits have fixed this problem

Nice catch though, this will need to be addressed for cross-frame link hints.

content_scripts/link_hints.coffee
```diff
@@ -673,5 +674,25 @@ class TypingProtector extends Mode
 
     @onExit callback
 
+# TODO(mrmr1993): Use event.deepPath to traverse into shadow DOMs, as part of a solution to #1861. This
```
gdh1995: We have append target node's ancestor to `registrationElement` if `document.contains` returns `false`, so I think, there won't be shadow DOMs on the path from listener to real target node. So what about you?

mrmr1993: > We have appended target node's ancestor to `registrationElement` if `document.contains` returns `false`

Only if `document.parentNode` is `null`, which is never the case for elements in a shadow DOM. This is intentional, so that we don't cause re-renders or trigger `MutationObserver`s by detatching and re-attaching shadow dom elements.

gdh1995: Oh I see. Sorry for my carelessness and poor English - you've added a note in `hook_addEventListener.coffee:81`.

gdh1995:
A feature request: add `setTimeout( (-> ElementRegistrar.init()) , 1000)`.

Here's my problem:
* if the page is too large or loaded too slowly, then DOMContentLoaded may be delayed for seconds
* but I know most page contents have shown, and I want to click something
* However, `ElementRegistrar.init` is still not called, making a few buttons not hintable.

This "race condition" often occurs when a page requires some javascripts files from other websites - for example, in China `ajax.googleapis.com` can not be visited.

mrmr1993:
@gdh1995 are events ever reliably passed between contexts before `DOMContentLoaded`? In my experience, they weren't, but I only tried for short page load times.

If your tests show that, after 1000ms but before `DOMContentLoaded`, messages are guaranteed to be sent and received between contexts as normal, then I'm happy to add a timeout that forces the message dispatch earlier.

gdh1995:
Oh, sorry I didn't know this problem. Then I'll make a test in a few days.

mrmr1993:
Rebased.

ilikebits:
Sorry to necropost, but is there any progress on this? Quite a few of my favourite sites (Feedly, in particular) are unusable with Vimium without onclick handler detection.

gdh1995:
> are events ever reliably passed between contexts before `DOMContentLoaded`? In my experience, they weren't, but I only tried for short page load times.

@mrmr1993 Sorry to bother you, but I wonder what was your test page as mentioned above (if you can remember it). Thanks anyway!

 I've looked through source code of Chromium and found this: https://bugs.chromium.org/p/chromium/issues/detail?id=449744 . Although this is a most likely reason which may event dispatching crash, but there're also lots of other checks before running into JavaScript. As a result I'm still confusing about this problem.

vnikk:
Would like this feature too

philc:
@smblott-github  I assume this was never merged because of the complexity of this approach and the discussion churn on this PR?

I rarely visit sites which don't have link hints, so I don't feel a strong personal motivation for this. Although it's possible that I'm just blind to those sites now. However, @mrmr1993 worked hard on this, and there are a few issues (perhaps many issues) mentioning this, so I want to check in to see if there's recent demand or debate for this.

@gdh1995 have you been using this in vimium-c without issue?

gdh1995:
Yes it's because of its length - my Vimium C is even much longer (500+ lines) to ensure safety.

Here're some points which need to be fixed:
1. (most urgent) hook `Function::toString` like what we have done in #2941
2. (urgent) save a copy of public names like `setTimeout` and `ShadowRoot`
- 1. a host page may override such names with their own functions
  - 1. For example, `zone.js` hooks all common functions including `setTimeout`, `addEventListener` and `Function::toString`
  - 2. In my Vimium C, there's a bug report that a page overrides `Element` with its own object
- 2. `ShadowRoot` may not exist before Firefox 63
- 3. member functions like `Document::contains` should also be saved

Some which may be improved:
1. put elements in a list and send them in a batch, to avoid so many timers
- 1. on heavy pages like `source.chromium.org`, there may be hundreds of click listeners
- 2. may also add a small timeout before real sending
2. clean listeners like `VimiumRegistrationElementEvent` and `DOMContentLoaded` after using 
3. in case of `<form>` elements, because its `<input name=...>` can override (mask) its html properties and member functions
4. add a new feature to collect and send all elements with non-null `.onclick` properties but without `[onclick]` attributes.

philc:
This has been open a long time. I've mulled over it and, while impressive and potentially impactful, I think this is too hairy for me to merge and maintain at this point.

PR #1864, opened 2015-10-23
Enter visual mode

borismattijssen:
I had a problem with entering visual mode in this specific case:
If searched the page for a string. The string got highlighted. I pressed "v" to enter visual mode, but then the message "no usable selection, switching to caret mode" appeared. 
I tried to figure out what happened and it appeared that the html of that specific text was like this:
```
<p>
Lorum<br>
Ipsum<br>
Dolor
</p>
```
(where the selected text was "Ipsum").

In the original code of getSelectedElement it, for some reason, returned the element that is before the selected element (`<br>` in this case). Of course selecting just the `<br>` isn't useful so it switched to caret mode. 

I replaced the method's contents by just returning the anchorNode of the selected text. It now works perfectly!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ dist
 jscoverage.json
 tags
 .cake_task_cache
+.idea/
+*.iml
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -280,7 +280,6 @@ DomUtils =
     event.initTextEvent "textInput", true, true, null, text
     element.dispatchEvent event
 
-  # Adapted from: http://roysharon.com/blog/37.
   # This finds the element containing the selection focus.
   getElementWithFocus: (selection, backwards) ->
     selection.anchorNode;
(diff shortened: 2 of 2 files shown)


smblott-github:
Thanks, @borismattijssen.  I'll try it for a bit and get back to you.

smblott-github:
Hmm. Seems broken here: http://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html

I searched for the word "designed" then entered visual mode (`v`).  And it breaks.  Looks like you're returning a node where we're expecting an element.

mrmr1993:
@borismattijssen unfortunately your fix isn't a fix, it just hides the issue (and changes the code so it doesn't quite do what was intended). The issue is that `<br>` isn't considered a visible element by `getVisibleClientRect`, since it's 0-width, and that's the element which `getElementWithFocus` (seemingly correctly) chooses.

The obvious fix is to not use `getVisibleClientRect` on the closest preceding element, but instead to use [`Range.getBoundingClientRect`](https://developer.mozilla.org/en-US/docs/Web/API/Range/getBoundingClientRect). I'll submit a PR with this shortly.

mrmr1993:
Pushed #1865 as a fix.

borismattijssen:
Ah I see! Thanks for fixing it the right way ;)

PR #1865, opened 2015-10-23
Properly check whether the selection is visible in the current viewport

mrmr1993:
This fixes the issue menioned in PR #1864, by using `Range.getBoundingClientRect`.

For more accurate results, it might be worth instead doing

```coffee
visibleSelectionRects = (@selection.getRangeAt(0).getClientRects().map (rect) ->
  Rect.intersect rect, (Rect.create 0, 0, window.innerWidth,
    window.innerHeight)
).filter (rect) -> rect.height >= 0 and selectionRect.width >= 0
if visibleSelectionRects.length > 0
...
```

which will mean that Visual Mode will fall back to Caret Mode if and only if there is no selection visible in the viewport.

Edit: I've also pushed a branch ([`visual-mode-range-getClientRects`](https://github.com/mrmr1993/vimium/tree/visual-mode-range-getClientRects)) with a version containing this instead.

### DIFF
--- a/content_scripts/mode_visual_edit.coffee
+++ b/content_scripts/mode_visual_edit.coffee
@@ -484,8 +484,11 @@ class VisualMode extends Movement
         @extendByOneCharacter(forward) or @extendByOneCharacter backward
       else
         if @selection.type in [ "Caret", "Range" ]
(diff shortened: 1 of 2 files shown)


smblott-github:
Thanks, @mrmr1993.

Issue #636, opened 2012-09-02
Deprecate/remove the custom CSS for link hints feature


philc:
This is an advanced option which allows you to override the CSS styling of the linkhints.

I can't remember why we added it, but it's pretty unintuitive and currently semi-broken since we restyled the linkhints (see #626). What are some current compelling use cases for this feature?

smblott-github:
> I propose we disable this (but leave the code and preference in place) in the next-next release and see what kind of feedback we get.

We could leave everything in place, and just remove the entry from the options page...  And see what people say.  That way, those who are married to the idea will still be using their custom CSS.

(And we place a *heads-up* in the release notes.)

philc:
PR #3346 disabled this feature. I closed that with this comment:

For now, I don't want to deal with removing something existing people might be using, and I think hiding it for new users isn't a great option: we still have the code complexity -- we're just reducing the UI complexity.

A more robust way to evaluate deprecating features is to implement some basic analytics collection and have some stats/counters on how many people are using each feature. If only X% of the population is using a given feature, then we can more cleanly debate its deprecation.

PR #1884, opened 2015-11-09
CSS overrides for Vomnibar + optional overlay behind it

bezhermoso:
Nothing particularly game-changing -- just added a white translucent overlay behind the Vomnibar to increase visibility. Also softened the shadow up a bit (mimicks OS X Spotlight bar closely).

vimium omnibar focus

I understand that this change is purely aesthetic. Just putting it in here in case people are interested.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -44,5 +44,6 @@ Contributors:
   Shiyong Chen <billbill290@gmail.com> (github: UncleBill)
   Utkarsh Upadhyay <musically.ut@gmail.com) (github: musically-ut)
   Michael Salihi <admin@prestance-informatique.fr> (github: PrestanceDesign)
+  Bez Hermoso <bezalelhermoso@gmail.com) (github: bezhermoso)
 
 Feel free to add real names in addition to GitHub usernames.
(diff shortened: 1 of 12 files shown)


gdh1995:
I'm sorry but, I often use Vomnibar to query a English word on the page, and this PR does hide the background text.

bezhermoso:
Thanks for the feedback @gdh1995 

So I changed how I go about accomplishing this. Instead of modifying the existing CSS styles for the Vomnibar, I added a few settings under Advanced Options:

* `userDefinedVomnibarCss` - This will be loaded into `vomnibar.html` when it is first built. If people wish to change how the Vomnibar looks, they can override the CSS here.
* `overlayBehindVomnibar` - A checkbox that will toggle the overlay behind the Vomnibar. Defaults to `false`.
* `userDefinedVomnibarOverlayCss` - If people want to override the CSS style of the overlay, they can do it here.

smblott-github:
The added functionality here doesn't seem to warrant the extra noise on the options page.

Shout if you'd like to argue otherwise.

bezhermoso:
@smblott-github Even considering that its under advanced category?

We have `CSS for link hints` under advanced options, and this would fall under the same category. It is exactly the same functionality, just styling different aspects of Vimium.

smblott-github:
@bezhermoso.  There's actually a desire knocking around with some to get rid of the `CSS for link hints` option.  If it weren't there, it wouldn't be added.  And if weren't for the inconvenience to some passionate existing users, it probably would go.

bezhermoso:
@smblott-github 

Are you referring to this? https://github.com/philc/vimium/issues/636

It is not clear to me if the reason for removing the feature is because of:

A. the chance that user-defined overrides _may_ break because Vimium isn't promising non-breaking changes to the markup to UI elements, or
B. something more technical?

PR #1899, opened 2015-11-20
Fix indentation so that `openerTabId` takes effect

gaborluk:
An indentation mistake prevents every new tab opened via `TabOperations.openUrlInNewTab` from remembering its opener tab. This patch fixes that.


smblott-github:
Thanks, @gaborluk.  My bad, in this case.

gaborluk:
No problem, @smblott-github, thank you for the merge.

Issue #1906, opened 2015-12-02
exclude ctrl-alt key


mihkell:
I need to use ctr-alt key for some of the special characters in my language.
I have custome made this key binding. But Vimium doesn't let me use that key combination.(ctrlAlt + [)

how could I exclude it from all of the sites? What kind of exclusion pattern should I use?

code-hunger:
Going to insert mode should solve your problem, I think. Just press `i` and then enter your key-combination.

mihkell:
If focuse is on text area and I press 'i' it will type 'i'. If I'm out side of text area and press 'i' then how can I get focuse to text area without leaving insert mode?

smblott-github:
For the specific key combination you're talking about, this looks like a bug. `C-[` is intentionally `Escape`.  Unfortunately, the code isn't checking that `Alt` isn't depressed too.

PR #1908, opened 2015-12-04
For Ctrl-[, also require NOT Alt

smblott-github:
See #1906.

It is not obvious that this is in fact correct.  In particular, it's not clear how `Ctrl-[` should work on Mac keyboards.

Anyone with a Mac know what the right thing to do is?

### DIFF
--- a/lib/keyboard_utils.coffee
+++ b/lib/keyboard_utils.coffee
@@ -53,7 +53,7 @@ KeyboardUtils =
 
   isEscape: (event) ->
     # c-[ is mapped to ESC in Vim by default.
-    (event.keyCode == @keyCodes.ESC) || (event.ctrlKey && @getKeyChar(event) == '[')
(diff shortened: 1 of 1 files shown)


mihkell:
I use karabiner with setting 'Command_R to Option_L' and use it as I use altCtrl+[ on windows.

smblott-github:
@mrmr1993.

Is this the correct thing to do?

mihkell:
Only easy way I know. But I'm not familiar with OS X as a developer.

smblott-github:
@philc is considering doing a new release soon (1.54).  To be on the safe side, we should merge this only afterwards.

mrmr1993:
This is right, and arguably should exclude alt and shift as well.

We would already have all this for free if this was handled by the normal key handling code..

smblott-github:
1.54 is out, so we can try this now.

Mac users (that's you, @philc) please watch out for problems.

Issue #1919, opened 2016-01-01
Feature request: move to previously active tab


jdpopkin:
When I have a lot of tabs open and I check a pinned tab like Gmail, I often waste time finding my way back to the tab I was in previously, especially when it's one of many similarly-titled tabs. So it'd be nice if Vimium had a command to go back to whatever tab I was in before the current one - something like `:b #` in Vim.

I've got this working locally and will submit a pull request soon.

mrmr1993:
`T<enter>` should achieve this (tabs are listed in order of recency). There is some discussion about this in another issue, where the verdict was not to add a new dedicated command for now.

jdpopkin:
Fair enough. I'd like to have this command be remappable (without messing with my `T` binding) but it's not very important.

Would there be more interest in commands analogous to Slack's Cmd-[ and Cmd-], which navigate backwards and forwards in tab history (such that calling the back command twice wouldn't put you back on the original tab)?

smblott-github:
> Would there be more interest in commands analogous to Slack's Cmd-[ and Cmd-], which navigate backwards and forwards in tab history (such that calling the back command twice wouldn't put you back on the original tab)?

We get feature requests for everything, and for every variation on everything.  At some point we have to pick what is most generally of interest.  Sorry -- but  I'm not sure that ideas is above threshold.

PR #1920, opened 2016-01-01
Add go to previously visited tab command

jdpopkin:


jdpopkin:
See #1919 - I'd like to have a command that works on tabs the way Vim's `:b #` works on buffers.

smblott-github:
@jdpopkin.

Thanks for this.

There's been considerable discussion about whether to add this command previously.  The current state of that discussion is to stick with `T<Enter>` (which always picks the most recently-visited tab), and `T<Tab><Enter>` for the next one, and so on.

If were were to do this, then:
- We already maintain a list of tabs in recently-visited order ([here](https://github.com/philc/vimium/blob/master/background_scripts/completion.coffee#L354)), so we'd want to re-use that code.
- We would have to support `count` prefixes.

gdh1995:
#1984 have implement this needs  in another way, so this PR can be closed.

Issue #1921, opened 2016-01-01
CONTRIBUTING instructions issue: `cake test` won't work


jdpopkin:
CONTRIBUTING.md tells me to run `npm install path`, which installs a version of path that won't work in phantomjs. Instead, people should do what the CI build does (`npm install path@0.11`).

(The latest version of path assumes that `process` is defined globally, which isn't true outside node. I'll report that as a bug in path, but for now we probably want to update CONTRIBUTING.md to tell people to use the old version.)

PR #1922, opened 2016-01-01
Update dependencies in CONTRIBUTING

jdpopkin:


jdpopkin:
See #1921 - someone who installs as instructed won't be able to run `cake test` because the current version of path is broken.

smblott-github:
Thank you, @jdpopkin.

Issue #1924, opened 2016-01-06
Feature plan: A simple way to map custom key to exit global InsertMode


gdh1995:
I get an idea of using custom combined keys to exit global InsertMode.

Now Vimium supports extra options,so we may add two options to command `enterInsertMode`, and they will be like: `keyCode=27` and `modifiers=0`.
* `27` is used as the only one valid keyCode to exit InsertMode on handling a keydown event
* `0` represents controller keys's status: `altKey | (ctrlKey << 1) | (metaKey << 2) | (shiftKey << 3)`

That means, if I `map I enterInsertMode keyCode=27 modifiers=8`, then
* `<Esc>` won't make InsertMode exit and will be passed to the page's other listeners
* `<Shift-Esc>` will make InsertMode exit
* You know, `<Shift-Esc>` originally shows Task Manager of Chrome, so it does no harm to re-map it.

I suppose that we need just ONE method to exit global `InsertMode`, but if there has to multi ways, you may map several keys to `InsertMode` with different options.

gdh1995:
Closed since #1925 is rejected.

PR #1925, opened 2016-01-07
rework command enterInsertMode so that we can use custom keys to exit it

gdh1995:
This implement #1924, and I think it's simpler and easier than #1372


smblott-github:
@gdh1995... It's not clear what real problem this is solving.

gdh1995:
Some want to pass `<esc>` key to the page even on `InsertMode`, mainly to activate the page's own action, but now `enterInsertMode` will suppress a `<esc>`, exit and blur current `activeElement`.

The old idea is to let `InsertMode` deal with a special "passKey" list and use some customizable key-bindings to exit. With this PR, we just need to provide another key-binding for exiting, and then other keys will directly become passKeys.

smblott-github:
OK.  Thanks, @gdh1995.
I think we have this covered with `passNextKey`, now.

Issue #1911, opened 2015-12-13
Scroll gets stuck until page refresh


jamespullar:
Once you get vimium in this state, any scroll command (j,k,u,d, etc) does not release so scrolling never stops. Refreshing the page corrects the issue. Otherwise you end up limited to using the spacebar and arrow keys to page up and down.

OS: OSX 10.10.5
Browser: Chrome 47.0.2526.80

Steps to reproduce:
1. Open any website
2. Press f
3. Open a background tab using shift + link shortcut
4. Go to new tab using shift + k
5. Close tab (Doesn't matter how, but I use cmd + w)
6. Use any scroll command (j,k,u,d) and notice scrolling does not stop

jamespullar:
This is a new issue instead of adding to #1549 because the title is referencing an older version.
Just noticed this has the same steps to reproduce as #1174. Most likely the same as #1905.

smblott-github:
Thanks, @jamespullar.  Yes, this does look like a different issue.

jdpopkin:
I tried adding console.log statements in the keydown/keyup handlers in `scroller.coffee` and the keyup event isn't firing at all after I follow @jamespullar's steps to reproduce.

I've also noticed that pressing the shift key solves this problem.

WChargin:
Ooh, neat, thanks for fixing! I had gotten used to just pressing shift all the time :)

PR #1926, opened 2016-01-08
Fix endless scrolling problem (#1911)

jdpopkin:
Link hints mode would, if the user held shift and then left the tab,
wait endlessly for a shift keyup, preventing other keyups from taking
effect.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -309,6 +309,9 @@ class LinkHintsMode
             if event.keyCode == keyCode
               handlerStack.remove()
               @setOpenLinkMode previousMode if @isActive
+          blur: (event) =>
+            handlerStack.remove()
+            @setOpenLinkMode previousMode if @isActive
 
     else if event.keyCode in [ keyCodes.backspace, keyCodes.deleteKey ]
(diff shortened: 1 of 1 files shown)


smblott-github:
@jdpopkin... Thanks for this.  It looks like you've correctly identified the source of the problem.  However, there'll have to be more to the solution.  It seems you can trigger the same bug this way:

- go to any old page with links
- use `f`, then select a link while holding down `Shift`
- hit `j` ... oops!

In this case, there's no `blur` event.

There are two problems.  The first is that the handler is blocking propagation of `keyup` events.  That's easy enough to fix...


```Coffeescript
handlerStack.push
  keyup: (event) =>
    if event.keyCode == keyCode
      handlerStack.remove()
      @setOpenLinkMode previousMode if @isActive
    true # Continue bubbling the event.
```

However, the second (and real) problem is that -- for some reason -- we're not seeing the relevant `keyup` event, so we're not removing the handler.  I'll look into what's going on.

smblott-github:
@jdpopkin. I think #1930 is the way to go.

smblott-github:
Closing (in light of #1930).  Thanks, @jdpopkin.

Issue #1917, opened 2015-12-22
Proposal browser action icon


brianvanburken:
As @smblott-github asked in #1916 I've looked at the browser action icon. Things I looked at for improvement are:
- visual distinction between different states;
- familiarity for current users;
- visual improvement for users with color vision deficiency.

The current logo uses a color blue that is similar to the disabled icon in the grayscale. Users with a color vision deficiency will not see any difference between enabled, partial or disabled state (all look the same). So I looked at changing to different shape of icon to make the states more clear. Then I looked at other populair icons and noticed a monochrome trend. Both finding are used in the proposal.

Below you can see the proposed icon next to the current icon. From right to left the icons are as following:
- current icon;
- proposed icon;
- Chromecast icon;
- Pocket icon;
- 1Password icon.

**Enabled state**
![screen shot 2015-12-22 at 18 50 55](https://cloud.githubusercontent.com/assets/1044316/11962455/390e4170-a8e0-11e5-8eaf-8e156f7c1ac7.png)

**Partial state**
![screen shot 2015-12-22 at 18 51 22](https://cloud.githubusercontent.com/assets/1044316/11962497/8a470cb6-a8e0-11e5-8633-f07a5ff21c2d.png)

**Disabled state**
![screen shot 2015-12-22 at 18 51 29](https://cloud.githubusercontent.com/assets/1044316/11962502/8cf7304e-a8e0-11e5-81e1-f2a7c11924a2.png)

It is a bit of a radical change so I love to hear feedback about this before I would make a PR. So let me know what you think.

For the people interested in this, here you can download the PSD file: 
[vimium_proposal.psd.zip](https://github.com/philc/vimium/files/70059/vimium_proposal.psd.zip)

smblott-github:
Thanks, @brianvanburken.

Here are my thoughts...

- The status quo is very 1990s.
- Your designs are nice, @brianvanburken.  If the decision were the other way around -- so, to replace your designs with the status quo -- we'd never buy it.
- Looking **only** at the partially-disabled icon, it's not obvious what it means.  The meaning is only clear when you can see the other icons at the same time... which we never do when actually using Vimium.

@philc... What are your thoughts.

brianvanburken:
Thank you for your thoughts @smblott-github. I must honestly say I'm not a designer by far :) (rather love to code). So this was an attempt so far my skills go ;).

smblott-github:
Closing.
We went with #1932.

PR #1932, opened 2016-01-10
Update browser action icons

brianvanburken:
I took the original icon and style to create it almost exactly as it was and made it sharper :). 

This should look better on higher DPI devices. Also fixes the weird dark border for the partial browser action icon (or has this any intention?). This would also make the icons ready as `@2x` version for high DPI devices as Apple's Retina MacBook Pro.

Let me know if you like it. Here is the PSD file for anyone interested: [vimium.psd.zip](https://github.com/philc/vimium/files/85111/vimium.psd.zip)


smblott-github:
See also #1917.

Thanks @brianvanburken.  Both of these proposals are improvements.  #1917 is -- to my eye -- more modern, but this retains more of the established look.

We should go with one of them.

@philc... If you don't provide any input, I'll merge this one (because it's less disruptive).

smblott-github:
I tried these icons live, then when back to the status quo -- which felt like stepping back into 1980s graphics.

So, I'm merging this.  Thanks, @brianvanburken.

@philc... You've seen this, right?

brianvanburken:
Great to hear! Thanks alot! 😄

PR #1938, opened 2016-01-21
add chinese pattern

qwIvan:


smblott-github:
Thanks, @qwIvan.

I'm not sure we should do this.  There are a **lot** of languages around, and the default list could become pretty long.

You know that you can change these on the options page, yes?

snapshot

Issue #1312, opened 2014-12-14
Proposal: implement a Vimium-specific right-click menu for links


philc:
Hey everyone, what do you think about a Vimium-specific right-click UI as a means to gracefully scale the number of commands that one may want to perform on a link?

You would "right click" on a link using link hints, and then get a popup, where each action has a reasonable hotkey:

![image](https://cloud.githubusercontent.com/assets/8598/5428924/242b3256-838c-11e4-9b53-a46b33595094.png)

Comments

* We can handle lots of commands this way without introducing new top-level commands.
* Since you don't need to bind any of these sub-commands to specific keys, it means you'll have to remember fewer keybindings to be able to use these commands. This is a nice property since most of these would see relatively rare usage compared to other top level commands.
* This naturally mirrors the interaction model Chrome uses for this use case.
* The UI elements are based on Chrome's right click menu (at least what they look like in OSX). The link which is being operated on is highlighted as it is when you right click using the mouse.
* The specific use case of hovering probably requires more thought, since you'll want to be able to unhover as well. Perhaps this is how you hover, and then hitting esc at any time unhovers.

Related issues: #484 and likely many others.

dragon788:
OMG YES PLEASE! I've actually been thinking about this since I frequently save
documents with Pocket or want to send them with Pushbullet, and it is
difficult to activate and send those normally via the keyboard (unless using
the default mapped Ctrl+Shift+X for a single endpoint).

Rajwin:
Yes Please, I cant believe there is no function for this

wcpines:
I would pay for this

smblott-github:
Move discussion to #1948.

PR #1947, opened 2016-01-27
Copy link text support

svgsponer:
Since I often wished to copy the link text I implemented it maybe it is of interest for other users too. I'm not too familiar with coffee script so if there are any flaws please let me know. I mainly adopted the `CopyLinkUrl` procedure. Besides existence of the text it additionally checks if the trimmed text is nonempty.

At the moment the command is bind to `yl`

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -91,6 +91,7 @@ Commands =
       "toggleViewSource",
       "copyCurrentUrl",
       "LinkHints.activateModeToCopyLinkUrl",
+      "LinkHints.activateModeToCopyLinkText",
       "openCopiedUrlInCurrentTab",
       "openCopiedUrlInNewTab",
       "goUp",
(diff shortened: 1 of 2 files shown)


smblott-github:
Thanks, @svgsponer.

Two comments:

- You *could* search for the link text then yank it with visual mode, just like any other text
- Personally, I wouldn't be too keen on adding more link-hint commands; I'd prefer to pursue the idea in #1312.

svgsponer:
Thanks, @smblott-github.

Yes this approach (search, yank) is the method I used until now and a separate command might be a too specific need indeed. 
#1312 looks very interesting is somebody already working in this direction?

smblott-github:
@svgsponer... Actually, in response to your earlier issue, I started looking at this...

snapshot

smblott-github:
Closing.

This is probably too niche to warrant a new command.  It's in #1948, though.

klammbueddel:
Comming from VimFX (which was killed by FF Quantum) I'd like to open this discussion again. Yanking text by hints was one of the most valuable feature for me. IMHO the "search" and "yank" suggestion is not an alternative, since it consumes to many key strokes.

In VimFX it was mapped to "yv". The hints where shown and selecting one would yank the content of the text node. I'd appreciate if we could get this feature in vimium, which for now is the recommended alternative to VimFX.

#2601 still needs to be solved then...

PR #1948, opened 2016-01-28
Link hints context-menu mode

### LINKED ISSUE #1312: Proposal: implement a Vimium-specific right-click menu for links

smblott-github:
This implements @philc's idea from #1312, a link-hints mode which activates a menu.

We get a lot of a requests for new link-hints modes.  Each new mode requires a new command, and we arguably already have too many commands.  The idea here is that - for less common modes - we can add a new menu operation instead.

Details:
- Default binding: `gm`.
- This PR also adds a hover option (borrowed from @mrmr1993's #1032).
- And it adds a "yank-link-text" option (borrowed from @svgsponer's #1947).

Currently, it looks like this...

snapshot

To do:
- internationalisation?
- with queue

Fixes #1312.
Replaces #1032.
Replaces #1947.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -106,6 +106,7 @@ Commands =
       "LinkHints.activateModeWithQueue",
       "LinkHints.activateModeToDownloadLink",
       "LinkHints.activateModeToOpenIncognito",
+      "LinkHints.activateModeForMenu",
       "goPrevious",
       "goNext",
       "nextFrame",
@@ -199,6 +200,7 @@ defaultKeyMappings =
   "f":     "LinkHints.activateMode"
   "F":     "LinkHints.activateModeToOpenInNewTab"
   "<a-f>": "LinkHints.activateModeWithQueue"
+  "gm": "LinkHints.activateModeForMenu"
 
   "/": "enterFindMode"
   "n": "performFind"
@@ -290,6 +292,7 @@ commandDescriptions =
   "LinkHints.activateModeWithQueue": ["Open multiple links in a new tab", { noRepeat: true }]
   "LinkHints.activateModeToOpenIncognito": ["Open a link in incognito window", { noRepeat: true }]
   "LinkHints.activateModeToDownloadLink": ["Download link url", { noRepeat: true }]
+  "LinkHints.activateModeForMenu": ["Open a link's context menu", { noRepeat: true }]
 
   enterFindMode: ["Enter find mode", { noRepeat: true }]
   performFind: ["Cycle forward to the next find match"]
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -187,7 +187,7 @@ getCompletionKeysRequest = (request, keysToCheck = "") ->
   completionKeys: generateCompletionKeys(keysToCheck)
   validFirstKeys: validFirstKeys
 
-openUrlInNewWindow = (request, callback = (->)) ->
+openUrlInNewWindow = (request, sender, callback = (->)) ->
   chrome.tabs.query { active: true, currentWindow: true }, (tabs) ->
     chrome.windows.create { url: request.url, incognito: tabs[0].incognito }, callback
 
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -187,6 +187,10 @@ getCompletionKeysRequest = (request, keysToCheck = "") ->
   completionKeys: generateCompletionKeys(keysToCheck)
   validFirstKeys: validFirstKeys
 
+openUrlInNewWindow = (request, callback = (->)) ->
+  chrome.tabs.query { active: true, currentWindow: true }, (tabs) ->
+    chrome.windows.create { url: request.url, incognito: tabs[0].incognito }, callback
+
 TabOperations =
   # Opens the url in the current tab.
   openUrlInCurrentTab: (request, callback = (->)) ->
@@ -653,6 +657,7 @@ sendRequestHandlers =
   openUrlInNewTab: TabOperations.openUrlInNewTab
   openUrlInIncognito: TabOperations.openUrlInIncognito
   openUrlInCurrentTab: TabOperations.openUrlInCurrentTab
+  openUrlInNewWindow: openUrlInNewWindow
   openOptionsPageInNewTab: openOptionsPageInNewTab
   registerFrame: registerFrame
   unregisterFrame: unregisterFrame
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -710,7 +710,6 @@ class ContextMenuMode extends Mode
         metaKey: KeyboardUtils.platform == "Mac", ctrlKey: KeyboardUtils.platform != "Mac"
 
     openNewWindow = text: "Open in new window", key: "w", requireLink: true, handler: =>
-      # Not yet implemented.
       chrome.runtime.sendMessage handler: 'openUrlInNewWindow', url: link.href
 
     openIncognito = text: "Open incognito", key: "i", requireLink: true, handler: =>
@@ -752,19 +751,21 @@ class ContextMenuMode extends Mode
     menuEntries =
       for group in [openMenuEntries, yankMenuEntries, otherMenuEntries]
         for entry in group
-          if entry.requireLink and not link.href?
-            # If there's no href, then don't include menu entries which require one.
-            ""
-          else if entry.requireLinkText and not link.text?.trim().length
-            # If there's no link text, then don't include menu entries which require it.
-            ""
-          else
+          # If there's no href, then don't include menu entries which require one.
+          continue if entry.requireLink and not link.href?
+          # If there's no link text, then don't include menu entries which require it.
+          continue if entry.requireLinkText and not link.text?.trim().length
+          entry
+
+    menuGroupHTMLs =
+      for group in menuEntries
+        for entry in group
             "<li><span class=\"vimiumHintContextMenuItem\">
               <span class=\"vimiumHintContextMenuKey\"n>#{entry.key}</span>
               <span class=\"vimiumHintContextMenuText\">#{entry.text}</span>
              </span></li>"
 
-    menuGroupHTMLs = (group.join "" for group in menuEntries)
+    menuGroupHTMLs = (group.join "" for group in menuGroupHTMLs)
     menuHTML = menuGroupHTMLs.join "<li><hr class=\"vimiumHintContextMenuHR\"/></li>"
     menuElement.innerHTML = "<ul>#{menuHTML}</ul>"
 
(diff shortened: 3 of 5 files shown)


mrmr1993:
Nice work at implementing #1312. I'm quite disappointed though; I was expecting to like this much more. My initial thoughts:
* the order in which you make decisions with the menu &mdash; choosing the object before the action &mdash; feels very un-vim-y. It might just be me, but I feel as though I've lost power, asking Vimium what I *can* do with the link rather than telling it what *to* do.
* this 'solves' the 'problem' of having too many commands by
  - still adding them
  - giving them restricted bindings
  - hiding them and their descriptions away in a menu (that itself is hard to find in the help dialog).

After using this, I feel like we're trying to solve a documentation problem with program design. Sorry Steve, but I think we should bin this and sort out #1269 (command zoo) another way. I'm going to look at implementing #1280 again over the next 2-3 days.

nfischer:
I'm ok with having something like "open in incognito" in a sort of right-click menu (as long as I can "right click" with link hints of course). These all seem obscure enough that it's ok to put them in a menu (as long as f and F still work as before). Just my 2 cents

philc:
I have some comments, I'll chime in later.

smblott-github:
> These all seem obscure enough that it's ok to put them in a menu

Yes, @nfischer.  This is a win for me too.  I use the lesser known link-hint modes so infrequently that I can never remember the binding, so it's quicker to reach for the mouse.

philc:
* "Right clicking" is an important action in web browsing that everyone is already familiar with. When I think
  about download a link's content or copying its URL, the first thing I think about is right clicking on that
  link.
* The actions taken via the right click menu are rare enough that the keybindings won't be remembered by most
  people. I still use the mouse to download links even though we have a command for it (which feels like a
  fail). Presenting a menu will really help with this problem. Now you need to remember just one command
  rather than many.
* You can still efficiently invoke these commands if you've memorized the keybinding: f; type the hint;
  immediately type the action you want.
* While you could argue that most Vim commands don't ask you to select the object before specifying the
  command, that is the way visual mode interactions work. @mrmr1993, regarding your feeling of "loss of
  power", one way to think of this is that it's progressive disclosure of (somewhat rare) actions which are
  available on a link. Kind of like listing a directory before selecting the file to open. Necessary and
  helpful.

Let's debate whether we make available standalone commands for some of these actions like `copy link address`. What do you guys think? I can see the value in keeping separate bindable commands if you perform one of the right click actions very often, like downloading files.

Some feedback on the UI:

* I think the aesthetics of this dialog should reflect the mock from #1312. In particular, the hints should be
higher contrast to the captions, e.g. bold and bluish, similar to the help dialog.
* The hrs should have less contrast
* Use the same margin above and below the menu's content as you use for the sides.
* (I'm happy to tweak the styles once merged)
* I haven't played with this yet, but just to preemptively QA this, the menu should be positioned fully
  on-screen even if the link is near the bottom of the view port

nfischer:
> Let's debate whether we make available standalone commands for some of these actions like copy link address.

I think each of those commands are obscure enough that, if they don't already have a binding, then a three-key binding (`gmt`, `gmd`, etc.) is probably fast enough. Anyone who has memorized the binding probably won't mind seeing the menu pop up on screen for the split second between pressing `m` and pressing `d`. If it turns out to be a major concern, you could expose the command (i.e. download link) without providing a default binding, letting the user decide without cluttering the docs by default.

I think the binding would be better if it was related to `f`. I'm guessing `gm` stands for something like "go to menu," which is intuitive enough, but I think it'd be better to press something like modifier-`f` and type in a link hint. I just can't think which modifier is left remaining.

mrmr1993:
I disagree on a fundamental level @philc. Screen real estate already has semantic meaning, and a context menu is a tool to repurpose it to provide more options beyond the descriptive power of saying "that, there". You need a context menu to change actions into visual action 'objects', otherwise they wouldn't be accessible from the mouse.

The keyboard is (for us at least) fundamentally different: the meaning of the keys change based on some not-necessarily-visual state. Briefly,
* the power in *expressing* any of these actions is unchanged by using a context menu; hide the menu and the keybindings are as accessible or inaccessible as any other.
* the menu represents a documentation hack, insofar as it
  - hides some commands separately from the others, so that frontline documentation is less dense
  - adds no extra contextual or other information to the commands by putting them there
  - preventd the documentation from being accessible at the same point as all other command documentation.

To address your comments,

> * "Right clicking" is an important action in web browsing that everyone is already familiar with. When I think about downloading a link's content or copying its URL, the first thing that comes to mind is right clicking on that link.

To have a dedicated visual button is a waste of space: visually a context menu is better. But for Vimium, it saves no space, apart from in documentation.

> * The actions taken via the right click menu are rare enough that the keybindings won't be remembered by most people. I still use the mouse to download links even though we have a command for it (which feels like a fail). Presenting a menu will really help with this problem. Now you need to remember just one command rather than many.

This is also a documentation issue. Can you remember the command to scroll all the way to the right/left? I can't. But I might still need it on occasion.

Actually, it takes long enough to identify that from cold in the help dialog that a command search would be a large improvement. Hell, enter from that search could launch the command, so a moment's forgetfulness barely slows you at all. I can't see that hiding it in a special scroll context menu &mdash; that I also need to remember the command for &mdash; is going to help anything except the number of commands the help dialog opens with.

> * You can still efficiently invoke these commands if you've memorized the keybinding: gm; type the hint; immediately type the action you want.

There's two context changes here: command, object, command. Which seems like a waste of mental effort when it could be just command, object. Otherwise, I agree that these are just mappings in disguise.

> * While you could argue that most Vim commands don't ask you to select the object before specifying the command, that is the way visual mode interactions work.

Agreed. Visual mode is better for some harder-to-describe (in Vim movement terms) cases. Are any of these actions hard to describe in simple, atomic terms?

> @mrmr1993, regarding your feeling of "loss of power", one way to think of this is that it's progressive disclosure of (somewhat rare) actions which are available on a link. Kind of like listing a directory before selecting the file to open. Necessary and helpful.

I agree, progressive disclosure (or interactive documentation, as you might call it) is necessary and helpful. But this is relegation with interactive documentation, and I'm not sure that we need to relegate these commands, nor that the interactive documentation to be so restrictive and restricted.

Please lets not fuck up commands over a documentation issue.

nfischer:
> Actually, it takes long enough to identify that from cold in the help dialog that a command search would be a large improvement. Hell, enter from that search could launch the command, so a moment's forgetfulness barely slows you at all.

^ That's actually a pretty neat idea. It might be worth opening that as an additional PR

mrmr1993:
> It might be worth opening that as an additional PR

Thanks @nfischer. I want to see if @smblott-github and @philc  approve first, otherwise I'll have to fully commit to forking to make that PR worthwhile.

nfischer:
@mrmr1993 I went ahead and wrote the searchable help menu up as a separate issue, since I don't think it necessarily solves the same problem, but I think it's worth consideration regardless of whether this gets merged or not.

smblott-github:
Update:

- Better styling.
- Match layout of #1312.
- Menu should (?) now always be in viewport (although the position calculation could be more subtle).

snapshot

To do:

- There are still some styling issues on some pages.  (E.g. For some reason, the separators don't show on github pages.)  My CSS foo is weak.

philc:
I'm going to solicit some more external input on this feature early this week.

smblott-github:
Update...

I have used this link-hints menu precisely 0 times in anger.  This feature may not be as useful as first appeared.

smblott-github:
Unless a champion emerges for this I propose to close it soon on the basis that it's not as useful as we thought.

(@mrmr1993... One thing that I like about this is the behaviour of "hover", where `Escape` unhovers.  That might be worth remembering.)

nfischer:
I haven't tried this out yet (but I could over the next week). To me it would be useful if for no other reason than hover and unhover. One request I have though would be some sort of "click and immediately unhover" option. This makes sense when clicking the "make fullscreen" button for youtube videos, for example. It seems like this could often be the default behavior (although I suppose sometimes a "click and continue to hover" option is preferable).

smblott-github:
Closing.
@mrmr1993 is strongly opposed to this (see above).  I'm mildly on the *no* side (because it's not as useful in practice as anticipated).

We can reopen this if a champion emerges.

dragon788:
I'd still love to see something like this, if only to open links into another instance of Chrome logged in as a different user. On my work laptop this has become something of a daily occurrence where I am in my "work window" chugging along and then I see something I'd like to read later, I then right click and open it into my personal account's open session so that my Chrome tabs/history sync and I can read the page later. Using Chrome native features like this and not having to do a bunch of extra extension syncing (Pocket/Pushbullet/XMarks/etc) would be awesome. Maybe I just need to examine the source and see if there is a command already defined for this, though I'm using currently 3 different users in 3 different Chrome sessions, so it would have to have a way to present an option to select the user.

Anak2016:
I agree that right click on a link feature is not as useful as it seem. Though, I feel the need to point out that right "clicking on the page" features would be extremely helpful when using along with other plugins. 
As of now, I have the following option when right clicking on a page. 
1. rename tab. really useful when doing tab search
2. highligther. This is really really useful when reading on a random page that you frequently visit.
3. Web Clipper. such as evernote. it is extremely useful in my case when i am doing research because I organized my notes within evernote.
4. using it with "copy page url" plugin. it allows you to copy url in various way. copy url and text, or copy text and embed url in it. imagine using it with EVernote mention above. This is a killer in my book. 

This is just plugins that i have. but i can imagine that It will be really useful in some other cases as well.

mcp292:
The only reason I want right click is to copy ~~text~~ emails. For example, an email hyperlink gives the option to copy the email on the context menu. That is pretty much my main use case as it's hard to argue for copying text, because there's no way to highlight with just the keyboard and vimium.

espoelstra:
> The only reason I want right click is to copy ~~text~~ emails. For example, an email hyperlink gives the option to copy the email on the context menu. That is pretty much my main use case as it's hard to argue for copying text, because there's no way to highlight with just the keyboard and vimium. 

Have you met visual or caret mode? You can totally highlight text using just the keyboard using the normal vim shortcuts to switch which direction to expand your selection.

gdh1995:
@mcp292 Maybe the Vimium command named `LinkHints.activateModeToCopyLinkUrl` can help.

mcp292:
> Have you met visual or caret mode? You can totally highlight text using just the keyboard using the normal vim shortcuts to switch which direction to expand your selection.

No actually! Will have to check it out!

mcp292:
@gdh1995 I don't know if that would work since I don't want to copy the `mailto:` portion of the url.

gdh1995:
Vimium will auto trim a `mailto:` prefix.

mcp292:
@gdh1995 Ah shit! Nice!

mcp292:
@gdh1995 It doesn't auto-trim `http://`. Try the `yf` on the email links on this site:
https://sendcutsend.com/design-partners/

PR #1871, opened 2015-10-29
Show as many short hint markers as possible

mrmr1993:
The previous formula was sub-optimal; this updates the formula to get the maximum number of short hints that we can.

Explanation from the commit messge:

```
Let linkCount be the number of links,
    characterCount be the number of characters.

Let digitsNeeded be such that
  characterCount^digitsNeeded >= linkCount > characterCount^(digitsNeeded - 1),
  as in the code.

Let longLen = characterCount^(digitsNeeded - 1),
    shortLen = characterCount^(digitsNeeded - 2),
  such that they represent the number of hints a letter can prefix,
  depending on whether they are used for short or long hint strings.

Then the maximum number of links we can show hints for, when shortCount
is the number of short hint prefixes, is

maxCount = shortCount * shortLen + (hintCharacters - shortCount) * longLen
         = shortCount * (shortLen - longLen) + hintCharacters^digitsNeeded

We want that linkCount <= maxCount, so

linkCount <= shortCount * (shortLen - longLen) + hintCharacters^digitsNeeded

shortCount * (longLen - shortLen) <= hintCharacters^digitsNeeded - linkCount
shortCount * shortLen * (characterCount - 1) <= hintCharacters^digitsNeeded - linkCount
shortCount * shortLen <= (hintCharacters^digitsNeeded - linkCount) / (characterCount - 1)

Since (shortCount * shortLen) is the largest number of short hints we
can show, we get that shortHintCount can be as in the code.
```

PR #1949, opened 2016-01-28
Simplify hint string generation

smblott-github:
This is considerably easier to understand than the existing hint-string generation code (in my opinion).

Fixes #1871 (probably).

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -440,11 +440,6 @@ class AlphabetHints
   # may be of different lengths.
   #
   hintStrings: (linkCount) ->
-    return [] if linkCount == 0
-
-    # In the following:
-    #   - at no point is any hint in hints a prefix of any other hint
-    #   - the shorter hints are always at the start of the list
     hints = [""]
     offset = 0
     while hints.length - offset < linkCount or hints.length == 1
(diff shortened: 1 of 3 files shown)


philc:
Nice simplification; I appreciate the brevity. "+12 −36" is always a solid change.

content_scripts/link_hints.coffee
```diff
@@ -440,39 +440,16 @@ class AlphabetHints
...
+    hints = [""]
+    offset = 0
+    while hints.length - offset < linkCount or hints.length == 1
+      hint = hints[offset++]
+      hints.push ch + hint for ch in @linkHintCharacters
+    hints = hints[offset...offset+linkCount]
+
+    # Shuffle the hints so that they're scattered; hints starting with the same character are spread evenly
+    # throughout the array.
+    return hints.sort().map (str) -> str.reverse()
```
mrmr1993: @smblott-github you haven't reversed the strings before returning them; won't this make them consecutive (rather than shuffled)?

Arguably, since the hints are already sorted, shouldn't we be able to reverse them without any re-ordering and achieve the same shuffling effect?

smblott-github: @mrmr1993 Note:

    hints.push ch + hint

We build the hints *already* reversed.  This ensures that the shorter hints have (initially) different prefixes.  So when sorted, they get spread roughly evenly throughout the list.  Then, when we reverse them again, the hint prefixes get spread roughly evenly too.

smblott-github:
If we're simplifying hint generation for alphabet hints, then we can substantially simplify the same code for filtered hints too... d6a7ea0094f8071f1785539ccd167ed371bd3bb9.

PR #1950, opened 2016-01-29
For filtered hints, wait for <Enter> before activating link

smblott-github:
This affects filtered hints only.

There's a usability issue with filtered hints which has bothered me for some time.

The link is activated:

- when the user types the hint text
- when the user hits `<Enter>`, or `<Tab><Enter>`
- or when only one link remains

The latter case is problematic.  We add a delay in case the user is still typing.  But that delay could be too long, or too short.  Who knows.  In practice, I find myself typing a couple of characters then hitting `<Enter>` or `<Tab><Enter>` just to *avoid* having the link automatically activated.

The idea here is to get rid of the race condition in the latter case.  If a hint is triggered because the user typed the link text, then:

- highlight the link (using the same blue "flash" as we use currently),
- but wait until the user types `<Enter>` before actually activating the link (consuming any other characters typed).

With this approach, typing a few characters from the link then hitting `<Enter>` nearly always gets it right.

When the link is highlighted, it looks like this...

snapshot

Because this would be a change to the established behaviour, it would probably (and unfortunately) require a new option.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -367,7 +367,7 @@ class LinkHintsMode
     if linksMatched.length == 0
       @deactivateMode()
     else if linksMatched.length == 1
-      @activateLink linksMatched[0], keyResult.delay ? 0
+      @activateLink linksMatched[0], keyResult.delay ? 0, keyResult.waitForEnter and Settings.get "waitForEnterForFilteredHints"
     else
       @hideMarker marker for marker in hintMarkers
       @showMarker matched, @markerMatcher.hintKeystrokeQueue.length for matched in linksMatched
@@ -375,7 +375,7 @@ class LinkHintsMode
   #
   # When only one link hint remains, this function activates it in the appropriate way.
   #
-  activateLink: (matchedLink, delay = 0) ->
+  activateLink: (matchedLink, delay = 0, waitForEnter = false) ->
     clickEl = matchedLink.clickableItem
     if (DomUtils.isSelectable(clickEl))
       DomUtils.simulateSelect(clickEl)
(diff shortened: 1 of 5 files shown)


smblott-github:
Update...

This does indeed appear to be a better UX than the status quo. *If you're typing a hint's text, then you will always be hitting `Enter` too*.

The major concern is that it requires a new option (not yet implemented).  The option would only be visible it *Filtered Hints* is selected.  So that may not be too bad.

smblott-github:
Another update.  Two things...

- I tried going back to life before this way of activating (filtered) link hints and it was painful.  It seems I (for one) *do* find this approach considerably more usable.

- As noted previously, this changes the established behaviour; therefore, unless we're willing to force it through (bad idea), we need a new option.  The option (which defaults to `false`) looks as follows, and is hidden unless the *filetered hints* option is ticked.  So, many users will never even see it.

snapshot

philc:
If we need to add a new option for this and it's a better UX, it would be
better to default this to "true" for new installs and "false" for existing
users who are upgrading.

For other folks using filtered hints, it would be great to hear your
impressions of this UX.

danielyli:
> For other folks using filtered hints, it would be great to hear your
impressions of this UX.

I'm the reporter for #1988.  I think having the ability to press `Enter` is a very welcome feature.  The checkbox in the Settings window seems reasonable.

Regarding any possible objections to having another key press: in Vim, we're very accustomed to pressing `Enter` after commands already, so it isn't much of an issue.

smblott-github:
> If we need to add a new option for this and it's a better UX, it would be better to default this to "true" for new installs and "false" for existing users who are upgrading.

Done: 92be10810a273cf003a694c7c56423bd2ccb53fe.

Update... This is (still) a considerably superior UX `</opinion>`.

smblott-github:
Merging.
@philc... This is a big usability improvement and - modulo any bugs - is definitely the way to go.

PR #1952, opened 2016-01-30
Do not exit link-hints mode on scroll.

### LINKED ISSUE #1918: F/f don't work in this page http://www.cs.sjtu.edu.cn/en/ .
http://www.cs.sjtu.edu.cn/en/ .

smblott-github:
Fixes #1918.

See there for explanation.


gdh1995:
Does the title of this PR have any typo? I think it's (Do not) "exit", not "enter".

mrmr1993:
This also fixes a similar issue on messenger.com (injecting our hints causes a scroll). LGTM.

PR #1954, opened 2016-01-30
Exit link hints if there are no links.

### LINKED ISSUE #1395: No visual feedback if no link-hints on page
Find a page which has no links and nothing clickable ([here's one](http://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html)).  Type `f`.  We go into link-hints mode, but without any link hints.  And the user gets no visual feedback as to what's going on.  Only after the next (useless) keystroke do we drop back into normal mode.

What's the right thing to do?

Perhaps:
- Flash the HUD: "No links found".
- Drop immediately back into normal mode.

smblott-github:
If there are no links on a page, then link-hints mode should exit immediately.

Fixes #1395.


mrmr1993:
LGTM.

PR #1958, opened 2016-01-31
Add a logging page...

smblott-github:
This tweaks @mrmr1993's logging ideas discussed in #1629:

- Do not generate log entries from the logging page itself.
- Use the logger for *all* logging (including from `commands.coffee`, and from `bgLog`).

Edit... also:
- Echo the logging URL to the console on start up.

@mrmr1993's original idea is 91fb337.

Fixes #1629.


mrmr1993:
LGTM!

PR #1960, opened 2016-01-31
Better scoring for filtered hints.

smblott-github:
This affects filtered hints only.

On some sites (e.g. Facebook), there are a series of posts, and each post contains some text and a link.  And both the post itself and the link are clickable.

In this case, filtered hints often pick out both elements as clickable and, usually, prioritises the post over the link.  Which is usually wrong.

This PR takes the length of the link text into account when scoring links, prioritising shorter link texts, much as we do when scoring completion entries.


smblott-github:
Merging directly.
This is probably not controversial. And -- if there are any issues -- we'll spot them quite quickly.

PR #1961, opened 2016-01-31
Wholly remove edit-mode code.

smblott-github:
I am now of the opinion that we should not do this within Vimium, as:

- better solutions exist externally, and
- it's better to not have to maintain this.

Will merge immediately.  Please watch out for any unexpected behaviour from visual mode.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -98,7 +98,6 @@ Commands =
       "enterInsertMode",
       "enterVisualMode",
       "enterVisualLineMode",
-      # "enterEditMode",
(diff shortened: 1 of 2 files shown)


gdh1995:
> we don't have to maintain this.

I agree this. Here I just wonder what's the "better solutions" - for my personal interest.

smblott-github:
[This](https://chrome.google.com/webstore/detail/wasavi/dgogifpkoilgiofhhhodbodcfgomelhe) works for me.

gdh1995:
Thanks for your advice. It looks cool!

PR #1962, opened 2016-01-31
Fix visual indicator for `focusInput()`.

### LINKED ISSUE #1257: focusInput command overlays incorrect for off-screen inputs

smblott-github:
Currently, if an input is only partially in the view port, then the page may scroll when it is focused and the overlays for `focusInput()` are wonky.  See #1257.

Here, we draw the overlay around the *entire* input, instead of just around the visible part.  Being partially visible therefore is no longer relevant.

@mrmr1993... Does this look correct to you?

Fixes #1257.
Closes #1258.

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -116,6 +116,18 @@ DomUtils =
     null
 
   #
+  # Returns the smallest rectangle which encloses element.  Assumes that there is at least one rectangle.
+  #
+  getBoundingClientRect: (element) ->
+    [clientRect, rects...] = element.getClientRects()
+    for rect in rects
+      clientRect.top = Math.min clientRect.top, rect.top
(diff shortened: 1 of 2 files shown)


lib/dom_utils.coffee
```diff
@@ -116,6 +116,18 @@ DomUtils =
     null
 
   #
+  # Returns the smallest rectangle which encloses element.  Assumes that there is at least one rectangle.
```
mrmr1993: Isn't this exactly what `element.getBoundingClientRects` does already?

smblott-github: Thanks, @mrmr1993. I didn't know that existed.

PR #1963, opened 2016-01-31
Link hints with count

### LINKED ISSUE #1289: [Feature request] Use number prefix for link hints to activate that number of links

smblott-github:
Use a count with link hints.
    
Pass a count to link-hint commands, and the link-hint mode is started that many times (except "with queue").

This resolves the same issue as #1291.  However, all of the `count` handling is outside of the `LinkHints` class, which simplifies things somewhat.

Fixes #1289.
Closes #1291.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -270,7 +270,7 @@ commandDescriptions =
   toggleViewSource: ["View page source", { noRepeat: true }]
 
   copyCurrentUrl: ["Copy the current URL to the clipboard", { noRepeat: true }]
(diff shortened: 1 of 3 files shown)


mrmr1993:
LGTM.

PR #1956, opened 2016-01-30
Implemented closeWindow command with default binding :q

Reilithion:
Fixes #1872. This turned out to be easy.

PR #1965, opened 2016-02-01
Implement scrolling to percent (the '%' command).

leoj3n:
Idea stolen from Vrome:

`[count] %         Scroll to percent of the page`

https://github.com/jinzhu/vrome/blob/bf6340b9f9b6695aa9211180237aea02f04619b5/coffee/frontend/modules/scroll.coffee#L145

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Navigating the current page:
     j       scroll down
     k       scroll up
     l       scroll right
+    %       scroll to percent
     gg      scroll to top of the page
     G       scroll to bottom of the page
     d       scroll down half a page
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -87,6 +87,7 @@ Commands =
       "scrollPageUp",
       "scrollFullPageUp",
       "scrollFullPageDown",
+      "scrollToPercent",
       "reload",
       "toggleViewSource",
       "copyCurrentUrl",
@@ -147,6 +148,7 @@ Commands =
   advancedCommands: [
     "scrollToLeft",
     "scrollToRight",
+    "scrollPercent",
     "moveTabToNewWindow",
     "goUp",
     "goToRoot",
@@ -178,6 +180,7 @@ defaultKeyMappings =
   "zL": "scrollToRight"
   "<c-e>": "scrollDown"
   "<c-y>": "scrollUp"
+  "%": "scrollToPercent"
 
   "d": "scrollPageDown"
   "u": "scrollPageUp"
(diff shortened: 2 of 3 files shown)


smblott-github:
Thanks, @leoj3n.

As it happens, just yesterday we merged #1023, which does something very similar.  The difference being that it uses a multiple of `scrollStepSize` instead of a percentage.  It's probably not necessary to support both a percentage and an absolute amount.

leoj3n:
@smblott-github I respectfully disagree! It would be useful to have both, because I cannot possibly guess how many steps it would require to reach 50%.

smblott-github:
@leoj3n.  We get lots of proposals for new commands (e.g. #1956, where we're having pretty much the same discussion).  We could easily have hundreds and hundreds of commands.  However, nobody would know them all, and nobody would be able to remember the shortcuts.  So we have to pick a set of commands that we think serve the broad Vimium user base well.  In this case, the new `10gg` meets a very similar need to your proposal, is easily discoverable, and does not require a new command.  Having `10gg`, the need for your proposal is somewhat diminished.

(Notwithstanding all of that, we do appreciate your input.)

leoj3n:
While I understand your concern with adding new commands, I feel you're conflating issues and not being reasonable with respect to this very useful feature! So I will give my input here...

It is not possible to visually estimate the number of steps. Even if the current step count were displayed with some kind of overlay, nobody is going to remember the exact step count of that page they were on yesterday, especially if they have since updates their step count settings.

Anyone can visually estimate and remember an already existing feature of the web browser which is the percentage the scrollbar was scrolled down the page. Scrollbars and scrolling are fundamental parts of a web browser, or any application for that matter. It is already intuitive. I feel it is important that I am able to interface with the scrollbar for this reason. It allows me, in a reasonable and reproducible manner, to return to where I was on a page yesterday without anything more complex than what I am already used to using. I am sure all often find us returning to long documentation web pages or the like that do not contain a table of contents, or if they do we don't remember the exact anchor we were looking at. What we do remember is "I was scrolled about 75% of the way down the page". Now, what you're suggesting, is that it's more effective to guess a scroll step FROM THE TOP of a page, which in this case, or any case really, simply is not true! I hope my point is clear! :heart: 

Regardless, I do appreciate Vimium and hope it stays relevant/competitive with other projects like Vrome for both power users and normal users alike! I understand it is important to strike a balance, and it can be hard to do so, but in this case I feel this is a feature all users can enjoy.

leoj3n:
Aside: I think the continuous integration tests failed because I force pushed while they were running.

smblott-github:
Thanks, @leoj3n.  Not withstanding the reasonable points you make, I think we'll stick with what we've got currently.

BTW.  Some of the use cases you describe can be well solved with marks.  See the section on "Using marks", [here](https://github.com/philc/vimium#keyboard-bindings).

leoj3n:
I like to think I helped Humanity by at least trying. :boom:

mawkler:
It's a shame that `%` command wasn't added. To me it makes a lot of sense to include, for one, since it *does* exist in Vim but also because I often find myself wanting to scroll to example 50% of the page

PR #1977, opened 2016-02-04
Move help dialog to a UIComponent

mrmr1993:
This moves the help dialog to a `UIComponent`.

@smblott-github there's a few tweaks I'd like to make to how we exclude the help dialog iframe from creating its own help dialog iframe, but it'd be nice to get it in sooner rather than later.

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -137,6 +137,24 @@ div.vimiumHighlightedFrame {
 
 /* Help Dialog CSS */
 
+iframe.vimiumHelpDialogFrame {
+  background-color: transparent;
+  padding: 0px;
+  overflow: hidden;
+
+  display: block;
+  position: fixed;
+  width: 100%;
+  min-width: 400px;
+  height: 100%;
+  top: 0px;
+  left: 0px;
+  border: none;
+
+  /* One less than vimiumReset */
+  z-index: 2147483647;
+}
+
 div#vimiumHelpDialog {
   border:3px solid red;
   opacity:0.92;
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -151,8 +151,8 @@ iframe.vimiumHelpDialogFrame {
   left: 0px;
   border: none;
 
-  /* One less than vimiumReset */
-  z-index: 2147483647;
+  /* One less than hint markers and the vomnibar. */
+  z-index: 2147483645;
 }
 
 div#vimiumHelpDialog {
@@ -171,8 +171,6 @@ div#vimiumHelpDialog {
   top:50px;
   -webkit-box-shadow: rgba(0, 0, 0, 0.4) 0px 0px 6px;
   overflow-y: auto;
-  /* One less than vimiumReset */
-  z-index: 2147483647;
 }
 
 div#vimiumHelpDialog a { color:blue; }
@@ -349,8 +347,8 @@ iframe.vomnibarFrame {
   border: none;
   font-family: sans-serif;
 
-  /* One less than hint markers and the help dialog. */
-  z-index: 2147483645;
+  /* One less than hint markers. */
+  z-index: 2147483646;
 }
 
 div#vimiumFlash {
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -769,71 +769,32 @@ window.enterFindMode = ->
   new FindMode()
 
 window.HelpDialog =
+  helpUI: null
   container: null
-  dialogElement: null
   showing: false
 
   # This setting is pulled out of local storage. It's false by default.
   getShowAdvancedCommands: -> Settings.get("helpDialog_showAdvancedCommands")
 
   init: ->
-    return if @container?
-    @container = DomUtils.createElement "div"
-    @container.id = "vimiumHelpDialogContainer"
-    @container.className = "vimiumReset"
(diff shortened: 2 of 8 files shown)


smblott-github:
This looks nice, @mrmr1993.

An oddity....

Find and visual mode within the help page generally works as expected.  *Except* after clicking "show available commands" on the options page.

Any idea of the source of the difference?

mrmr1993:
> Find and visual mode within the help page generally works as expected. *Except* after clicking "show available commands" on the options page.

I've pushed a commit fixing this.

smblott-github:
Nice, @mrmr1993.  Here are some things which come to mind (some of which are not specific to this PR).

- We do `HelpDialog.init()` on every frame of every page.  However, unlike the vomnibar, say, most users rarely use the help dialog.  Might it be better to delay initialising the help dialog until it's first used?

- If the vomnibar is activated while the help dialog is active, it appears behind the help dialog.

- This `try localStorage` lark is pretty ugly.  Might there be a more coventional way to achieve the same thing?  Similarly, another `try` elsewhere.

- It looks like `showing` and `isReady()` (and possibly some other stuff) in `pages/help_dialog.coffee` aren't used.  Could you check, please?

mrmr1993:
I've pushed some changes to address your comments.

> We do `HelpDialog.init()` on every frame of every page. However, unlike the vomnibar, say, most users rarely use the help dialog. Might it be better to delay initialising the help dialog until it's first used?

Changed this behaviour to only load the frame as necessary (6bd8318fa1a7faf942cd56b8023f2947d84f3123).

> If the vomnibar is activated while the help dialog is active, it appears behind the help dialog.

This has been the case since 2012 (bb22d36d0ef0d50de3fc28162e18b292eaa4bea6), and the comment there seems to imply that it was intentional. I can't think of any good reason why that would be though, so I've changed it in 9a7456334fca81dff41fea6a2701365806f6c3b3.

> This `try localStorage` lark is pretty ugly. Might there be a more coventional way to achieve the same thing? Similarly, another `try` elsewhere.

2f73ad6c7357c9a7105e99a929991df735efc96e is the best I could manage for this. I can't find a way to do it without a `try` somewhere though...

> It looks like `showing` and `isReady()` (and possibly some other stuff) in `pages/help_dialog.coffee` aren't used. Could you check, please?

The `HelpDialog` object from there is used instead of the one from `vimium_frontend.coffee` when we're in the help dialog iframe. This seemed like the simplest way to avoid opening nested help dialog `<iframe>`s. I've added a comment about this (2922fdf80f21d36acd12501dd1b3264377a95895).

In a follow-up PR, I'd like to move the help dialog to the top frame only, and move the `<esc>` keyhandling for hiding the dialog into `pages/help_dialog.coffee`. Once that's done, we won't need to do this any more, and we can clear out those properties.

smblott-github:
Thanks, @mrmr1993.

smblott-github:
@mrmr1993. I've always disliked the logic in `Utils.isBackgroundPage()` and `Utils.isExtensionPage()`.  It involves string comparisons and just seems brittle.

What do you think about this as an alternative: 832daac31806688d2fe51ab5594970f8a7289d53?

mrmr1993:
> It involves string comparisons and just seems brittle.

That's a ridiculous thing to say, sorry Steve. The url is a string, and to extract information out of it, we have to do string comparisons. It's also the easiest way to check what type of context we're in, since:

* if the current window's origin is our extension's origin, then
  - if the top window's origin is our extension's origin, then
    + if the window object is equal to the background page window object, we're in the background page
    + otherwise we're in an 'extension page'
  - otherwise we're in a `UIComponent`
* otherwise we're in a normal page

> What do you think about this as an alternative: 832daac?

I think it's currently wrong on the options page (`isExtensionPage()` is false), and correcting the error ([see here](https://github.com/philc/vimium/commit/832daac31806688d2fe51ab5594970f8a7289d53#commitcomment-16034707)) breaks the help dialog. You can just add a script tag for `context/extension_page.js` to the options page and avoid messing with the content script loader.

Nonetheless, I recon it makes no sense to do this:
* `try` and string comparison aren't *so* bad that we can't use them once or twice.
* two lines of code have ballooned across multiple files.
* injecting the new files correctly involves either url string comparison anyway or manually marking which files should/shouldn't have them injected (which really is brittle).

I think you should bin this.

smblott-github:
The string comparison point is pretty moot, @mrmr1993.  I'm not too bothered about it.  It would be better to do it once rather than on every call, and better still not at all.  But it's not a big deal.

The "brittle" point is this...

```Coffeescript
win.document.location?.origin + "/" == chrome.extension.getURL ""
```

It's that extra `/`.  The URL is perfectly valid both with it and without it.  Chrome could arbitrarily decide one day to not include the trailing `/` in `...getUrl()` and Chrome would still be correct -- but Vimium would break.

It doesn't seem right depend on such an implementation detail when we could alternatively, just *know* what context we're in.

mrmr1993:
> Chrome could arbitrarily decide one day to not include the trailing `/` in `...getUrl()`

The documentation for `getUrl` explicitly qualifies that the url is 'fully-qualified' (and so can be relied upon to contain a slash).

> It doesn't seem right depend on such an implementation detail when we could alternatively, just *know* what context we're in.

Can you clarify how we're going to "just *know*" the difference between the help dialog `<iframe>` (no `localStorage`) and the options page (`localStorage`)?

PR #1980, opened 2016-02-06
Externally-implemented Vimium commands [PoC].

smblott-github:
Proof of concept (and thinking aloud).

This is a rough framework for how Vimium could call commands implemented in external extensions.  Which would allow others to implement Vimium commands and publish them, possibly on the Chrome Store.  The user then adds suitable bindings to their Vimium options; e.g.:

    map qq externalCommand cagbppkocaniehcbmobegafmmeeebgff.openVimiumIssues

(`cagbppkocaniehcbmobegafmmeeebgff` is the Id of another extension, and `openVimiumIssues` is a command implemented by that extension.)

In effect, this would allow Vimium users to bind Vimium key sequences to arbitrary Javascript, and - in particular - to niche commands which are unlikely to be of interest to the wider Vimium community.

As implemented here, there are two types of commands:

- asynchronous commands -- these just run asynchronously (e.g. close window).
- synchronous commands -- the idea is that we would block Vimium keyboard activity (not currently implemented) until we receive confirmation that the command has completed (e.g. some foreground command).

Here, we use [this API](https://developer.chrome.com/extensions/messaging#external).  And we do not open a port, because that would create problems when installed/activated extensions come and go.

This approach solves some problems, but not all.  It wouldn't, for example, be possible to implement new link-hints or vomnibar modes.


smblott-github:
From the Chrome Store...

snapshot

That's an interesting idea.

smblott-github:
The story... Joe wants a new Vimium command, but it's a bit obscure.  Joe copies the example in the examples directory here, tweaks the manifest and defines his new command.  Something like:

```Coffeescript
  # Focus the first tab encountered which is playing audio.
  goToAudible:
    key: "qa"
    description: "Focus the current audio tab"
    blocking: false
    run: (count) ->
      chrome.windows.getAll {populate: true}, (windows) ->
        for window in windows
          for tab in window.tabs
            continue unless tab.audible
            chrome.tabs.update tab.id, selected: true
            return
```

Joe then publishes his extension on the Chrome Store.  There, other users can also install and use Joe's new command:

```
# Focus the current audio tab:
map qa externalCommand abjmlgelklomcednhcplkfdfdbajhdbn.goToAudible
```

You can try this out.  Build and install the branch in this PR, then install this [demo extension](https://chrome.google.com/webstore/detail/vimium-api-example/abjmlgelklomcednhcplkfdfdbajhdbn) and add whatever key mapping you want.

Here are some things such an extension might want to do:

a. *Just do something now, asynchronously*.  That's how `goToAudible` works.
b. *Do something now, and block Vimium until its finished*.  Who knows what a user might want to implement.
c. *Here are some things, pick one with the Vomnibar and tell me which*.  For example, pick a window to focus.
d. *Pick an element with link hints, then tell me which one*.  E.g, to implement link hovering, or something.  (Not sure how to specify an element, but there's probably a way.

It would probably be better to use symbolic names rather than extension Ids.

issmirnov:
I like this. This would let vimium keep the core pure and allow hooks to add functionality. 

What are the concrete steps required to get this merged in? I'd love to help where I can.

smblott-github:
@issmirnov... This -- meaning the approach -- would need support from @philc.  Then, we'd need to figure out a few things...

What should the API be?  E.g.

1. Just do something now.

2. Send control to the external extension and block Vimium until that extension tells us it's done.

3. Pick a link via link hints, tell the external extension which link and then do one of 1 and 2.

4. In the Vomnibar, pick from a list of things (provided by the extension), tell the external extension which and then do 1 or 2.

We don't need to implement all of that immediately.  1 is the most common case.  But the API should be properly thought out.

Next is the UX.  How do users tell Vimium to trust and use an external extension?  And that must be done without disrupting the Vimium options page.  Here, that's achieved by using the extension Id, but that's pretty ugly.  We should really be using names; compare:

```
# Yank the current URL in markdown format:
map ym externalCommand kolfjpnjomekbkgpgipagmgcmbdghlmo.yankMarkdownUrl
map ym markdownUtils.yankMarkdownUrl
```

I'm imagining something whereby, there's an *Authorise Me* link in the external extension.  It launches a Vimium page: *Do you want to use `markdownUtils`?*  And so on.  But would we need a mechanism to de-authorise extensions?

If you want to think about this stuff then please do, @issmirnov.  You could propose something on top of this branch or separately -- which ever seems best.

smblott-github:
Closing favour of #2109 (which is a simpler version of roughly the same thing).

Issue #1955, opened 2016-01-30
New features...


smblott-github:
@mrmr1993 ... As you probably gathered, I did a long troll of the open issues and PRs today (ten hours, and I didn't even look at them all!).

In any case, here are the two things which stood out and that we might act on...

- Implement *pass-next-key* in normal and insert modes. (A bit like `Ctrl-v` in vim.)
- Implement *go-to-previous-tab* (with count).  We have the argument about using the vomnibar in tabs mode; however, reviewing the issues, this really stands out as something which people want.

philc:
Steve, that was epic. I woke up with 250 emails in my inbox this morning =) Looking through them all now.

+1 for both of those suggestions.

PR #1984, opened 2016-02-09
New command: visit previous tab.

smblott-github:
As discussed in #1955, this implements `visitPreviousTab`.

- 4ada22e85dcd061fa806a5fe72a9dc3f1cfe0442 is a no-op; it moves the tab-recency code from `completions.coffee` to a new `bg_utils.coffee`.  There are other things (like `logMessage()`) which could similarly be moved to a separate "utils" file for the background page.
- 60d33d20026cdcdd0c4ecef20410d38341c86633 is the actual new command.

Issues:

- Is `visitPreviousTab` a good name?  Unfortunately `previousTab` is already taken.
- Is `^` a good default binding?
- My sense is that this is _not_ an "advanced command".

### DIFF
--- a/background_scripts/bg_utils.coffee
+++ b/background_scripts/bg_utils.coffee
@@ -39,12 +39,11 @@ class TabRecency
     @cache[tabId] ||= 1
     if tabId == @current then 0.0 else @cache[tabId] / @timestamp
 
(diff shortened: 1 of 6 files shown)


smblott-github:
Merging this.
We've agreed that we need this command, and there doesn't seem to be anything unusual or controversial about the implementation.

PR #1985, opened 2016-02-10
New command: Pass next key

smblott-github:
As discussed in #1955.

Example:

    map <c-]> passNextKey

Works in normal mode *and* in insert mode:

- In normal mode, accepts a count prefix.
- In insert mode, only works for single-key mappings.  (In insert mode, multi-key mappings don't make sense, because we've already passed the first key.)

Currently, this is only one command.  Better bindings might be possible, however, if we were to have separate commands for normal and insert modes.

This should make it possible, for example, to close chat popups on Facebook: `<c-]><Esc>`.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -67,16 +67,10 @@ Commands =
           when "unmapAll"
             @keyToCommandRegistry = {}
 
-    # Push the mapping for passNextKey into Settings so that it's available in the front end for insert mode.
-    passNextKeyKey = null
-    for own key of @keyToCommandRegistry
-      if @keyToCommandRegistry[key].command == "passNextKey"
-        passNextKeyKey = key
-        break
-    if passNextKeyKey
-      Settings.set "passNextKeyKey", passNextKeyKey
-    else
-      Settings.clear "passNextKeyKey"
+    # Push the key mapping for passNextKey into Settings so that it's available in the front end for insert
+    # mode.
+    Settings.set "passNextKeyKeys",
+      (key for own key of @keyToCommandRegistry when @keyToCommandRegistry[key].command == "passNextKey")
 
   clearKeyMappingsAndSetDefaults: ->
     @keyToCommandRegistry = {}
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -67,6 +67,17 @@ Commands =
           when "unmapAll"
             @keyToCommandRegistry = {}
 
+    # Push the mapping for passNextKey into Settings so that it's available in the front end for insert mode.
+    passNextKeyKey = null
+    for own key of @keyToCommandRegistry
+      if @keyToCommandRegistry[key].command == "passNextKey"
+        passNextKeyKey = key
+        break
+    if passNextKeyKey
+      Settings.set "passNextKeyKey", passNextKeyKey
+    else
+      Settings.clear "passNextKeyKey"
+
   clearKeyMappingsAndSetDefaults: ->
     @keyToCommandRegistry = {}
     @mapKeyToCommand { key, command } for own key, command of defaultKeyMappings
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -96,6 +96,7 @@ Commands =
       "goUp",
       "goToRoot",
       "enterInsertMode",
+      "passNextKey",
       "enterVisualMode",
       "enterVisualLineMode",
       "focusInput",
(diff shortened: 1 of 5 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -487,50 +490,20 @@ handlerStack.push
...
+    if currentCompletionKeys.indexOf(keyChar) != -1 or isValidFirstKey keyChar
+      DomUtils.suppressEvent(event)
+      keyPort.postMessage keyChar:keyChar, frameId:frameId
+      return @stopBubblingAndTrue
 
-    if (keyChar)
-      if currentCompletionKeys.indexOf(keyChar) != -1 or isValidFirstKey(keyChar)
-        DomUtils.suppressEvent(event)
-        keyPort.postMessage({ keyChar:keyChar, frameId:frameId })
-        return @stopBubblingAndTrue
```
smblott-github: This code has been moved to KeyboardUtils.

content_scripts/vimium_frontend.coffee
```diff
@@ -487,50 +490,20 @@ handlerStack.push
...
-      if (event.ctrlKey)
-        modifiers.push("c")
-      if (event.altKey)
-        modifiers.push("a")
-
-      for i of modifiers
-        keyChar = modifiers[i] + "-" + keyChar
-
-      if (modifiers.length > 0 || keyChar.length > 1)
-        keyChar = "<" + keyChar + ">"
```
smblott-github: This code has been moved to KeyboardUtils (and tidied up there, a bit).

background_scripts/commands.coffee
```diff
@@ -67,6 +67,11 @@ Commands =
           when "unmapAll"
             @keyToCommandRegistry = {}
 
+    # Push the key mapping for passNextKey into Settings so that it's available in the front end for insert
+    # mode.
+    Settings.set "passNextKeyKeys",
```
mrmr1993: Using `Settings` for this feels a bit wrong, since it's a computed value rather than a setting. Seems like this value could be passed in `getCompletionKeysRequest` to be consistent with how we treat our other calculated mapping values.

smblott-github: On the other hand, it is a (fairly) static value, a function of the user's settings, and this approach requires no extra plumbing to maintain.

smblott-github:
Please note 0cf81be8f23d3eabdc0c85216f4d35c68d7a29a2 (the discussion in the commit).  This is an issue which could cause some user confusion.

Also:  Are there any clearly better approaches to achieve the same effect as this PR?

smblott-github:
I propose to merge this soon, unless we identify a significant issue with it.

gdh1995:
What's the use case of this command in `InsertMode`? I think it can only pass `<esc>` - other keys have been passed before this PR.

smblott-github:
> What's the use case of this command in InsertMode?

Go to Facebook.  Open a messenger popup (and focus where you type the text).  Like this...

snapshot

Now, without Vimium, `Escape` closes the popup.  With Vimium, `Escape`  exits insert mode (after which point you cannot close the popup with the keyboard).

With this PR, `<c-]><Esc>` closes the popup.

> other keys have been passed before this PR.

That's correct.  Although also `<c-[>`, I think.

gdh1995:
Okay. Thanks for your patience。

rlrosa:
This doesn't seem to work for me. I have conversations highlighted (ready to type) and neither ```<c-]><Esc>``` nor ```<c-[><Esc>``` close it, they just drop focus.
Just to be sure, they key combination in words would be "hold down ctrl key and bracket key, and Esc keys simultaneously". Also tried "hold dow ctrl key and brackey key simultaneously, then release both and press Esc key".
Thanks for the input, plugin is great but this is no-go for me, since it affects hangouts, fb, and other chat things.

smblott-github:
There's no default mapping for this command.  You have to choose your own mapping.  Could that be it?

dwighthouse:
I cannot get this to work, same as @rlrosa . I've added this to my Extension mapping:

    map <c-]> passNextKey

And no combination of Control, ], or Escape actually fires the Escape keydown event on my input.

I even tried a manual push of Escape:

    map <c-]> <Escape>

This does nothing when I hold Control and push ], except continuously fire onkeydown events for ].

In my use case, escape is used on the input field to hide the dropdown results of a livesearch component I built in React. Without the escape key, the livesearch results stay forever open when trying to activate another control using Vimium.

PR #1989, opened 2016-02-12
Remove link-hints with queue.

smblott-github:
Since link-hints modes now accept a count, we don't get any extra functionality from the *with-queue* mode.  We get the same effect by passing a rediculously large count.

This, therefore, removes a small amount of logic associated with the *with-queue* mode.

On the downside, we lose the associated HUD indicator text of the *with-queue* mode.

Do this?  Not do this?


smblott-github:
Closing this.

We could do it, or not, so let's just leave it for now.

I'm increasingly thinking that maybe we should just remove the *with queue* command entirely.

PR #1990, opened 2016-02-12
Detect page context

smblott-github:
This changes the mechanism by which we detect the extension-page/background-page context.  It avoids the need to do a rather weird calculation in every frame of every tab.

Continuing the discussion from #1977....

As far as I can see, this seems to work just fine.  And we *do* seem to have access to `localStorage` on the help page.

(Or perhaps I don't understand your point, @mrmr1993.)


mrmr1993:
> And we *do* seem to have access to `localStorage` on the help page.

Did you trest this on a page that *wasn't* the options page?

smblott-github:
> Did you trest this on a page that wasn't the options page?

Yes.  Well here's what I did (in case you have something else in mind):

- Go to any old page.
- Open the help page.  Test *scroll-step-size*.  Leave the help page open (or not).
- In another tab, go to the options page and change the *scroll-step-size*.  Save.
- Go back to the tab with the help page and scroll again.... The change takes effect.

Is there something else you have in mind?

mrmr1993:
I get this on the page for this PR:

screenshot from 2016-02-12 19 16 13

Note especially that no advanced commands are shown &mdash; even though I always have them visible &mdash; because `Settings` is undefined after erroring out when accessing `localStorage`.

Steps to repro:

* Compile and load this PR
* Reload this page
* Open the developer console (F12)
* Refocus the page and press `?` (or whatever your mapping for help is)
* Notice the errors in the console, and that clicking the "show advanced commands" link doesn't work.

smblott-github:
@mrmr1993.  At first I couldn't reproduce your issue.  Then, finding [this page](https://code.google.com/p/chromium/issues/detail?id=357625), it seems the problem is related to the Chrome *Block third-party cookies and data* setting.

So, we have to put the [try](https://github.com/philc/vimium/commit/4f894b44a06971012739726b2a43de9601f22e1f) back in.

smblott-github:
@mrmr1993.  I favour this over the status quo --- assuming it's correct.  It avoids a strange string comparison in *every frame*, and for which you have to look up the details of the relevant APIs to convince yourself it's correct.

If this *were* the status quo, then we certainly wouldn't consider changing it to the current implementation.  And that's usually a good indication of the correct decision.

smblott-github:
Closing.  7bb0455ac210aeea00f9d772eebcddbc4deed3a2 (not merged) would be considerably simpler.

PR #1991, opened 2016-02-12
Nuke some legacy migration code.

smblott-github:
These migrations are at least ten months old (since release), and will be older still be the time of the next release.  They can probably go.


smblott-github:
Merging.
The most controversial thing here is probably the method name `Settings.nuke()`.

PR #1993, opened 2016-02-14
Split filtered hints on non-word and hint characters.

smblott-github:
This affects the scoring system for filtered link hits, and therefore affects their usability.

Example link text:

    "We open all day", say 7Eleven

Previously:

    "We
    open
    all
    day",
    say
    7Eleven

With this PR:

    we
    open
    all
    day
    say
    Eleven

Previously: the typed text `we` and `day` would receive poor scores (not the start of a word, and not a whole word).  Now, these get high scores, so are more likely to be selected as the active link.  Also, `7Eleven` cannot be typed (because `7` is a hint character).

With this PR, the typed text `we` and `day` now get high scores, as they should.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -479,6 +479,9 @@ class FilterHints
     @linkTextKeystrokeQueue = []
     @labelMap = {}
     @activeHintMarker = null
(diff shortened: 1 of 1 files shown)


smblott-github:
Merging directly.

Probably (?) not controversial, and it's only a small behaviour tweak.  If there's a problem with this, I'll spot it soon enough.

PR #1997, opened 2016-02-18
Do not recognise digits as valid first keys if Shift is depressed.

### LINKED ISSUE #1757: cannot exclude '#' keybinding
I would like to use the Gmail keybinding '#' to delete mail, but
if I type '#' into the exclude field, it seems to comment out the rest
of the field instead.

Steps to reproduce:

- open vimium options
- exclude url mail.google.com/*
- exclude keys jknp#c
- C-RET to save
- open Gmail

At this point, I expect that
* j selects the next message
* # deletes it
* c opens a compose buffer

Instead I observe that
* j select the next message
* # does nothing
* c does nothing

If I instead move the 'c' before the '#', I find that 'c' opens the
compose buffer as I would expect if it were being excluded.

Are you maybe running the exclude field through a parser that
interprets Python/bash style comment syntax?

### LINKED ISSUE #1996: Asterisk is not being passed through as an excluded key
Hi,

When I want to pass asterisk through to the underlying page it doesn't seem to work. Here is an example of what I have set up for gmail:
screenshot from 2016-02-18 11 35 01
Other keys are being passed through, just not the asterisk.
I'm on Chrome 48.0.2564.109 (Official Build) (64-bit) on Ubuntu; vimium is version 1.54.

Thanks!

smblott-github:
Currently, we cannot set passkeys for symbols accessed with `Shift` depressed on a digit key: e.g. `$`, `*`.  This affects native keyboard shortcuts in GMail.

@mrmr1993.  This would be fixed if we had `event.key`, right?  In the meantime, is this fix correct?  Surely there are no keyboards with digits requiring the `Shift` key?

Fixes #1996.
Fixes #1757.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -585,7 +585,7 @@ window.refreshCompletionKeys = (response) ->
 
 isValidFirstKey = (keyChar, event) ->
   # Digits are valid first keys only if Shift is not depressed; see #1996 and #1757.
-  validFirstKeys[keyChar] || (not event.shiftKey and /^[1-9]/.test keyChar)
+  validFirstKeys[keyChar] || (KeyboardUtils.hasNoModifiers(event) and /^[1-9]/.test keyChar)
 
 window.handleEscapeForFindMode = ->
   document.body.classList.remove("vimiumFindMode")
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -491,7 +491,7 @@ handlerStack.push
 onKeypress = (event) ->
   keyChar = KeyboardUtils.getKeyCharString event
   if keyChar
-    if currentCompletionKeys.indexOf(keyChar) != -1 or isValidFirstKey keyChar
+    if currentCompletionKeys.indexOf(keyChar) != -1 or isValidFirstKey keyChar, event
       DomUtils.suppressEvent(event)
       keyPort.postMessage keyChar:keyChar, frameId:frameId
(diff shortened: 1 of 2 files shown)


mrmr1993:
Haven't tested (I'm away for a few days), but I expect this breaks `g$` on google (we leak the `$`'s `keydown`).

Adding a shift modifier (eg. `<s-1>`) and implementing #1368 or similar would be the way I would look to solve the problem.

> This would be fixed if we had `event.key`, right?

Yes, `event.key` fixes all of the edge cases for our key handling.

smblott-github:
> but I expect this breaks g$ on google (we leak the $'s keydown)

You mean Google's *instant predictions*?  Yup.  It breaks `g$` in that case.

To summarize:

- We're currently incorrectly blocking keydown events on one important Google site (GMail).
- But the fix - which is correct in principle, I think - breaks another important Google site (if instant prediction is enabled).

(Google instant prediction and Vimium aren't great bedfellows.)

lydell:
The widely used AZERTY layout has symbols on the number row and numbers in its shift layer. So there are _many_ users who must use shift to type numbers :)

Se programmers also like to swap the regular and shift layers of the number row because they feel that they use symbols more often than numbers, which reduces the need to press shift.

lydell:
AZERTY requires shift to type numbers, as far as I remember.

smblott-github:
Thanks for the input, @lydell.  So, I guess we're leaking a lot of `keydown` events on AZERTY keyboards.

In any case, the status quo wins.  In light of the points from @mrmr1993 and @lydell, we probably just stick with the status quo until `event.key` is available.

PR #2000, opened 2016-02-20
hideHUD option applies to insert mode only.

### LINKED ISSUE #487: Make the hideHUD preference only apply to insert mode, and rename it accordingly
With hideHUD enabled, the user can't see what he is searching for when doing find-in-page. I think hideHUD should be applied only to insert mode (and renamed accordingly). We could possibly extend it to hiding it during the display of link hints as well.

### LINKED ISSUE #1953: Find mode broken if HUD display is disabled
Try it.

(I guess it's been like that for a while, and nobody's noticed.  Do we even need to option to disable the HUD?)

smblott-github:
This makes the `hideHud` option apply only to insert mode (when entered with `i`).

Fixes #1953.
Fixes #487.

We could rename the option itself and add migration code, but that seems overkill.

An alternative would be to remove this option entirely.

### DIFF
--- a/content_scripts/hud.coffee
+++ b/content_scripts/hud.coffee
@@ -87,8 +87,7 @@ HUD =
     DomUtils.documentReady -> ready = true
     -> ready and document.body != null
 
-  # A preference which can be toggled in the Options page. */
-  enabled: -> !Settings.get("hideHud")
+  enabled: -> true
 
 class Tween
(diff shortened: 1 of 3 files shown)


smblott-github:
Merging.
If this turns out to be the wrong thing to do, then - it's trivial - so we can easily revert it.

PR #2001, opened 2016-02-20
Add count command option

smblott-github:
Examples:

    # This approximates a closeWindow command.
    map q removeTab count=9999
    
    # This approximates LinkHints.activateModeWithQueue.
    map <a-f> LinkHints.activateModeToOpenInNewTab count=9999
    
    # Download many links.
    map <a-d> LinkHints.activateModeToDownloadLink count=9999
    
    # Have multiple scroll speeds.
    map J scrollDown count=3

Currently, there are two commits:

- 61ce82be4acae20562bb2ccb93ff64772ba553c1 just refactors (and generalises) the command-option parsing.  Previously it was in the vomnibar code (which was silly), now it's in `commands.coffee`.

- 2c634d9b2324c760007d59d5cafd97a364125991 adds the `count=N` command option.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -56,7 +56,7 @@ Commands =
             if command? and @availableCommands[command]
               key = @normalizeKey key
               logMessage? "Mapping #{key} to #{command}"
-              @mapKeyToCommand { key, command, options: @parseCommandOptions optionList }
+              @mapKeyToCommand { key, command, options: @parseCommandOptions command, optionList }
 
           when "unmap"
             if tokens.length == 2
@@ -77,7 +77,7 @@ Commands =
   # Command options follow command mappings, and are of one of two forms:
   #   key=value     - a value
   #   key           - a flag
-  parseCommandOptions: (optionList) ->
+  parseCommandOptions: (command, optionList) ->
     options = {}
     for option in optionList
       parse = option.split "="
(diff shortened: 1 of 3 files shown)


background_scripts/commands.coffee
```diff
@@ -74,6 +74,27 @@ Commands =
     Settings.set "passNextKeyKeys",
       (key for own key of @keyToCommandRegistry when @keyToCommandRegistry[key].command == "passNextKey" and 1 < key.length)
 
+  # Command options follow command mappings, and are of one of two forms:
+  #   key=value     - a value
+  #   key           - a flag
+  parseCommandOptions: (command, optionList) ->
+    options = {}
+    for option in optionList
+      parse = option.split "="
```
mrmr1993: We can use `option.split "=", 2` here to ignore any later `=` matches, and then the `else` case below becomes redundant.

mrmr1993:
Very nice, LGTM.

smblott-github:
Thanks, @mrmr1993.

> We can use option.split "=", 2 here to ignore any later = matches, and then the else case below becomes redundant.

Done: 70b8645d61a56ea760864d222791494aa6031f92.

I think I'll just merge this.  I think it's a good idea. @mrmr1993 approves.  And if anybody else shows up with a concern (e.g. @philc), then it's small, so easy to unwind.

PR #2002, opened 2016-02-20
Move wholly to chrome.sessions.

smblott-github:
This strips out the legacy pre-chrome.sessions code.  The API has been in Chrome since version 37, and we probably don't really need to support older versions than that.


smblott-github:
Assuming uncontroversial.
Merging directly.

PR #2003, opened 2016-02-20
Make `moveToNewWindow` accept a count.

smblott-github:
This make `moveToNewWindow` accept a count.  For example, `3W` to move three tabs to (the same) new window.

The tabs chosen are the current tab, then those to the right of the current tab, and then those to the left of the current tab.

`999W` moves *all* tabs to a new window.  It's not clear why you would want to do that.  An alternative would be to leave the last tab behind.


mrmr1993:
Works perfectly. LGTM!

smblott-github:
Thanks for checking, @mrmr1993.

PR #2008, opened 2016-02-20
declare min(chrome.ver) as 37 since we moved to chrome.sessions

gdh1995:
Chrome's sessions functionality began from version 37, so we may declare it in Vimium's manifest.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2,6 +2,7 @@
   "manifest_version": 2,
   "name": "Vimium",
   "version": "1.54",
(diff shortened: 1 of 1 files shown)


smblott-github:
If we do this, @gdh1995, then Vimium won't install (or upgrade?) at all on older Chrome versions.  Whereas if we don't do this, then it's just the `chrome.sessions` commands which don't work.  The latter seems arguably better.

gdh1995:
Okay, you are right.

PR #2010, opened 2016-02-20
Use for (not for own) when iterating over an event's properties

mrmr1993:
Moving to `for own` broke find mode history (and thus also next/previous).

This reverts that part of commit 56fed2ac6663d99ca03023f3ffa313c51de5fe32.


smblott-github:
Thanks, @mrmr1993.

PR #2014, opened 2016-02-22
Add install date to logging page.

smblott-github:
This implements a poor-man's build info.

See #1352.  Unfortunately, that requires a separate build target, and does not work with `cake autobuild`.

This just records the *install date* and displays that info on the logging page.  "Install date" because we can reliably determine it, and because it does answer the question *have I upgrade Vimium on this machine since last week?*.  And on the logging page because that's out of the way, and not part of the regular Vimium interface.

snapshot

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -658,5 +658,10 @@ showUpgradeMessage = ->
 
 showUpgradeMessage()
 
+# The install date is shown on the logging page.
+chrome.runtime.onInstalled.addListener ({reason}) ->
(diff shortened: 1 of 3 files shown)


mrmr1993:
I've pushed [a branch](https://github.com/mrmr1993/vimium/tree/record-install-date) that also gets the current branch (via an `XMLHttpRequest` to `.git/HEAD`, no special build commands), failing gracefully when this file doesn't exist.

smblott-github:
That's nice, @mrmr1993.

Issue #1944, opened 2016-01-26
Tests fail/hang with latest PhantomJS (2.1.1)


stephanepoirier:
With the latest PhantomJS (2.1.1), running `cake test` fails then hangs:

``` bash
Running unit tests...
Pass (136/136)
Running DOM tests...
Fail "Mode utilities: not exit on escape if not enabled" -  function (message) {
  this.name = AssertionError;
  this.message = message;
}: Expected 0 but received 2
Fail "Normal mode: not suppress escape" -  function (message) {
  this.name = AssertionError;
  this.message = message;
}: Expected 0 but received 2
```

Execution hangs when the test named "Focus each input when its hint text is typed" indirectly invokes: `page.sendEvent "keypress", request.key` (phantom_runner.coffee: line 32)

I started investigating. I'm currently unsure about the root cause and how to fix. In the meantime, I'll update the CONTRIBUTING.md file to recommend using version PhantomJS 1.9.8, which works fine.

StephanePoirier:
Progress update: issue reproduced on both Windows 10 and Debian Jessie.

The hanging issue is specific to the HTML "color" input component. Under Windows, running tests opens the OS-native color picker dialog and waits for an user input to resume and complete tests. Under Debian, no dialog is shown, but PhantomJS still waits for the user input without showing the dialog.

smblott-github:
Thanks, @StephanePoirier.

Yup.  I reproduced this (although the `npm` version I get still seems to be fine).

So, there are two issues:

- The tests hang due to the HTML5 `color` input, as you've identified.  We can fix this just by dropping `color` from the `inputTypes` list.  That case isn't particularly adding much.
- And the two tests listed above now fail.  This needs more work.

Suggest we hold off on #1945 for a few days until we've had a better chance to look into this.

StephanePoirier:
Progress update:
- Color input: I'll file a bug on PhantomJS. As you suggest, @smblott-github, I'll remove the color input test for the moment.
- The 2 failing tests with 2.1.1 are a valid concern. Those same tests pass in 1.9.8, as a side-effect of the `setup`/`tearDown` invocation from `PostFindMode`. I'll be start working on this test isolation problem.

smblott-github:
Thanks, @StephanePoirier.

StephanePoirier:
Root issue for the hanging problem: https://github.com/ariya/phantomjs/issues/13979

smblott-github:
Fixed (well, workaround) in #2015.

[...]

PR #2015, opened 2016-02-22
Workaround for the hanging part of issue #1944

stephanepoirier:
Workaround for the hanging part of issue #1944 "Tests fail/hang with latest PhantomJS (2.1.1)".

URL: https://github.com/philc/vimium/issues/1944


smblott-github:
Thanks, @stephanepoirier.
This is a better workaround.

And this is [merged](https://github.com/philc/vimium/commits/masterhttps://github.com/philc/vimium/commits/master).

PR #2016, opened 2016-02-25
Treat mappings as arrays

mrmr1993:
This cherry picks some commits from my branch [here](https://github.com/mrmr1993/vimium/tree/rebase-normal-mode-mappings-in-frontend), which is a rebase of #1775.

This introduces `commandKeys`, an array of arrays of keys which represent the normal mode mappings. The main benefit of this is that we can avoid evaluating the regex `namedKeyRegex` 4 times (or more, for 2-key mappings).

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -13,6 +13,7 @@ keyQueue = null
 # The user's operating system.
 currentCompletionKeys = ""
 validFirstKeys = ""
+commandKeys = []
 
 # We track whther the current window has the focus or not.
 windowIsFocused = do ->
@@ -580,6 +581,8 @@ window.refreshCompletionKeys = (response) ->
 
     if (response.validFirstKeys)
       validFirstKeys = response.validFirstKeys
+    if (response.commandKeys)
+      commandKeys = response.commandKeys
   else
     chrome.runtime.sendMessage({ handler: "getCompletionKeys" }, refreshCompletionKeys)
 
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -13,7 +13,6 @@ keyQueue = null
 # The user's operating system.
 currentCompletionKeys = ""
 validFirstKeys = ""
-commandKeys = []
 
 # We track whther the current window has the focus or not.
 windowIsFocused = do ->
@@ -581,8 +580,6 @@ window.refreshCompletionKeys = (response) ->
 
     if (response.validFirstKeys)
       validFirstKeys = response.validFirstKeys
-    if (response.commandKeys)
-      commandKeys = response.commandKeys
   else
     chrome.runtime.sendMessage({ handler: "getCompletionKeys" }, refreshCompletionKeys)
 
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -407,9 +407,9 @@ populateValidFirstKeys = ->
 
 populateSingleKeyCommands = ->
   singleKeyCommands = []
-  for own key of Commands.keyToCommandRegistry
-    if (getActualKeyStrokeLength(key) == 1)
-      singleKeyCommands.push(key)
+  for keys in commandKeys
+    if (keys.length == 1)
+      singleKeyCommands.push(keys[0])
 
 populateCommandKeys = ->
   commandKeys = []
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -21,6 +21,7 @@ chrome.runtime.onInstalled.addListener ({ reason }) ->
 currentVersion = Utils.getCurrentVersion()
 keyQueue = "" # Queue of keys typed
 validFirstKeys = {}
+commandKeys = []
 singleKeyCommands = []
 frameIdsForTab = {}
 root.urlForTab = {}
@@ -196,6 +197,7 @@ getCompletionKeysRequest = (request, keysToCheck = "") ->
   name: "refreshCompletionKeys"
   completionKeys: generateCompletionKeys(keysToCheck)
   validFirstKeys: validFirstKeys
+  commandKeys: commandKeys
 
 TabOperations =
   # Opens the url in the current tab.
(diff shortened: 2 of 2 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -580,6 +581,8 @@ window.refreshCompletionKeys = (response) ->
 
     if (response.validFirstKeys)
       validFirstKeys = response.validFirstKeys
+    if (response.commandKeys)
+      commandKeys = response.commandKeys
```
smblott-github: Are you sure you've cherry picked correctly, @mrmr1993?  It looks like `commandKeys` is unused here.

mrmr1993: I have; if this is mergable, then I wanted to follow it with other commits from #1775's rebase, which would immediately put this to use. I can push a commit to remove this, and revert it in the next PR, if you'd prefer.

mrmr1993:
@smblott-github I've pushed a commit removing the code that exposes `commandKeys` to the frontend.

smblott-github:
Alternative idea...

A better data structure for modelling the keys-to-commands mapping might be a dictionary of dictionaries (in Python terminology).

```Coffeescript
keyMappings = 
   i: enterInsertMode
   g:
      u: goUp
      U: goToRoot
   # And it's not clear that we need to stop at just two levels...
   a:
      b:
         c: aThreeKeyBinding
```

Then we model the state as a list of "pointers" into this structure.  So, after `g`:

```Coffeescript
keyState = [ keyMappings['g'], keyMappings ]
```

Then, the key processing looks something like...

```Coffeescript
commandToRun = null
newKeyState =
   for mapping in keyState
      continue unless mapping[keyChar]
      if mapping[keyChar].description  # We would need a better test for this.
         commandToRun = mapping[keyChar]
         break
      else
         mapping[keyChar]

newKeyState = [] if commandToRun?
keyState = [newKeyState..., keyMappings]

validFirstKeys = []
for mapping in keyState
   validFirstKeys.push (key for own key of mapping)...
```

(All of that was written in this textarea, so there will be mistakes, but hopefully you get the idea.)

mrmr1993:
> A better data structure for modelling the keys-to-commands mapping might be a dictionary of dictionaries (in Python terminology).

I instinctively dislike it, since it obfuscates the information we're trying to store. As far as I can see

Pros:
* It's more efficient to look up the commands for the current key(s) than an array of arrays of strings

Cons:
* It mixes extended mappings and command information (in structure, not within the same object).
* It's harder to check that the values are as we expect by eye (when testing and debugging)
* It's harder to explain conceptually

Is this a strong/better alternative in your mind, or just something you were floating?

smblott-github:
Also on the pros side:

- You get multi-key (more than two) mappings for free.
- And it's nearly always better if you can replace logic with data.

> or just something you were floating?

I'm just floating it.  I have a demo implementation ([branch](https://github.com/smblott-github/vimium/tree/generalised-key-bindings), [commit](https://github.com/smblott-github/vimium/commit/17750b3d0fbf572b10f705fcb0a6a9ce11114599)).  Most of that is just the normal-mode key handling.  The new idea is just `keyCharIsKeyStatePrefix` and `advanceKeyState`.

However, I'm not convinced it's worth pursuing right now.  There would be some benefits, but also some very considerable disruption.  And I'm not sure the benefits are great enough to warrant that disruption.  The *if-it-ain't-broke* principle.

mrmr1993:
> Also on the pros side:
>* You get multi-key (more than two) mappings for free.
>* And it's nearly always better if you can replace logic with data.

These are  actually consequences of your new `keyState` array and associated logic, not of using a dictionary. A dictionary *can* only have a performance impact, since it's just a restructuring of the data.

To compare with `commandKeys`:

```coffeescript
newKeyState = keyState.filter (mapping) ->
    mapping.keys[keyQueue.length] == keyChar
keyQueue.push keyChar
[commandToRun] = newKeyState.filter (mapping) ->
  mapping.length == keyQueue.length

newKeyState = [] if commandToRun?
keyState = [newKeyState..., keyMappings]

validFirstKeys = []
for {keys} in keyState
   validFirstKeys.push (keys[keyQueue.length] for own key of mapping)...
```

(I haven't fixed any code errors from above, just substituted in `commandKeys` code.) Both of the pros you list still apply here.

> I'm not convinced it's worth pursuing right now. 

Can this PR be merged then?

smblott-github:
FYI, @mrmr1993.
I'm now more serious about considering [this](https://github.com/smblott-github/vimium/tree/generalised-key-bindings).

mrmr1993:
> I'm now more serious about considering [this](https://github.com/smblott-github/vimium/tree/generalised-key-bindings).

Am I correct in understanding that you've essentially re-done #1775, but thrown out all of the tried and tested code and rewritten it yourself?

smblott-github:
> Am I correct in understanding that you've essentially re-done #1775, but thrown out all of the tried and tested code and rewritten it yourself?

A bit, but mainly no.  I'm mainly interested in improving the key handling.  And, since that requires re-writing the key handling, we might as well do it in the front end, and - if we're doing it in the front end, then we should do it in a general-purpose way.

If you look at the key handling, there's very little similarity between #1775 and [this](https://github.com/smblott-github/vimium/tree/generalised-key-bindings).

mrmr1993:
> I'm mainly interested in improving the key handling.

So am I, and have been for quite some time.

> And, since that requires re-writing the key handling, we might as well do it in the front end, and - if we're doing it in the front end, then we should do it in a general-purpose way.

I disagree: they should be done separately. Moving code is hard enough without also changing it as it goes. If you want to move it, then do that, along with the necessary refactoring. If you want to change it, change it in place. But don't do both in one go.

lydell:
> ```coffee
> keyMappings =
>    i: enterInsertMode
>    g:
>       u: goUp
>       U: goToRoot
>    # And it's not clear that we need to stop at just two levels...
>    a:
>       b:
>          c: aThreeKeyBinding
> ```

I took this approach in VimFx almost a year ago, and it has been great! The old shortcut matching code had become very difficult to understand and was a bit buggy, so for us the choice was easy. I really like how the key handling logic becomes essentially this:

```
Is the pressed key in the current tree?
Yes:
    Is it a leaf node?
    Yes:
        Run the command!
    No:
        Set the tree it points to as the current tree.
No:
    Reset to the toplevel tree.
```

Note that we don’t store the keyboard shortcuts this way. Instead, we store them in the simplest possible way that makes sense. When VimFx starts up, as well as any time the user customizes a shortcut, this “key tree” is (re-)built. When building it, it is also very easy to find and save all conflicting shortcuts in an array and then present them to the user in our options UI.

Something that has helped me when working on VimFx is “cheating” by checking how Vimium, cVim, Vimperator and Pentadactyl does things. So I just wanted to “give back” a bit here, by sharing our experience :)

Now, I don’t know enough about Vimium’s internals and Chrome extensions, so I probably shouldn’t say more :) Move along.

smblott-github:
Thanks for the input, @lydell.  My implementation is over in #2022.  It removes 200 very tricky LoC (net) and is - I think - simpler to understand and maintain.

PR #2020, opened 2016-02-26
Remove outdated blacklist in DomUtils.isEditable, use DomUtils.isSelectable

### LINKED ISSUE #1964: Hint/input mode "bug".
Take a look at this situation, here on Github...

snapshot

"Click" the link numbered `22` here.  The link is activated correctly, however, we also drop into insert mode, because there's a `focus` event and `DomUtils.isFocusable event.target` is true.

mrmr1993:
This fixes #1964.


gdh1995:
Does `input[type="range"]` need to be excluded?

PR #2022, opened 2016-02-27
Key handling in content scripts.

smblott-github:
~~Work in progress, and probably still not working correctly.~~ Edit... It's working well now.

This:

- Moves all key handling to the front end (so no race conditions in key handling).
- Introduces a general-purpose key-handling class (and uses it for normal mode).
- Allows multi-key mappings (more than two keys).

The idea for this started in [this comment](https://github.com/philc/vimium/pull/2016#issuecomment-189135639) and is here implemented in the `advanceKeyState` method in the `KeyHandlerMode` class.  This approach to managing the key state greatly simplifies the key-handling logic and - because it's now in the front end - eliminates a fair amount of plumbing.


mrmr1993:
Copying in my [comment](https://github.com/philc/vimium/pull/2016#issuecomment-189679510) here:

> > I'm mainly interested in improving the key handling.

> So am I, and have been for quite some time.

> > And, since that requires re-writing the key handling, we might as well do it in the front end, and - if we're doing it in the front end, then we should do it in a general-purpose way.

> I disagree: they should be done separately. Moving code is hard enough without also changing it as it goes. If you want to move it, then do that, along with the necessary refactoring. If you want to change it, change it in place. But don't do both in one go.

smblott-github:
> I disagree: they should be done separately

In this case I respectfully disagree.  Because the (old) key handling code is so spread around and effectively everything changes, it's clearer (to me) at least like this.

smblott-github:
Update...

This is working well.  Currently, it loses over 200 lines of pretty gnarly key-handling code.  In addition, it makes it possible to write key-handling tests which weren't possible previously (not yet done).

Unless a serious issue emerges, I intend to merge this soon.  The main user-facing win is multi-key mappings (more than two keys).  That's nice, but not such a big deal.  The really big win is that this is substantially simpler than the status quo, and with considerably fewer LoC.

background_scripts/commands.coffee
```diff
@@ -94,6 +101,24 @@ Commands =
     @keyToCommandRegistry = {}
     @mapKeyToCommand { key, command } for own key, command of defaultKeyMappings
 
+  generateKeyStateMapping: ->
```
philc: What this does is actually explained in mode_key_handler.coffee. It would be nice if this function had a docstring either referencing `mode_key_handler.coffee` or explaining what the method does.

philc:
Haven't tried it, but I'll take your word that it works. Deleting 200 LOC is a solid win in my book.

smblott-github:
Here are three (corner) use cases which are different in this PR (vis-a-vis `master`).

(*CTWM* -> *change-tab-with-mouse*)

1. `9g5gg`.  In `master`, this becomes `1gg`, here it becomes `5gg`.
2. `g`, *CTWM*, `g`.  In `master`, this is `scrollToTop`.  Here, it's just a `g` sitting in the "key queue".
3. `g`, *CTWM*, *CTWM* back to original tab, `g`.  In `master` this is `scrollToTop`, here it's again just `g` sitting in the "key queue".

The second two look very similar, but are actually different.  2 happens because we're now doing key handling on a per-tab basis.  3 happens because we (currently) set the passkeys every time we change tab/frame (when we also reset the key state).

My sense is to treat the first two as bug fixes, and the latter as a corner case where both the previous behaviour and the new behaviour are arguably correct, so not worth spending any extra logic on.

mrmr1993:
A few things are still a bit off AFAIC:
* the 'state' is communicated via `chrome.storage` instead of a messaging API. Since it is being recomputed every time the bg page is loaded, it could easily be a variable there and communicated by message.
* a mapping containing a number 'consumes' the number, even if the mapping ends up not matching
  - e.g. `map g2e`. Then `g2gg` becomes `gg`, when it should be `2gg`, as in Vim.

Otherwise, code LGTM

smblott-github:
> a mapping containing a number 'consumes' the number...

Right enough.  That's an interesting example - and one I hadn't thought of.  I'll take your word for it that vim behaves as you describe.

It's pretty far in the "corner case" Ecke, though.  And bindings like that aren't supported in `master`, so there aren't really any legacy issues.  I'd say we fix it only if it can be done with practically zero extra code complexity.

> ... it could easily be a variable there and communicated by message

It could.  Or it could not.  It's in `chrome.storage.local`, so the space is pretty much free.  I kind of like having `chrome.storage` doing the messaging for us.  That way we don't have to maintain any code.  We're already using `chrome.storage` to do the "messaging" for `Settings` -- and that implementation is a big improvement over the old version, where we did the messaging ourselves.

Thanks for the input.

smblott-github:
Note 811d6e864de485a11bb46d4a0618b9a27027f3c7 (merged).

We can now write tests for normal-mode key handling.  And writing those tests uncovered this bug: 7774beb6643c0d905f9caba4345453790af948ad.  We were handling `7gj` as `7j` (not `1j`, as it should be).

PR #2025, opened 2016-02-28
Pass to count to scroll commands.

smblott-github:
Currently, `10j` keeping `j` held down scrolls quickly for a time then reduces back the regular hold-`j` scroll speed.  Therefore, the user cannot use a count to influence the smooth-scrolling scroll speed.

This PR fixes that by passing the count to the scroll functions.  Consequently, we adjust the actual scroll amount (which affects the scroll speed) rather than calling the scroll commands several times (which doesn't).


smblott-github:
Merging.
Not controversial (I think).

PR #2026, opened 2016-02-28
Refactor focusFrame - Fix #2023.

### LINKED ISSUE #2023: Help page and vomnibar focus issue.
Try the following...

- Open any page.
- `?` to open the help page.
- `o` to open the vomnibar.
- `Esc` to close the vomnibar.

The focus ends up back in the main page, and not in the help window.

The problem seems to be that [this stub](https://github.com/philc/vimium/blob/master/content_scripts/vimium_frontend.coffee#L304) for `focusThisFrame` is being called.  So, apparently, the `documentReady` listener (above that) is not being triggered.

smblott-github:
Only the `flashFrame` part needs to be guarded against the DOM being ready.  So we can take the `focusFrame` part out as a regular function.

Fixes #2023 (although I don't fully understand why that's happening).

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -269,9 +269,7 @@ setScrollPosition = ({ scrollX, scrollY }) ->
         Marks.setPreviousPosition()
         window.scrollTo scrollX, scrollY
 
-#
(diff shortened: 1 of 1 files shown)


smblott-github:
Merging.
Not controversial (I think).

PR #2029, opened 2016-03-01
A small refactor link hints.

smblott-github:
Link hints is in need of a major refactoring.  But here are two small changes:

- Fix the hint-activation sequence.  We have several ways of activating a hint (immediately, with delay, wait-for-enter).  And the logic previously was complicated (and wrong in at least one case).  This puts all of the logic in one place: 3671382a6a3c14d1f834b6ebcf10605921201c53.  This also ensures that the links are hidden at the correct point in the exit sequence.

- Previously, we re-used the `LinkHints` object, so it was necessary to reset its state on exit.  For quite some time, we have not been reusing the object, so we do not need to reset its state: 0761521e45af560453abb10b35dfae410ae8bce9.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -43,8 +43,6 @@ class LinkHintsMode
   mode: undefined
   # Function that does the appropriate action on the selected link.
   linkActivator: undefined
-  # Lock to ensure only one instance runs at a time.
-  isActive: false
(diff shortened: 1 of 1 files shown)


smblott-github:
Merging.
We get 20 fewer LoC, a cleaner exit sequence on "click", and it fixes some incorrect exit sequences with filtered hints.

See also #2032 (proposing a larger refactoring of link hints).

PR #2030, opened 2016-03-03
move keyboard shortcut dialog to right side

brandonjp:
This anchors the keyboard shortcuts help dialog / modal to the right side of the page (was previously centered), so that it is out of the way when you're trying to edit shortcuts

image

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -163,12 +163,14 @@ div#vimiumHelpDialog {
   border:2px solid #b3b3b3;
   border-radius:6px;
   padding:8px 12px;
-  width:640px;
+  width: 45%;
+  max-width:640px; 
   max-height: calc(100% - 80px);
-  left:50%;
+  /* left:50%; */
   /* This needs to be 1/2 width to horizontally center the help dialog */
-  margin-left:-320px;
-  top:50px;
+  /* margin-left:-320px; */
+  top: 2%;
+  right: 1%;
   -webkit-box-shadow: rgba(0, 0, 0, 0.4) 0px 0px 6px;
   overflow-y: auto;
 }


smblott-github:
Thanks, @brandonjp.  Interesting idea.  Some comments...

Your idea makes sense on the options page, but not anywhere else.  In fact, it looks pretty weird on normal pages.  Also, it looks weird on the options page too, when that's opened in a half-width window.

Have you tried clicking in the *Custom key mappings* box with the help dialog active?  The help dialog hides, which kind of defeats your intention.  Although I guess that could be changed.

@philc has strong views on how these aspects of the Vimium UI look.  He'd have to be on board.

(Off topic:  @mrmr1993... Try the following. Open any old page, and open the help dialog.  Try `gf` a couple of times.  The frame focus works, but the "flash" is happening in the main frame for the help dialog.  Any ideas?)

smblott-github:
Thanks, @brandonjp.  Sorry, but I don't think we should do this.  However, inspired by your use case, we now have #2031, which can help avoid the need to type command names.

brandonjp:
@smblott-github - Thanks for the reply.  That all makes sense, so no prob.  

A) I didn't even realize you *could* open the help view outside the options page.  (I've just started digging in... so thanks for that tip!) 

B) I'd guess screen size is probably an important factor.  I've also set a `min-width` on mine, which helped combat the weirdness ([*gif](https://gyazo.com/9b32db3f2cf7e72582187468fd5124b8))

C) Side note: Weird.. I've never once had the help dialog disappear when I click away from it.  I've changed nothing but a couple css lines in my local copy.  Not sure why it's not functioning as intended, but I quite prefer it "broken"  :)  https://gyazo.com/8566b7a1ccc18a83d33360f677940efe

smblott-github:
> I've never once had the help dialog disappear when I click away from it.

My mistake.  I'm running off `master` which does behave as described (for now).  However, we'll need to think about your use case.

(@mrmr1993... As things stand in `master`, the user cannot type into the `keyMappings` box on the options page with the help dialog open.  We probably need to fix that.)

PR #2031, opened 2016-03-04
Make command names in the help diablog clickable.

smblott-github:
Clicking on command names in the help dialog copies the command name to the clipboard.  Moreover, command names can be clicked (hence copied) with link hints.

Looks like this, ... here, I've already typed the text `vomn`, ...

snapshot

~~As a a side effect, any element on any page with the `vimiumClickable` class becomes clickable.  That's zero elements on zero pages, currently.  But there was an issue some time ago where a web developer was looking for a way to make their elements vimium clickable (#1157).~~ Edit... No longer true (58e332fe942d01c8f4633b3415c460f678b2c75b).

On the negative side, this is pretty undiscoverable.  Also, the `indexOf()` class-name test within a string isn't exactly precise.

### DIFF
--- a/pages/help_dialog.coffee
+++ b/pages/help_dialog.coffee
@@ -42,7 +42,7 @@ HelpDialog =
     # clicked with link hints).
     for element in @dialogElement.getElementsByClassName "commandName"
       do (element) ->
-        element.classList.add "vimiumClickable"
+        element.setAttribute "role", "link"
         element.addEventListener "click", ->
           commandName = element.textContent.replace("(","").replace ")", ""
           chrome.runtime.sendMessage handler: "copyToClipboard", data: commandName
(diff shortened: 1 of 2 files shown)


pages/help_dialog.coffee
```diff
@@ -38,6 +38,16 @@ HelpDialog =
 
     @showAdvancedCommands(@getShowAdvancedCommands())
 
+    # When command names are shown, clicking on them copies their text to the clipboard (and they can be
+    # clicked with link hints).
+    for element in @dialogElement.getElementsByClassName "commandName"
+      do (element) ->
+        element.classList.add "vimiumClickable"
```
mrmr1993: Why not use the actual standard way to mark the element as a link (`element.setAttribute "role", " link"`) rather than create a new one?

smblott-github: Ah,  OK.  I didn't know that that's how it works.  Thanks.

smblott-github:
Merging.
Uncontroversial, I think.  And pretty much self contained.

PR #2033, opened 2016-03-04
Do not initialise UI components when disabled.

### LINKED ISSUE #1838: Can Vimium be more friendly to users of Chrome Dev Tools?
I haven't done much research on this, but when I'm using the dev tools to work on non-extension web apps, it's annoying that Vimium adds many entries to the network requests tab.

Workarounds for this that I'm familiar with are to filter out requests from extensions by setting `-scheme:chrome-extension` to your request filter, or using Incognito mode for development. Both of those are annoying when you're just having a one-off look at the network tab.

This review on the Chrome store resonated with me:

> Everything great, love it except that js files are not concatenated, it pollutes network log with 9 files for each frame.

The ideal solution is probably for Chrome to have a persistent checkbox to "hide requests from extensions" in the dev console. Barring that, any other ideas for ways to improve the situation? What do you think about concatenating JS in production builds?

smblott-github:
We delay initialising UI components until we know that Vimium is enabled.  In addition to being a reasonable thing to do, this suppresses the resulting `XMLHttpRequest` requests on pages where Vimium is disabled, and hence doesn't clog up the dev console with useless Vimium network requests.

The idea is that devs for whom these requests are a problem can simply disable Vimium.

Fixes #1838 (mostly).  Devs have to disable Vimium on pages where they're working.  But Vimium can be enabled and disabled on a page (via the popup) with just a couple of keystrokes.

### DIFF
--- a/content_scripts/hud.coffee
+++ b/content_scripts/hud.coffee
@@ -13,9 +13,10 @@ HUD =
   # it doesn't sit on top of horizontal scrollbars like Chrome's HUD does.
 
   init: ->
-    @hudUI = new UIComponent "pages/hud.html", "vimiumHUDFrame", ({data}) =>
-      this[data.name]? data
-    @tween = new Tween "iframe.vimiumHUDFrame.vimiumUIComponentVisible", @hudUI.shadowDOM
+    unless @hudUI?
+      @hudUI = new UIComponent "pages/hud.html", "vimiumHUDFrame", ({data}) =>
+        this[data.name]? data
+      @tween = new Tween "iframe.vimiumHUDFrame.vimiumUIComponentVisible", @hudUI.shadowDOM
 
   showForDuration: (text, duration) ->
     @show(text)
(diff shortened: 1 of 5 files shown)


es6Test:
sweet!

smblott-github:
@mrmr1993.  Could you take a look at this, please?  In particular, because we delay initialising the Vomnibar (in the top/main frame) until we know it's needed, I think there's a race condition.  See this commit message 3542db7b6c322d803c263db641ae0b02327447ca.

Basically, is it possible that the message to initialise the vomnibar is delivered to the top frame before the top frame is listening?

Is there something which we know to be always true which prevents this from happening?

smblott-github:
OK, @mrmr1993.  Sorry to bother you.  I think this does it: 5584c8222608908b124d845a59aa445b7d4d0598.

mrmr1993:
Could we not just add an option for lazy loading of the Vomnibar?

smblott-github:
> Could we not just add an option for lazy loading of the Vomnibar?

I'm not in favour of that, @mrmr1993.  This is probably only worth fixing at close-to-zero cost.

leoj3n:
@smblott-github mentions that

> Vimium can be enabled and disabled on a page (via the popup) with just a couple of keystrokes

See how at https://github.com/philc/vimium/issues/2154#issuecomment-224245127

Issue #2032, opened 2016-03-04
Link hints, a larger refactor


smblott-github:
@mrmr1993.  I'm about to merge #2029 (a small link-hints refactor).

However, the link-hints code is in need of a larger refactor. Specifically, each mode should be an object with:

- a `name` (as currently)
- a `linkActivator` method
- a `filter` method (to reject some clickable elements)
- (possibly) an `isClickable(element)` method (to pick elements which would wouldn't otherwise be clickable).  @mrmr1993 - you have an example of such a thing in your link-hints for frames PR.

The idea is that we do this now, and freeze other link-hints activity until it's done.  The work should be a no-op functionality wise.  It may add LoC, but not new logic.

Is that something you'd be interested in contributing, @mrmr1993?

mrmr1993:
Yes, I can do that.

PR #2036, opened 2016-03-05
Move link hint activators into the link hint mode objects

mrmr1993:
This is the first part of a resolution to #2032, moving `linkActivator`s into the link hint mode objects.

@smblott-github I've taken some liberties with activators that use simulated clicks, only storing the modifiers in the modes and generating the appropriate activator on the fly. Is that okay, or should they be explicit activators?

(See also 97c98649a0dc28ce040a1cbe512d8f023499335c, which moves the HUD indicator text into the mode objects, but is outside the scope of #2032.)

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -10,13 +10,20 @@
 #
 # The "name" property below is a short-form name to appear in the link-hints mode's name.  It's for debug only.
 #
-OPEN_IN_CURRENT_TAB = name: "curr-tab"
-OPEN_IN_NEW_BG_TAB = name: "bg-tab"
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @mrmr1993.  Should the indicators not be in the mode objects too?

mrmr1993:
@smblott-github 97c98649a0dc28ce040a1cbe512d8f023499335c, but I didn't include it here in case you wanted to keep on spec with #2032.

smblott-github:
OK. Shall I just merge 97c98649a0dc28ce040a1cbe512d8f023499335c?

mrmr1993:
Please

smblott-github:
Done.

PR #2037, opened 2016-03-05
Help dialog to top frame

smblott-github:
This moves the help dialog such that it is always shown in the main/top frame, just like the Vomnibar.

**Edit: The following issue has been fixed**... 7c78211bdcc4e92acdc9ab66b1ebe11f707e1c93.  Leaving the comment below, though, for the record.

This works.  But...

There is a problem.  When a UI component closes, Chrome focuses some other frame.  Previously, (with just the Vomnibar) we could predict that it would be the main/top frame, and we could wait for that frame to receive the focus before re-focusing the original frame.

Here, however, we can open the Vomnibar from within the help dialog.  In this case, if we then close all of the frames with `Esc`, the focus ends up in the (hidden!) help dialog.  This is not actually a new problem.  It exists in `master` too.

In the first commit here, there is a big "FIXME".  We get around the issue by waiting an aribitrary short time (150ms) before sending the request to re-focus the original frame.  So there's a pretty serious race condition.  Also, during that waiting period, keyboard events are being sent to some unknown frame.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -298,7 +298,7 @@ defaultKeyMappings =
 # If the noRepeat and repeatLimit options are both specified, then noRepeat takes precedence.
 commandDescriptions =
   # Navigating the current page
-  showHelp: ["Show help", { background: true }]
+  showHelp: ["Show help", { topFrame: true }]
   scrollDown: ["Scroll down", { passCountToFunction: true }]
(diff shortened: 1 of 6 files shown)


smblott-github:
OK.  With 7c78211bdcc4e92acdc9ab66b1ebe11f707e1c93 (#2038), this seems to be working fine.

PR #2040, opened 2016-03-06
Resize page icons to 48x48.

### LINKED ISSUE #2035: It does not work fine on codeforces.com.
The icon on the toolbar become gray. And on some webpages, such as [http://codeforces.com/problemset/problem/635/A](http://codeforces.com/problemset/problem/635/A), it doesn't work.
OS version:Windows 10 x64
Vimium Version:1.54

smblott-github:
Chrome 49.0.2623.75 won't accept our 300x300 page icons, so the page icon stays greyed out, despite Vimium being active.

This resizes those icons to 48x48 (and keeps a copy of the original 300x300 icons).

@philc... We might need to push a new version for this.  Basically, the page icon isn't updating (it stays greyed out).  We could just cherry pick the commit here (4a4ea96dbb5a77b0539fd745f00066d7fd141f47) onto [this branch](https://github.com/philc/vimium/tree/1.54). (`master` isn't really good to go right now.)

Fixes #2035.


philc:
So they just decided to regress this? That's annoying. Does 48x48 cover retina resolution for high dpi displays?

On vacation ATM but I can push to store of needed

smblott-github:
@philc.  I'll experiment with the largest version of the icon that I can get to work, then prepare a branch to release, then let you know.  Your call vis-a-vis the priority of doing this.  It's a pretty bad bug, and Chrome 49 is very close.  I guess it could wait a week or two, though.

smblott-github:
Changed to 128x128.  That's the best resolution that seems to work.
Merging.
Instructions for possible 1.55 release to follow in new issue.

PR #2044, opened 2016-03-06
Make [[ and ]] also inc/dec URL digits.

### LINKED ISSUE #664: Added increment/decrement commands
This was implemented in pull request #634, but we've decided not to merge it for now.

smblott-github:
If `[[` or`]]` doesn't find anything to follow, then try incrementing or
decrementing any digits found in the URL instead.

This implements an idea from #664.

We should either do this (or similar) or close that issue.

Fixes #664.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -620,7 +620,6 @@ findAndFollowRel = (value) ->
 # If count is 1, then navigate from http://example.com/issue/4/x to http://example.com/issue/5/x.
 changeUrlDigits = (count) ->
   chrome.runtime.sendMessage {handler: "getCurrentTabUrl"}, (url) ->
-    regexp = /(.*)([0-9]+)([^0-9]*)/
     if 0 == url.search /(.*[^0-9])([0-9]+)([^0-9]*)/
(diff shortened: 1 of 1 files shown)


smblott-github:
Is there any interest in this?  I'm not hugely enthusiastic myself, so - absent support - I'll just kill it off.

maxauthority:
I think the idea is great and actually reusing ]] to just go to anything "next" would also support that use case for many things like image galleries or github issues where you don't have a "next" string. 
An alternative would be to use the potential that ]] was desigend as a two-key stroke, so one could use ]u and [u to go to the next/previous url based on the numbers (so using ] and [ as a general way to navigate to new urls based on the current page/url).

gdh1995:
We may also use an number-increased url to verify urls found by `next/previous`.

PhilippWendler:
I would like to see this added. It was present in vimperator, and after switching to vimium, this is the main feature I find missing. I use it for example with my continous integration system, to jump to the previous or next build to compare something.

boris-petrov:
@smblott-github - I also think this is great functionality and I've been missing it since I migrated from vrome. Why not merge it? I don't see any negative impact this could have - it's a feature that many people will use. Would you consider reopening a new PR?

yogi81:
Please add it !

yarikoptic:
I would have loved to see it too

Issue #2045, opened 2016-03-08
Help dialog issues (release blockers)


smblott-github:
There are two issues with the help dialog...

1. The help dialog closes if the user clicks outside of it (that is, elsewhere on the page).  This makes it not possible to enter new command mappings (on the options page) with the help dialog open (e.g. to read the command names).

2.  Try this... `?`, `gf`, `Esc`.  Previously, this closed the help dialog.  Now, with the help dialog in an iframe (and in part due #2022), the help dialog doesn't see the `Esc`, so doesn't close.

We need to fix these things.

Edit.  Not a show stopper, but the way in which the help dialog is populated incrementally at the moment is ugly.

smblott-github:
Re-opening.  Even with #2066, we still can't type into the key-mapping input on the options page with the help dialog open.

PR #2048, opened 2016-03-12
Global link hints

smblott-github:
This implements global link hints (so, link hints can be used to activate hints across *all* frames).

Observation:

- `LinkHintsMode` has very little to do with links.  It does three things: 1. find clickable things, 2. show hints and pick one, 3. "click" the thing we picked.
- 1. and 3. happen at the start and end.  Everything in between (2.) has *nothing* to do with links.  We're just picking something.  And that's where much of the link-hints logic is.
- Therefore, we can equally pick things which are not in the current frame.  So, we can pick hints from the current frame along with hints from other frames all at the same time.  We just don't display the hints from other frames.
- Therefore, if we give all frames *exactly the same list of things*, and run `linkHintsMode` in all frames, and feed them all the same key state, then they will all end up picking the same thing.
- When we pick a thing, we activate it in its own frame and silently exit in all other frames.

So that's the approach here.  We generate a list of hint descriptors (frame Id, rect, and for filtered hints the link's text and "show" state), and pass that list to a `LinkHintsMode` instance in _every frame_.  Each frame has everything it needs to know about every hint in every frame.  We then run those link-hint modes in lock step until a hint descriptor is chosen.

Use cases:

- You can pick conversations from the Hangouts frame in Google Inbox while the main frame has the focus.  (I haven't tried GMail, but it's probably the same.)
- You can select the little iframe popups that appear in Google inbox (e.g., "Confirm Sweep", "Undo Discard").

Test pages: [Google Inbox](https://inbox.google.com/u/0/), a [frames](http://www.quackit.com/html/templates/frames/frames_example_5.html) test page.

To do:

- ~~Fix tests.~~
- Many of the variable names and comments are out of date.  However, I've left them as is for now to keep the diff as simple as possible.
- There is some code which is in the wrong place.  In particular, the code for finding clickable things is physically in the middle of the (old) link-hints class.  This needs to be moved out.  However, here we go through some gymnastics to avoid doing it right now, since it would make the diff unnecessarily difficult to read.
- ~~It's broken on the help dialog, but that looks like another artefact #2045.~~

Background... I've tried several approaches to implementing global link hints (as have others, #1182).  The problem is that they all become pretty complicated very quickly, and end up having a massive footprint.  The observation at the top of this post seems to point the way to the a relatively simple implementation.  It's still complicated, but probably more manageable.

Edit... More "to do":

- ~~Reinstate filtering for hints with hrefs for modes which only use those.~~
- ~~Reinstate Shift/Ctrl to toggle hint mode.~~

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -322,25 +322,26 @@ cycleToFrame = (frames, frameId, count = 0) ->
 HintCoordinator =
   tabState: {}
 
-  onMessage: (request, sender) ->
+  onMessage: (request, {tab: {id: tabId}}) ->
     if request.messageType of this
-      this[request.messageType] extend request, tabId: sender.tab.id
+      this[request.messageType] tabId, request
     else
-      # The message is not for us.  It's for all frames, so we bounce it there.
-      @sendMessage request.messageType, sender.tab.id, request
+      # If there's no handler here, then the message is bounced to all frames in the sender's tab.
+      @sendMessage request.messageType, tabId, request
 
   sendMessage: (messageType, tabId, request = {}) ->
     chrome.tabs.sendMessage tabId, extend request, {name: "linkHintsMessage", messageType}
 
-  activateMode: ({tabId, frameId, modeIndex}) ->
-    @tabState[tabId] = {frameIds: frameIdsForTab[tabId], hints: [], modeIndex, frameId}
-    @sendMessage "getHints", tabId
+  prepareToActivateMode: (tabId, {frameId: originatingFrameId, modeIndex}) ->
+    @tabState[tabId] = {frameIds: frameIdsForTab[tabId], hintDescriptors: [], originatingFrameId, modeIndex}
+    @sendMessage "getHintDescriptors", tabId
 
-  postHints: ({tabId, frameId, hints}) ->
-    @tabState[tabId].hints.push hints...
+  # Receive hint descriptors from all frames and activate link-hints mode when we have them all.
+  postHintDescriptors: (tabId, {frameId, hintDescriptors}) ->
+    @tabState[tabId].hintDescriptors.push hintDescriptors...
     @tabState[tabId].frameIds = @tabState[tabId].frameIds.filter (fId) -> fId != frameId
     if @tabState[tabId].frameIds.length == 0
-      @sendMessage "activateLinkHintsMode", tabId, @tabState[tabId]
+      @sendMessage "activateMode", tabId, @tabState[tabId]
       delete @tabState[tabId] # We won't be needing this any more.
 
 # Port handler mapping
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -323,14 +323,14 @@ HintCoordinator =
   tabState: {}
 
   onMessage: (request, sender) ->
-    if request.name of this
-      this[request.name] extend request, tabId: sender.tab.id
+    if request.messageType of this
+      this[request.messageType] extend request, tabId: sender.tab.id
     else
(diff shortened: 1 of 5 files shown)


smblott-github:
This is fairly stable, now. Two risks...

1. We rely on knowing the frameIds of all frames, and wait for responses from all frames.  If a frame disappears and we don't learn about it, then we'd block indefinitely.  I've not seen this happen, and it seems to be fundamental to all approaches to global link hints.  We could mitigate this by bailing and proceeding with whatever hints we have after a timer expires (e.g. 300ms).  Or, we could maintain a separate port for link hints.  Then, we'd probably get an `onDisconnect` callback, which should be accurate.  (In fact, this might be a better method of maintaining frame state anyway.)

2. With floating iframes, we see hints for links underneath the iframe.  This applies to the help dialog, and also on Google Inbox (example below).  We could encounter cases where the number of hints is just overwhelming and confusing.  (Nevertheless, in the Google Inbox example and with filtered hints, `f`, `un`, `Enter` undoes an operation beautifully.)

snapshot

smblott-github:
Note 5b347e706a055d2c48ca3cbd26aad41c7fea3625.  This makes frames selectable (focusable) with link hints.

smblott-github:
I intend to merge this soon (e.g. at the weekend) unless an issue emerges (@mrmr1993, @philc):

- It seems to be working well in practice.
- It's a big win functionality wise (it already feels weird going back to single-frame hints).

Although it has a fairly big footprint, there's really only about 70 or 80 extra LoC, and some of those relate to peripheral additional functionality (e.g. 5b347e706a055d2c48ca3cbd26aad41c7fea3625 adds ten lines to make frames selectable with link hints).

(Separately, I'm working on an extended version of #2053 which mitigates my concern re. our dependence on `frameIdsForTab` being maintained correctly.)

philc:
This is awesome and long-overdue! I'll try to take a look at the
implementation this weekend.

smblott-github:
Hey Phil,

I guess you didn't get a chance to look at this.

Should I hold off on merging?

Steve

smblott-github:
@philc... Any input on this?  Otherwise merging v. soon (e.g. tomorrow morning, local time).

philc:
Ship it!

jesusvazquez:
Hello , I just installed the extension and I love it. I'm using this servicenow portal which is full of iframes and the only links that get hints are those within the iframe that has the focus. Is there anything i can do to help? Is this feature available in the stable release I can find in the chrome extensions gallery?

smblott-github:
Thanks, @jesusvazquez.  This feature is not yet in the Chrome Store release.  I expect there'll be a new release in the coming 5/6 weeks.

jesusvazquez:
Thanks a lot for your fast response. Is there a way to install the latest release (master) ? I can give you feedback while using it.

smblott-github:
@jesusvazquez ... You can [install from source](https://github.com/philc/vimium/blob/master/CONTRIBUTING.md#installing-from-source), or try [my working copy](https://drive.google.com/open?id=0B-PWDBroAYxrNFJQaHFBbDVuMkU).  But, in either case, you won't receive updates automatically.

jesusvazquez:
Thank you, I'll give it a try and I'll let you know.

jesusvazquez:
Oh god this is awesome! It works perfectly! Good job man

anilanar:
Any updates on this feature making its way to Chrome Web Store?

ProbonoBonobo:
@smblott-github This is fantastic. I'd also be interested to know if you have an update re: Chrome Web Store. Chrome keeps disabling it in an attempt to protect me...

PR #2052, opened 2016-03-13
Add options-page button to sort exclusion rules.

smblott-github:
After a while, the list of exclusion rules on the options page can become quite long, and it can be difficult to find the rule (or rules) pertaining to a particular site.

This adds a button to the options page to sort the rules.  The sort order currently is...  strip any leading `protocol://`, then strip any leading `www.`, then sort what's left.  So `https://github.com` and `http://www.github.com` sort beside each other.

It currently looks like this...

snapshot

(Perhaps it would look better if the buttons were of equal width?)

### DIFF
--- a/pages/options.coffee
+++ b/pages/options.coffee
@@ -133,8 +133,8 @@ class ExclusionRulesOption extends Option
     rules = @readValueFromElement()
     for rule in rules
       key = rule.pattern
-      key = RegExp.$1 if 0 == key.search /[a-z?*]+:\/\/(.*)/
-      key = RegExp.$1 if 0 == key.search /www\.(.*)/
+      key = RegExp.$1 if 0 == key.search /^[a-z?*]+:\/\/(.*)/
+      key = RegExp.$1 if 0 == key.search /^www\.(.*)/
       rule.key = key
 
     rules.sort (a,b) -> if a.key < b.key then -1 else if b.key < a.key then 1 else 0
--- a/pages/options.coffee
+++ b/pages/options.coffee
@@ -80,6 +80,8 @@ class ExclusionRulesOption extends Option
     super(args...)
     $("exclusionAddButton").addEventListener "click", (event) =>
       @addRule()
+    $("exclusionSortButton").addEventListener "click", (event) =>
+      @sortRules()
 
   # Add a new rule, focus its pattern, scroll it into view, and return the newly-added element.  On the
   # options page, there is no current URL, so there is no initial pattern.  This is the default.  On the popup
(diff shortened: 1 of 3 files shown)


smblott-github:
Do this?  Don't do this?

I'm leaning towards do this (and will do it soon, unless somebody raises an issue or alternative proposal, e.g. @philc).

(Actually, now I think about it, an alternative would be just to *always* sort the exclusion rules when the options page loads -- that might be better.)

pages/options.coffee
```diff
@@ -80,6 +80,8 @@ class ExclusionRulesOption extends Option
     super(args...)
     $("exclusionAddButton").addEventListener "click", (event) =>
       @addRule()
+    $("exclusionSortButton").addEventListener "click", (event) =>
```
gdh1995: Ah, please use `$("...")?.addEvent...` to fix the crash on the popup page.

pages/options.coffee
```diff
@@ -127,6 +129,19 @@ class ExclusionRulesOption extends Option
...
+    for rule in rules
+      key = rule.pattern
+      key = RegExp.$1 if 0 == key.search /^[a-z?*]+:\/\/(.*)/
+      key = RegExp.$1 if 0 == key.search /^www\.(.*)/
+      rule.key = key
+
+    rules.sort (a,b) -> if a.key < b.key then -1 else if b.key < a.key then 1 else 0
+    @element.innerHTML = ""
+    @appendRule rule for rule in rules
+    @onUpdated()
```
gdh1995: Maybe we should change the button's text to indicat we have done this work.

My exclusion panel has only 2 rules and both urls are too long for the pattern inputs. So I could not see any change after I clicked it.

philc: Having a sort button feels like we're offering too much of an
implementation detail to the user. Chrome's "custom search engines"
preferences page sorts everything by default. What do you think of doing
that?

smblott-github: @gdh1995.  Thanks for your input.  Your point re. the popup page (above) is a great catch.

Yes, @philc.  Just "always sort" seems best.

@gdh1995.  Have you thought at all about what the comparison function should be?  Straight string comparison?  What's here (which is a bit of a hack)?  Something better?

gdh1995:
You sort function is very suitable, but says there're:
* 1. `http://www.yahoo.com/` => `yahoo.com`
* 2. `http://i.yahoo.com/` => `i.yahoo.com`
* 3. `http://www.amazon.com/` => `amazon.com`
* 4. `http://i.amazon.com/` => `i.amazon.com`

And then, we get `1 < 2` but also `3 > 4` - this sort function may be not "stable" enough.

A better way is to `.split(".").reverse().join(".")`, though it's a little costly.
* For domains having no `"."` (e.g. written in `hosts`), I think we should leave them bottom.

gdh1995:
Well, after many times of forgetting and delaying, I succeed in finding where Chrome compare two exclusion patterns: The comparator is `ContentSettingsPattern::Relation ContentSettingsPattern::Compare(const ContentSettingsPattern& other) const;`, in `components/content_settings/core/common/content_settings_pattern.cc`.

* the order pattern's parts are compared is: `host` > `port` > `schema` > `path`
* what we may want are `IsSubDomainOrEqual` and `CompareDomainNames` at the top of the cc file
* `IsSubDomainOrEqual` is an `endsWith` with one more check of `[index - 1] == '.'`
* `CompareDomainNames` splits both hosts and compare items in reverse order.

### Here's how to track a rule change

If we changes an exception rule:
* The JavaScript code in `chrome://settings/content` calls
  * `chrome.send('setException', [this.contentType, this.mode, newPattern, newSetting]);`
  * or `chrome.send('removeException', ...);`. Says it sends `setException`, and then:
* The command is received by ContentSettingsHandler::SetException in `chrome/browser/ui/webui/options/content_settings_handler.h`
* the handler calls `settings_map->SetContentSettingCustomScope`, in which `HostContentSettingsMap::SetWebsiteSettingCustomScope` is called
* `SetWebsiteSettingCustomScope` searchs for a `provider` to deal with the change action
* `ContentSettingsPref::SetWebsiteSetting` updates `map_to_modify` (`OriginIdentifierValueMap*`), and may call `UpdatePref` to save changes into Chrome's preference file.
* then, changed `value_map_` or `incognito_value_map_` is read to show on screen

The class `OriginIdentifierValueMap` is in `components/content_settings/core/browser/content_settings_origin_identifier_value_map.h`, its iterator will sort patterns first by `primary_pattern` (its type is `ContentSettingsPattern`) and second by `secondary_pattern`.

PR #2053, opened 2016-03-14
Use port for frames

smblott-github:
This is a bundle of small changes related to managing frames and frame Ids.

**Register/unregister**.  Use the `domReady` port connect/disconnect events to register and unregister frames.  As far as I know, the existing logic is sound.  However, it's always looked a bit flakey (e.g., the fact that [this](https://github.com/philc/vimium/blob/1.54/background_scripts/main.coffee#L687) is needed).  (This is related to point 1 in  [this comment](https://github.com/philc/vimium/pull/2048#issuecomment-196176731).)

**Framesets**.  Unlikely previously, here we *do* register framesets (and skip over them in `focusThisFrame` instead).  This just seems like a better approach.  It means that `frameIdsForTab` *really is* a list of all of the frame Ids in which we're running.

**Frame Ids**.  This changes from using a random frame Id (which has always seemed dodgy), to using Chrome frame Ids -- which are guaranteed to be unique.  Also, a Chrome frameId of `0` is the top frame, which might prove helpful.

Going forward, I can see us moving more messages onto the port (e.g. `checkIfEnabledForUrl`) and possibly reducing our dependence on our own `frameId`.  And there are probably other ways in which the frame-handling logic can be simplified.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -423,7 +423,7 @@ bgLog = (request, sender) ->
 # Port handler mapping
 portHandlers =
   completions: handleCompletions
-  domReady: Frames.onConnect.bind Frames
+  frames: Frames.onConnect.bind Frames
 
 sendRequestHandlers =
   runBackgroundCommand: runBackgroundCommand
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -383,14 +383,13 @@ Frames =
     (frameIdsForTab[tabId] ?= []).push frameId
     port.postMessage name: "registerFrameId", chromeFrameId: frameId
 
-    port.onDisconnect.addListener ->
-      if frameId == 0 # This is the top frame in the tab.
-        delete frameIdsForTab[tabId]
-      else
-        if tabId of frameIdsForTab
+    port.onDisconnect.addListener listener = ->
+      if tabId of frameIdsForTab
+        if frameId == 0 # This is the top frame in the tab.
+          delete frameIdsForTab[tabId]
+        else
           frameIdsForTab[tabId] = frameIdsForTab[tabId].filter (fId) -> fId != frameId
-          # Sub-frames can connect before the top frame; so we can't rely on seeing the top frame at all.
-          delete frameIdsForTab[tabId] if frameIdsForTab[tabId].length == 0
+      port.onDisconnect.removeListener listener
 
 handleFrameFocused = (request, sender) ->
   [tabId, frameId] = [sender.tab.id, sender.frameId]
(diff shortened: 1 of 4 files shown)


smblott-github:
Here are two issues with how we handle frame Ids, both here and in `master`.

**One**.  We invoke `checkIfEnabledForUrl()` in every frame, and the response is broadcast to all frames.  Therefore, on start up, we have O(n^2) messages for n frames in a tab.  Google likes tiny iframes, some pages have on the order of 13 to 14 iframes.  So we're creating quite some unnecessary and non-negligible messaging overhead on start up.  Therefore, we should seriously consider moving `checkIfEnabledForUrl()` to a port.

**Two**.  There are possible race conditions here and in `master`.  When, within a tab, we navigate from one frame to another, Chrome re-uses the tab Id.  We unregister all frames of the old page, and then register all frames of the new page.

Because of [this line](https://github.com/philc/vimium/blob/1.54/background_scripts/main.coffee#L613), we remove the frame entry entirely when the top from closes.  The equivalent in this PR is [here](https://github.com/philc/vimium/pull/2053/files#diff-eb68fda05b0a59fbd1482418fd0223e8R388).  We can rely on getting the necessary messages eventually, but it's not clear that we can rely on getting them in the correct order.  If the "unregister" message for the previous page's top frame arrives after the first "register" message for the new page, then `frameIdsForTab` ends up incorrect (we're missing at least one frame Id).

In `master`, the effect is just that `gf` skips a frame.  With #2048, the effect would be more serious, link hints wouldn't work in the missing frame, which is probably most likely to be the top/main frame -- oops!

A proposed fix for this is 5e546c9a058ed65c9e03b79659d3ae46dc3094a1 (which is not yet in this PR, or anywhere else important).  It takes advantage of the fact that the frame Id for the top frame is always 0.  I think we should go ahead with that "fix".

smblott-github:
This has become a considerably larger endeavour.  The original motivation was to ensure that `frameIdsForTab` was correct, because #2048 depends on that.  Now:

- Use Chrome's frame Ids instead of our own.  So there's now no possibility of duplicates.
- Eliminate a possible race condition in how frames are registered and unregistered when a tab navigates from one page to another.
- Get rid of the `domReady` port (which wasn't used much).
- Add a new `frames` port.
- Handle the `domReady` state change via the new `frames` port, and simplify the logic for the `domReady` callbacks (for marks).
- Handle `checkIfEnabledForUrl()` via the `frames` port.  This takes the messaging overhead from O(n^2) on start up to O(n).  And from O(n) on focus change to O(1).  The benefit of moving other communication to the `frames` port would be marginal, so it's probably not worth doing.  There aren't any O(n^2) issues elsewhere.
- We register *all frames* with the background page, and filter out frame sets in `focusThisFrame()`.  This seems to make more sense, since the frame-set issue really only applies to `focusThisFrame()`.

PR #2054, opened 2016-03-16
Implement yanking tab url in markdown form. Mapped to 'ym' be default.

issmirnov:
Hello. Please review a small patch that adds the capability to yank a url to markdown form. I find myself often using `yy` and then manually adding the title - this will greatly speed up the workflow of all people working with markdown tools.

If you're happy with the code quality, docs message and default mapping please merge!

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -45,5 +45,6 @@ Contributors:
   Utkarsh Upadhyay <musically.ut@gmail.com) (github: musically-ut)
   Michael Salihi <admin@prestance-informatique.fr> (github: PrestanceDesign)
   Dahan Gong <gdh1995@qq.com> (github: gdh1995)
+  Ivan Smirnov <ismirnov@google.com> (github: issmirnov)
 
 Feel free to add real names in addition to GitHub usernames.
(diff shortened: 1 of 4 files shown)


smblott-github:
Thanks, @issmirnov.  My first reaction is that this is a bit niche, and - for me - I'd usually end up having to edit the title anyway.

However, things like this would be candidates for #1980 (e.g. the newly added 1de1f019e789b80df09cb3c60d2fb766faa5965f, there).

#1980 is just a proof of concept.  It does not contain a properly thought out API.  However, it would provide a way for users to write and share niche commands without polluting core Vimium with too many commands.

issmirnov:
Understandable. I was hoping this would be above that threshold, but you have a better statistic than I :)

What are the steps needed to make the plugin mechanism a reality? I'd be happy to contribute some time and code - there's a myriad of commands that people with a similar workflow to mine would likely use.

edrex:
This is an essential feature for users of a markdown-accepting note-taking system (vimwiki, Notion, ...). None of us have numbers, but I would guess that the intersection with vimium users is rather large. Is this really niche, in the context of vimium?

PR #2069, opened 2016-03-28
Link hints: false positives and scrollable divs

### LINKED ISSUE #425: Make it possible to focus a scrollable div without using the mouse

### LINKED ISSUE #2056: Link-hint false positives
Here's an idea for reducing the number of link-hint false positives.  See f6925630a7e22b4483bc872d9242776bce5337c1; and from there:

```
    Filter out link-hint false positives.
    
    We recognise elements with a class names containing the text "button" as
    clickable.  However, often they're not, they're just wrapping a
    clickable thing, like a real button.
    
    Here, we filter out such false positives.
    
    This has two effects:
    
    - It eliminates quite a number of real false pasitives in practice.
    - With fewer hints close together, fewer hint markers are obscured by
      the hints from (non-clickable) wrappers.  This reduces the need for
      rotating the hint stacking order, e.g #1252.
```

What do you think, @mrmr1993?

smblott-github:
This builds on #2048 (global link hints) adding two new features:

###### Filter out false positives

We mark any element with a class name containing the text `button` as clickable.  Many of these are wrappers around real clickables.  E.g....

```html
<span class="buttonWrapper"><input type="button"/></span>
```

Here, we treat the outer `span` as a false positive (because a close descendent is a real clickable).  This has two effects:

1. We get fewer useless hints.
2. Because we have fewer hints, it is less common that hint markers overlap, so the hint markers for the real clickables are less likely to be obscured by useless hint markers.

###### Make scrollable divs selectable

Here's an example (using filtered hints):

snapshot

Fixes #425.
Fixes #2056.

(I intend to merge this directly; this PR is just for visibility.)

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -189,6 +189,7 @@ LocalHints =
     tagName = element.tagName.toLowerCase()
     isClickable = false
     onlyHasTabIndex = false
+    possibleFalsePositive = false
     visibleElements = []
     reason = null
 
@@ -229,7 +230,6 @@ LocalHints =
     # Check for attributes that make an element clickable regardless of its tagName.
     if (element.hasAttribute("onclick") or
         element.getAttribute("role")?.toLowerCase() in ["button", "link"] or
-        element.getAttribute("class")?.toLowerCase().indexOf("button") >= 0 or
         element.getAttribute("contentEditable")?.toLowerCase() in ["", "contentEditable", "true"])
       isClickable = true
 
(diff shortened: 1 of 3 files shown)


vnikk:
hi, could you please comment on how to use it. I cant find any documentation on this and it's not clear from the code. thanks

smblott-github:
There's nothing that you (the user does).  The text `Scroll` appears if you're using filtered hints (under advanced options on the options page, *Use the link's name and characters for link-hint filtering*).

vnikk:
thank you. is it possible to apply without hint filtering?

smblott-github:
The hint is still there with alphabet hints, you just don't have the text "Scroll" next to it.

PR #2071, opened 2016-03-28
Remove Ctrl modifier for link-hints.

smblott-github:
In 1.54, we (I) accidentally broke the `Ctrl` modifier for link-hints activation.

And nobody has complained.

So, perhaps we should just remove that feature.

Now is you last chance to complain...


smblott-github:
Doh!
Sorry for the noise.

It's only broken for filtered hints.

PR #2072, opened 2016-03-28
Do not grab back focused input on URL change.

smblott-github:
Currently, if grab-back-focus is enabled, then we grab back the focus on a URL change -- including blurring any already focused input.

On some pages, e.g. Google Inbox, it is that act of focusing an input which triggers a URL change, and in such cases, we should not blur the input.

This is a refinement of #1622.  We still grab back the focus after a URL change if it is the page which focuses the input.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -39,14 +39,19 @@ bgLog = (args...) ->
 # If an input grabs the focus before the user has interacted with the page, then grab it back (if the
 # grabBackFocus option is set).
 class GrabBackFocus extends Mode
-  constructor: ->
+  constructor: (isUrlChange) ->
     super
       name: "grab-back-focus"
+      singleton: "grab-back-focus"
       keydown: => @alwaysContinueBubbling => @exit()
 
     @push
       _name: "grab-back-focus-mousedown"
       mousedown: => @alwaysContinueBubbling => @exit()
+
+    # If it's just that the URL has changed, then bail if an input is already focused.  On some pages (e.g.
+    # search on Google Inbox) it's the act of focusing the input which triggers the URL change.
+    @exit() if isUrlChange and document.activeElement and DomUtils.isEditable document.activeElement
 
     Settings.use "grabBackFocus", (grabBackFocus) =>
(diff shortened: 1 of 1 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -39,15 +39,20 @@ bgLog = (args...) ->
...
+      singleton: "grab-back-focus"
       keydown: => @alwaysContinueBubbling => @exit()
 
     @push
       _name: "grab-back-focus-mousedown"
       mousedown: => @alwaysContinueBubbling => @exit()
 
+    # If it's just that the URL has changed, then bail if an input is already focused.  On some pages (e.g.
+    # search on Google Inbox) it's the act of focusing the input which triggers the URL change.
+    @exit() if isUrlChange and document.activeElement and DomUtils.isEditable document.activeElement
```
gdh1995: I cann't understand this line. Here `document.activeElement` should still be an `<a>` node, shouldn't it?

For an `<input>` clicked, `return true if DomUtils.isFocusable document.activeElement` at line 78 should have stopped.

smblott-github: If `document.activeElement` happens to be an `<a>`, then 

```Coffeescript
DomUtils.isEditable document.activeElement
```

 is `false`, which is what we want.  Specifically, we *do not* want to exit in that case.

smblott-github:
Closing this.
On reflection, I don't think this is the source of the problem.  Here's the story...

I've seen cases where we get locked in grab-back-focus mode on Google inbox.  There's simply nothing you can do to select the "search" input.

I now think it's a race condition, and it's fixed here... 77a2c6b41959f978147bd953ac94f5484394b276 (hopefully).

Inbox is a pretty busy page, so likely to trigger race conditions.  What I think is happening is that we're seeing a `keydown` (and exiting grab-back-focus mode) *before* the settings are ready, so *before* we install the grab-back-focus `focus` listener.  So that listener keeps kicking us out of insert mode.  And, because the mode has already exited, the `keydown` and `mousedown` event listeners have already been removed, so there's nothing left to remove the `focus` listener.  Hence we're stuck.

PR #2089, opened 2016-04-09
Simplify javascript: URLs in vomnibar.

### LINKED ISSUE #961: Vomnibar: don't show the javascript contents of bookmarklets
We shouldn't show or index the javascript of bookmarklets.

screen shot 2013-11-22 at 2 54 15 pm

smblott-github:
Show `javascript:` URLs as `javascript:...` in the Vomnibar.

Fixes #961.


gdh1995:
In my option, it's somehow useless to match query words with a `javascript:` bookmarklet - so there may be more thing that can be done.

PR #2094, opened 2016-04-13
add pre tag for link_hints

MoenCertiK:
Will add specs if this small feature proved usable, thanks.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -620,9 +620,9 @@ LocalHints =
         isClickable ||=
           if element == document.body and document.hasFocus() and Scroller.isScrollableElement element
             reason = "Scroll."
-      when "div", "ol", "ul"
+      when "div", "ol", "ul", "pre"
         isClickable ||=
-          if element.clientHeight < element.scrollHeight and Scroller.isScrollableElement element
+          if (element.clientHeight < element.scrollHeight or element.clientWidth < element.scrollWidth) and Scroller.isScrollableElement element
             reason = "Scroll."
 
     # An element with a class name containing the text "button" might be clickable.  However, real clickables


MuhanZou:
https://jsfiddle.net/w9vasqf5/ for demo purpose

content_scripts/link_hints.coffee
```diff
@@ -625,6 +625,9 @@ LocalHints =
           if element.clientHeight < element.scrollHeight and Scroller.isScrollableElement element
             reason = "Scroll."
 
+      when "pre"
+        isClickable ||= !!(element.clientWidth - element.scrollWidth) || !!(element.clientHeight - element.scrollHeight)
+
```
smblott-github: @MuhanZou The code immediately above is doing the same thing, except for the y-axis only.  We should do this once only.  Probably we'd be better off generalising the code above rather than adding a new case.

MuhanZou: Got it, no problem. I will update the pr by your suggest to generalize the case with tests covered, thanks!

MuhanZou:
@smblott-github updated based on your comment, thanks

content_scripts/link_hints.coffee
```diff
@@ -620,9 +620,9 @@ LocalHints =
         isClickable ||=
           if element == document.body and document.hasFocus() and Scroller.isScrollableElement element
             reason = "Scroll."
-      when "div", "ol", "ul"
+      when "div", "ol", "ul", "pre"
         isClickable ||=
-          if element.clientHeight < element.scrollHeight and Scroller.isScrollableElement element
+          if (element.clientHeight < element.scrollHeight or element.clientWidth < element.scrollWidth) and Scroller.isScrollableElement element
```
smblott-github: Have you tested this, @MuhanZou?

The `isScrollableElement()` test only checks the `y` axis, I think.

MuhanZou: I tested with current prod version v.s. my local and it works...https://jsfiddle.net/w9vasqf5/1/
But yeah you are right the default direction is y, plz give me some time to find it out. BTW, I didn't see similar test cases for this, could you help advise on it? thanks!

philc:
Closing due to age.

PR #2101, opened 2016-04-17
Move help dialog to top frame

smblott-github:
Replaces #2037.


smblott-github:
Added two things to this:

- Do not initialize the HUD until it's needed.  On sites like GMail or Google inbox, this means we avoid initializing something like fifteen unused HUDs.

- Do not initialize the Vomnibar until it's needed.  This doesn't seem to introduce any sluggishness.

Both of these area easily rolled back if they turn out to be problematic.

PR #2107, opened 2016-04-19
Modernize help dialog styling

smblott-github:
Calling @philc...

The current help dialog styling is a bit cramped, and doesn't match the approach used by sites Google Inbox, GMail or Github.

This is a proposal for modernising the help dialog.  Here's what it looks like currently...

snapshot
 
The changes are:

- When the dialog is showing, the background page is darkened.  The color and opacity are taken from Google inbox.

- The background is white and the fonts are larger.

- The links, ("Wiki", "Options", etc) are all in a dark font. (They were blue before.)
- The dialog window in 200px wider (but still seems to work ok on narrower screens).

- The keys are no longer blue, but instead use a grey background like this: `j`.  The background color was taken from either Google Inbox or Github (can't remember).

- When command names are shown, they use a fixed-width font.  See below.  It looks like this was originally intended, just somebody got the CSS wrong.

- When the help dialog is showing, link hints only select links from within the help dialog.  This is a big win when selecting command names on the options page.  (Because the hint dialog isn't polluted with useless hints from the options page itself.)

On the options page, clicking *Show available commands*, it looks like this...

snapshot

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -118,27 +118,26 @@ Commands =
     pageNavigation:
       ["scrollDown",
       "scrollUp",
-      "scrollLeft",
-      "scrollRight",
       "scrollToTop",
       "scrollToBottom",
-      "scrollToLeft",
-      "scrollToRight",
       "scrollPageDown",
       "scrollPageUp",
       "scrollFullPageUp",
       "scrollFullPageDown",
+      "scrollLeft",
+      "scrollRight",
+      "scrollToLeft",
+      "scrollToRight",
       "reload",
-      "toggleViewSource",
       "copyCurrentUrl",
       "openCopiedUrlInCurrentTab",
       "openCopiedUrlInNewTab",
       "goUp",
       "goToRoot",
       "enterInsertMode",
-      "passNextKey",
       "enterVisualMode",
       "enterVisualLineMode",
+      "passNextKey",
       "focusInput",
       "LinkHints.activateMode",
       "LinkHints.activateModeToOpenInNewTab",
@@ -156,32 +155,33 @@ Commands =
     vomnibarCommands:
       ["Vomnibar.activate",
       "Vomnibar.activateInNewTab",
-      "Vomnibar.activateTabSelection",
       "Vomnibar.activateBookmarks",
       "Vomnibar.activateBookmarksInNewTab",
+      "Vomnibar.activateTabSelection",
       "Vomnibar.activateEditUrl",
       "Vomnibar.activateEditUrlInNewTab"]
     findCommands: ["enterFindMode", "performFind", "performBackwardsFind"]
     historyNavigation:
       ["goBack", "goForward"]
     tabManipulation:
-      ["nextTab",
+      ["createTab",
       "previousTab",
+      "nextTab",
       "visitPreviousTab",
       "firstTab",
       "lastTab",
-      "createTab",
       "duplicateTab",
+      "togglePinTab",
       "removeTab",
       "restoreTab",
       "moveTabToNewWindow",
-      "togglePinTab",
       "closeTabsOnLeft","closeTabsOnRight",
       "closeOtherTabs",
       "moveTabLeft",
       "moveTabRight"]
     misc:
-      ["showHelp"]
+      ["showHelp",
+      "toggleViewSource"]
 
   # Rarely used commands are not shown by default in the help dialog or in the README. The goal is to present
   # a focused, high-signal set of commands to the new and casual user. Only those truly hungry for more power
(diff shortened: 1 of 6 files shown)


smblott-github:
Update... Many small tweaks.  Here, showing command names (on the options page):

snapshot

- Add a tip that clicking on command names copies them.  This is only visible when we are actually showing command names.
- Cursor hover over command names now uses a pointer.
- Use the same font colour which we use for links for command names.  (Not whether this looks better or not.)  Do not use underlining: that's just too busy.
- Revert to italic font for command names (this looks better that a fixed-width font).

smblott-github:
There's a problem case with this.

First, open the help dialog on any old page.  Then make the window narrower and narrower.  The help dialog copes fairly well with pretty narrow windows.

Next, do the same thing after clicking "Show available commands" on the options page.  This time, the left-hand column ends up over-writing the right-hand one; like this:

snapshot

The problem is those *really-long* command names; they overflow their column.  It's not clear what we can do about this.  One options is to increase the minimum dialog width.  Currently it's `600px`; we'd have to increase that to `750px` to make the problem go away.

philc:
These are some nice and long-needed improvements. Nice work so far. Feedback:

* The darkening of the page behind the dialog is an excellent improvement!
* White as the background color is good; higher contrast, matches chrome://settings
* Making links look gray instead of blue is dubious. You can see in chrome://settings and other dialogs that the Chrome team uses blue for links.What's the motivation of making them less noticeable by removing the blue?
* I'm not too worried about the dialog being too wide when command names are showing; that's typically only used on the options page, and I think people can widen their browser if it's sized tightly. It would be nice if the overflow happened a little more gracefully (overflowing down rather than wide), but I don't know if there's an easy way to do that. We probably shouldn't have originally introduced such a long command name (`LinkHints.activateModeToOpenInNewForegroundTab`).
* Shouldn't we present keys with modifiers like `<c-e>` as `<C-e>`, etc.?
* All other changes you've made LGTM.
* The X button should be larger and vertically centered within the title bar, similar to the vertical alignment of the Options link. See Github's help dialog for reference.
* Making the shortcuts look like a little more like keys is a nice touch. How about giving them some depth by using the same styles for keys that the github help dialog uses? This should make our dialog much more attractive. Although note that our dialog is much more vertically dense than theirs, so we may need to adapt Github's styles rather than simply copying them.

image

smblott-github:
> What's the motivation of making them less noticeable by removing the blue?

The blue looks ugly to me, very 1990s.  But it's a matter of taste, so it's been reverted now.

I've fixed everything else you suggest, @philc, except:

> Shouldn't we present keys with modifiers like <c-e> as <C-e>, etc.?

To my taste, this would not be an improvement; it's unnecessarily loud.

Will merge pretty soon, if there's no further feedback.  We can always tweak details later anyway.

smblott-github:
Screenshots...

snapshot

snapshot

philc:
Looking good.

To make the blue less loud, you could use the same shade as the "Vim" in
the Vimium text in the header of this dialog.

I think using boxes with depth looks better than before, but now I realize
we have a short-coming in our help dialog which isn't present in Github's.
They are using one square per key typed, which is aesthetically nice and
maps cleanly to the visualize metaphor of keys on a keyboard. That visual
metaphor breaks down if you have multiple letters within one "key" (box),
as we do.

To take Github's styling all the way, we'd need to draw one box around each
character (e.g. `[y] [f]`), and in the case where multiple key bindings map
to a single command, separate each binding with a slash or the word "or",
e.g.: `[K] or [g][t]`. We'd also want each key to be fixed-width (so "w"
isn't wider than "i") to create nice alignments.

That will make this dialog wider. It may also be unnecessarily heavy. What
do you think?

(BTW after you merge, I may play with a few small line-height tweaks to see
how it affects the size of the dialog).

smblott-github:
Changes:

- Use Vimium blue for links (05b2289ab0bb2278a60d77c5e08b7603e49e014d).  That's considerably better, in my view.

- Use a fixed width font for keys (0b6e9b806df828864eac1fc3dda038238881c4bf).  Also a big improvement.

- Re-order and re-word some commands and their descriptions (8449ad05ebc2b08e7575d6090a3d80daa6c466bc).  Over time, it seems we've not taken enough care to keep like with like, keep the important stuff near the top, and use consistent wording.  This also moves a couple of commands to "Advanced commands".

Re. your point, @philc... Specifically, that we're abusing the visual metaphor of a "keyboard key" for something which is actually a key sequence.

Fair point, but I'm not sure it's likely to confuse users at all.  And we have to trade that issue off against screen real estate, which is quite tight in our case.  Using `,` as a separator is helpful in that it's small and almost tucks underneath the preceding key. `/` would have to be centred, and would require more padding than between individual keys.

Example:

- `J` &#8202;&#8202;&#8202;&#8202;/&#8202;&#8202;&#8202; `g` `T`
- `J`, `gT`

The difference is striking and - to my eye - the bottom version is clearer.

Screenshot of the current version...

snapshot

philc:
This is looking great. Ship it!

Issue #1562, opened 2015-04-13
Turn home row (asdfghjkl) into number row (1234567890) when link hinting


leoj3n:
I currently use [Karabiner.app](https://pqrs.org/osx/karabiner) to turn my home row into a number row by holding SPACE:

![homerow](https://cloud.githubusercontent.com/assets/990216/7126288/afa918c8-e1fd-11e4-8cfd-3f633427a2a3.png)

In Vimium I set link hints to numbers so I can hit `f` then hold SPACE and press f.ex `a`+`h` for `16`. This way my fingers never move off the home row.

This could be a builtin Vimium setting, then I wouldn't have to hold SPACE.

smblott-github:
Closing.
Lack of interest.

PR #2109, opened 2016-04-21
Add sendMessage command to message external components.

smblott-github:
This is a simpler alternative to #1980.

Example:

    map p sendMessage name=SomeName secret=SomeSecret

Now, `3p` invokes `window.PostMessage()` with the message:

    name: "SomeName"
    secret: "SomeSecret"
    count: 3

It is a simple matter to write a  TamperMonkey script to receive and handle the message.  And - of course - it would be relatively simple for users to share TamperMonkey commands.

The names `name` and `secret` have no special meaning to Vimium.

Another example:

    map p sendMessage name=SomeName secret=SomeSecret extension=abjmlgelklomcednhcplkfdfdbajhdbn

If the `extension` option is truthy, then we use `chrome.runtime.sendMessage()` instead, using the cross-extension messaging API.  In this case, users could share new commands by publishing chrome extensions.

A possible extension of this idea would be to have a new link-hints mode (not a command) which selects a link then posts it as above to the page.  We can identify (and communicate)  elements by their index within:

    document.documentElement.getElementsByTagName "*"

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -140,6 +140,7 @@ Commands =
       "enterVisualMode",
       "enterVisualLineMode",
       "focusInput",
+      "sendMessage",
       "LinkHints.activateMode",
       "LinkHints.activateModeToOpenInNewTab",
       "LinkHints.activateModeToOpenInNewForegroundTab",
@@ -207,7 +208,8 @@ Commands =
     "closeTabsOnLeft",
     "closeTabsOnRight",
     "closeOtherTabs",
-    "passNextKey"]
+    "passNextKey",
+    "sendMessage"]
 
 defaultKeyMappings =
   "?": "showHelp"
@@ -320,6 +322,7 @@ commandDescriptions =
 
   enterInsertMode: ["Enter insert mode", { noRepeat: true }]
   passNextKey: ["Pass the next key to Chrome"]
+  sendMessage: ["Post a message to the page or an extension"]
   enterVisualMode: ["Enter visual mode", { noRepeat: true }]
   enterVisualLineMode: ["Enter visual line mode", { noRepeat: true }]
 
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -364,10 +364,10 @@ HintCoordinator =
     for own frameId, port of @tabState[tabId].ports
       @postMessage tabId, parseInt(frameId), messageType, port, request
 
-  prepareToActivateMode: (tabId, originatingFrameId, {modeIndex}) ->
+  prepareToActivateMode: (tabId, originatingFrameId, {modeIndex, options}) ->
     @tabState[tabId] = {frameIds: frameIdsForTab[tabId][..], hintDescriptors: {}, originatingFrameId, modeIndex}
     @tabState[tabId].ports = extend {}, portsForTab[tabId]
-    @sendMessage "getHintDescriptors", tabId, {modeIndex}
+    @sendMessage "getHintDescriptors", tabId, {modeIndex, options}
 
   # Receive hint descriptors from all frames and activate link-hints mode when we have them all.
   postHintDescriptors: (tabId, frameId, {hintDescriptors}) ->
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -45,9 +45,19 @@ DOWNLOAD_LINK_URL =
   name: "download"
   indicator: "Download link URL"
   clickModifiers: altKey: true, ctrlKey: false, metaKey: false
+SEND_MESSAGE =
+  name: "send-message"
+  indicator: "Select link"
+  linkActivator: (link) ->
+    index = do ->
+      elements = document.documentElement.getElementsByTagName "*"
+      for element, i in elements
+        return i if element == link
+    message = extend @options, {index}
+    window.postMessage message, "*"
 
 availableModes = [OPEN_IN_CURRENT_TAB, OPEN_IN_NEW_BG_TAB, OPEN_IN_NEW_FG_TAB, OPEN_WITH_QUEUE, COPY_LINK_URL,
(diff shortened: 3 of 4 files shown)


smblott-github:
I created a demo repo for this, [here](https://github.com/smblott-github/vimium-external-commands-demo).  It's extraordinarily simple to add commands: [background](https://github.com/smblott-github/vimium-external-commands-demo/blob/master/background.coffee) commands, [foreground](https://github.com/smblott-github/vimium-external-commands-demo/blob/master/foreground.coffee) commands.

Here's the full text...

```CoffeeScript
# Background commands.
Commands =
  # Mute all tabs.
  muteAllAudio: ->
    chrome.tabs.query {audible: true}, (tabs) ->
      for tab in tabs
        chrome.tabs.update tab.id, muted: true

  # Taggle the mute status in the current tab.
  toggleMute: ->
    chrome.tabs.query {currentWindow: true, active: true}, (tabs) ->
      for tab in tabs
        chrome.tabs.update tab.id, muted: not tab.mutedInfo.muted

  # Focus the first audible tab.
  focusAudio: ->
    chrome.tabs.query {audible: true}, (tabs) ->
      if 0 < tabs.length
        tab = tabs[0]
        chrome.tabs.update tab.id, active: true, ->
          chrome.windows.update tab.windowId, focused: true

chrome.runtime.onMessageExternal.addListener (request) ->
  Commands[request.name]? request
```

```CoffeeScript
# Foreground commands.
Commands =
  sayHello: ->
    alert "Hello!"

window.addEventListener "message", (request) ->
  Commands[request.data?.name]? request
```

Bindings:

    map em sendMessage name=muteAllAudio extension=ambcbpdpgjihleiflgbmelmbikncjmcf
    map et sendMessage name=toggleMute extension=ambcbpdpgjihleiflgbmelmbikncjmcf
    map ef sendMessage name=focusAudio extension=ambcbpdpgjihleiflgbmelmbikncjmcf
    map eh sendMessage name=sayHello

smblott-github:
f58de384c1d4c89236ef4b8d7a99620129a70fe3 adds the ability to select links with link hints and pass them to an external command.

[This](https://github.com/smblott-github/vimium-external-commands-demo/blob/master/foreground.coffee) now contains an example:

```CoffeeScript
# map el sendMessage hints name=showLinkText
Commands =
  # Use link hints to select a link then do something with it (here, just show its link text).
  showLinkText: ({index}) ->
    element = document.documentElement.getElementsByTagName("*")[index]
    alert "The link text is #{element.textContent}."
```

And there's a similar example using TamperMonkey [here](https://github.com/smblott-github/vimium-external-commands-demo/blob/master/tamper-monkey/showLinkUrl.js):

```Javascript
// ==UserScript==
// @name         Vimium external command demo
// @namespace    http://tampermonkey.net/
// @version      0.1
// @description  Demonstrate Vimium's external commands
// @author       Stephen Blott
// @match        http://*/*
// @grant        none
// ==/UserScript==

// Example Vimium binding:
//   map eu sendMessage hints name=showLinkUrl

(function() {
    'use strict';

    window.addEventListener("message", function(event) {
        if (event.data && event.data.name == "showLinkUrl") {
           var element = document.documentElement.getElementsByTagName("*")[event.data.index];
           alert(element.href);
        }
    });
})();
```

gdh1995:
I wonder if Vimium may accept messages from some special extensions, such as TamperMonkey, execute some commands in messages and then post back responses - even a frontend messaging mechanism should also be very useful.

If there're only messages from Vimium to others, I'll have to add too many commands to Vimium keyMappings, which is not graceful at all, and it's most likely that many of those external commands can only work on pages with special urls.

smblott-github:
@gdh1995.  This feature, if we go with it, will never be core Vimium functionality.  Take a look at the design goals.  This feature doesn't fit there at all.  Therefore, nothing we do should involve a complicated user interface, or probably any user interface at all.

So why consider this at all?  Folks can always use TamperMonkey (or something similar).

One reason is that Vimium maps a lot of keys and does key handling well (multi-key mappings, counts, etc.).  It could be difficulty to integrate users' own custom JS with Vimium's key handling (we map most of the likely keys already).  And we get lots of feature requests which we just aren't going to implement, but which could easily be implemented by users (and possibly shared) with a framework like this.

The attractiveness of the approach here is that it's very light weight and catches most of the use cases (which I've seen).  Not all, but most.  And if we can get 90% of the way there with very little cost, then that seems like a reasonable thing to do.

I'd not be too keen on going further unless it can be done with almost no additional complexity.

gdh1995:
What I means is not to import external key bindings but a simple interface for other extensions to call Vimium's (frontend) commands directly. It's easy because Vimium just needs a `runtime.onMessageExternal` listener.

smblott-github:
@gdh1995... Yes, but Vimium needs to know whether it can trust messages from the outside, for example via a shared secret.  And that requires some extra user-interface machinery.

leoj3n:
I imagine it would require too much new machinery to make the Vomnibar be extensible to allow for extensions like #2191 and #2196?

I would personally try to implement #1562 and #1965 if this were made available.

The ability to trigger Vimium frontend commands as @gdh1995 mentioned would help with #2197.

leoj3n:
Additionally, the ability to manipulate Vimium from the outside (not just the Vomnibar) would allow for interesting things like you mentioned over here https://github.com/philc/vimium/pull/1980#issuecomment-183223423

leoj3n:
As @smblott-github mentioned at the end of https://github.com/philc/vimium/pull/1980#issuecomment-183643476:

> d. Pick an element with link hints, then tell me which one. E.g, to implement link hovering, or something. 

This would also probably make possible implementing #2174 and #2188.

smblott-github:
@philc... I think we should consider fleshing the idea in this PR out a bit.  @leoj3n... Input would be welcome (e.g., on the Vomnibar).

If we do this, it would be a "very-advanced feature", and not involve any new options or the like.  It should be invisible to new users.

Issue #2081, opened 2016-04-02
Global link-hints issues


smblott-github:
With #2048, we merged global link hints.  I'm continuing to work on tweaks and fixes, and others may uncover further issues.  We need somewhere to document this without having a new issue for every detail.  That's the purpose of this issue; unwatch if you're not interested in updates.

Here are the issues identified (and fixed) to date:
1. 39adee9090fc5aadfd5dd681f91b80025084858a fixes a race condition when link-hints mode launches.  In particular, because launching link-hints mode requires messaging all frames, it takes some time.  39adee9090fc5aadfd5dd681f91b80025084858a prevents other Vimium commands from being launched during that interval.
2. 8574dec533e5c2b3d4623de359c30a16846cd0e6 also fixes a race condition, this time on page load.  It is possible that the settings have loaded in one frame, but not in another.  Since we fall back to the default value if the settings have not yet loaded (and we collect extra hint information for filtered hints), we could end up assuming different hint modes in different frames.  8574dec533e5c2b3d4623de359c30a16846cd0e6 ensures that we don't collect hint information until the settings have loaded.
3. [This branch](https://github.com/smblott-github/vimium/commits/use-ports-for-global-link-hints) is merged; see the last five commits.  This uses ports for all messaging related to link hints (as opposed to `sendMessage()`).  This is considerably faster in practice.  We may still need to look at other ways of speeding up the link-hints initialisation sequence.

smblott-github:
See the sequence of three commits ending with 175f4b275c49ba332000ab773061777af820a87f.

This refactors the frame registration sequence such that we do not register tiny frames.  Some sites (e.g. Google Inbox) use many tiny iframes (on the order of 12 or 15).  Here, we don't register those tiny frames at all.  Because they're not registered, we don't request hint descriptors from them, which speeds up the link-hints activation sequence.

As a side effect, it also speeds up `gf` on frames with many tiny iframes.

smblott-github:
In the last 24 hours, a combination of the commits above left `master` unstable.  Sorry about that.

9d3e6cc399e69aec49ff59e78e58d403fe0bdf4b and 96c74e4aa7e39a99bf5511440ba7a4155f1e2db8 should - hopefully - bring `master` back to a stable state.

Future potentially unstable changes will be in [this branch](https://github.com/smblott-github/vimium/tree/global-link-hints-continued).

smblott-github:
Status:

[...]

PR #2116, opened 2016-04-25
Make global link hints optional (Vimium Labs)

smblott-github:
Add a Vimium-labs section to the options page where users can opt in to use global link hints.  Defaults to "no".

See this post:

- https://github.com/philc/vimium/issues/2081#issuecomment-210980903

On the options page, it looks like this:

snapshot

Some of the issues mentioned [here](https://github.com/philc/vimium/issues/2081#issuecomment-210980903) are quite concerning, and could break key Vimium features and  be disruptive to users.  On the other hand, just sitting on this feature won't help shake these issues out.  If we can get the code out there, then we might get feedback which provides a  better handle on how to fix them.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -55,7 +55,23 @@ HintCoordinator =
   suppressKeyboardEvents: null
 
   sendMessage: (messageType, request = {}) ->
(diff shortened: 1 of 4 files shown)


smblott-github:
Closing in light of #2125.

Issue #1935, opened 2016-01-19
The action "Follow the link labeled next or >" does not follow the <input type="button">


kalokng:
Hi,
Currently I am using 1.53, in the following page the popup did not show when I press ]]
```html
<html>
<body>
<input type="button" value="next" onclick="prompt('test')" />
</body>
</html>
```
Is it intended or a bug?
Thanks in advance.

smblott-github:
It looks like we don't "click" buttons currently.

A PR would be welcome.  It should just be a matter of fixing up the XPath in `findAndFollowLink`.

smblott-github:
Fixed in #2117.

sco-tt:
Sorry I wasn't clear in my PR -- I haven't actually fixed this particular issue yet, that has to do with looking for matches in the element value attribute, not innerText.

I just wanted to make a separate commit for similar issues I encountered while debugging this. I plan on getting to this particular issue soon.

PR #2117, opened 2016-04-26
Improves next/previous links by targetting button elements and adding…

sco-tt:
I was attempting to address issue #1935 but I encountered two other issues with the previous/next shortcut.

1. Changes to vimium_frontend.coffee: previous/next don't work when the element is a `<button>`. See https://codepen.io/khilnani/pen/qEWojX as an example

2. Changes to settings.coffee: The previous/next patterns don't include single angle quotation marks, so on an example like this http://codepen.io/rasx/full/xGadEB the default behavior was going to the double angle quotation marks, usually first page/last page.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -45,5 +45,6 @@ Contributors:
   Utkarsh Upadhyay <musically.ut@gmail.com) (github: musically-ut)
   Michael Salihi <admin@prestance-informatique.fr> (github: PrestanceDesign)
   Dahan Gong <gdh1995@qq.com> (github: gdh1995)
+  Scott Pinkelman <scott@scottpinkelman.com> (github: sco-tt)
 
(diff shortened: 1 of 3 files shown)


smblott-github:
Nice.  Thanks, @sco-tt.

philc:
👍

smblott-github:
@sco-tt.

I reverted this.  See 22f0ed9c3c63ecbc4fc2413dc5235a9489d9cf25.

Reddit is important.  If we want to do this, then we have to think again.

sco-tt:
Thanks, I'm going to take a look at this and try to fix it. For the time being I've split out the changes to the next/prev string pattersn from the original PR. See the [new PR](https://github.com/philc/vimium/pull/2132)

PR #2118, opened 2016-04-28
Fix UI-component initialization issues (maybe).

smblott-github:
This fixes some UI component initialization issues.  It's a long story...

The problem.

- Go to this page: http://www.thejournal.ie/seanad-election-results-2016-2737768-Apr2016/
- Click one of the links from the "Most Popular" box on the right.
- Navigate back (`H`) and, as the original page is loading, activate the Vomnibar.

In 1.54 this renders Vimium unusable.  In `master` this renders the Vomnibar unsable, but the rest of Vimium usable.  Depending on how you set things up, you can generate these errors...

snapshot

(Another error is that the `load` event never fires on the `@iframeElement`, but I cannot screenshot that.)

It seems the source of the issue is that we're initializing UI components too soon.  We need to wait until the `readyState` is "complete".  It's not clear *why* this is the case.  However, the example page above does use rather a lot of iframes, and I'm guessing their loading somehow interferes with the loading of the Vomnibar UI component.

With this PR:

- The Vomnibar is initialised when the `readyState` is "complete" (in the top frame only, and only if Vimium is enabled).  Requests arriving prior to then are silently discarded.
- The HUD is also initialized only when the `readyState` is "complete"; however, requests arriving before then are queued.  So, if the user immediately enters insert mode, then the "Insert mode" indicator will eventually be displayed.
- The help dialog silently discards requests until the `readyState` is "complete".

I'm posting this as a PR because:

1. It needs some visibility.
2. With this, if the `readyState` *never* reaches "complete", then the Vomnibar would be unusable.  And that's pretty serious.

I'll run this live for a while and see how it gets on, then report back.

This would be a release blocker, except that the same (and worse) issues exist in 1.54.

### DIFF
--- a/content_scripts/hud.coffee
+++ b/content_scripts/hud.coffee
@@ -22,15 +22,17 @@ HUD =
     @_showForDurationTimerId = setTimeout((=> @hide()), duration)
 
   show: (text) ->
-    @init()
-    clearTimeout(@_showForDurationTimerId)
-    @hudUI.activate {name: "show", text}
-    @tween.fade 1.0, 150
+    DomUtils.documentComplete =>
+      @init()
+      clearTimeout(@_showForDurationTimerId)
+      @hudUI.activate {name: "show", text}
+      @tween.fade 1.0, 150
 
   showFindMode: (@findMode = null) ->
-    @init()
-    @hudUI.activate name: "showFindMode"
-    @tween.fade 1.0, 150
+    DomUtils.documentComplete =>
+      @init()
+      @hudUI.activate name: "showFindMode"
+      @tween.fade 1.0, 150
 
   search: (data) ->
     @findMode.findInPlace data.query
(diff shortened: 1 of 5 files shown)


smblott-github:
Update...  With this PR, I can't crash the Vomnibar.  That's good.

However, all is not particularly well:

- It's not clear *why* the existing logic is wrong; that's not encouraging.

- With `master` and the test described above, it's possible to generate the situation illustrated below.  It's not at all clear *how* we can end up with our Vomnibar injected into one of the page's iframes:

snapshot

- Also with `master`, we can generate this error:

snapshot

- That error is difficult to understand.  Somehow we cannot post a message to our own iframe, one we just created.  It's as if the page has somehow taken over that frame.

- These errors only occur if we go *back* (and if we go *forward* again).  We can't trigger them if we only navigate forwards.  So, what is that's different about going back and forward?

All of this makes it unclear what we should actually do, and suggests that there's something else at play.

smblott-github:
Merged.  I've been using this for a week and haven't seen any issues.

PR #2122, opened 2016-05-01
Make findAndFollowLink look at element values so <input> can be used …

### LINKED ISSUE #1935: The action "Follow the link labeled next or >" does not follow the <input type="button">

sco-tt:
Fixes #1935 

In order to use <input> elements as next/prev links, you need to look at the <input> value because there is no innerText. This change change `findAndFollowLink` to add <input> values to `canidateLinks` if they exist. <a> elements that match a `linkString` are still preferred if both appear on the page.

Demos:

http://codepen.io/scott0/pen/MyzexY
<input> only, doesn't work on current master, works with these changes.

http://codepen.io/scott0/pen/grQMEj
<a> only, 'standard' situation, continues to work after changes

http://codepen.io/scott0/pen/reQLRx
<input> and <a>, current master ignores <input>; after changes <a> preferred in the event that there are both on a single page.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -563,8 +563,8 @@ findAndFollowLink = (linkStrings) ->
 
     linkMatches = false
     for linkString in linkStrings
-      if (link.innerText.toLowerCase().indexOf(linkString) != -1 || link.value &&
-          link.value.indexOf(linkString) != -1)
+      if link.innerText.toLowerCase().indexOf(linkString) != -1 ||
+          link.value?.indexOf(linkString) != -1
         linkMatches = true
         break
     continue unless linkMatches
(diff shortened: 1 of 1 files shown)


content_scripts/vimium_frontend.coffee
```diff
@@ -563,7 +563,8 @@ findAndFollowLink = (linkStrings) ->
 
     linkMatches = false
     for linkString in linkStrings
-      if (link.innerText.toLowerCase().indexOf(linkString) != -1)
+      if (link.innerText.toLowerCase().indexOf(linkString) != -1 || link.value &&
+          link.value.indexOf(linkString) != -1)
         linkMatches = true
```
smblott-github: Thanks, @sco-tt.

Could we get rid of the unnecessary parentheses here?  (I know they were there before, but we can tidy the code up as we touch it.)

Also, here (and below) the dependency on the precedence of `||` and `&&` should be made explicit.  Here, it can just be `link.value?.indexOf...`, below it will need parentheses.

sco-tt:
Would something like this work? My apologies, I'm a newbie to Coffeescript.

smblott-github:
Thanks, @sco-tt.

PR #2124, opened 2016-05-05
install the real selectionchange listener after window.find

gdh1995:
The old code would get `window` as `@`, whose `restoreDefaultSelectionHighlight` is `undefined`, so this commit fixes it.

### DIFF
--- a/content_scripts/mode_find.coffee
+++ b/content_scripts/mode_find.coffee
@@ -179,7 +179,7 @@ class FindMode extends Mode
 
     if options.colorSelection
       setTimeout(
-        -> document.addEventListener("selectionchange", @restoreDefaultSelectionHighlight, true)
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @gdh1995.  Your observation looks correct.  However, I can't seem to trigger the error you'd expect.  Can you produce the error?  And, if so, how?

gdh1995:
Here is how to reproduce the error:
* open a page, press '/', and type a word which matches.
* now the page should have a highlighted word, whose color is #ff9632
* press down your mouse and move to select another word
* then the new selected one is also in #ff9632.
* use the mouse to select the third word, and now the selected is in the default color.

`FindMode` adds a `click` listener to the main frame, so this error only occurs when the first click event means to select any chars, but not to activate a blank area.

Issue #2128, opened 2016-05-09
KeyboardEvent.keyIdentifier is deprecated


lucianoar:
'KeyboardEvent.keyIdentifier' is deprecated and will be removed in M53, around September 2016.
See [here](https://www.chromestatus.com/features/5316065118650368) for more details.

It must be changed to this:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

https://github.com/philc/vimium/blob/0ddc7a28d17e94296f209d3d8d06357de281e1d1/lib/keyboard_utils.coffee#L35

Chrome version: Chromium - 52.0.2726.0 (Developer Build) (64-bit)
Vimium version: 1.54

smblott-github:
@lucianoar...  I took a look at this: #2133.

lucianoar:
Fixed by #2133

PR #2129, opened 2016-05-09
Change KeyboardEvent.keyIdentifier to KeyboardEvent.key

lucianoar:
Fix issue #2128

### DIFF
--- a/lib/keyboard_utils.coffee
+++ b/lib/keyboard_utils.coffee
@@ -32,14 +32,14 @@ KeyboardUtils =
 
   getKeyChar: (event) ->
     # Not a letter
-    if (event.keyIdentifier.slice(0, 2) != "U+")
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @lucianoar.

There are a couple of problems with this.  First, the [key API](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) doesn't land until Chrome version 51, apparently.  I think 50 is the current stable version.  So we have to wait a bit yet.  In fact, it's not obvious how we're supposed to time this.  We'll probably have to support both the old API and the new API for a while.

(And second, the tests are failing.)

lucianoar:
What a rookie mistake! I rushed to make the PR, sorry about that.
I'll dive deeply into the codebase and I'll try to fix it.

lucianoar:
Fixed by #2133

PR #2133, opened 2016-05-14
Use event.key (not event.keyIdentifier).

### LINKED ISSUE #2128: KeyboardEvent.keyIdentifier is deprecated

smblott-github:
`event.keyIdentifier` is depricated and will be removed soon.  It is being replaced by `event.key`.  Unfortunatelty, `event.key` is not yet available in the stable Chrome version.

Here, we use whichever API is available.  In due course, we can remove the `event.keyIdentifier` implementation (and a considerable amount of machinery surrounding it).

~~For the time being, if both APIs are available, then we verify one against the other and show a warning message if they do not match.  This should help us track down any issues which arise.~~ Edit... Don't do that.  It's too noisy (e.g. every time `Alt` is pressed).

Using `event.key` has the additional benefit of correctly detecting shifted characters on the numbers row on keydown, which has been a problem for some users.

Fixes #2128.

Note:  We have a problem with the tests.  phantomjs does not currently support event.key, and possibly never will.


smblott-github:
Merging.  This seems sound, and I can't find a case which breaks the existing key-handling behaviour.

PR #2152, opened 2016-06-04
Add chrome.commands support for simple actions

mrkishi:
Hello,

As awesome as Vimium is, it's very annoying to get stuck in certain tabs. *The Great Suspender*, for example, produces a lot of extension tabs, constantly breaking my flow to remember to use ctrl/cmd-tab.

While this patch requires user configuration, I think it far benefits those that go to that length. The chrome.commands API lets us register a set of commands for user-configurable keyboard shortcuts. Once they choose a key combination, it works across the browser (including `chrome-extension://` and `chrome://` pages). At +11 lines, it's also a very simple modification so I don't think it'll cause too much headache.

I'm unfamiliar with Vimium's codebase, so I've only added the obvious ones: `nextTab`, `previousTab`, `visitPreviousTab`, `firstTab` and `lastTab`. I'll gladly improve this, should you have any objections or additions.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -74,6 +74,10 @@ chrome.runtime.onMessage.addListener((request, sender, sendResponse) ->
   # Ensure the sendResponse callback is freed.
   return false)
 
+# Allow simple commands to be run from restricted tabs
+chrome.commands.onCommand.addListener (command) ->
+  chrome.tabs.getSelected null, (tab) -> BackgroundCommands[command] { count: 1, tab: tab }
+
 onURLChange = (details) ->
   chrome.tabs.sendMessage details.tabId, name: "checkEnabledAfterURLChange"
 
--- a/manifest.json
+++ b/manifest.json
@@ -71,6 +71,13 @@
     "default_icon": "icons/browser_action_disabled.png",
     "default_popup": "pages/popup.html"
   },
+  "commands": {
+    "nextTab": { "description": "Go one tab right" },
+    "previousTab": { "description": "Go one tab left" },
+    "visitPreviousTab": { "description": "Go to previously-visited tab" },
+    "firstTab": { "description": "Go to the first tab" },
+    "lastTab": { "description": "Go to the last tab" }
+  },
   "web_accessible_resources": [
     "pages/vomnibar.html",
     "content_scripts/vimium.css",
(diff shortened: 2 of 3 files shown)


gdh1995:
+1.

But I think `visitPreviousTab`, `firstTab` and `lastTab` might be replaced with `createTab` and `removeTab`, which means we can always switch to another Vimium-enabled tab.

mrkishi:
I see no reason to replace them instead of simply adding `createTab` and `removeTab`. Care to elaborate?

gdh1995:
Oops, `removeTab` is not what I want. If a tab is failed to load, then I want `reloadTab` to retry at once.

The reason I want other keys to reload / create tabs is that the default shortcuts (`Ctrl+T`, `Ctrl+R` and `F5`) are a little hard to press by only one hand - on Windows, it's easier to press `Alt+T` or `Alt+R`.

mrkishi:
I certainly see why you'd want those *added*. What I don't understand is why you'd want to *remove* `visitPreviousTab`, `firstTab` and `lastTab`.

I'll add the create/remove/reload actions once we get confirmation there's interest to merge this.

smblott-github:
I'm still thinking about what to do with this, @mrkishi.  On the one hand, it's just a few lines of code.  On the other hand, it's hard to discover, and outside of the regular Vimium workflow.  It's also outside of the Vimium keyboard model (e.g., no count prefixes).  It might be something that would be better implemented as a separate extension.

mrkishi:
I see. To be honest, if I were to use an external extension for this, I'd rather replace Vimium as a whole.

While this pull-request is completely outside Vimium's keyboard model, it's still *possible* to upgrade `chrome.commands` in Vimium-enabled pages -- which is not the case for third-party extensions. I'm not intimate enough with Vimium to comment on whether it'd be *feasible* to implement, though. 

I'm guessing it shouldn't be too hard, though: instead of calling `BackgroundCommands`, `chrome.commands` need to be translated to the regular key-chords in the active tab's current input sequence. For non-Vimium-enabled tabs, they'd drop full-feature-compatibility and execute the raw background command.

I completely agree the discoverability of this feature would be an issue, but another way to look at it is this: users who don't need it won't ever see it; users who do would consider its abcence a blocking issue.

In other words, if you use Vimium's shortcuts for moving between tabs and come across `chrome://` tabs, you'll either look for a solution or call it buggy and uninstall the extension. Currently, that solution is nonexistent.

leoj3n:
This is impossible because <kbd>Shift+K</kbd> is not a valid/allowed combo:

image

You can only really something like <kbd>Command+Shift+K</kbd> or <kbd>Alt+K</kbd>.

So, as far as I'm concerned, I'll be sticking with good 'ol <kbd>Command+Shift+]</kbd>.

Issue #2163, opened 2016-06-14
Hint Marker font is serif on facebook.com


sco-tt:
Facebook added the following to their stylesheet:

`#facebook span {
    font-family: inherit;
}`

This overrides the font-family declaration for the rule `div.internalVimiumHintMarker span` in vimium.css, causing the hint marker letters to appear as a serf font, not Helvetica / sans-serif.

philc:
Confirmed. It looks pretty bad.

PR #2164, opened 2016-06-14
Adds important to hint marker font-family declaration

sco-tt:
Addresses #2163

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -88,7 +88,7 @@ div.internalVimiumHintMarker {
 
 div.internalVimiumHintMarker span {
   color: #302505;
-  font-family: Helvetica, Arial, sans-serif;
+  font-family: Helvetica, Arial, sans-serif !important;
   font-weight: bold;
   font-size: 11px;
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @sco-tt.

I can't reproduce this myself (possibly it's geo-dependent), however I'll take your and @philc's word for it.

smblott-github:
@sco-tt.  I'm no expert on CSS.  However, I've been wondering whether this is the correct fix.  Somehow, Facebook's style is leaking into Vimium's style.  The fix here looks like patch.  Is there a better way to prevent the styles leaking in, e.g. using `vimiumReset`?

sco-tt:
It might not be necessary to merge these changes; it looks like FB has updated the rule to 
`#facebook ._-kb span {
    font-family: inherit;
}`

and the rule without !important is sufficiently specific; I'm not longer seeing the issue on 1.56. I've checked on Chromium Canary Channel on OSX - and Chrome stable on Win 10. @philc are you seeing this? Perhaps it had already been pushed to @smblott-github 's part of the world when I made this PR.

smblott-github:
Thanks for the update, @sco-tt.  I'll take the commit out again.

PR #2166, opened 2016-06-15
Better detection of elements that aren't visible

poacher2k:
Ref: [discussion in PR 1745](https://github.com/philc/vimium/pull/1745#issuecomment-150222798)

In my testing, the changes I made dramatically improves the detection of inactive/non-visible items.

In the case of partly covered elements, the next step would possibly be to place the hint labels at the visible area of the element.

Without vimium:
no-vimium

Before changes:
before

After changes:
after

Test coverage remains the same.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -228,51 +228,38 @@ class LinkHintsMode
       visibleElement = @getVisibleClickable element
       visibleElements.push visibleElement...
 
-    # TODO(mrmr1993): Consider z-index. z-index affects behviour as follows:
-    #  * The document has a local stacking context.
-    #  * An element with z-index specified
-    #    - sets its z-order position in the containing stacking context, and
-    #    - creates a local stacking context containing its children.
-    #  * An element (1) is shown above another element (2) if either
-    #    - in the last stacking context which contains both an ancestor of (1) and an ancestor of (2), the
-    #      ancestor of (1) has a higher z-index than the ancestor of (2); or
-    #    - in the last stacking context which contains both an ancestor of (1) and an ancestor of (2),
-    #        + the ancestors of (1) and (2) have equal z-index, and
-    #        + the ancestor of (1) appears later in the DOM than the ancestor of (2).
-    #
-    # Remove rects from elements where another clickable element lies above it.
     # This loop will check if any corner or center of element is clickable
     # document.elementFromPoint will find an element at a x,y location.
-    # Node.contain checks to see if an element contains another.
+    # Node.contain checks to see if an element contains another. note: someNode.contains(someNode) === true
     # If we do not find our element as a descendant of any element we find, assume it's completely covered.
     nonOverlappingElements = []
     while visibleElement = visibleElements.pop()
-        rect = visibleElement.rect
-        elem = visibleElement.element
-        leftTop = document.elementFromPoint(rect.left, rect.top)
-        leftTopContains = leftTop && leftTop.contains(elem)
-        if leftTopContains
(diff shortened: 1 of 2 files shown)


poacher2k:
@philc Had a chance to look at this? I'd be happy to provide any explanation or further fixes that is necessary to get this merged.

IMO, this would be useful for a lot of Vimium users.

mrmr1993:
The `contains` statements look inverted to me: if the element is occupying the rect, then its children -- not its parents -- will be the only elements can overlap it, if it's not obscured by another element.

I also have slight pause over elements with non-zero `border-radius`, as an element overlapping only their centre renders them invisible under these changes. However, as I had already relegated these to be `secondClassCitizen`s (as `tabIndex` is a poor approximator of clickable elements), any victims of this should already be rare enough that we shouldn't have to worry. However, it's probably worth a comment to this extent in case future issues arise around it (and indeed, in the rare general case of corner and centre overlaps).

poacher2k:
Note: `someDiv.contains(someDiv) // true`

> The contains statements look inverted to me: if the element is occupying the rect, then its children -- not its parents -- will be the only elements can overlap it, if it's not obscured by another element.

This is correct, but `elementFromPoint` will sometimes return a parent element instead of our target. As long as our target is a descendant of the returned element, we can safely assume that a click at that point (screen coordinates) will click our target. If our target is not contained in the returned element, the target is most likely overlapped by another element, not visible at the given point, and we can assume that a click at that point will click another element instead of our target.

> I also have slight pause over elements with non-zero border-radius, as an element overlapping only their centre renders them invisible under these changes.

As we're using `getBoundingClientRect` to get our coordinates, and borders are included, won't we detect these targets as the changes attempts to find the target at the target center and all four corners?

content_scripts/link_hints.coffee
```diff
@@ -241,23 +241,38 @@ class LinkHintsMode
...
+        rightTop = document.elementFromPoint(rect.right, rect.top)
+        rightTopContains = rightTop && rightTop.contains(elem)
+        if rightTopContains
+            nonOverlappingElements.push visibleElement unless visibleElement.secondClassCitizen
+            continue
+        rightBottom = document.elementFromPoint(rect.right, rect.bottom)
+        rightBottomContains = rightBottom && rightBottom.contains(elem)
+        if rightBottomContains
+            nonOverlappingElements.push visibleElement unless visibleElement.secondClassCitizen
+            continue
```
smblott-github: Could we not roll the four (almost identical) cases above into a loop?

smblott-github:
@poacher2k... Thanks for this.  It looks pretty promising and we should try it out.

I made one small line comment.  Also, could you take a look at the two larger blocks of comments above your changes.  Some of them look like they may no longer be accurate.

poacher2k:
Hi @smblott-github ! Thank you for looking into this PR :)

I made changes according to your comment, hope it looks good now.

smblott-github:
@poacher2k... It looks like this branch is based off a very old copy of `master` (before a large number of changes in the last release).  That makes the rebase/merge difficult and error prone.  Since your changes affect really just a single block of code, it would easier just to copy them into a fresh copy of `master`.

I can do that, but then the change will appear to have been made by me.

Would you prefer to do that (and then point me towards the new branch, and I'll merge it)?

poacher2k:
I understand! I'd be happy to do this some time after work today if that's
okay :)

Den ons. 14. sep. 2016, 07:56 skrev Stephen Blott <notifications@github.com

poacher2k:
New PR @ #2251

Issue #1138, opened 2014-08-21
navigation keys don't work on most pages


timfeirg:
I installed vimium 1.45 but find out only very few of the keys work as excepted, like (x, f, r, /), navigation keys don't work at all on whatever pages I opened.
**The same under incognito mode with only vimium loaded.**
### update

For now, the only pages that navigation keys work is the vimium options tab: 
`chrome-extension://dbepggeogbaibhgnhhndojpepiihcmeb/pages/options.html`
### update 2

upgraded my chrome yesterday
### chrome version

Google Inc.
Copyright 2014 Google Inc. All rights reserved.
Google Chrome   37.0.2062.94 (Official Build 290621) 
OS  Mac OS X 
Blink   537.36 (@180557)
JavaScript  V8 3.27.34.14
Flash   14.0.0.177
User Agent  Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/37.0.2062.94 Safari/537.36
Command Line    /Applications/Google Chrome.app/Contents/MacOS/Google Chrome --new-profile-management --flag-switches-begin --enable-automatic-password-saving --enable-cloud-devices --enable-download-resumption --enable-experimental-web-platform-features --enable-fast-unload --fast-user-switching --google-profile-info --js-flags=--harmony --new-avatar-menu --new-profile-management --enable-scroll-prediction --enable-simplified-fullscreen --enable-search-button-in-omnibox-for-str-or-iip --flag-switches-end
Executable Path /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
Profile Path    /Users/timfeirg/Library/Application Support/Google/Chrome/Default
Variations  e950616e-37fb3cc2
8afebf76-164ff01c
c70841c8-4866ef6e
195ce1b5-901eb5f9
9e5c75f1-e1c7cd46
24dca50e-837c4893
ca65a9fe-91ac3782
8d790604-9cb2a91c
4ea303a6-3d47f4f4
9736de91-ca7d8d80
b2612322-f8cf70e2
5a3c10b5-e1cc0f14
244ca1ac-4ad60575
f47ae82a-746c2ad4
5e29d81-cf4f6ead
3ac60855-486e2a9c
246fb659-6754d7b7
f296190c-b7d97d3d
4442aae2-6bdfffe7
ed1d377-e1cc0f14
75f0f0a0-d7f6b13c
e2b18481-75cb33fc
[...]

Issue #962, opened 2013-11-23
Movement commands not working,


leoj3n:
```
33.0.1716.0 (Official Build 236646) canary
Mac OS X (10.8.5)
Vimium 1.44
```

I've tried using the latest Chrome Store package and cloning. I tried checking out `1.42` and disabling all other plugins. Most command like `f` and `gs` work. The ones that don't however are movement related like `j`, `k`, or `gg`, `G`... 

Is the plugin working fine for others with the same or similar setup? I'm at a loss.

PR #2168, opened 2016-06-16
Use document.scrollingElement for scrolling

dpogue:
This fixes Vimium with the 'Experimental Web Platform Features' flag enabled, and ensures compatibility across quirks mode scrolling (via `document.body`) and standards-compliant scrolling (via `document.documentElement`).

Includes a polyfill for `document.scrollingElement` that returns `document.body`.

### DIFF
--- a/content_scripts/scroller.coffee
+++ b/content_scripts/scroller.coffee
@@ -4,6 +4,9 @@
 #
 activatedElement = null
 
+if !('scrollingElement' of document)
+  Object.defineProperty(document, 'scrollingElement', { get: () -> document.body })
+
 # Return 0, -1 or 1: the sign of the argument.
 # NOTE(smblott; 2014/12/17) We would like to use Math.sign().  However, according to this site
 # (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign) Math.sign() was
@@ -37,7 +40,7 @@ getDimension = (el, direction, amount) ->
     name = amount
     # the clientSizes of the body are the dimensions of the entire page, but the viewport should only be the
     # part visible through the window
-    if name is 'viewSize' and el is document.body
+    if name is 'viewSize' and el is document.scrollingElement
       # TODO(smblott) Should we not be returning the width/height of element, here?
       if direction is 'x' then window.innerWidth else window.innerHeight
     else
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @dpogue.  Could you flesh out your explanation a bit please; e.g. why does the status quo break? and why does this fix that?

(FYI: Vimium development is going into summer recess soon; I won't get time to think about this properly until August.)

leoj3n:
@smblott-github One more day! :D

I'm also curious about an explanation @dpogue; What exactly breaks or what exactly changes about scrolling with the experimental flag enabled?

dpogue:
Historically, in quirks mode, browsers applied scrolling the `body` element. In standards mode, scrolling is applied to the document element (`html` in most cases). However, Webkit (and thus Blink) applied scrolling to `body` regardless of the mode.

This led frameworks trying to measure scrolling to resort to trying to handle all cases:
```javascript
scrollTop = (document.documentElement.scrollTop + document.body.scrollTop);
```

`document.scrollingElement` was added to the CSSOM spec to provide a consistent way of accessing scroll position, regardless of mode.
  
  
The reason we need to use it here is because Chrome has started following the spec and applying scrolling to the document rather than to the body, when the "Experimental Web Platform Features" flag is enabled.
  

This PR provides a safe (for Chrome) polyfill, in that it defines `document.scrollingElement` to be `document.body` if it does not already exist. Thus existing behaviour is preserved for people running without the experimental flag.

leoj3n:
Ah, so this _is_ related to #962 and #1138.

smblott-github:
Thanks for the explanation, @dpogue.

Something doesn't seem to be right yet.  On [this simple page](https://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html), I seem to have to click the page to get scrolling to work.  Any ideas, @dpogue?

smblott-github:
Ping @dpogue... (see my comment above).

gdh1995:
I approve this idea, because Chrome has "invented" a flag `#enable-experimental-web-platform-features` to allow `document.scrollingElement` to be `<html>` on some pages.

In my recent tests with this flag enabled, some pages did scroll `<html>`, although the other pages scrolled `<body>` as before. I haven't found any rule to predict which one they choose, and Vimium's options page scrolled `<body>`.

BTW, on my Chrome 53 x64 stable with the flag disabled, Vimium `master` can scroll [The Xlib page above](https://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html) directly (no click needed). But after applying this PR, a click becomes needed.

gdh1995:
The bug in this PR is:
* if `document.scrollingElement` is not defined, it setups it using `document.body` on `scroller.coffee` is being executed, but at that time the page is beginning to load and `document.body` is always `null`.

A fix may be:
* define `.scrollingElement` using a `getter` function which always return `.body`.

dpogue:
I've fixed the issues with scrolling not working immediately on page load. It was due to CoffeeScript's `in` operator being different than JavaScript's `in` :(

dpogue:
I wasn't able to reproduce that case with any of Chrome stable, dev, or canary, with the experimental-web-platform-features flag both on and off; but I've updated the PR to use a getter just to be safe.

gdh1995:
(I don't know why, but) Not all pages use `document.documentElement` as their scrolling root if `#experimental-web-platform-features` is enabled, and in my tests, Vimium's options page is just such a page which still insists using `document.body`.

It seems that a valid `<!DOCTYPE >` is needed to make `<html>` the scrolling root. You may try https://jsfiddle.net/86bpy7pw/.

smblott-github:
Thanks, @dpogue.  I'm going to merge this, but like this: 07bc4fbe820ff17b5a6b559a2ebcceb7d52700a2.

PR #2179, opened 2016-06-27
Mention edit URL keybindings

gaving:
Tracked down some keybindings I was after under #1004

Added them to the main documentation.


philc:
Nice. Thank you! Also note, for future reference, that you can find these in Vimium's help dialog under advanced commands.

PR #2181, opened 2016-07-03
Alter search appearance

cdosborn:
This is my first contribution to vimium. 

I wanted to open some discussion about the `vomnibar` and its appearance. I was hoping someone could clarify the usage of ↪  in the search results. From the source it looks like its suggestions that can be completed. This was unintuitive to me, and at the beginning of each suggestion it throws off the alignment with the input bar. In the demo below, I pulled out the arrow to demonstrate. I reduced the shadows and placed them more intentionally, if you zoom in on google's chrome search bar and the `after` gif, you'll see similar highlights. 
# Before:

# After:

### DIFF
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -46,15 +46,13 @@ class Suggestion
   generateHtml: (request) ->
     return @html if @html
     relevancyHtml = if @showRelevancy then "<span class='relevancy'>#{@computeRelevancy()}</span>" else ""
-    insertTextClass = if @insertText then "vomnibarInsertText" else "vomnibarNoInsertText"
-    insertTextIndicator = "&#8618;" # A right hooked arrow.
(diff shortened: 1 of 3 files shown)


smblott-github:
Yes, the hook indicates that the suggestion comes from a previous search-engine query, and Vimium can fill in the query terms so that the user can re-run or tweak the query.

Would we not want a better reason than "alignment" to lose that?

Re. styling... Hmm.  There is some difference.  The main thing I see is the different colour of the background shade.  Neither seems obviously better, to my eye.  Does the color in "before" match the background of the HUD?

philc:
Closing this due to inactivity. Note that refreshing the alignment of the results has some recent discussion in #2656.

(and to validate @cdosborn 's comment, I didn't know what those return icons alluded to until I read this comment thread 😮)

PR #2210, opened 2016-08-06
Complex key mapping

### LINKED ISSUE #2206: Using multiple modifiers for a mapping: <C-A-j>
I just tried to add the following mappings in order to change tabs with `Ctrl+Alt+[j or k]`.

```
map <C-A-j> previousTab
map <C-A-k> nextTab
```

It doesn't seem to work. Anything I could do about it?

gdh1995:
Allow Vimium to handle key presses like `<Ctrl+Alt+F1>` (`<c-a-f1>`) / `<Ctrl+Alt+Shift+Left>` (`<c-a-LEFT>`).

This should fix #2206 completely.


smblott-github:
Thanks, @gdh1995. Some questions:

1. Why the change from using `f1` to using `F1`?  (That will break `getKeyCharUsingKey`, in keyboard utils.)

2. If we allow `<c-a-X>`, then the user might put the modifiers in either order (e.g., also `<a-c-X>`).  In `getKeyCharString`, the order we use is fixed, so one of those will work but one will not.

Separately:

- These big regular expressions are very difficult to work through.
- Is there really a demand for these types of complex mappings?

smblott-github:
@gdh1995.  Thanks for this, but the regexps here were hurting my head, so I did something like this but in a different way (#2289).

Issue #986, opened 2014-01-03
unable to unset ']' key


Akenaide:
Hello,
browser : Chrome

I can not unset ] key. I don't even know what is exactly do, I just see it unfocus when I typing.

I add "unset ]" in the config but I don't see any change.

Great plugin I starting using vim a lot and it feel great to use the same way to navigate in my browser.

sorry for my poor english

niwox:
any solution for this partical issue ?

mrmr1993:
Sorry this has been sat so long. I'd been hoping that a fix for #301 would resolve this along the way, but #1140 didn't make the grade.

I've pushed 956145d as my attempt at a (very low quality) fix. @niwox (or somebody else), could you test it? If it works, I'll put it in a PR.

Akenaide:
Sorry for the late, your change do the job. I can type "]" again with vimium activate o/

Thanks

Akenaide:
oops

niwox:
No fix on the official release yet for this issue ?

roessland:
I cannot type [ in Vimium on my Norwegian keyboard (I had to copy paste it in here). It is typed by pressing AltGr+8, or equivalently Ctrl+Alt+8. I lose focus from the textarea when I press it.

It's also impossible to unmap. I've tried

unmap [  
unmap <c-a-8>  
unmap <c-[>

and some other combinations.

zieren:
Same here (German layout).

smblott-github:
OK.  Added *next-version* label.  We should fix this.

(It happens because `Ctrl-[` is hardwired to be "Escape".)

niwox:
I believe `Alt Gr - [` is hardwired to be Escape too. Any fix for that ?

andersoyvind:
This will be a noticable issue, as users of online javascript editors (jsbin, jsfiddle etc) or the interactive training sites (codeschool etc) will face the <insert bracket here>-issue, eventually.
Glad that it is worked on.

smblott-github:
Closing. #2215 is merged (and released), which should fix most of the issues here.

PR #2215, opened 2016-08-10
Consider the alt-key status when trying to detect <c-[>

colin-kiegel:
see https://github.com/philc/vimium/issues/986#issuecomment-53955175

### DIFF
--- a/lib/keyboard_utils.coffee
+++ b/lib/keyboard_utils.coffee
@@ -75,7 +75,8 @@ KeyboardUtils =
 
   isEscape: (event) ->
     # c-[ is mapped to ESC in Vim by default.
-    (event.keyCode == @keyCodes.ESC) || (event.ctrlKey && @getKeyChar(event) == '[' and not event.metaKey)
+    (event.keyCode == @keyCodes.ESC) ||
+    (event.ctrlKey && @getKeyChar(event) == '[' and not event.metaKey and not event.altKey)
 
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @colin-kiegel.

colin-kiegel:
Thank you so much - I had to disable vimium on a lot of pages due to this! I am looking forward to re-enabling it. :-)

smblott-github:
As a temporary measure (and if you trust me), you could use [this](https://drive.google.com/file/d/0B-PWDBroAYxrNFJQaHFBbDVuMkU/view?usp=sharing) version of Vimium.  It already has this fix.  Drag and drop onto the extensions page.  You might have to enable developer mode.  And don't forget to disable the Chrome Store version.

(It won't pick up your existing Vimium configuration.)

krokofant:
@smblott-github Any idea when this will be pushed out as stable? I didn't see any consistency about release schedule so I thought I'd ask :)

colin-kiegel:
works now with vimium 1.57, finally I can type it again [[[[ :+1: 

Thank you! :-)

noelpuru:
Im using version 1.59 and cannot map <c-[>. Is this a regression bug?

smblott-github:
This is probably a very old bug, I think. `<c-[>` is hardwired as `Escape`, the check for `Escape` is done before the check for other mappings.

PR #2217, opened 2016-08-15
replace \xA0 on copying and pasting

gdh1995:
The char `\xA0` may be contained in a string if we use `y` to yank a line which contains `&nbsp;`.

### DIFF
--- a/lib/clipboard.coffee
+++ b/lib/clipboard.coffee
@@ -8,7 +8,7 @@ Clipboard =
   # http://groups.google.com/group/chromium-extensions/browse_thread/thread/49027e7f3b04f68/f6ab2457dee5bf55
   copy: ({data}) ->
     textArea = @_createTextArea()
-    textArea.value = data
+    textArea.value = data.replace /\xa0/g, " "
 
     document.body.appendChild(textArea)
     textArea.select()
@@ -22,7 +22,7 @@ Clipboard =
     document.execCommand("Paste")
     value = textArea.value
     document.body.removeChild(textArea)
-    value
+    value.replace /\xa0/g, " "
 
 
 root = exports ? window


smblott-github:
Thanks, @gdh1995.  Could you provide a pointer to an example page, please?

gdh1995:
Like this: https://jsfiddle.net/807khgcw/

Some pages may use &nbsp; to force there're spaces before text.

smblott-github:
Thanks, @gdh1995.  But I cannot recreate the problem.  I select an area of text including `&nbsp;`, then use visual mode (say) to yank it: `vy`.  I'm getting regular spaces in the copied text.

Am I trying this wrong?  Might it be platform specific?

gdh1995:
Oh I forgot to explain where `\xa0` chars are. You will only see "regular" spaces if you paste it to Notepad.exe or some other simple text editors, because `\xa0` is shown as ` ` by default. This char can be picked easily if you has a HEX editor:

image

How to repeat:
1. select and press `vy`.
2. paste the text to Notepad.exe, and save to a file **using Unicode encoding** .
3. use a HEX editor to open the file, and you can see `A0 00 A0 00` which represents two `\xa0` chars.

Although `\xa0` is usually shown as ` `, it will affects most text toolsets including `sed`, `grep` and so on.

Env: Windows 10 x64, Chrome 53 stable x64, Vimium-master
Tools: Notepad.exe, WinHex / BeyondCompare / ..., and a snapshot tool.

smblott-github:
OK, @gdh1995.  But that makes me wonder whether this is the correct thing to do.  What does Chrome (regular `<Ctrl-C>`) do?

gdh1995:
image
This snapshot is what Chrome copys into the system clipboard on `Ctrl+C` (converted into Unicode encoding: UCS-2 LE with BOM).

As far as I know, a '\xa0' is only used to enforce there's a space in HTML pages, and usually in a code box or a `<textarea>`.

And if a Chrome's version is not larger than 50 (up to 50.0.2661.75, stable, 64bit):
* in a `<textarea>`, if a line starts with a few spaces, then the first space char must be transformed into `\xa0`. Otherwise the spaces will not be shown.

smblott-github:
So, it looks like Chrome does not do the mapping you're proposing, is that right?  Id so, then Vimium probably shouldn't either.

gdh1995:
> Chrome does not do the mapping you're proposing

No. I'm sorry that I gives two little related points in a comment. Now let me give more explanation.

1. The snapshot
In the snapshot I gave above, `20 00 20 00` is highlighted, and they means Chrome converts all `&nbsp;` in HTML text into `\xa0` on copying.

smblott-github:
OK.  So we should do same.  Which means *not* merging this PR, right?

gdh1995:
@smblott-github  I've updated my last comment, because it had wrong meanings because of my poor English. Therefore please review it again. Thanks.

ykgmfq:
I took a look at this. When you yank the space in front of "Here" on JSFiddle and paste it in the Unicode table search, you get the No-Break Space: https://unicode-table.com/en/search/?q=%C2%A0.
Copying with Chrome gives gives you a normal Space: https://unicode-table.com/en/search/?q=+.

This PR looks good to me.

smblott-github:
Not sure, myself.  I can't seem to generate any problematic cases.  However, this PR is simple, so let's just merge and give it a go.

Issue #2214, opened 2016-08-10
How do I assign 'backspace' to goBack?


rikkilake:
Hello,

In it's on-going power trip Google Chrome in an overnight update has stopped my backspace being used as Go Back... this shortcut has been around since the year dot.  Idiots!

How can I correctly map this key?  Thank you.

PS. Yes, I've looked at the Wiki, but cannot find any information on how to do this.

gdh1995:
#2207 also mentioned this. Vimium hasn't supported mapping `backspace` now.

jaekyeom:
How about using [this](https://chrome.google.com/webstore/detail/go-back-with-backspace/eekailopagacbcdloonjhbiecobagjci)?

gdh1995:
> How about using this?

So I've written a PR to implement mapping `<backspace>`, which borrows some code from the extension `Go Back With Backspace`.

smblott-github:
What @jaekyeom said (and I'll take a look at #2223 soon).

iahu:
+1

PR #2223, opened 2016-08-18
support mapping <backspace>

### LINKED ISSUE #2207: "Custom key mappings" does not work.
The  chrome version I using is 52.0.2743.82－32 bit.
I want to make pressing "backspace" in the keyboard to go back in history，just like press "H".
So I add a rule--map backSpace goBack.
But It does not work.
So What's the right rule?
Can you help me?

gdh1995:
Chrome stop using <backspace> to go back, but some still wants this
feature.

This should fix #2207 and #2214.


lib/keyboard_utils.coffee
```diff
@@ -49,6 +49,8 @@ KeyboardUtils =
       event.key.toLowerCase() # F1 to F9.
     else if event.key.length == 3 and "F10" <= event.key <= "F12"
       event.key.toLowerCase() # F10 to F12.
+    else if event.key.length > 3 and event.key in ["Backspace"]
+      event.key.toLowerCase() # F10 to F12.
     else
```
gdh1995: Here's a typo... I forgot to remove the tail comment.

background_scripts/commands.coffee
```diff
@@ -97,7 +97,7 @@ Commands =
     # Keys are either literal characters, or "named" - for example <a-b> (alt+b), <left> (left arrow) or <f12>
     # This regular expression captures two groups: the first is a named key, the second is the remainder of
     # the string.
-    namedKeyRegex = /^(<(?:[amc]-.|(?:[amc]-)?[a-z0-9]{2,5})>)(.*)$/
+    namedKeyRegex = /^(<(?:[amc]-.|(?:[amc]-)?[a-z0-9]{2,})>)(.*)$/
```
smblott-github: Is there a reason why we have the same (complex) regex twice?  Here, and on line 44.

lib/keyboard_utils.coffee
smblott-github: Also, we don't need `event.key.length > 3 and`.

background_scripts/commands.coffee
gdh1995: They're used for different reasons:
* the one on line 44 is to make non-printable key names (length >= 2) lower-case, and only matches at most one status keys (`altKey/metaKey/ctrlKey`) and a non-printable key,
which means shiftKey is not supported.
* the one on line 100 is to split all key names which is wrapped by `<` and `>`.

In fact, I think we should support `shiftKey` for non-printable keys and combination of status keys, you may see #2210 for what I want.

lib/keyboard_utils.coffee
```diff
@@ -49,6 +49,8 @@ KeyboardUtils =
       event.key.toLowerCase() # F1 to F9.
     else if event.key.length == 3 and "F10" <= event.key <= "F12"
       event.key.toLowerCase() # F10 to F12.
+    else if event.key == "Backspace"
+      event.key.toLowerCase() # Backspace.
```
smblott-github: Should this not be handled via `@keyNames` (rather than as dedicated logic)?

lib/keyboard_utils.coffee
```diff
@@ -60,6 +62,7 @@ KeyboardUtils =
       if (event.keyCode >= @keyCodes.f1 && event.keyCode <= @keyCodes.f12)
         return "f" + (1 + event.keyCode - keyCodes.f1)
       return ""
+    return "backspace" if event.keyIdentifier == "U+0008"
```
smblott-github: This should probably also be in `@keyNames`.

gdh1995: No, because `keyNames` can only be searched `if (event.keyIdentifier.slice(0, 2) != "U+")`

smblott-github: So?  Why can it **only** be searched in those circumstances?

(I'm not a fan of hard wiring key names into the logic.  We do it for the function keys, but only because there is a range of them.)

gdh1995: I've no enough knowledge about the line `if (event.keyIdentifier.slice(0, 2) != "U+")` - it has been there since commit 50a4fd9, with a comment `Not a letter` by Phil Crosby.

But it seems safe (or little harmful) to remove / delay this statement, for `keyIdentifier` has been deprecated and is to be removed on Chrome 54.

Do I need to make a new commit to delay it?

smblott-github:
But this won't work now, @gdh1995, no?  Because, as you pointed out previously, the `keyIdentifier` for `Backspace` **does** start with `U+`.  Do we not need something like [this](https://github.com/smblott-github/vimium/commit/b82e0c8f87bc0e6650e38e4ca7901e21d9ef327f)?

lib/keyboard_utils.coffee
gdh1995: These two lines has been removed, so may the `Changes requested` flag be removed?

PR #2242, opened 2016-09-09
hint a label only if its control is not disabled

gdh1995:
example:
``` html
<label><input type=checkbox disabled />exlpanation</label>
```

This may occurs for some settings page which includes some settings not allowed to change:
* e.g. Fuel (OpenStack deploy tools) settings page

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -621,7 +621,8 @@ LocalHints =
       when "button", "select"
         isClickable ||= not element.disabled
       when "label"
-        isClickable ||= element.control? and (@getVisibleClickable element.control).length == 0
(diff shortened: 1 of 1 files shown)


smblott-github:
Looks good, @gdh1995 - thanks.  Is there a page where I can test this?

gdh1995:
here's a test: https://jsfiddle.net/hw1vag2d/

image

gdh1995:
Some pages hide checkboxs/radios and draw their own controls for beauty, and then hints for labels are useful. But Vimium now treated disabled items as clickable, which is bad, so this PR wants to filter out them.

smblott-github:
Thanks, @gdh1995.

PR #2250, opened 2016-09-14
Use prefix matching for global marks.

### LINKED ISSUE #2248: Suggestion: Change behaviour of global marks
With global marks, there's a usability issue that really bugs me...

I use global marks for the big important sites: email, facebook, twitter, etc.  Now, on all of those sites, I'll often navigate away from the initial landing page, so the loaded URL changes.  Then, when you use a global mark, you get a new tab. Bah!  Not what I want.  I want to be back at that original tab.

Might it not be better for global marks to choose the tab with the shortest URL which has the mark's URL as a prefix?

E.g. if the mark URL is

```
https://inbox.google.com/u/0/
```

and I have a tab

```
https://inbox.google.com/u/0/sent
```

then visiting that existing tab is a better UX than creating a new tab.

Yes?  No?

Edit: As a tweak, you could exclude the current tab for consideration.

smblott-github:
This changes the logic for selecting an existing tab when using global marks.

Previously, an exact URL match was required.  Here, we only require a prefix match.

For example, if the global-mark URL is:

    https://inbox.google.com/u/0/

Then a tab with the URL

    https://inbox.google.com/u/0/sent

will be selected.

This is a more usable approach when the user uses global marks to visit important sites like gmail, Facebook or Twitter.  On these sites, the URL changes, but the user still thinks of the tab as their "gmail tab", for example.

Also, when choosing between multiple candidate tabs:

- If there is at least one candidate in the current window, then only consider candidates in the current window.
- If there are more than one candidates, then don't select the current tab.
- Finally, select the remaining candidate with the shortest URL.

Closes #2248.

### DIFF
--- a/background_scripts/marks.coffee
+++ b/background_scripts/marks.coffee
@@ -66,7 +66,7 @@ Marks =
     # If we're not going to be scrolling to a particular position in the tab, then we choose all tabs with a
     # matching URL prefix.  Otherwise, we require an exact match.
     query = if markInfo.scrollX == markInfo.scrollY == 0 then "#{markInfo.url}*" else markInfo.url
-    chrome.tabs.query { url: "#{markInfo.url}*" }, (tabs) =>
+    chrome.tabs.query { url: query }, (tabs) =>
       if 0 < tabs.length
         # We have at least one matching tab.  Pick one and go to it.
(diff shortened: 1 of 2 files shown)


philc:
Title of this PR is incorrect.

PR #2251, opened 2016-09-14
Adds new check to see if an element is covered by another

poacher2k:
Replaces #2166

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -46,5 +46,6 @@ Contributors:
   Michael Salihi <admin@prestance-informatique.fr> (github: PrestanceDesign)
   Dahan Gong <gdh1995@qq.com> (github: gdh1995)
   Scott Pinkelman <scott@scottpinkelman.com> (github: sco-tt)
+  Daniel Skogly <daniel@pnkt.no> (github: poacher2k)
 
 Feel free to add real names in addition to GitHub usernames.
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -709,34 +709,38 @@ LocalHints =
           false # This is not a false positive.
         element
 
-    # TODO(mrmr1993): Consider z-index. z-index affects behaviour as follows:
-    #  * The document has a local stacking context.
-    #  * An element with z-index specified
-    #    - sets its z-order position in the containing stacking context, and
-    #    - creates a local stacking context containing its children.
-    #  * An element (1) is shown above another element (2) if either
-    #    - in the last stacking context which contains both an ancestor of (1) and an ancestor of (2), the
-    #      ancestor of (1) has a higher z-index than the ancestor of (2); or
-    #    - in the last stacking context which contains both an ancestor of (1) and an ancestor of (2),
-    #        + the ancestors of (1) and (2) have equal z-index, and
-    #        + the ancestor of (1) appears later in the DOM than the ancestor of (2).
-    #
-    # Remove rects from elements where another clickable element lies above it.
+    # This loop will check if any corner or center of element is clickable
+  	# document.elementFromPoint will find an element at a x,y location.
+  	# Node.contain checks to see if an element contains another. note: someNode.contains(someNode) === true
+  	# If we do not find our element as a descendant of any element we find, assume it's completely covered.
     localHints = nonOverlappingElements = []
     while visibleElement = visibleElements.pop()
-      rects = [visibleElement.rect]
-      for {rect: negativeRect} in visibleElements
-        # Subtract negativeRect from every rect in rects, and concatenate the arrays of rects that result.
-        rects = [].concat (rects.map (rect) -> Rect.subtract rect, negativeRect)...
-      if rects.length > 0
-        nonOverlappingElements.push extend visibleElement, rect: rects[0]
-      else
-        # Every part of the element is covered by some other element, so just insert the whole element's
-        # rect. Except for elements with tabIndex set (second class citizens); these are often more trouble
-        # than they're worth.
-        # TODO(mrmr1993): This is probably the wrong thing to do, but we don't want to stop being able to
-        # click some elements that we could click before.
-        nonOverlappingElements.push visibleElement unless visibleElement.secondClassCitizen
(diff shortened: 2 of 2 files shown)


smblott-github:
The right-hand *Suggested Groups* pane on Facebook....

Before...

snapshot

After...

snapshot

We're now missing a lot of hints.  Any ideas, @poacher2k?

poacher2k:
Good find! It seems that the case with these missing links is that the wrapping anchor tag itself doesn't have a any box size, and therefore only its children are found with `document.getElementFromPoint`.
anchor tags:
image
direct children:
image

I'll have to do some thinking here, as this is in theory correct behavior (click candiate is not in fact a visible element), but in practice not quite so.

Let me know if you come up with any ideas as well!

poacher2k:
@smblott-github I added a fix, and as far as I can see, things are working really smoothly now. Did some more extensive testing on facebook, reddit and github, and it seems to improve things a lot.

smblott-github:
@poacher2k... See 5a32ecdd858cc95e504129747a5580dc27590b68.

I'm going to try it out for a bit before merging.

(Tests aren't passing for now.)

poacher2k:
I'm also trying it out, but yeah - I see tests aren't passing. Any clue on why, or hints on how to debug the tests that are failing?

smblott-github:
@poacher2k... I'm still using [this branch](https://github.com/smblott-github/vimium/tree/poacher2k-master-squashed) which includes your work tweaked and squashed.  Are you too?

Do you have any sense as to whether it's faster or slower than the status quo?

I good test is to scroll a looooong way down on Twitter or Google Plus and then activate hints.

poacher2k:
@smblott-github: I'll do some tests on this after work today!

I believe the tweaked version (with only one check instead of two) is definitively faster than the one with two, but you might be talking about the main vimium branch?

smblott-github:
> is definitively faster than the one with two

That's what I think.

poacher2k:
Then I think we should go for that one. The reason I reverted the tweak was because Travis failed to build with the tweak and succeeded without it.

smblott-github:
Tests are fixed over in my branch of this: 1c66503d1d2d1352ffacf0d9753d8310ea35aaf9.

The problem was that the tests were making too many assumptions about the implementation.

poacher2k:
That's great! I feared I had broken something. Will be testing with your branch soon.

poacher2k:
I tested on twitter.com scrolled down far enough to make document.body.scrollHeight ~157k, and did some timings from F was pressed until hints appeared.

On master: ~0.7s
On poacher2k-master-squashed: ~0.84s
On poacher2k-master-squashed with tweak: ~0.7s

Seems pretty solid to me, will do some more testing when I come back home.

poacher2k:
Tried going as far down I could on Twitter, Reddit and Facebook - all sites with loads of elements - and it never really goes more than ~0.9s for the link hints to show.

It should be said that at the bottom of Reddit/Twitter, the entire browsing experience gets sluggish - not just vimium.

smblott-github:
@poacher2k...  Thanks for that.

Now, a problem.  From Facebook....

snapshot

`master` finds the link labelled "44" (probably the most important link!).  Your approach (as implemented over in my other branch), *does not*.  This happens consistently (for all Guardian posts), but I cannot find other examples.

poacher2k:
Hmm, I can't reproduce this it seems.

image

Do you experience it every time? Definitively worrisome, as that's the most important link for sure!

poacher2k:
Is there any interest for me to work this into the current codebase? As far as I can see, this is still an issue that when solved could improve the user experience on several websites.

Some examples:
github
GitHub
reddit
Reddit
YouTube
YouTube
Medium
Medium (only 4 of these are actually clickable with the mouse)

poacher2k:
Looking at the https://github.com/philc/vimium/pull/2251#issuecomment-249766333 I realise I actually _did_ reproduce it in https://github.com/philc/vimium/pull/2251#issuecomment-249983140 . 

Not really sure what is happening here (and the post is long gone), but I believe the "44" label may have been underneath the main post thumbnail and not actually visible although correctly placed. 

I'll take extra notice of this if/when doing more work on this.

philc:
I looked at this and followed the comment history, but couldn't find a the original rationale, or recent examples. However, I did reproduce this modern example from medium. 1st is before this patch, 2nd is after. You can see that the 2nd screenshot omits many link hints which aren't clickable by the mouse because they're under an overlay, so they probably shouldn't be exposed by link hints (even though they are indeed clickable if the overlay is removed).

I've squashed this to one commit, tidied up some comments, and merged it into master in caf13c2dbfa9b3c5f35aae76dfcffa4cfea14bac. We can now easily test and revert this if it surfaces many suboptimal edge cases once it hits a wider audience.

I didn't want to lose this PR in the upcoming migration away from coffeescript.

CC @gdh1995 in case you've modified Vimium-C's linkhints in this way and have any commentary.

gdh1995:
I haven't implemented this, maybe because of concerns about speed / code size. But this indeed looks well, and now I agree that it should be executed.

poacher2k:
Wow @philc , this is certainly a pleasant surprise! Thank you so much for looking at this! I've been eagerly wanting this ever since I made the PR in 2016! :)

gnarendran:
This seems to break  "Traverse shadow DOM to find links #3406" - particularly if the test on element.contains rejects shadow elements...

philc:
Follow-up discussion about @gnarendran's comment is in #3474

RomainBitard:
I believe this change makes me miss a lot of links that I could navigate before without the mouse, now I have to partially use the mouse some times (😞). 
Am I the only one with this issue now ?

poacher2k:
@RomainBitard : would be great with some links and screenshots so that those issues can be ironed out in that case :)

gnarendran:
@RomainBitard : You are not alone, there are already followup issues  #3492 #3493 #3501(closed). But I do not know if these three are the only regressions, and if you have encountered a different one. I guess that is why @poacher2k is requesting your problem links. Regards.

PR #2253, opened 2016-09-17
Add escape command: 'map <c-c> exitMode' [WIP]

### LINKED ISSUE #301: Allow any key to mapped to "exit mode" instead of just <esc>

smblott-github:
This adds:

    map <c-c> exitMode

Any such mappings are used to exit Vimium's various modes (insert mode, visual mode, find mode, etc.)

Limitations:

- Only non-printable, **single-key** mappings are supported.

So, for example, the following **does not** work:

    map jk exitMode

Fixes #301.

**Edit**...

This doesn't allow mappings like `jk` for a number of reasons:

- With the current Vimium code base, we need to know that the key is to be treated as `Escape` on `keydown`, and we don't know the printable character at that point.

- Consider insert mode.  For multi-key mappings, what do we do with the first key (`j`, in the example) event?  We can't hold it back, because the underlying page might do something with it.  However, when we later see the second key (`k`, in the example), we cannot undo the effect of the first key.  (We've no idea what the page itself did with the `j`).  Similarly for other Vimium modes.  E.g., what if `jk` appears as the hint for a link hint?  Does `jk` mean activate the hint?  Or does it mean `Escape`?  Regardless, the logic becomes complex, and it's not obvious to users what is happening.

So, I don't think we should allow mappings like `jk`.  What we could do, however, is log a warning on the console if a user attempts to establish such a mapping.

**Edit**...

Changed new command name from `escape` to `exitMode`.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -63,7 +63,9 @@ Commands =
               delete @keyToCommandRegistry[key]
 
           when "unmapAll"
-            @keyToCommandRegistry = {}
+            # We do not unmap "<c-[>" because it was previously hardwired into KeyboardUtils.isEscape(), and
+            # therefore unaffected by "unmapAll".
+            @keyToCommandRegistry = "<c-[>": {command: "exitMode"}
 
     # Push the key mapping for passNextKey into Settings so that it's available in the front end for insert
     # mode.  We exclude single-key mappings (that is, printable keys) because when users press printable keys
@@ -303,6 +305,9 @@ defaultKeyMappings =
 
   "m": "Marks.activateCreateMode"
   "`": "Marks.activateGotoMode"
+
+  # <c-[> is mapped to <Escape> in Vim by default.
+  "<c-[>": "exitMode"
 
 
 # This is a mapping of: commandIdentifier => [description, options].
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -73,13 +73,15 @@ Commands =
       (key for own key of @keyToCommandRegistry when @keyToCommandRegistry[key].command == "passNextKey" and 1 < key.length)
 
     # Push any key mappings for escape into Settings so that they are available in KeyboardUtils.isEscape(),
-    # and remove those bindings from the key-to-command registry (they're not needed).
+    # and remove those bindings from @keyToCommandRegistry (they're not needed).
     escapeKeyBindings =
       (key for own key of @keyToCommandRegistry when @keyToCommandRegistry[key].command == "escape")
-    for own key of escapeKeyBindings
-      delete @keyToCommandRegistry[key]
 
-    Settings.set "escapeKeyBindings", escapeKeyBindings
+    if 0 < escapeKeyBindings.length
+      delete @keyToCommandRegistry[key] for own key of escapeKeyBindings
+      Settings.set "escapeKeyBindings", escapeKeyBindings
+    else
+      Settings.clear "escapeKeyBindings"
 
   # Command options follow command mappings, and are of one of two forms:
   #   key=value     - a value
@@ -219,8 +221,7 @@ Commands =
     "enterVisualLineMode",
     "toggleViewSource",
     "passNextKey",
-    "escape"
-  ]
+    "escape"]
 
 defaultKeyMappings =
   "?": "showHelp"
@@ -337,7 +338,7 @@ commandDescriptions =
   enterVisualLineMode: ["Enter visual line mode", { noRepeat: true }]
 
   focusInput: ["Focus the first text input on the page"]
-  escape: ["Escape"]
+  escape: ["Escape (exit the current Vimium mode)"]
 
   "LinkHints.activateMode": ["Open a link in the current tab"]
   "LinkHints.activateModeToOpenInNewTab": ["Open a link in a new tab"]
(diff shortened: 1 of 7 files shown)


smblott-github:
Update, with dc31d6c3fe78a0b195478b2b04c5d77412522dcf, <c-[> is treated as a regular binding, therefore users can `unmap` it on the options page.

gdh1995:
This usage looks good to me.

I want to discuss if we may add a new setting item directly for this `exitMode`, just like exclusion passKeys, which means  we don't need to change background key mapping logic and add an extra keybinding limit to this special "command".

The setting item's value should be some non-printable, single-key mappings joined with `` (empty string) or   ` ` (space char):
* default to `<c-[>`.
* `<c-[> <c-c>` is valid and means `<c-[>` and `<c-c]` are all valid "escape" keys.

smblott-github:
Here's a bit of a show stopper for this PR.

What are the obvious bindings users might want for `exitMode`?

My guess is `jj`, `jk` and `<c-c>`.

We cannot support the first two of those.

For the third, a mapping:

    map <c-c> exitMode

means that the user cannot select text and *copy-to-clipboard* (`<Control-c>`) in insert mode, for example while editing and email.

That leaves no useful bindings.

What to do?  We could merge this and let the user decide, or we could accept that - while it might seem like a great idea -  remapping `exitMode` doesn't work that well in a browser environment.

Are there other common bindings?

y2kbugger:
I don't understand why `jk` woulda be an issue. Can't you just do what vim and readline(with correct setup) -- capture the `j`, wait for up to 250ms(or whatever) and if a k follows within that time execute exitMode, else pass the keystroke to Chromium.

For what it's worth, I quit using vimium because I couldn't `jk`

smblott-github:
@y2kbugger... 

Printable characters are a problem because we don't know the printable character when we get the `keydown` event, so we *have to* leak that event (which means the page sees it and might do something).  This may be fixed with `event.key` which is rolling through at the moment, but that has bugs too.

Two printable characters are a problem because we have no idea what the page did with the first one.  Sure, normally it will be sitting in an input just in front of the cursor.  But who really knows?  The page could do anything with that character.  For example, if Vimium is in insert mode on GMail and you use GMail's native `j` binding followed by `k` (down message and up message), suddenly you're popped out of insert mode.  And not knowing what the page did with the `j`, we have no way of undoing it.

The problem, I think, is that vim is a text editor and Chrome is a browser.  Vim knows exactly what happens on every keystroke, Vimium does not.

My sense is to ditch this PR.  It's true that you could, for example, allow `jk` to work in text inputs only (say).  But that mucks up the consistency of the UX.

danijar:
Regarding `<c-c>`: There is noting like Vim's normal mode for text input fields right now, correct? Otherwise we could switch to that on `<c-c>` and use visual selection plus yank to copy text.

bmcilw1:
My personal favorite, might be worth considering: ;; I use double semi-colons

smblott-github:
I'm going to bin this.  Here's a summary of why.  Basically, it's a "good idea" that turns out to be not such a great idea in practice.

Assume:

    map jj exitMode
    map ee exitMode
    map <c-c> exitMode

- On GMail, the user hits `i` so that they can use GMail's native bindings. `j` (next message) and `j` drops them back into Vimium's normal mode.

- Similarly on Gmail, the user hits `ee` to return to normal mode,  but GMail grabs the first `e` and archives a message.

- For users of alphabet hints (most users, I think), `jj` can be a hint (or a substring of a hint).  Or in find mode, the user could search for "jj".  These can be fixed, but there are *lots* places in the code base which would require special handling.  There's a cost to that, and I'm not sure it's worth it.

- If there is selected text within an input, then `<c-c>` no longer copies the text, instead blurring the input and unselecting the text.

All of this just creates a confusing experience for users.

danijar:
@smblott-github If there was a vim editor for input fields (which would be awesome and some Vim plugins for Chrome have that), I would use `y` for copy and `<c-c>` is free to go back to normal mode.

hypnoglow:
`exitMode` mapping should definitely be configurable. This can be default as it now, but it should not be hardwired.

smblott-github:
I'll reopen this for discussion, because there is long-standing and current interest in this.

Here are some things we cannot have.

We cannot have users, particularly new users, configuring

    map jj exitMode

and then wondering why it doesn't work.

Similarly, we cannot have new users configuring

    map <c-c> exitMode

and then getting themselves in a mess because they cannot use `<c-c>` to copy text within inputs.

I'm not too keen on it, but perhaps a compromise would be...

- Merge this, but tweak the help page such that `exitMode` is *never* listed.
- Add a description of the feature (including its limitations) to an "Advanced usage" section on the wiki.

That would avoid creating difficulties for new users, and would ensure that users who choose to use `exitMode` understand its limitations.

danijar:
@smblott-github That seems like a reasonable compromise to me.

y2kbugger:
I can imagine a person who configures vim and wants to go as far as
configuring their browser and being surprised how `map jj exitMode` will
cause side affects and interactions.

smblott-github:
See #2306.

#2306 is less disruptive and more flexible than this, and solves pretty much the same problem.

smblott-github:
Closing in favour of #2306.

PR #2260, opened 2016-09-20
Make vomnibar suggestions clickable

### LINKED ISSUE #615: Make entries in the vomnibar clickable
I like Vimium very much!
I mapped the keys only under my left hand ( **QWERTASDFGBVC** ) for link hint
so that i could use my left hand for link hinting,and,my right hand holding on the mouse for something else.I like using Vimium this way.
But when I want to open a bookmark ,I should hint **B** , text some key words ,and then move one of my hands to hint **ENTER**.
That makes me feel uncomfortable.
If you could let the matched bookmark items clickable ,it Vimium be more easy and fun to use! :-)

smblott-github:
This makes Vomnibar suggestions clickable with the mouse.

It would be nice to also make them selectable with the keyboard (without having to `Tab` all the way down.  The most obvious bind for that would be `<c-N>` (e.g. `<c-7>`).  However, unfortunately Chrome already uses those bindings.

Fixes #615.

### DIFF
--- a/pages/vomnibar.coffee
+++ b/pages/vomnibar.coffee
@@ -202,7 +202,16 @@ class VomnibarUI
         @completions = results
         @selection = if @completions[0]?.autoSelect then 0 else @initialSelectionValue
         # Update completion list with the new suggestions.
-        @completionList.innerHTML = @completions.map((completion) -> "<li>#{completion.html}</li>").join("")
+        @completionList.innerHTML = @completions.map(
+          (completion, index) -> "<li data-index=\"#{index}\">#{completion.html}</li>"
+        ).join("")
+        # Add click listeners for each completion.
+        for element in @completionList.childNodes
+          do (element) =>
+            element.addEventListener "click", (event) =>
+              openInNewTab = event.shiftKey or event.ctrlKey or event.altKey or KeyboardUtils.isPrimaryModifierKey event
+              completion = @completions[parseInt element.dataset.index]
+              @hide -> completion.performAction openInNewTab
         @completionList.style.display = if @completions.length > 0 then "block" else ""
         @selection = Math.min @completions.length - 1, Math.max @initialSelectionValue, @selection
         @updateSelection()
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -53,21 +53,29 @@ class Suggestion
     @html =
       if request.isCustomSearch
         """
-        <div class="vimiumReset vomnibarTopHalf">
-           <span class="vimiumReset vomnibarSource #{insertTextClass}">#{insertTextIndicator}</span><span class="vimiumReset vomnibarSource">#{@type}</span>
-           <span class="vimiumReset vomnibarTitle">#{@highlightQueryTerms Utils.escapeHtml @title}</span>
-           #{relevancyHtml}
-         </div>
+        <div class="vimiumReset">
+          <div class="vomnibarSuggestion">
+            <div class="vomnibarTopHalf">
+              <span class="vomnibarSource #{insertTextClass}">#{insertTextIndicator}</span><span class="vomnibarSource">#{@type}</span>
+              <span class="vomnibarTitle">#{@highlightQueryTerms Utils.escapeHtml @title}</span>
+              #{relevancyHtml}
+            </div>
+          </div>
+        </div>
         """
       else
         """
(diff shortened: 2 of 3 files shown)


tobimensch:
> The most obvious bind for that would be <c-N> (e.g. <c-7>). However, unfortunately Chrome already uses those bindings.

No, the whole point of vim is to keep your fingers on the home row as much as possible, and numbers aren't ideal for navigation. It should work the same way the link feature (f) works, actually it should be possible to reuse that code and simply apply it to this set of links.

smblott-github:
> It should work the same way the link feature (f) works, actually it should be possible to reuse that code and simply apply it to this set of links.

That may not be quite a simple as it sounds:

- How does the user activate link-hints mode?  It cannot be `f`.  And it cannot be `Esc` `f`.
- How does the user *know how to* activate link-hints from the Vomnibar?

If we were to do something like this, then it needs to be intuitive and easily discoverable.  And ideally not require anything extra on the options page.

tobimensch:
@smblott-github : 
I updated my text, I think it answered your question.

smblott-github:
> Tab isn't needed there anyway, since we can also use ctrl+j/k

That is the long-established way the Vomnibar works.  We should not change it.

There currently isn't (or, at least, shouldn't be) a way of getting the focus out of the Vomnibar input.  And most of the obvious keystrokes mean something already.  In particular, `Escape` means something. Also, there should either be a strong vim analogy or it should be easily discoverable.

I don't see any attractive solutions to that problem at the moment.

tobimensch:
> That is the long-established way the Vomnibar works. We should not change it.

I don't like that argument. Because something used to work this way, it should continue to work this way, although there're other needs and better solutions.

In the current work flow, I would p refer ctrl+j/k anyway, and I assume most would, since as I said it is more vimish.

One hit on tab gets you on the first link, but doesn't open it yet. So with one hit and my proposed method, you would probably get an "A" tooltip on the first link, a "B" tooltip on the second and so forth, which means for the first link you would need to hit two keys in total (Tab+A), just like now. For the second link you would still need to hit two keys in total (Tab+B), whereas with our current method you need to hits (Tab+Tab+Enter). So from the second link forward the usability is worse. And I would argue that usability is also worse for the first link now, since you have to hit (Tab+Enter), and Enter is generally more difficult to hit than a, where your left pinky rests.

philc:
Ctrl-j and k work fine. Note that ctrl-n and p would work fine on Mac since
they're not shadowed by anything, but not Windows or Linux. Ctrl-n and
ctrl-p has precedence because it's used in Bash for this purpose.

smblott-github:
> So with one hit and my proposed method, you would probably get an "A" tooltip on the first link, a "B" tooltip on the second and so forth, which means for the first link you would need to hit two keys in total (Tab+A), just like now. For the second link you would still need to hit two keys in total (Tab+B)

The Vomnibar input is (and should still be) active when the user has selected a suggestion.  In tab and bookmarks modes, and when a custom search engine is being used, we automatically select the first suggestion.  And, when a search-engine suggestion is selected, we add the text to the vomnibar input so that the user can edit it or append to it.  In these cases, "A" and "B" have meaning already, so we cannot use them as a form of link hint.

(Also, regardless individual preferences, we cannot repurpose `Tab`.  It's too core to the Vimium UX, and is used with a similar meaning elsewhere.)

pages/vomnibar.coffee
```diff
@@ -202,7 +202,16 @@ class VomnibarUI
...
         # Update completion list with the new suggestions.
-        @completionList.innerHTML = @completions.map((completion) -> "<li>#{completion.html}</li>").join("")
+        @completionList.innerHTML = @completions.map(
+          (completion, index) -> "<li data-index=\"#{index}\">#{completion.html}</li>"
+        ).join("")
+        # Add click listeners for each completion.
+        for element in @completionList.childNodes
+          do (element) =>
+            element.addEventListener "click", (event) =>
+              openInNewTab = event.shiftKey or event.ctrlKey or event.altKey or KeyboardUtils.isPrimaryModifierKey event
```
gdh1995: What about`@forceNewTab or`?

smblott-github:
Closing.  I don't think that we should do this.  There's very little demand for clickable vomnibar entries.  A better solution would be to find a keyboard-based method of achieving the same thing.  However, we don't really have a good proposal for doing that at the moment.

PR #2269, opened 2016-09-23
New feature/command: Mute/unmute tab.

tobimensch:
Mute/Unmute feature/command analogous to pin/unpin command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Manipulating tabs:
     X       restore closed tab (i.e. unwind the 'x' command)
     T       search through your open tabs
     <a-p>   pin/unpin current tab
+    <a-m>   mute/unmute current tab
 
 Using marks:
 
@@ -161,6 +162,7 @@ Release Notes
 
 Changes since previous release (not in the Chrome Store version)
 
+- New commands: "mute/unmute tab".
 - You can now map `<backspace>` to a Vimium command (e.g. `map <backspace> goBack`).
 - Bug fixes:
     - `<c-a-[>` is no longer handled (incorrectly) as `Escape`.  This also affects `<Alt-Gr-[>`.
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -171,6 +171,7 @@ Commands =
       "lastTab",
       "duplicateTab",
       "togglePinTab",
+      "toggleMuteTab",
       "removeTab",
       "restoreTab",
       "moveTabToNewWindow",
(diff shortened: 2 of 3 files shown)


smblott-github:
Thanks, @tobimensch.

I believe I may have said "no" to this feature in the past.  However, it is frequently requested, and the implementation is very simple, so I'm minded to merge this.

Just let me think a bit.

See ffebb6dd15a9320842603854109149bc26078ae5.

tobimensch:
@smblott-github:

Well, I'm hopeful you'll merge it. That fix you made in main.coffee line 213 I also had locally, but forgot to push it.

smblott-github:
@tobimensch.  Here's my current thinking on this.

If we release `toggleMuteTab`, then users will immediately request `toggleMuteAllTabs`.

One of the problems we face with Vimium is keeping the feature set small, discoverable and manageable (both for maintainers, and so that it's not overwhelming for new users).  If we were to add every command which users request, then things would rapidly get out of hand.  So, my preference is to be pretty selective when adding new commands.  Read more [here](https://github.com/philc/vimium/blob/master/CONTRIBUTING.md#vimium-design-goals).

One way around this is to have simple, clear commands which are useful to a large number of users, but with more advanced features too, but which are described only in the wiki (for example, [here](https://github.com/philc/vimium/wiki/Tips-and-Tricks#keyboard-shortcuts-for-custom-search-engines) and [here](https://github.com/philc/vimium/wiki/Tips-and-Tricks#custom-key-mappings-with-a-count)).

In this case, I'm wondering whether we shouldn't add what you have here:

    map <a-m> toggleMuteTag

as is, but also add

    map <a-M> toggleMuteTag all

as an advanced feature (with no default binding, and described only on the wiki).

If we do the latter, then we would have to think carefully about what it means to mute all tabs.  Here's a proposal:

- If there is at least one tab which is `audible` ([here](https://developer.chrome.com/extensions/tabs#method-query)) and not muted, then mute all `audible` tabs.
- otherwise, unmute all muted tabs.

tobimensch:
@smblott-github 
I like your proposal.

Especially toggleMuteTab other would be extremely useful.

tobimensch:
I'd actually map
<a-M> to toggleMuteTag other

by default. Since toggleMuteTab all is essentially toggleMuteTag other + toggleMuteTag (this).

And leave toggleMuteTab all for the user to map himself.

smblott-github:
OK.  I'm going to tweak this PR and merge it.  I'll take a look at the other features we've discussed (unless you'd prefer to, @tobimensch).

PR #2276, opened 2016-09-27
Treat key mappings and custom search engines textareas as code (monospace font treatment)

yandavid:


smblott-github:
@yandavid...  Thanks for this.

I cherry picked ba7302fd7b486a06ac9721287c88a6b4daf0eb1e.  We need that anyway.  I hope you don't mind.

Regarding the formatting: yes, those inputs should be using the `code` class.

But there's another problem with those inputs, and that problem is exacerbated by this PR.  Specifically, the input is so narrow that - for many lines - the user cannot see the whole line.  And that gets worse here because the fixed-width font takes up even more space.  My custom-search-engine input looks like this:

snapshot

Editing it is a pain.

It's not clear what to do about this.  Any ideas?  Perhaps we should make the inputs (and thus the columns) wider by some amount, subject to there being sufficient horizontal space on the page.

Mention: @philc.

yandavid:
@smblott-github You're very welcome, and 👍 on the cherry picking.

I have pushed up a css change that will ensure that the options page will spread as wide as your screen will allow. It's definitely not the most elegant or maintainable solution (for that, I think a responsive, flexbox-based layout would be best) but should help in the immediate future in giving users more space to work with when entering their strings.

I have also checked that the other pages with the same div (`div#wrapper`) are not affected, as there exists more css that handles their widths with more specificity.

gdh1995:
I use this function to auto-resize an element by click a button:
(says there's a `<button data-auto-resize="keyMappings">a</button>`)
``` javascript
  a.onclick = function() {
    var target = $(this.getAttribute("data-auto-resize")), delta, height = target.scrollHeight;
    if (height <= target.clientHeight) { return; }
    target.style.maxWidth = Math.min(window.innerWidth, 1024) - 120 + "px";
    delta = target.offsetHeight - target.clientHeight;
    delta = target.scrollWidth > target.clientWidth ? Math.max(26, delta) : delta + 18;
    height += delta;
    delta = target.scrollWidth - target.clientWidth;
    if (delta > 0) {
      target.style.width = target.offsetWidth + delta + "px";
    }
    target.style.height = height + "px";
  }
```

yandavid:
@gdh1995 I think this is fundamentally a visual issue, and should be solved with css. Your proposal would also require user action, which is not ideal.

It also looks like your code touches the height property of the textarea. If you're attached to using JavaScript, I think a much stronger (and easier) solution would be to edit the `rows` property of the textarea instead.

gdh1995:
> edit the `rows` property

To compute `rows` number, we need to get line height and divide `scrollHeight` by it, which I think not elegant, either.

What I want is:
* when the page is firstly loaded, all textarea are small and aligned to others
* if I want to edit a text box for a while, I'll click the button to zoom in it.

yandavid:
@gdh1995 Perhaps a better user experience would be to update on input/focus events instead of requiring a button click.

In any case, my goal for this PR was to change the font on those textareas to be monospace. I've made the textareas wider to accomodate the increase is font width, but I'm not convinced that the height needs to also change. Let's move unrelated discussions away from this PR.

gdh1995:
Oh yes I should not discuss auto resizing here.

But your PR makes the options page look too **ugly** for me on a large screen  (e.g. 1920x1080). Have you tested it on large screens?

yandavid:
@gdh1995 Could you attach a screenshot of what you mean? I personally use a 1080p monitor, and things look as I intend it to look. I want to make sure there are no layout issues.

If it's an issue of the page not being aesthetically pleasing at full width, perhaps capping the width with a `max-width` of 1200px would be a good idea.

gdh1995:
> `max-width` of 1200px

Yes, it's just what I want, though I think 1024px may be better.

But when I set `max-width` in your version's options page, it looked like:
image

This is because the `body {margin: 0 auto}` you set also prevents `body`'s descendants from filling the page's width.

yandavid:
@gdh1995 I think you've set the max-width on body, but I meant a max-width on `div#wrapper`.

yandavid:
@smblott-github Any thoughts on this? It'd be great if those fields could be monospaced.

smblott-github:
@yandavid...

> Any thoughts on this?

Yes...

1. The inputs should indeed be monospaced.
2. However, I'm worried too about the fact that there's even less of the inputs' contents visible.
3. I'm not yet happy with either of the proposals (what's here, and the "resize" button) to widen the options page.  We shouldn't need a resize button.  And your "fix" is too aggressive, I think.

If we do this, then I'd propose the version in [this branch](https://github.com/smblott-github/vimium/tree/yandavid-monospaced-fonts-on-options-page), specifically cc2fff970cf84f67c0006bbe6c90d294850b4e45.

yandavid:
Sure thing @smblott-github, I'm sure whatever width value you come up with will be fine, I'll close this PR and let you take care of the rest.

smblott-github:
Let's not close it.

Having it open signals (to me, at least), that the issue is still under consideration.

PR #2280, opened 2016-09-28
add button to auto resize option textarea

gdh1995:
Note: Scrollbars will be taken into consideration, and after resizing, they will disappear, so we can see and edit all the lines without scrolling anything.

This is an modified implement of https://github.com/philc/vimium/pull/2276#issuecomment-250064984.

### DIFF
--- a/pages/options.coffee
+++ b/pages/options.coffee
@@ -265,6 +265,33 @@ initOptionsPage = ->
 
   maintainLinkHintsView()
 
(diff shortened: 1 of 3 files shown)


smblott-github:
Thanks, @gdh1995; but I don't think we should do this.  Having an extra link to click is too weird.

gdh1995:
Um... So closed.

PR #2284, opened 2016-10-01
Enable silent/patch releases.

smblott-github:
Currently, all new releases trigger a notification.

This changes that behaviour such that if the previous and current releases have the same major and minor release numbers, then no notification is shown.

We could then push bug-fix and minor releases without bugging the user.

Such releases would have a number like `1.56.1`.

Edit.  Most of the diff is just an indentation change; the real change is just the logic starting [here](https://github.com/philc/vimium/pull/2284/files#diff-eb68fda05b0a59fbd1482418fd0223e8R481).


tobimensch:
I approve of this.

gdh1995:
LGTM.

A small advice: `ver[...2]` may be reworked into `parseFloat(ver)`.

PR #2300, opened 2016-10-07
Change initialisation sequence; do not inject into small frames

smblott-github:
Pages like Google Inbox, include between ten and fifteen tiny invisible iframes.

Vimium takes more than 50ms to load (for me) in *every* frame.  So, for Google Inbox, Vimium is accounting for almost a second of the total load time.

This PR:

- changes the initialisation sequence such that frames themselves request Vimium's content scripts to be injected (as opposed to listing them all in the manifest).
- and we *do not* inject content scripts into tiny frames.

There may still be some issues around ~~CORS~~ CSP that we need to consider.

Edit: I've checked a page with a strict CSP ([here](https://smblott.firebaseapp.com/secure)) and my bank, and it doesn't seem to be a problem.


gdh1995:
Chrome's origin injecting process for content scripts has such advantages:
* there's a special cache for these scripts, which means less cost
* extension code is able to run exactly on `document_start`, and exactly one by one
and on the contrary:

First, this PR uses `chrome.tabs.executeScript` which always reads new file content from disk. In my test on Chrome 53, if I changed a file, the second `executeScript` would use the new content, so I think there's no such a cache.

Second, this PR needs a message named `injectVimium` to call `executeScript`, so the main script files will be ran after `document_start`. Although this may not make any difference, I think we should take it into consideration.

Third, according to https://developer.chrome.com/extensions/tabs#method-executeScript, this API seems not to ensure execution sequence of many (called) files when we calls it many times directly and not using callback functions.

Therefore, it may be better to execute less JS functions (`do` keyword in coffeescript) and install less listeners if a window is too small, but this is just an immature idea, and more works will be needed by it.

smblott-github:
Thanks, @gdh1995.

> First, this PR uses `chrome.tabs.executeScript` which always reads new file content from disk...

This matters only insomuch as it affects the load time.  Some numbers on that below.

> so the main script files will be ran after `document_start`...

The *only* part which **has** to run at document start is the installation of the event listeners (so that we beat the page at installing our listeners).  That's unchanged here.  Even in `master`, Vimium doesn't actually become active until after at least two asynchronous events (a round trip to the background page to check whether Vimium is enabled and a lookup in `chrome.storage.local` to get the key bindings).  So, `master` doesn't really activate at `document_start` either.  What matters is how long it takes for Vimium to be active, to wit, see below.

> ...this API seems not to ensure execution sequence...

The documentation for `sendMessage` also doesn't guarantee in-order delivery (although that is what it does), and we already rely on in-order delivery there, I think.  If this is a problem, we'll find out pretty soon (I'm running this PR live at the moment, just to check it out).

> Therefore, it may be better to execute less JS functions

If we list files in the manifest, then they will be loaded in every frame.  That's what motivates this PR.

> and install less listeners if a window is too small

Then the logic around initialisation would be spread throughout the content scripts, which would be bad.  Whereas here, all of the logic is in one file.

> but this is just an immature idea, and more works will be needed by it.

Yes.

Here are the numbers.  In all cases, we're measuring the time from when Vimium is first active (`document_start`) until normal mode becomes active.  There are two sets of numbers, one measures the initialisation time in *all* frames, and one in just the top frame.  In each case, there is a simple (blank) page, the BBC News site and Google Inbox.  `I.` stands for "inject Vimium" (this PR). `M.` stands for `master`.  Groups of numbers in square brackets are the times for all of the frames in a single page.

```
All frames:
   Blank page:
      I. 101 106 93 91 101
      M. 89 105 76 82 86

   BBC News:
      I. [196 166 255 208] [148 204 366 191]
      M. [127 60 63 364 135 107 160] [137 129 98 76 138 108 152]

   Google Inbox:
      I. [153 905] [172 862]
      M. [158 270 295 154 149 159 144 120 160 94 74 84 68] [147 303 291 163 142 135 150 103 153 99 70 83 81]

Just top frame:
   Blank page:
      I. 103 98
      M. 86 91

   BBC News:
      I. 141 157
      M. 103 113

   Google Inbox:
      I. 194 174
      M. 142 155
```

This PR is slower, perhaps about 20-30% slower in simple cases.  However, the BBC and Inbox numbers show the amount of wasted work we're doing on more complicated sites.  It's not clear whether it's meaningful or not, but adding up the numbers amounts to one whole second for the BBC, and two whole seconds for Inbox.

That "wasted work" is the time that the user spends watching the the "loading" icon flicker.  On Inbox, we load Vimium in just two frames, instead of 13.

So, to summarise, it's a trade off.  Do we want do a little bit more work on simple pages in order to save a large amount of wasted work on more complicated pages?  That's difficult to say.  The only thing which matters, really, is the UX.  So I'm going to try this out for a bit and see whether it makes a noticeable difference, positive or negative.

(Sorry, that was long.)

gdh1995:
Well. most of your explanations are reasonable.

In fact, I think there're 3 simpler methods to speed up Vimium:
* use less `chrome.runtime.sendMessage` - use `Frame.port` to send handler messages
* use less `bgLog`, which is too heavy
* merge requests on initing, that means:
  * in `initializePreDomReady`, merge `checkIfEnabledForUrl document.hasFocus()` into `Frame.init`
  * in `Frame.init`, `port.name` uses `if document.hasFocus() then "frames" else "frames2"`
  * `main.coffee` gets the focused status by checking `port.name`, and reads the frame url from `port.sender.url`

smblott-github:
I have found a use case where this PR interacts badly with other Vimium features.

On Google Hangouts ([here](https://hangouts.google.com/)), the chat windows are initially loaded but tiny (so, with this PR, Vimium is not initialised).  When a chat is selected, its window (frame) is resized and Vimium is injected.  That's all fine.

The problem is that grab-back-focus (if enabled) is activated at that time too, whereas the user began interacting with the page quite some time ago!

It's possible to fix this, but that would require additional logic, for example, in `main.coffee`, and then the logic for the injection process - which is relatively localised in this PR - starts to leak into other parts of the code base.

Most likely this PR is for the bin.

philc:
Yes, this approach is a minefield. Sorry =(

PR #2301, opened 2016-10-07
handle documentElement's top/left margin correctly

### LINKED ISSUE #2291: Bug: Link hints not in correct position
See here:

http://debbugs.gnu.org/

I'm attaching a screenshot in case you can't reproduce it.
screenshot-vimium-link-hint-bug

gdh1995:
This PR fixes #2291 .

`document.documentElement` will be the offset parent of Vimium's divs if its `.style.position` is not `"static"`, but others (like `"absolute", "relative"`).

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -811,9 +811,16 @@ LocalHints =
         nonOverlappingElements.push visibleElement unless visibleElement.secondClassCitizen
 
     # Position the rects within the window.
+    if getComputedStyle(document.documentElement).position == "static"
+      top = window.scrollY
+      left = window.scrollX
+    else
+      rect = document.documentElement.getBoundingClientRect()
+      top = -rect.top
+      left = -rect.left
     for hint in nonOverlappingElements
-      hint.rect.top += window.scrollY
-      hint.rect.left += window.scrollX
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @gdh1995.

Would this be same: commit 4e2e86aac324b1510eea78fbf177be2d18ef131c, [code](https://github.com/smblott-github/vimium/blob/gdh1995-documentElement-offset/content_scripts/link_hints.coffee#L814)?

gdh1995:
No, they are not same.

If `document.documentElement.style.positon` is `"static"`, then:
* it's bounding rect takes its margin into consideration. For example, rect may be `[100,100,800,800]`
* while a `<div>` with `position: absolute` calcs its offset from the window's top-left corner, that is `[0, 0]`.

Only other `position`s except `"static"` can make `documentElement` the `offsetParent` of descendants, although `document.body` does not follow this rule, but usually keeps its offsetParent `null`.

xiaoronglv:
Hi this problem happens randomly in `Google Chrome is up to date
Version 61.0.3163.79 (Official Build) (64-bit)`.


image

xiaoronglv:
oh. this bug is relating to `zoom`

PR #2304, opened 2016-10-09
With grab-back-focus, main frame keeps focus.

### LINKED ISSUE #2303: Vimium loses the focus on load...
See [this page](http://stackoverflow.com/questions/2496710/writing-files-in-node-js).

On load, an `embed` is focused in a non-top frame.

To get back control (e.g. just to scroll) the user has to `Esc` then `gF`.

This happens even if *grab-back-focus* is enabled.

It's simple enough to extend *grab-back-focus* such that it also prevents `embed`s from stealing the focus.  But the user still ends up in the non-top frame (so, `gF` would still be required).

smblott-github:
If *grab-back-focus* is enabled, then also ensure that the focus stays in the main frame.  That is, if a non-top frame gets the focus, then ask the main frame to (re-)focus itself.

Fixes #2303.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -65,6 +65,11 @@ class GrabBackFocus extends Mode
   grabBackFocus: (element) ->
     return @continueBubbling unless DomUtils.isFocusable element
     element.blur()
+    # Do not let non-top frames acquire the focus (see #2303).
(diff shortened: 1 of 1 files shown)


smblott-github:
Binning this.  It causes bad effects when the page creates a new frame dynamically and that frame acquires the focus.

PR #2306, opened 2016-10-09
Add "mapkey" command for key mappings.

### LINKED ISSUE #2305: Change Visual maping
Its an issue because, Brazilian keyboards have the __ç__ and this single key, changes the way your brain works lol, making using the movement from __hjkl__ keys really annoying to use, I have faith i'm not the only Brazilian who changes the movement biding to __jklç__, so I'd really love to be able to change this :D

smblott-github:
*Edit: The name "translate", here, has been replaced with "mapkey".*

Under *Custom key mappings* (on the options page), this implements:

    translate x y

Whenever the users types `x` in normal mode or in visual mode, the `x` is replaced by `y`.

For example:

    translate ç l

now, `ç` maps to `scrollRight` (which apparently would be helpful on Brazilian keyboards).

Issues:

- Is there an established vim name for this (to use instead of `translate`)?

- Do we want yet another hack like this?  This would be documented only on the wiki.

- With `translate <c-c> <c-[>` (and with `isEscape()` extended to use the translation), we'd get the `exitMode` command almost for free (#2253).

- Instead of adding a new directive called `translate`, we could alternatively overload the existing `map` directive.  Since these are single-key mappings, there's no ambiguity.  (Although, I guess there's a risk that some user has junk in their key mappings, and they would be taken somewhat by surprise).

Inspired by isssue posted by @vhoyer (#2305).

Fixes #2305.

**Edit...**

With 9c1012ad3a731b015b8a70b58828fbcd0acb7db0, `Escape` also uses `translate`, so you can say things like:

    translate <c-c> <c-[>

Also, to disable `<c-[>`:

    translate <c-[> <anyOldNonesense>

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -9,7 +9,7 @@ Commands =
     @clearKeyMappingsAndSetDefaults()
     @parseCustomKeyMappings customKeyMappings
     @generateKeyStateMapping()
-    chrome.storage.local.set keyTranslationRegistry: @keyTranslationRegistry
+    chrome.storage.local.set mapKeyRegistry: @mapKeyRegistry
 
   availableCommands: {}
   keyToCommandRegistry: {}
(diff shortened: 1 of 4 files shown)


lydell:
Small typo: The commit message says `map ç l` instead of `translate ç l`.

danijar:
Thanks!

calebplum:
How can this be used to pass the `Escape` key through to a web page?

sumanth-dnbk:
> How can this be used to pass the `Escape` key through to a web page?

I mapped Escape to Ctrl + ; using the below custom key mapping.
**mapKey <c-;> <c-[>**

PR #2308, opened 2016-10-10
Don't set the update_url by default.

achew22:
Setting the update_url by default can lead to accidentally updating to
the version you host on github. Setting this to a safe default of "" and
leaving a comment explaining why.

### DIFF
--- a/Cakefile
+++ b/Cakefile
@@ -86,7 +86,11 @@ task "package-custom-crx", "build .crx file", ->
   # ugly hack to modify our manifest file on-the-fly
   origManifestText = fs.readFileSync "manifest.json"
   manifest = JSON.parse origManifestText
-  manifest.update_url = "http://philc.github.com/vimium/updates.xml"
+  # Update manifest fields that you would like to override here.  If
+  # distributing your CRX outside the Chrome webstore in a fork, please follow
+  # the instructions available at
+  # https://developer.chrome.com/extensions/autoupdate.
+  # manifest.update_url = "http://philc.github.com/vimium/updates.xml"
   fs.writeFileSync "manifest.json", JSON.stringify manifest
 
   pem = process.env.VIMIUM_CRX_PEM ? "vimium.pem"


achew22:
@philc Do you have any thoughts on this change?

acornejo:
+1 to get this one in.

smblott-github:
But if you're maintaining a fork, then surely you just change the update URL to whatever you want.  That's the nature of forks.

achew22:
That's totally true, but it if this is merged it is one less place that you need to remember. I'm doing this because the security auditing team asked me to remove it and I would like to keep the code as in sync as possible to ease contributing back.

smblott-github:
It's fairly safe to change `manifest.json` and `Cakefile` in your fork, they don't change all that often.  Perhaps you could:

1. Remove the update URL in your fork.  That will almost never conflict with upstream changes.
2. Add a script to your fork which fails if the update URL is defined in `manifest.json`.  That will never conflict with upstream changes.
3. Add your script to the `Cakefile` build process; if the script fails, then the build fails.  That will almost never conflict with upstream changes.

We can't just remove the update URL here.  We need it.  If we removed it from `manifest.json`, then we'd just have to add it somewhere else (perhaps in a `sed` script), and your problem would still remain.

achew22:
I think I must be missing something. @smblott-github, can you help me understand. When I go to the url in question "http://philc.github.com/vimium/updates.xml" it gives a 404. Further, the app is distributed through the Chrome Web Store which means that the value is ignored. To quote [the chrome extension documentation](https://developer.chrome.com/extensions/autoupdate) at length:

> We want extensions and apps to be autoupdated for some of the same reasons as Google Chrome itself: to incorporate bug and security fixes, add new features or performance enhancements, and improve user interfaces.
>
> If you publish using the Chrome Developer Dashboard, you can ignore this page. You can use the dashboard to release updated versions to users, as well as to the Chrome Web Store.

Later in the page:

> If you're hosting your own extension or app, you need to add the "update_url" field to your manifest.json file, like this:

It seems to me that this url is only used in plugins that are not available via the Chrome Web Store. Can you help me understand where my assumptions about the world do not meet reality?

achew22:
@smblott-github, friendly ping. I'm interested in understanding what I'm missing. Thanks so much

smblott-github:
Sounds like you're right, @achew22.  Looks like this is just a legacy of some earlier distribution scheme.

Issue #1986, opened 2016-02-11
Feature Request: Leader key support


danielyli:
For those of us whose muscle memories are accustomed to the leader key (e.g., the comma), supporting a leader key would be a very welcome feature.

smblott-github:
Thanks, @danielyli.  My sense would be that `<leader>` support would be overkill for Vimium.  We have a far fewer commands, and we don't support plugins.  And users can always re-map their keys to achieve whatever mapping best meets their needs.

danielyli:
Alright, that's unfortunate, though very understandable.  Thanks.

Aaditay:
Can we please rethink on this ? Because vimium now have so much options to customize.

james-gour:
Seriously, this should be re-thought @smblott-github.

The value proposition of this feature has nothing to do with the limited number of available commands in vimium.

It has _everything_ to do with consistency of developer workflow across platforms. The leader key is a core feature in vim. Vimium is inspired by vim. Every tool that emulates vim should offer the leader key so that users can simply use all of their bindings that they know and love :).

The mental overhead of having to learn yet another set up arbitrary keymaps, instead of just using my our beloved set from vim makes me hesitant to pick up vimium.

PR #2311, opened 2016-10-15
New command option to enter normal mode

smblott-github:
Here's the problem...

Many sites define their own keyboard shortcuts, for example Google Play Music defines `gh` for "go home".

On such sites, it is normal to set up pass keys for `g` and `h`, but then Vimium bindings which begin with `g` are inaccessible.  And this can be frustrating.

Here, we add a new command `enterNormalMode` which installs a new normal-mode instance (without any pass keys).  This executes a single normal-mode command (or `count` commands) and then exits.

Examples:

```
map \ enterNormalMode
map | enterNormalMode count=999999
```

Assuming `g` and `o` are pass keys:
- `gh` or `o` - use the page's binding
- `\gg` - scroll to top
- `2\ggo` - scroll to the top and open the Vomnibar
- `\g<Escape>o` - open the Vomnibar
- `\<Escape>o` - use the page's bindings
- `\\\\\\<Escape>o` - use the page's bindings (new normal-mode instances
  displace previous ones, and `Escape` exits)

This required some changes to the scroller.  Previously, we only ever had one normal-mode instance, and could arrange statically that the scroller's key listeners were above normal-mode's key listeners in the handler stack.  Here, we fix this by adding and removing the scroller's listeners dynamically, so they're always at the top of the handler stack.

(This might be similar to what the poster of #1986 had in mind.)

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -166,6 +166,7 @@ Commands =
       "enterVisualMode",
       "enterVisualLineMode",
       "passNextKey",
+      "enterNormalMode",
       "focusInput",
       "LinkHints.activateMode",
       "LinkHints.activateModeToOpenInNewTab",
@@ -238,6 +239,7 @@ Commands =
     "closeOtherTabs",
     "enterVisualLineMode",
     "toggleViewSource",
+    "enterNormalMode",
     "passNextKey"]
 
 defaultKeyMappings =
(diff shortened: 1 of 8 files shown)


smblott-github:
I've used this a lot in the last week, so it's definitely useful - particularly on Google Inbox.

smblott-github:
I use this *a lot*, so it's a useful feature.

In 3019e38d6db4d56036eff8aaa2d2d37f85e4fb07 I tweaked it such that instead of saying...

    map \ enterNormalMode

we say...

    map \ passNextKey normal

In other words, *pass the next key to normal mode*.  This seems natural, and avoids the need for a new command.

PR #2312, opened 2016-10-16
Add sendMessage command to message external components (version two)

smblott-github:
This adds a new command `sendMessage` for implementing Vimium commands in external extensions of via TamperMonkey userscripts. 

See the explanation [here](https://github.com/smblott-github/vimium/blob/sendMessage-command-v2/examples/external-commands/README.md).

The most disruptive aspect of this PR is the plumbing through link-hints mode.  The `sendMessage` implementation itself is pretty much self contained.

Replaces #2109.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -210,7 +210,8 @@ Commands =
       "moveTabRight"]
     misc:
       ["showHelp",
-      "toggleViewSource"]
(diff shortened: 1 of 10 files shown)


smblott-github:
Added a command/options page to the demo extension:

snapshot

smblott-github:
Closing.  I think we should do *Option B* from #2319.

Issue #2229, opened 2016-08-22
Capital Letters Set in Hints causes Vimium to send keybindings


nabsiddiqui:
I seem to have noticed a bug where placing the capital letters in the set character hints option results in Vimium attempting to send the actual key bindings. This is solved by making the letter lower case again.

sco-tt:
I was able to reproduce this bug (set any number of hints on the options page to a capital letter, reload, press f, and then try to follow a link where the link hint starts with that letter).

Submitted a PR.

PR #2320, opened 2016-10-18
Force custom link hints to be lowercase

sco-tt:
Addresses #2229 

Test if link hints has been set to something other than the default; if that is the case, this forces them all to be lowercase.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -444,7 +444,17 @@ class LinkHintsMode
 # Use characters for hints, and do not filter links by their text.
 class AlphabetHints
   constructor: ->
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @sco-tt.  I think this is the right thing to do.

However:

- Why treat the default differently?  They're all lower case anyway.
- Why make characters lower case character by character? `str.toLowerCase()` just makes all characters in string lower case.

sco-tt:
Thanks for the feedback @smblott-github. The loop doesn't make sense at all. I put in PR for a cleaner/simpler implementation here: https://github.com/philc/vimium/pull/2327.

PR #2323, opened 2016-10-20
Fix event.key error

yermie:
In certain websites, the error `event.key.length` is undefined and throws an error. Example: https://www.dropbox.com/s/9bwdey1cfos9xfs/Screenshot%202016-10-20%2016.45.36.png?dl=0

Second screenshot: https://www.dropbox.com/s/9bwdey1cfos9xfs/Screenshot%202016-10-20%2016.45.36.png?dl=0

This prevents that error from occurring.

### DIFF
--- a/lib/keyboard_utils.coffee
+++ b/lib/keyboard_utils.coffee
@@ -41,6 +41,7 @@ KeyboardUtils =
       @getKeyCharUsingKey event
 
   getKeyCharUsingKey: (event) ->
+    return unless event.key?  # Avoid error if event.key is undefined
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @yermie.  I cannot reproduce this issue.  You say *on some web sites*.  Can you give an example?

It might be better to track down the source of the problem, if possible, rather than simply mask it.

smblott-github:
Do you have an example page, @yermie?

mrmr1993:
I think this can be closed; #2605 hopefully deals with the underlying issue.

PR #2327, opened 2016-10-24
Makes sure all custom link hint characters defined in settings are lowercase

sco-tt:
Cleaner/simpler implementation of https://github.com/philc/vimium/pull/2320

Addresses #2229

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -444,7 +444,8 @@ class LinkHintsMode
 # Use characters for hints, and do not filter links by their text.
 class AlphabetHints
   constructor: ->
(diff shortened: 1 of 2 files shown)


smblott-github:
Thanks, @sco-tt.  This is going to be a bit weird when there are capital letters showing on the options page.  Should we force them to lower case there too (either on `blur` or on save).

(Sorry for the slow response, I've been tied up with other matters.)

sco-tt:
I think this is a good idea. 

I added this change to the saveOptions method as it seemed a bit cleaner from a UX as well as code perspective, but I'm open to other ideas.

smblott-github:
Thanks, @sco-tt.

Issue #2287, opened 2016-10-01
Feature request: "clickable" comboboxes


tobimensch:
Here's an example, of what doesn't work.

http://www.anandtech.com/show/10697/the-lenovo-thinkpad-x1-yoga-review/2

There's a combobox that you can use to navigate between the pages in the article.

Hitting f you get a hint on the combobox, but pressing the key combination results
in nothing. What's expected here is that the combobox expands, so that the user
can make a selection.

gdh1995:
It's a Chrome's feature that disallow Vimium to activate a `<select>`.

This feature is imported since Chrome 53 (https://www.chromestatus.com/features/5718803933560832), and as far as I know, there's no solution to work it around until now.

Issue #2257, opened 2016-09-19
Vimium cannot "click" select elements


daemedeor:
Previously when I used vim on Chrome 51 or thereabouts the dropdown used to open when I hit f and then the associated tag that vimium tagged the field with. Now vimium still tags the field but it doesn't seem to open the select boxes. I tried it on several pages by adding my own select boxes but I'm not quite sure why its not working. Any workaround?

smblott-github:
@daemedeor... Could you provide an example page?

(This works on Facebook, for me.)

daemedeor:
Here: http://store.steampowered.com/agecheck/app/234140/?snr=1_239_4__122, if you have steam. also note: I'm on Chrome 53, Seirra, macOS. 

<img width="1285" alt="screen shot 2016-09-19 at 7 10 30 am" src="https://cloud.githubusercontent.com/assets/6037411/18631579/6a02df4c-7e38-11e6-8ed2-a0863333969d.png">

daemedeor:
Ah, the thing is. Even when I go to other pages (not just this one, this is just one of the examples) it seems to do it. Even when I just make my own select box. I'll test on facebook one minute.

daemedeor:
Hmmmm Wait... its not the drop down at the top of the page on the dropdown menu, I mean like where it says "Feburary" or the date picker.

romgrk:
Hello,
I have been seeing warnings pop up in the console (when clicking through vimium), saying that some non-standard behaviors would be removed from Chrome sometime during September, so I'm guessing that this is it. I'm using version 54 and those warnings have disappeared, along with the ability to click through vimium.

romgrk:
After a bit of investigation, it seems that it is indeed unstandard to be able to open a select element through javascript. Given that Chrome/ium is lately pushing to more standardization, I think it is unlikely they will go back to the old behavior. I would recommend at least triggering a 'focus' event on such elements, so the user can trigger it with space subsequently.

smblott-github:
@romgrk... If you're referring to the following console message regarding `keyIdentifier`, then we already have code in place to cover its removal:

![snapshot](https://cloud.githubusercontent.com/assets/2641335/18655817/d452fcfa-7ee3-11e6-9e76-f7f84d3ef2ef.png)

Regarding Chrome 54... I tried it, and didn't have any issues with link hints.  Please post a link and a description of the steps to reproduce.

romgrk:
Oh, well,
1. Find any select element
[...]

PR #2338, opened 2016-11-09
Partial fix for selecting SELECT elements

ramiroaraujo:
This issue has been reported and verified in #2257 and #2287, and was introduced with Chrome v53.
It's been noted in #2257 that calling `focus()` on the `select` element could do the trick, and indeed it does. It doesn't expand the `select` as it used to be, but at least it's focused.

The fix treats the `select` elements the same as `input` elements, and simply focus on them.
I've also updated the test to verify the focus on the element.

I've been using this fix for a few days now, since I couldn't go by without hitting selects any more. Hope it helps!

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -48,5 +48,6 @@ Contributors:
   Scott Pinkelman <scott@scottpinkelman.com> (github: sco-tt)
   Darryl Pogue <darryl@dpogue.ca> (github: dpogue)
   tobimensch
+  Ramiro Araujo <rama.araujo@gmail.com> (github: ramiroaraujo)
 
 Feel free to add real names in addition to GitHub usernames.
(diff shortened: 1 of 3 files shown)


gdh1995:
+1

Some ideas:
* I think we may also `focus` `<embed>`s and `<object>`s.
* It seems that `clickEl.type not in ["button", "submit"]` is not needed, because `else if DomUtils.isSelectable clickEl` in Line 390 has checked the same thing.

ramiroaraujo:
@gdh1995 I would prefer to keep this PR as simple as possible, "focusing" solely on the `<select>` issue

smblott-github:
That's really nice, @ramiroaraujo, thanks.

(And sorry for the delay.)

AndrewSouthpaw:
Any chance we could get a version bump on the Chrome extensions store, so we can all more easily benefit from this partial fix? :)

vmindru:
+1 need a version bump

PR #2369, opened 2016-12-11
getViewportTopLeft: support zoomed static documentElement

gdh1995:
This is a completion of #2301. If `document.documentElement` is zoomed, Vimium's hints `<div>` are also zoomed, and then `scrollY` may be not equal with viewport client rect's `top`. Therefore the zoom level must be taken into consideration.

Example:
* make the tab zoom level is 1, `documentElement`'s `zoom` style is 2
* open a page, scroll to the top left corner
* press `f`, and all things are right
* exit LinkHints mode, scroll down for 100px, and then press `f`
* this time all hints has moved down 200px from the correct place

Comment: if `documentElement.style.position` is not `"static"`, then the old code works well.

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -289,13 +289,14 @@ DomUtils =
     setTimeout((-> DomUtils.removeElement flashEl), 400)
 
   getViewportTopLeft: ->
-    style = getComputedStyle document.documentElement
+    box = document.documentElement
+    style = getComputedStyle box
     if style.position == "static" and not /content|paint|strict/.test(style.contain or "")
       zoom = +style.zoom || 1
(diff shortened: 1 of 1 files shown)


smblott-github:
Is there a page where I can test this, @gdh1995?

gdh1995:
https://jsfiddle.net/zh4foqp9/

image

gdh1995:
Oops, I find that a zoomed `body` or other element can make a mess of all computed bounding rects, and there seems no good method to detect them, since `getBoundingClientRect` does not compute the CSS property `zoom`.

But this fix to `documentElement` still works if without any other zoomed elements.

gdh1995:
I've updated the demo page: https://jsfiddle.net/zh4foqp9/1/

In face, all CSS properties which can affect `documentElement`'s box should be taken into consideration, and this PR just includes what I've found.

PR #2371, opened 2016-12-14
Add "swap" command option for marks.

smblott-github:
In a browser, I find global marks considerably more useful than local marks; for example, I use marks to quickly jump back to my email tab.

Unfortunately, inheriting vim's approach, global marks require capital letters, so the `<Shift>` key.

This PR implements:

    map m Marks.activateCreateMode swap
    map ' Marks.activateGotoMode swap

which inverts the `isGlobalMark()` test for the `<Shift>` key.

Edit: So, you can use lower-case letters for global marks, which is much more useful in practice. End edit.

Is the name `swap` the best one?  It could be `invert` instead, or perhaps there's something better.

### DIFF
--- a/content_scripts/marks.coffee
+++ b/content_scripts/marks.coffee
@@ -2,6 +2,7 @@
 Marks =
   previousPositionRegisters: [ "`", "'" ]
   localRegisters: {}
+  currentRegistryEntry: null
   mode: null
 
   exit: (continuation = null) ->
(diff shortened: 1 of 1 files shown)


gdh1995:
So a mark created by `Marks.activateCreateMode swap` only has a lower-case letter, and we can goto it only by `activateGotoMode swap`?

Do we need to ensure letters of global marks upper-case? (So that we may activate them whatever the `swap` option is)

smblott-github:
> So a mark created by Marks.activateCreateMode swap only has a lower-case letter, and we can goto it only by activateGotoMode swap?

Good observation.  To "fix" this, we would also toggle the case of the mark.  That works for letters, but not for other characters (e.g. `9` versus `(`).

My sense is to leave this as is, although that's not a high-confidence position.

Issue #2275, opened 2016-09-26
Feature request: Search box (optionally) at top.


tobimensch:
I'm used to search boxes being at the top right in chrome/chromium. When I type / it can happen that I think for a few seconds that it didn't work, because I expect the search input box in the upper right corner from years of getting used to that work flow.

It would be much appreciated if there was an option to set the location of the search box.

sergeyzozulya:
Best option would be to use nice vomnibox

smblott-github:
I'm not sure there's a strong enough to add an extra option for this.  Arguably, bottom-left would be the correct place (a la vim), but the status quo seems ok for most users.

sergeyzozulya:
Yes, right now it's similar to vim but looks a bit ugly comparing to vomnibox

smblott-github:
> bit ugly comparing to vomnibox

I agree.  I'd like to see it restyled to better match the look of the vomnibar.

philc:
Closing in favor of #3468, which more narrowly captures the latest proposal by @ghost and @smblott-github.

PR #2380, opened 2016-12-21
Re-style HUD to look similar to Vomnibar

sergeyzozulya:
According to issue #2275, I tried to change tune styling a bit

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -303,43 +303,60 @@ a:link.vimiumHelDialogLink, a:visited.vimiumHelDialogLink,
 
 div.vimiumHUD {
   display: block;
-  position: fixed;
-  bottom: 0px;
-  /* Keep this far enough to the right so that it doesn't collide with the "popups blocked" chrome HUD. */
-  right: 0px;
-  color: black;
-  height: auto;
-  min-height: 13px;
-  width: auto;
-  max-width: 400px;
-  min-width: 150px;
-  text-align: left;
-  background-color: #ebebeb;
-  padding: 3px 3px 2px 3px;
-  margin: 0;
-  border: 1px solid #b3b3b3;
-  border-radius: 4px 4px 0 0;
-  font-family: "Lucida Grande", "Arial", "Sans";
-  font-size: 12px;
-  text-shadow: 0px 1px 2px #FFF;
-  line-height: 1.0;
+	position: fixed;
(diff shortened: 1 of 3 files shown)


smblott-github:
Well, that's interesting, @sergeyzozulya.

However, it's probably too much of a divergence from the status quo.

Some requirements would be:

- familiar to use for existing users; that probably means it needs to be in the same place
- unlikely to obscure relevant content; that probably means it stays small (although possibly a bit bigger than previously.

Remember, the HUD is also used for informational messages (like "Visual mode", "Yanked SOMETHING").  So it needs to be unobtrusive too.

I'd suggest...

- move it back to roughly where it was
- it can be a bit bigger than previously
- it should be attached to the edge of the viewport
- and it should be styled more like the Vomnibar (or possibly the Chrome Find box).

This will take some experimentation to get right.

sergeyzozulya:
Thanks for comments, I do agree. Now it's smaller and on bottom right. Also, when it shows just info - it's grey, when it's in edit mode - it's white to match Chrome Find.

smblott-github:
Screenshots, for the curious...

Find mode:

snapshot

Any other HUD mode:

snapshot

These are both tight up against the bottom right of the viewport.

@sergeyzozulya... Looks nice.  Perhaps a touch too wide, but I can't decide whether it looks better tucked against the left of the page or floating slightly further left (the status quo).

@philc... What do you think?

sergeyzozulya:
That was the idea, something like it pop up from the bottom edge. I put it to the very right side to match Chrome Find. Width: it's 20% right now and 400px minimum.

rsperezn:
Hi there, any updates on this pr? Was also thinking of submitting something to make the search box more visible

philc:
Sorry for coming to this late. The screenshot @smblott-github posted looks good, although I would like to weak the styles further to make it less wide. Also, the diff is using tabs, not spaces.

However, I tried this patch recently and the find box isn't showing up with this patch applied. I didn't debug it further. If anyone wants to tune this to get it working again, I'd like to preview it.

gdh1995:
@philc Sorry, the HUD is disappearing because of one of my typos, and you may merge #3467 then.

philc:
Thanks @gdh1995. I tried this out and it looks like an improvement to me, so I've merged it in 74710b743119efaf525a63fee3ad5d5e8bb98bb1.

PR #2382, opened 2016-12-22
make getSelectionFocusElement always return an element or null

gdh1995:
The old function may return `undefined`, because when current selection is just a caret and inside a plain text node, it will return something like `new Text().childNodes[focusOffset]`, which will always be `undefined` since a `Text` can not own children.

This PR returns `Text::parentElement`, so that it's more likely to succeed in focusing a current selection.

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -328,15 +328,13 @@ DomUtils =
 
   getSelectionFocusElement: ->
     sel = window.getSelection()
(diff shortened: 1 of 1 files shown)


philc:
@gdh1995 is this still relevant, and why is it important?

gdh1995:
@philc I've updated the explanation in the 1st floor (the old message seemed total irrelevant to the commit).

Issue #2378, opened 2016-12-20
Link Hints "click" doesn't focus button/role="button"


nicholasguyett:
When using link hints (`f`) to "click" a `button` or `[role="button"]` the element does not have focus afterwards.  When using the mouse to directly click, the mouse does have focus.

Is there a particular reason why vimium's link hints deviate from natural clicks in this instance or could this be corrected?

smblott-github:
> Is there a particular reason why vimium's link hints deviate from natural clicks in this instance or could this be corrected?

Not sure, @nicholasguyett. Vimium does not actively blur the element, the element just doesn't get the focus when we simulate clicking it.

Do you have a positive reason for why that would be the *right thing to do*?

nicholasguyett:
@smblott-github I think consistency with a real click is a positive reason for giving focus to the clicked element.  It also helps tab navigation as the focus means that you'll be able to tab from that button after clicking on it.

smblott-github:
Somebody (not me) made a decision in the past that a button should not receive the focus: [here](https://github.com/philc/vimium/blob/master/content_scripts/link_hints.coffee#L396).  I wonder why.

gdh1995:
@smblott-github The first person who decided not to focus an `input[type=button]` is @philc in [the initial commit of Vimium](https://github.com/philc/vimium/blob/f33de347f0de1448446eac462892addbe7a3c461/content_scripts/link_hints.coffee#L193), so maybe it's just a typo?

I think it's quite reasonable to actively focus any visible `input` element, but we can not know if an `[role=button]` needs focus. Therefore, I suggest focusing those with `[tabindex] >=0` .

smblott-github:
That sounds reasonable, @gdh1995.  Do you fancy putting in a PR?

philc:
Thanks for the sleuthing guys. I think the element should get focus for the
reasons mentioned.

gdh1995:
OKay I'll make two PRs about this issue and the 351489c7c5e093531aa8cd5870a3f78c99fb2923 some hours later.

nicholasguyett:
Resolved by #2383

PR #2383, opened 2016-12-22
LinkHints: focus select elements, and elements with nonnegative tabIndex

gdh1995:
Selectable (and also editable) `<input>`s have been handled on L390,
so here we just need to focus `<input>`s of other types
    and those whose tabIndex is nonnegative.

This PR is a complement of #2338 , and changes have been beyond what #2378 needs.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -394,7 +394,7 @@ class LinkHintsMode
             clickActivator = (modifiers) -> (link) -> DomUtils.simulateClick link, modifiers
             linkActivator = @mode.linkActivator ? clickActivator @mode.clickModifiers
             # TODO: Are there any other input elements which should not receive focus?
-            if clickEl.nodeName.toLowerCase() in ["input", "select"] and clickEl.type not in ["button", "submit"]
+            if clickEl.nodeName.toLowerCase() in ["input", "select"] or clickEl.tabIndex >= 0
               clickEl.focus()
             linkActivator clickEl
 
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -398,7 +398,7 @@ class LinkHintsMode
             #     focus it to let user press space to activate the popup menu
             # <object> & <embed>: for Flash games which have their own key event handlers
             #     since we have been able to blur them by pressing `Escape` 
-            if clickEl.nodeName.toLowerCase() in ["input", "select", "object", "embed"] or clickEl.tabIndex >= 0
+            if clickEl.nodeName.toLowerCase() in ["input", "select", "object", "embed"]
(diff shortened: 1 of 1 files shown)


content_scripts/link_hints.coffee
```diff
@@ -394,7 +394,7 @@ class LinkHintsMode
             clickActivator = (modifiers) -> (link) -> DomUtils.simulateClick link, modifiers
             linkActivator = @mode.linkActivator ? clickActivator @mode.clickModifiers
             # TODO: Are there any other input elements which should not receive focus?
-            if clickEl.nodeName.toLowerCase() in ["input", "select"] and clickEl.type not in ["button", "submit"]
+            if clickEl.nodeName.toLowerCase() in ["input", "select"] or clickEl.tabIndex >= 0
```
smblott-github: Thanks, @gdh1995.  Two things...

1. I think we should stick with `and` here (not `or`).  I'm concerned that we could be with the focus in an `embed` (or something) and unintentionally in insert mode.  Would that happen?
2. We can lose the comment on line 396, or perhaps replace it with a comment that better describes what is going on.

philc:
Looks like this does two things: allow `embed` and `select` elements to be focused, and anything with a non-negative tabIndex. For the latter, @gdh1995, what prompted this change? Do you recall the example page which had an element with a positive tab index that you wanted to click on which wasn't covered by our current link hint logic?

In any case, we're long past due on merging this or #2338, so I'll do that soon.

content_scripts/link_hints.coffee
philc: Bump @gdh1995 to resolve this conversation or respond to each point (in particular, the first comment).

gdh1995:
Well, this PR only comes from some of my occasional ideas, and now I've given the point of "focus those whose tabIndex >=0".

But as far as I've tested:
* `<embed>` may be used to show a PDF (just like an iframe) directly, so I still expect `<embed>` to be marked clickable
  * if Vimium focus it, then I can press PageDown/ArrowDown on it to scroll down PDF pages
* once `<object>` is often used to show Flash animations and games, but recently the case is very rare, so it may be removed from this PR.
* if a plain `<span>`'s `tabindex` attribute is an empty string, then its `tabIndex` property is `-1`.
  * the old logic is imported in commit 3a688f754ebd647ce56b33d18c5744759c5efe95 , which was only to fix those `[tabIndex=-1]`
  * but change in this PR may indeed make some elements with empty `tabindex` lose hints.

philc:
This part is the one I'm most curious about: `or clickEl.tabIndex >= 0`. Do you recommend merging with this in, or not?

The rest LGTM.

gdh1995:
Oh yes I should remove this part.

I've pushed a new commit to do so.

PR #2384, opened 2016-12-22
InsertMode should not handle key events if document.body is editable

gdh1995:
For example, the host JavaScript may create an "about:blank" iframe with a `content-editable` `body`, and then:
* the parent frame may handle `Escape` key events by itself
* but now, Vimium always grabs `Escape` events
  * and tries to exit `InsertMode`
  * although `document.body` is still current `activeElement` after `body.blur()`
* as a result, neither the parent can receive and handle wanted keyevents,
  nor Vimium will succeed in returing back to NormalMode

### DIFF
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -10,6 +10,7 @@ class InsertMode extends Mode
 
     handleKeyEvent = (event) =>
       return @continueBubbling unless @isActive event
+      return @passEventToPage if @insertModeLock is document.body
 
       # Check for a pass-next-key key.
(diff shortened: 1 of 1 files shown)


gdh1995:
In fact, I think Vimium should destroy itself whenever it finds `location.href == "about:blank" and document.body.isContentEditable`, and what troubles me very much is that:
* I'm developing another version of Vimium which use `"about:blank"` to support `FindMode`.
* so if I open both Vimium and my customized version, the "about:blank" will be hooked by Vimium,
* and my reworked `FindMode` can not get a correct `Escape` keydown event to exit itself.

And I've updated the commit and use `@passEventToPage` to pass those key events, so now it does work.

PR #2385, opened 2016-12-26
Guard against element.tagName not being a string.

### LINKED ISSUE #2035: It does not work fine on codeforces.com.

smblott-github:
Example page: http://codeforces.com/contest/752/problem/B; see #2035.

Here's what you get if you `console.log element.tagName`:

snapshot

There, `element.tagName` is an element with `name` `tagName` (not a string).  Here, we guard against that case.

Fixes #2035.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -626,7 +626,9 @@ LocalHints =
   # image), therefore we always return a array of element/rect pairs (which may also be a singleton or empty).
   #
   getVisibleClickable: (element) ->
-    tagName = element.tagName.toLowerCase()
+    # Get the tag name.  However, `element.tagName` can be an element (not a string, see #2305), so we guard
+    # against that.
+    tagName = element.tagName.toLowerCase?() ? ""
     isClickable = false
     onlyHasTabIndex = false
     possibleFalsePositive = false


philc:
Nice find and nice fix. How is that even possible?

content_scripts/link_hints.coffee
```diff
@@ -626,7 +626,9 @@ LocalHints =
   # image), therefore we always return a array of element/rect pairs (which may also be a singleton or empty).
   #
   getVisibleClickable: (element) ->
-    tagName = element.tagName.toLowerCase()
+    # Get the tag name.  However, `element.tagName` can be an element (not a string, see #2305), so we guard
+    # against that.
+    tagName = element.tagName.toLowerCase?() ? ""
```
philc: Minor: while yours is more succinct, I think this is more explicit:

```
# It's possible that element.tagName is an element, not a  string. See #2305.
tagName = if typeof(element.tagName) == "string" then element.tagName.toLowerCase() else ""
```

PR #2388, opened 2016-12-29
Key mappings for Shift + [non-printable char key]

alexei-zayakin:
Currently it is impossible to create a key mapping for combinations like `Shift + Left` or `Shift + F1`.

Proposed changes make it possible using `<s-left>` syntax.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -62,7 +62,7 @@ Commands =
   # We sort modifiers here to match the order used in keyboard_utils.coffee.
   # The return value is a sequence of keys: e.g. "<Space><c-A>b" -> ["<space>", "<c-A>", "b"].
   parseKeySequence: do ->
-    modifier = "(?:[acms]-)"                             # E.g. "a-", "c-", "m-".
+    modifier = "(?:[acms]-)"                            # E.g. "a-", "c-", "m-".
     namedKey = "(?:[a-z][a-z0-9]+)"                     # E.g. "left" or "f12" (always two characters or more).
     modifiedKey = "(?:#{modifier}+(?:.|#{namedKey}))"   # E.g. "c-*" or "c-left".
     specialKeyRegexp = new RegExp "^<(#{namedKey}|#{modifiedKey})>(.*)", "i"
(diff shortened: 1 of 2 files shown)


gdh1995:
I agree that `<Shift+Fn>` should be supported, but I think `<f1> / <F1>` and `<left> / <LEFT>` are better than a new modifier `s`.

alexei-zayakin:
Currently, `<left> / <f1>` are case insensitive, thus I thought a new modifier would be better than changing behavior for some existing combinations.

Anyways, I am completely okay with any implementation, just want to use that feature.

smblott-github:
@gdh1995... I prefer this to #2390.  Suggest we do this a close the other one.

gdh1995:
Well, this indeed seems a good solution.

My PR is from my project [Vimium++](https://github.com/gdh1995/vimium-plus), and it has supported keys like `<LEFT>` for years. @smblott-github, since users of Vimium are so much more than those of mine extension, what Vimium chooses is up to you. I'll change Vimium++ to make it support `<s-*>` too.

sluongng:
could we pls have this?

gdh1995:
Now I've given my idea of "<LEFT>" and  turned to this PR of "<s-left>".

philc:
@gdh1995 @alexei-zayakin I'll merge this if it can be updated to merge cleanly (the code has changed since original authoring). Ideally we also have a test case for this in commands_test.coffee.

gdh1995:
Oh this is safe to merge.

The conflicting line is `keyChar = keyChar.toUpperCase() if event.shiftKey and keyChar.length == 1`, which "has been added" into the master branch.

philc:
Merged in ebc57ba7b8c45a0ccfb6710a6989cce73ac19f11. Nice. Thanks!

politicus:
What a great news! More mappings available!

Can we get a list of the keys that now can be mapped the following way <s-NameOfTheKey>?

I tried and successfully mapped the four arrow keys: 
map <s-left> LinkHints.activateMode action=copy-text
map <s-down> LinkHints.activateMode action=hover
map <s-up> LinkHints.activateMode action=focus
map <s-right> LinkHints.activateMode

PR #2390, opened 2017-01-03
use upper-case key names to represent shiftKey in <([acm]-)?.*> key mappings

gdh1995:
This is another solution to #2388, and allows users to make such key mappings:
```
# press "Space"
map <space> scrollDown
# press "Shift+Space"
map <SPACE> scrollUp

# just treated as "<space>"
map <Space> scrollDown
```

The added logic is that if a `keyChar` (background_scripts/commands.coffee#L75) has only upper-case chars, then it means `Shift` should be pressed.

This does not influence most old mappings like `<Left>`, but will treat `<f1>` and `<F1>` differently.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -66,13 +66,14 @@ Commands =
     namedKey = "(?:[a-z][a-z0-9]+)"                     # E.g. "left" or "f12" (always two characters or more).
     modifiedKey = "(?:#{modifier}+(?:.|#{namedKey}))"   # E.g. "c-*" or "c-left".
     specialKeyRegexp = new RegExp "^<(#{namedKey}|#{modifiedKey})>(.*)", "i"
+    lowerCaseRegexp = /[a-z]/
     (key) ->
       if key.length == 0
         []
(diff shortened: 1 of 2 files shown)


smblott-github:
Treating `<f1>` and `<F1>` differently would be problematic, would it not, since a fair few users will already have mappings for `<F1>`, which do not require the `Shift` key?

Is #2388 not more vim like too?

gdh1995:
But the difference between `<a>` and `<A>` is just the `shiftKey`, and Vimium does not accept `<s-a>`.

I have no good idea about `<F1>`. Maybe we can add a function in `runtime.onInstalled` to detect old mappings and convert them?

lydell:
Shameless plug: You could also use https://github.com/lydell/vim-like-key-notation for inspiration (which is used by [VimFx](https://github.com/akhodakivskiy/VimFx/)).

gdh1995:
I'm giving up this idea, because now I find that, usually a key like <kbd>Alt+Shift+C</kbd> should not be influenced by the Caps Lock key...

Issue #2170, opened 2016-06-19
Add a timer between multiple-character actions, like "]]" "[[" "yy"


ccaapton:
In the current implementation, two consecutive ']' clicks will always trigger "next page" action, no matter how long in between these two clicks. Even there is a one-minute gap, it will always redirect to the next page. However most of time when people use these multi-character actions, they will click twice within a second, and they don't want surprises, like click "y" twice change the clipboard without any visual clue.  

It will be nice to add a timeout action to the state-machine, so these accidents will not happen. Thanks!

smblott-github:
Agree, we should fix that.  PR welcome.  Fix would go somewhere [here](https://github.com/philc/vimium/blob/master/content_scripts/mode_key_handler.coffee).

mrshu:
Since #2393 has been merged, can this issue be closed?

ajwadjaved:
Hi, is this issue still open. Would love to take it up. I see that the path that Stephen mentioned isn't valid anymore @smblott-github

philc:
While it seems like a reasonable proposal, it also seems like a rare sharp edge. Is this really worth fixing?

Also, is there a precedent of a timeout in Vim?

gdh1995:
I like this idea, and I suggest a timeout of 30 seconds.

PR #2393, opened 2017-01-06
Add timer between multiple character actions

andrewseaman35:
This adds a one second timeout for multiple character actions.

Addresses #2170


smblott-github:
@andrewseaman35.  Thanks for your input.

I'm not sure whether we should fix #2170 or not.  I don't actually every remember encountering this as a real problem in the wild myself.

(Separately, one second is certainly two short a timeout.)

(Also, I know that my view above is the opposite of what I said some time ago in #2170.)

andrewseaman35:
@smblott-github Thanks for the comments. #2170 isn't really an issue that I've encountered either, but I could see it causing confusion for newer Vim users. 

Would you like me to increase the timeout duration? or should we close this one up?

smblott-github:
@andrewseaman35....  I'd leave this sitting for a bit, and then close it if nobody else chips in.

PR #2406, opened 2017-01-23
Fix visual line mode line selection behavior

pjhades:
The current line selection behavior in the visual line mode is not
the same as in vim, where we cannot select lines upwards.

Steps to reproduce:

- Press `v` and `c` to enter the caret mode.
- Press `k`.

This is because when only the base line (the line selected when we
enter the visual line mode) is selected, and its direction is different
from the direction of the command, that command will actually unselect
the base line.

Suppose we have the following page, and we enter the visual line mode from Line 3:

```
aaaaa
bbbbb
ccccc  <-- selected
ddddd
```

The current selection has `forward` direction. Now if we press `k`, the `commandHandler()` of the visual mode will be called, which executes a `backward line` command, unselecting the line. After that `extendSelection()` is called, Line 3 is selected again, but in fact the `k` command should reverse the direction of Line 3 and continue to select Line 2.

To fix this, when executing line selection commands, we check if the
base line is unselected. If so, we restore it and reverse its direction.

Test cases are added.

### DIFF
--- a/content_scripts/mode_visual.coffee
+++ b/content_scripts/mode_visual.coffee
@@ -326,8 +326,25 @@ class VisualLineMode extends VisualMode
     super extend options, name: "visual/line", indicator: "Visual mode (line)"
     @extendSelection()
 
-  commandHandler: (args...) ->
-    super args...
+  commandHandler: ({command: {command}, count}) ->
+    switch typeof command
+      when "string"
+        for i in [0...count] by 1
+          @movement.runMovement command
+          # If the current selection
+          #  * has only 1 line (the line that is selected when we # enter the visual line mode), and
+          #  * its direction is different from the command,
+          # then the command will in effect unselect that line. In this case, we restore that line and
+          # reverse its direction, keeping that line selected.
+          if @selection.isCollapsed
+            @extendSelection()
+            [direction, granularity] = command.split(' ')
+            if @movement.getDirection() != direction and granularity == "line"
+              @movement.reverseSelection()
+            @movement.runMovement command
+      when "function"
+        command count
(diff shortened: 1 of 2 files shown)


smblott-github:
@pjhades.  Thanks for this.  I can't seem to reproduce the problem...

- Single-click on `cccc` above (this causes no visible selection, the selection is collapsed).
- `V`
- `k` ... and I get two lines selected.

I think that's your intended behaviour, is it not?

pjhades:
Oops...I'm confused >_<

I just tested again with your steps, first click on the `cccc` line with my mouse, then press `V`, finally `k`. On my browser I cannot select lines upwards, the selected line is still the `cccc` line.

I tested with both the `v1.57` tag and `master`. My Chrome version is `56.0.2924.87 (64-bit)`.

philc:
The bug @pjhades reports repros for me, and it's annoying. Not sure why @smblott-github couldn't reproduce it.

I'd like to merge this because it's a nice fix, but the logic is a little complicated. @pjhades or @gdh1995 can you think of a way to simplify it further? Otherwise I'll merge as-is.

philc:
I committed this in 13f3b82dc117a8ead5c37b25f7128a2f2a25d7a9 -- I took the logic, but not the tests, since the visual mode tests are currently broken. Thanks for the fix!

gdh1995:
Um, I could not reproduce it on a fresh Chrome 80 or 56 (Win 10, x64). Maybe I was still misunderstanding some steps?

gdh1995:
Well, I reproduced it on macOS only - neither Windows 10 nor Ubuntu/X11 . This PR does work.

Issue #2141, opened 2016-05-26
The first time the help dialog is opened on a page, it flickers


philc:
If you open it again before refreshing the page, it displays smoothly.

innaterebel:
The favicon of the active tab inhabits the same phenomenon when link hint is activated on the page the first time.

smblott-github:
We could fix this by pre-loading the help dialog.  That's what we do with the Vomnibar (to avoid a similar flicker).  However, once a user is familiar with Vimium, they rarely use the help dialog, so we would be doing a lot of work (and using a lot of memory) in every frame for no reason.

A better approach would be to make the help dialog visible only after it is fully rendered.  I'm adding a "Help wanted" label for now.

philc:
I believe this is no longer an issue. The help dialog looks snappy to me.

PR #2417, opened 2017-02-05
Show help dialog only after it has rendered

andrewseaman35:
This change makes sure the help dialog is only shown after it has rendered. It gets rid of the flicker that is seen during the first load of the help dialog.

Addresses #2141


andrewseaman35:
closing to fix

PR #2428, opened 2017-02-11
Use event argument instead of global event object

mrmr1993:
Cherry-picked as suggested by @gdh1995.

This should have no effect in Chrome/Edge, but fixes errors for undefined `window.event` in Firefox.


smblott-github:
Thanks, @mrmr1993.

PR #2448, opened 2017-03-10
Minor fix: Use the appropriate handler for hashChange

mrmr1993:
The `hashchange` event should recheck whether the URL is enabled with `checkEnabledAfterURLChange`, not `onFocus`, otherwise we falsely notify the background page/UI components of a change of frame.


smblott-github:
That sounds correct.  Thanks!

PR #2456, opened 2017-03-22
Always use keydown and event.key (partial PDF-tab fix)

### LINKED ISSUE #1243: vimium not working for pdf files
I cannot use j/k bindings for scrolling through pdfs, when open by the native pdf reader in chrome on ubuntu. The vimium icon shows blue (loaded).
Please help.
Vimium version : 1.45
Chrome Version : 37.0.2062.94

smblott-github:
This moves all (almost) key handling to `keydown`, and uses `event.key`.

This fixes a number of inconsistencies and makes Vimium's background-page and Vomnibar commands work on PDF tabs.

There is a possibility that there could be unintended consequences for some non-Latin, non-QUERTY keyboards.

Fixes #1243 (partly).

I will:

- Try this myself for a bit.
- If that seems ok, then I'll merge this and we'll see if anybody else identifies issued.

It would be great to get the PDF-tab issue fixed.

### DIFF
--- a/content_scripts/mode_key_handler.coffee
+++ b/content_scripts/mode_key_handler.coffee
@@ -33,9 +33,6 @@ class KeyHandlerMode extends Mode
       # We cannot track keyup events if we lose the focus.
       blur: (event) => @alwaysContinueBubbling => @keydownEvents = {} if event.target == window
 
-    @mapKeyRegistry = {}
-    Utils.monitorChromeStorage "mapKeyRegistry", (value) => @mapKeyRegistry = value
-
     if options.exitOnEscape
       # If we're part way through a command's key sequence, then a first Escape should reset the key state,
       # and only a second Escape should actually exit this mode.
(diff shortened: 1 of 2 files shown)


smblott-github:
This is working well for me.  I'll merge it and see how we get on.

sco-tt:
@smblott-github very nice, thank you!

Xeite:
Does this patch correct that problem that could not use hjkl in PDF tab?

It does not work in my environment, so I wonder this patch is what I am looking for. 

I used
vimium version - 1.59
os - macOS sierra
chrome - Version 58.0.3029.110 (64-bit)

smblott-github:
> Does this patch correct the problem that I could not use hjkl in PDF tab for scrolling?

No.  There is no API to implement those operations on PDF documents.

This PR means that you can use `J`, `J`, `J` and so on, through PDF tabs, for example.  Previously, you always got stuck when you hit a PDF tab.

PR #2470, opened 2017-04-13
Rework key handling to keydown

smblott-github:
This is a follow up to #2456.  It moves (almost) all key handling to `keydown` instead of `keypress` (and loses almost 200 LoC).  `event.key` is used throughout.

There's more we can do.  For example, we could easily get rid of `KeyboardUtils.keyCodes` too.  I might add that.

This fixes one bug whereby marks do not work on PDF tabs (because previously marks were using `keydown`.

We lose a small number of tests because phantomjs does not support `event.key`.

Edit:

- ee3ad44ed78c5c5939a2e7bac20dde5e24be8684 now completely removes the use of `event.keyCode` (which is deprecated).


lib/dom_utils.coffee
```diff
@@ -315,6 +315,25 @@ DomUtils =
...
+        return handlerStack.continueBubbling unless keyChar == KeyboardUtils.getKeyCharString event
+        @remove()
+        handlerStack.suppressEvent
+      # We cannot track keyup events if we lose the focus.
+      blur: (event) ->
+        @remove() if event.target == window
+        handlerStack.continueBubbling
+    callback?()
+    handlerStack.suppressEvent
+
```
mrmr1993: Will this not leak `keyup`s when multiple keys end up being held at the same time?

For example:
- typing `F`,
  * `[keydown shift][keydown f][keyup shift][keyup f]`
  * the `keyup` for `f`leaks
- typing `zH`,
  * `[keydown z][keydown shift][keyup z][keydown h][keyup shift][keyup h]`
  * slightly exaggerated example, but will leak the `keyup`s from `z` and `h` will both leak

lib/dom_utils.coffee
```diff
@@ -318,19 +318,20 @@ DomUtils =
...
-        @remove()
-        handlerStack.suppressEvent
-      # We cannot track keyup events if we lose the focus.
-      blur: (event) ->
-        @remove() if event.target == window
-        handlerStack.continueBubbling
+    unless event.repeat
+      handlerStack.push
+        _name: "dom_utils/consumeKeyup"
+        keydown: (event) ->
```
mrmr1993: Shouldn't this handler also reject repeats?

mrmr1993:
Been using this (on Firefox, admittedly) and it's working really nicely.

Might be worth re-introducing the old `keyup` suppression, but adopting it into dom_utils in place of `consumeKeyup`.

lib/dom_utils.coffee
mrmr1993: On further reflection, this code (or any alternative) should use `event.code`, not `event.key`. Otherwise, the `keyup` may have a different `key` (eg. `keydownEvent.key == "F"`, `keyupEvent.key == "f"`) if a modifier is released before the key itself, whereas we only care about the physical key, as described by `code`.

smblott-github:
Thanks, @mrmr1993.  I changed it to use `event.code`.  That seems to be the right thing to do.  See 7a50f4f7e2a03206b549d49e251751ac00758c27.

These two sequences still leak `keyup` events (for different reasons):

- `/`, `Escape`
- `i`

They both leak events in master/1.59 too.

(If you're testing, then [this page](http://smblott.org/keyboard.html) can be helpful.)

smblott-github:
An alternative version is:

- https://github.com/smblott-github/vimium/blob/rework-key-handling-to-keydown-v2/content_scripts/mode.coffee#L211
- 21afa69df1301a9b5fb3ad4f710035cae86c2314

lib/dom_utils.coffee
```diff
@@ -305,15 +305,26 @@ DomUtils =
...
-        return true unless KeyboardUtils.isEscape event
-        @remove()
-        false
-    handlerStack.suppressEvent
+  consumeKeyup: do ->
+    handlerId = null
+
+    (event, callback = null) ->
+      unless event.repeat
+        handlerStack.remove handlerId ? "not-an-id"
```
mrmr1993: Why not initialize `handlerId` directly to `"not-an-id"` above?

mrmr1993:
Changes LGTM.

philc:
Nice work getting this in! Any idea if this will impact non Latin keyboards?

smblott-github:
> Any idea if this will impact non Latin keyboards?

I think not, but am not 100% certain.

For special characters, we've actually been using `event.key` for quite some time (because `event.keyIdentifier` as removed about six months).  So that should be no change.

For printable characters, we were using `String.fromCharCode event.charCode`.  As far as I can tell, that should be the same as `event.key`.  And `event.charCode` is deprecated (although no date has been set for its removal).

Issue #2425, opened 2017-02-09
Support for Firefox


mrmr1993:
Firefox now supports WebExtensions. It would be nice to tweak Vimium to work on Firefox.

This is related to #2076.

So far, I've developed some basic support in my fork:
* branch [no-depreciated-apis](https://github.com/mrmr1993/vimium/tree/no-depreciated-apis) changes code using depreciated APIs (which FF will not implement) to use their replacements.
* branch [firefox-compat](https://github.com/mrmr1993/vimium/tree/firefox-compat) makes some changes to support Firefox, where the current code does not run correctly.
* branch [firefox-51-compat](https://github.com/mrmr1993/vimium/tree/firefox-51-compat) works around some unimplemented features in FF51, the current release. They are fixed in FF52.

You can use branch [firefox-merger](https://github.com/mrmr1993/vimium/tree/firefox-merger) to test, which is simply all of these merged into master.

smblott-github:
Thanks, @mrmr1993. Reviewing the final branch mentioned above, it looks like the required changes aren't really that disruptive.  How complete is this?  (I don't have the setup to test this.)  I know @philc is interested in adding Firefox support.

mrmr1993:
I probably still need to look at non-normal mode key handling. Other than that, it seems complete, but Firefox isn't.

Some example problem Firefox bugs:
* disabling a element while it is focused doesn't blur it, and results in (nearly) all key events being suppressed.
  - in practice, this means that we get stuck on the save button in the options page. A `.blur()` before disabling should fix that, but we can still get stuck on other pages.
  - general escape strategy is `[tab][esc]`, but still bad UX.
* no support for the clipboard API, which stops several commands working
* all simulated mouse clicks with modifiers (think opening tabs in foreground/background) do not trigger the default option
  - combined with the clipboard issues, this means that it is difficult to open a link in a new window. (Suggested nasty workaround is `fytJH`)
* `keypress` *always* fires, so we can't distinguish printing and non-printing keys.

philc:
This is really cool; great start @mrmr1993. I would love to get a viable Firefox version out onto the store since the current crop of Vim-style extensions are going to stop working once the classic extension APIs are fully deprecated. Hopefully Firefox support won't incur a large QA/maintenance burden.

I haven't stressed tested it with days of usage, but I will try to. However, there were a couple of things I noticed right away:

[...]

PR #2476, opened 2017-04-16
Improve firefox support

mrmr1993:
This merges all of the code mentioned in [my comment](https://github.com/philc/vimium/issues/2425#issuecomment-294355905) on #2425.

As listed there:
* link hints for opening links anywhere *except* the current tab do not work
* clipboard operations do not work.

Also, there are still issues with focus in the find mode HUD.
(Branch [hud-focus](https://github.com/mrmr1993/vimium/tree/hud-focus) seems to improve the situation, by using separate persistent elements for normal HUD and the HUD find input. Not sure if this is a direction we want to take.)


content_scripts/mode_find.coffee
```diff
@@ -178,7 +178,10 @@ class FindMode extends Mode
       # ignore the selectionchange event generated by find()
       document.removeEventListener("selectionchange", @restoreDefaultSelectionHighlight, true)
 
-    result = window.find(query, options.caseSensitive, options.backwards, true, false, true, false)
+    try
+      result = window.find(query, options.caseSensitive, options.backwards, true, false, true, false)
+    catch # Failed searches throw on Firefox.
+    options.postFindFocus?.focus()
```
smblott-github: Is the indentation correct here?  Looks like `XXX.focus()` should happen only if the exception is raised.

pages/content_script_loader.coffee
```diff
@@ -0,0 +1,28 @@
+injectContentScripts = ->
```
smblott-github: Is this file actually being used?

content_scripts/mode_find.coffee
mrmr1993: Yes. `window.find` in Firefox focuses the window that we call it on, so this gives us an opportunity to refocus the HUD window.

Should I add a newline between the `try ... catch` and this line, to clearly separate them?

pages/content_script_loader.coffee
mrmr1993: No, sorry. The revert commit reintroduced it and I didn't notice. I'll remove it

mrmr1993:
@smblott-github I've pushed some commits to address your comments.

smblott-github:
@mrmr1993 OK.  Merged this.  I'm about to merge #2470 too (move to `keydown`, `event.key`).  I don't use Firefox, so the ball is largely in your court in terms of figuring out what works and what doesn't.

PR #2487, opened 2017-04-21
Added hard reload binding - resolves #2486

### LINKED ISSUE #2486: Hard Refresh
I find that I regularly want to perform a hard refresh `Shift+Command+r`. That 🖕 claw style binding really pulls your hands away from the home row. `Shift+r` aka `R` would be more comfortable. 

**Useful for lots of Vimium users**
- There are many times (especially for developers) where you want to make a request from the server and not potentially what is cached in the browser.
- Chrome does not allow for creating this binding.

**Require no/little documentation**
- just a single line

  F       open a link in a new tab
  r       reload
  **R       hard reload (not from cache)**
gs      view source

**Useful without configuration**
- no configuration required

**Intuitive or leverage strong convention from Vim**
- leverages the existing lowercase `r` keyword pattern

**Work robustly on most/all sites**
- MDN shows full browser compatibility
- same as `r` implementing `.reload()` but with `true` passed in to force request to server

| Feature | Chrome | Edge | Firefox (Gecko) | Internet Explorer | Opera | Safari |
|---|---|---|---|---|---|---|
| Basic support | (Yes) | (Yes) | (Yes) | (Yes) | (Yes) | (Yes) |

MrJadaml:


smblott-github:
Thanks @MrJadaml.  A better way of doing this might be via a command option.  Something like...

    map R reload hard

There's an example of using a command option [here](https://github.com/philc/vimium/blob/master/content_scripts/vimium_frontend.coffee#L390).  With the example above, `options.registryEntry.options.hard` would be `true`.

gdh1995:
MDN use `forcedReload` to name this option: https://developer.mozilla.org/en-US/docs/Web/API/location/reload, so I support a new option `"forced"` / `"force"`.

krlooss:
Please accept this PR

PR #2494, opened 2017-04-24
Fix #2430

### LINKED ISSUE #2430: [Future] add shortcut for copy decoding current url
There is a shortcut "yy" for copy the current url to the clipboard. It copies encoding url, for example on page https://ru.wikipedia.org/wiki/Заглавная_страница it will copy https://ru.wikipedia.org/wiki/%D0%97%D0%B0%D0%B3%D0%BB%D0%B0%D0%B2%D0%BD%D0%B0%D1%8F_%D1%81%D1%82%D1%80%D0%B0%D0%BD%D0%B8%D1%86%D0%B0

It would be nice if you add shortcut for copying decoding url, "yd" for example.

Thanks for vimium!

dima74:
This pull request perform future request #2430

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -371,7 +371,8 @@ extend window,
     # TODO(ilya): When the following bug is fixed, revisit this approach of sending back to the background
     # page to copy.
     # http://code.google.com/p/chromium/issues/detail?id=55188
-    chrome.runtime.sendMessage { handler: "getCurrentTabUrl" }, (url) ->
+    chrome.runtime.sendMessage { handler: "getCurrentTabUrl" }, (urlEncoding) ->
+      url = decodeURI(urlEncoding)
       chrome.runtime.sendMessage { handler: "copyToClipboard", data: url }
       url = url[0..25] + "...." if 28 < url.length
       HUD.showForDuration("Yanked #{url}", 2000)


smblott-github:
OK -- thanks -- but two comments:

1. With `yy`, the result should be something that you can paste into either the vomnibar or the omnibox.  With this PR, you can no longer do that.
2. Vimium's current behaviour seems to match the behaviour of the Chrome omnibox.  We'd normally want a pretty good reason to do our own thing with a feature like this.

So, I suspect the status quo is where we want to be.

dima74:
OK. Can I open another pull request, that adds new shortcut for copying decoding url?

smblott-github:
> Can I open another pull request, that adds new shortcut for copying decoding url?

I'm not sure there's much demand for this.  Why not leave it sitting for a bit and see what/whether other's chip in?

dima74:
OK, thank you.

dima74:
@smblott-github, could you say, please, if It can't be this pr, and it can't be pr with new shortcut, maybe it will be an option in vimium settings, which controls whether copied url is decoded or not? (Thank you)

ght:
I can definitely see this coming in useful sometimes, but it should not be the default behavior, as currently implemented in this PR. It would be nice to have this available as a command to bind though.

philc:
Thinks for working on this @dima74. Closing due to age and low demand -- I think this as its own command is too niche for general usage, and it shouldn't be the default behavior of yy.

gdh1995:
Um, we may add an option like `options.registryEntry.options.decode` to toggle this behavior.

PR #2496, opened 2017-04-24
Make Mode::exit idempotent

mrmr1993:


smblott-github:
Is this not idempotent already?  Some parts *could* be executeed twice, e.g.:

```Coffeescript
Mode.modes = Mode.modes.filter (mode) => mode != this
@modeIsActive = false
@setIndicator()
```

... but these are idempotent already, are they not?

mrmr1993:
> these are idempotent already, are they not?

No. Especially `setIndicator`, since it sends an event through `handlerStack` every time it is called. But my main concern is that an `onExit` handler calling `exit` again makes these things happen early, which is just asking for race conditions.

I also figured this would be unobjectionable, since current behaviour makes no sense: why should calling `exit`again makes these things happen immediately when they wouldn't otherwise?

PR #2505, opened 2017-05-06
Fix focus event handling within shadow DOM.

### LINKED ISSUE #2504: Use of shadowRoot in page makes each focus event register another focus handler, potentially building up to massive slowdowns
I believe this line is the culprit: https://github.com/philc/vimium/blob/master/content_scripts/mode_insert.coffee#L69

Repro:

1) Open http://openfin.github.io/fin-hypergrid-polymer-demo/components/fin-hypergrid/demo.html
2) Add a breakpoint in the `else if (event.target.shadowRoot) {` block of mode_insert.js
3) Click a cell, observe that the event handler gets invoked a couple times (not sure why it's already more than 1).  Click another cell, observe that it now gets invoked 3 times.  Click another, 4 times, etc etc.

Vimium 1.59

smblott-github:
Fixes #2504.

For some reason, the test removed here never fires.

However, it appears that *any* `blur` event has to result in an input
within the shadow DOM (which previously had the focus) losing the focus.
Therefore, we don't need this test.

Aside:

- This code is firing *twice*.  So, we add two handlers and remove two
  handlers each time.  This seems to be harmless.  I think what's
  happening is that we first see the event when it's bubbling down, so
  our newly-added handler on the shadow-DOM element fires again as the
  event continues bubbling down.

### DIFF
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -70,11 +70,10 @@ class InsertMode extends Mode
 
           handlerStack.push
             _name: "shadow-DOM-input-mode"
-            blur: (event) ->
-              if event.target.shadowRoot == shadowRoot
-                handlerStack.remove()
-                for own type, listener of eventListeners
-                  shadowRoot.removeEventListener type, listener, true
+            blur: (event) => @alwaysContinueBubbling ->
+              handlerStack.remove()
+              for own type, listener of eventListeners
+                shadowRoot.removeEventListener type, listener, true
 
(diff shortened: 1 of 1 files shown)


mrmr1993:
Unfortunately we can't do this.

[Minimal example](https://jsfiddle.net/s2p2en5t/). Focus one input, then change focus to the other. We are no longer in insert mode, and typing activates Vimium commands.

philc:
Closing in favor of #2506.

PR #2506, opened 2017-05-06
Refine insert mode to correctly capture all shadow DOM focuses/blurs

### LINKED ISSUE #2504: Use of shadowRoot in page makes each focus event register another focus handler, potentially building up to massive slowdowns

mrmr1993:
This replaces #2505 to (hopefully) fully fix #2504.

### DIFF
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -102,6 +102,7 @@ class InsertMode extends Mode
                     handlerStack.remove()
                     for own type, listener of eventListeners
                       shadowRoot.removeEventListener type, listener, true
+                  return handlerStack.continueBubbling
 
     # Only for tests.  This gives us a hook to test the status of the permanently-installed instance.
     InsertMode.permanentInstance = this if @permanent
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -18,14 +18,14 @@ class InsertMode extends Mode
         return @suppressEvent
 
       return @passEventToPage unless event.type == 'keydown' and KeyboardUtils.isEscape event
-      target = event.target
+      target = event.path?[0] ? event.target
       if target and DomUtils.isFocusable target
         # Remove the focus, so the user can't just get back into insert mode by typing in the same input box.
         target.blur()
       else if target?.shadowRoot and @insertModeLock
         # An editable element in a shadow DOM is focused; blur it.
         @insertModeLock.blur()
-      @exit event, event.target
+      @exit event, target
       DomUtils.consumeKeyup event
 
     defaults =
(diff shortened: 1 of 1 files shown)


content_scripts/mode_insert.coffee
```diff
@@ -47,34 +47,61 @@ class InsertMode extends Mode
...
-            _name: "shadow-DOM-input-mode"
-            blur: (event) ->
-              if event.target.shadowRoot == shadowRoot
-                handlerStack.remove()
-                for own type, listener of eventListeners
-                  shadowRoot.removeEventListener type, listener, true
+        # NOTE(mrmr1993): The first element of event.path gives us the element being focused, even when it is
+        # in an (open) shadow DOM.
+        target = event.path?[0] ? event.target
+        if ShadowRoot?
```
gdh1995: Here `window.ShadowRoot?` is better: Chrome has `window.ShadowRoot` as early as version 36, but Firefox only declares a global `ShadowRoot` if `dom.webcomponents.enabled` is `true` in `about:config`.
BTW, now Vimium can not work if no `ShadowDOM` APIs, so in face this line is not needed.

mrmr1993: > Here `window.ShadowRoot?` is better

Fair enough.

> Firefox only declares a global `ShadowRoot` if `dom.webcomponents.enabled` is `true` in `about:config`.

I know.

> BTW, now Vimium can not work if no `ShadowDOM` APIs, so in face this line is not needed.

It works fine in Firefox and in PhantomJS, both with no Shadow DOM.

gdh1995: Oh it's my mistake to think VImium can not work without ShadowDOM. So this line is indeed necessary.

philc:
This no longer merges cleanly, but it doesn't look too hard to merge. I'm not sure if the underlying issue is still a problem. @gdh1995 what do you think?

gdh1995:
Oh yes this has been replaced by #2851 , which is from the idea of https://github.com/philc/vimium/issues/2841#issuecomment-347995300

philc:
Thanks for tying those all together @gdh1995. Closing this.

PR #2507, opened 2017-05-08
Update linear gradient syntax to avoid deprecation

bingocaller:
This updated syntax avoids the following deprecation warning in Chrome Canary v60.0.3090.0:
```
[Deprecation] -webkit-gradient is deprecated. Please use linear-gradient or radial-gradient instead.
```

This syntax is also supported in Firefox (and [all other modern browsers](https://caniuse.com/#feat=css-gradients)), which could help transitioning the Firefox version, which is in the works.


gdh1995:
Um, Chrome 60.0.3107.4 does not report such a deprecation warning. Anyway, this PR is still OKay.

Issue #2521, opened 2017-05-26
FF: clipboard copy does not work 


jabranham:
Hello - 

Copying the current URL (`copyCurrentUrl`) to the clipboard in Firefox does not work for me. Glancing through the manifest, I notice that the `clipboardWrite` permission isn't listed. I'm guessing that adding that will do the trick?

Thanks!

mrmr1993:
I'm sure this will be part of the solution, but we'll also have to change *where* we're doing the copying to clipboard (currently from the background page). From [this document](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Interact_with_the_clipboard):

> you can write to the clipboard like this in all execution contexts *except background pages*. In Firefox you can't select text or focus an input field in background pages, so you can't write to the clipboard from a background page

Should still be an easy fix; I'll give it a look now.

Mendess2526:
Status on this issue? I have this issue, but it's very inconsistent. Sometimes it works, sometimes it doesn't. `yf` seems to be the most problematic one, `yy` seems to work fine most of the time.

Firefox: v71 
OS: Arch
Using Xorg

PR #2524, opened 2017-05-26
WIP: Enable Firefox clipboard commands

mrmr1993:
This enables clipboard copying on Firefox 51+ and pasting on Firefox
54+, addressing #2521.

* A `.blur()` has been added to each of the commands in Clipboard, since (at
least) Firefox does not fire a blur event when the focused element is
removed from the document, and thus we get stuck perpetually in insert
mode.
  - A more sophisticated fix would be to not enter insert mode in the
    first place.
  - This also means that our behaviour is wrong on webpages which remove
    a user-focused text input. We should probably fix this.
* Commands using the clipboard now clear the user's selection in the
  page. For the most part, people will probably not notice this, but
  it's still worth bearing in mind.
  - We could mitigate this by doing the copy in an iframe, especially
    since we already have some active. However, this will make the code
    confusing (i.e. "why is the content script talking to the HUD to
    copy/paste text?!").
  - Users who use "p" to paste bookmarklets that act upon their existing
    selection will be upset by this. I expect there are exactly 0 of
    these users.
* The page scrolls to the bottom on paste.

I recommend against merging this (except b916562), but I'll leave the PR here
in case anybody has any great ideas on how to proceed.

/cc @jabranham


gdh1995:
This PR is **unsafe**: In this way, the page's host scripts may **change and get what users copy and paste** by installing `document.onbeforecopy` and `document.onbeforepaste` on Chrome, although Firefox has no such listeners.

So it's a must to execute different code on Chrome and Firefox.

langzisas:
.mozilla.org[](url)

mrmr1993:
> * What is the status of this?

This was just me getting the idea out there, to see if we could do anything with it. I don't think we should merge it.

> * What is your response to the comment re. safeness above?

He's right and I was aware of it; I just wanted a base to build off as Firefox's implementation evolves.

mrmr1993:
Closing in favour of #2601. Pasting doesn't work there yet, but it uses the HUD for the clipboard management, which is an improvement.

PR #2525, opened 2017-05-26
Change keyEventSuppressor to a handler

mrmr1993:
There's no need for `keyEventSuppressor` to be a `Mode`, since it doesn't use any of its mechanisms beyond being a handler. This is a no-op.


philc:
Thanks @mrmr1993

PR #2533, opened 2017-06-06
font color append "!important"

qwIvan:
image

In a few cases, vimium will conflict with other css rule.


philc:
This broke the link styles in Vomnibar, so I reverted it. The suggestion of using shadowdom in #2946 will be much more robust when implemented.

PR #2543, opened 2017-06-20
add options to make global InsertMode more compatible with other tools

### LINKED ISSUE #2542: Remove/Rework Passthrough to remove interference with other extensions
Hi there...

Is there a way to modify the passthrough for this extension such that it doesn't hijack some other Chrome extensions I use? (e.g. gleebox, typeahead etc.). Unfortunately, it's practically unusable for me in it's current state.

**My mapping file looks like this:** 
unmapAll
map  dj scrollDown
map dk scollUp
map dh scrollLeft
map dl scrollRight
map yy  copyCurrentUrl)
map , Vomnibar.activate
map i focusInput

The [surfkeys](https://chrome.google.com/webstore/detail/surfingkeys/gfbliohnnapiefjpjlpjnehglfpaknnc) extension does a good job at a VIM-like approach but ununfortunately takes up 100% CPU everytime I use it. 

Thanks,

gdh1995:
This should fix #2542 .

All options are off by default, so this PR doesn't change old behaviors of `enterInsertMode`.

### DIFF
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -8,6 +8,8 @@ class InsertMode extends Mode
     # If truthy, then we were activated by the user (with "i").
     @global = options.global
 
+    @passExitKey = options.passExitKey
+
     handleKeyEvent = (event) =>
       return @continueBubbling unless @isActive event
       return @passEventToPage if @insertModeLock is document.body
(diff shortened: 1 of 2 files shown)


gdh1995:
Seems that no other people report a same bug. Then this may not be so useful.

PR #2555, opened 2017-07-02
findAndFollowLink(): look at title and id of links

matthew-cline:
This patch makes it so that the commands `goNext` and `goPrevious` (`]]` and `[[`) will also examine the value of each link's `title` attribute and `id` attribute (if they exist) for the desired patterns.  The use case for this is pages where the next/previous links are images without a `rel` attribute.  An example is the [*Girl Genius*](http://www.girlgeniusonline.com/comic.php?date=20021104#.WVhYunUrKHs) webcomic.

Here is the [patch as a diff](https://github.com/philc/vimium/files/1117646/diff.txt).

[Here is the output of `cake test`](https://github.com/philc/vimium/files/1117648/test.txt), though I don't think those are due to me.

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -595,9 +595,14 @@ findAndFollowLink = (linkStrings) ->
       continue
 
     linkMatches = false
-    for linkString in linkStrings
-      if link.innerText.toLowerCase().indexOf(linkString) != -1 ||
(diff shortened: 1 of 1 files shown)


philc:
Thanks for exploring this @matthew-cline. I'm going to close this due to lack of interest. I don't think the added complexity is worth solving this edge case.

PR #2558, opened 2017-07-04
Simplify rectsOverlap

mrmr1993:
This PR does several things (each its own commit):
* Simplifies the definition of `Rect.rectsOverlap` to 4 comparisons (the old one used 16).
* Rename `Rect.contains` to `Rect.intersects`
  - since `Rect.contains` is almost identical to (the reduced) `Rect.rectsOverlap`, poor naming must have motivated creating `Rect.rectsOverlap`
  - add a comment (actually the comment accompanying `Rect.rectsOverlap`) to make it more discoverable
* Rename `Rect.rectsOverlap` to `Rect.intersectsStrict`
  - I don't like the name `rectOverlaps`, and `intersectsStrict` gives a clearer impression of what it's doing and how it relates to `intersects`
  - change the comment to reflect that it considers boundary-only intersections (which `intersects` does not)
* Use `Rect.intersects` instead of `Rect.intersectsStrict` at the sole use-point
  - We don't care about boundary intersections in this situation, and so it is wasteful to consider them.

This should have no user-visible impact.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -321,7 +321,7 @@ class LinkHintsMode
   rotateHints: do ->
     markerOverlapsStack = (marker, stack) ->
       for otherMarker in stack
-        return true if Rect.intersectsStrict marker.markerRect, otherMarker.markerRect
+        return true if Rect.intersects marker.markerRect, otherMarker.markerRect
       false
 
     ->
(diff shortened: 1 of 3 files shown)


mrmr1993:
Rebased onto master. @smblott-github any chance of merging this?

smblott-github:
Done.

mrmr1993:
Thanks!

PR #2573, opened 2017-07-20
Update README.md

alphaCTzo7G:
add W to list of default commands to make current tab into a new window..


wetlife:
This seems like a good idea on first blush. Vi's power is in its ability to make arbitrary commands from composeable lingual elements. This seems to come from the land of a hotkey for command. I think this may specialize vimium's interface too far away from that of vi for inclusion in the trunk.

PR #2601, opened 2017-08-15
Enable Firefox clipboard commands

mrmr1993:
This replaces #2524.

* Copying from the clipboard works in both Chrome and Firefox.
* Pasting from the clipboard only works in Chrome.
  - I can't work out why it's not working in Firefox.
  - the method [here](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Interact_with_the_clipboard) for pasting is almost exactly what we're using, so I'm pretty low on ideas.
* ~~I haven't converted the `LinkHints.activateModeToCopyLinkUrl` command to use the new handler, since I'm focusing on getting Firefox to paste.~~ Done

NB: this directs the copy/paste commands via the HUD. This feels (and is) weird, but it's the best we've got if the background page doesn't do this for us on FF.

/cc @smblott-github

### DIFF
--- a/content_scripts/mode_normal.coffee
+++ b/content_scripts/mode_normal.coffee
@@ -91,7 +91,7 @@ NormalModeCommands =
 
   copyCurrentUrl: ->
     chrome.runtime.sendMessage { handler: "getCurrentTabUrl" }, (url) ->
-      chrome.runtime.sendMessage { handler: "copyToClipboard", data: url }
+      HUD.copyToClipboard url
       url = url[0..25] + "...." if 28 < url.length
       HUD.showForDuration("Yanked #{url}", 2000)
 
--- a/content_scripts/mode_normal.coffee
+++ b/content_scripts/mode_normal.coffee
@@ -95,6 +95,15 @@ NormalModeCommands =
       url = url[0..25] + "...." if 28 < url.length
       HUD.showForDuration("Yanked #{url}", 2000)
 
+  openCopiedUrlInNewTab: (count) ->
+    HUD.pasteFromClipboard (url) ->
+      for i in [0...count] by 1
+        chrome.runtime.sendMessage { handler: "openUrlInNewTab", url }
+
+  openCopiedUrlInCurrentTab: ->
+    HUD.pasteFromClipboard (url) ->
+      chrome.runtime.sendMessage { handler: "openUrlInCurrentTab", url }
+
   # Mode changes.
   enterInsertMode: ->
     # If a focusable element receives the focus, then we exit and leave the permanently-installed insert-mode
--- a/pages/hud.coffee
+++ b/pages/hud.coffee
@@ -100,6 +100,7 @@ handlers =
     Clipboard.copy data
     focusedElement?.focus()
     window.parent.focus()
+    UIComponentServer.postMessage {name: "unfocusIfFocused", data}
 
   pasteFromClipboard: ->
     focusedElement = document.activeElement
--- a/pages/hud.coffee
+++ b/pages/hud.coffee
@@ -95,5 +95,16 @@ handlers =
       " (No matches)"
     countElement.textContent = if showMatchText then countText else ""
 
+  copyToClipboard: (data) ->
+    focusedElement = document.activeElement
+    Clipboard.copy data
+    focusedElement?.focus()
+
+  pasteFromClipboard: ->
+    focusedElement = document.activeElement
+    data = Clipboard.paste()
+    focusedElement?.focus()
+    UIComponentServer.postMessage {name: "pasteResponse", data}
+
 UIComponentServer.registerHandler ({data}) -> handlers[data.name ? data]? data
 FindModeHistory.init()
--- a/pages/hud.coffee
+++ b/pages/hud.coffee
@@ -99,11 +99,13 @@ handlers =
     focusedElement = document.activeElement
     Clipboard.copy data
     focusedElement?.focus()
+    window.parent.focus()
 
   pasteFromClipboard: ->
     focusedElement = document.activeElement
     data = Clipboard.paste()
     focusedElement?.focus()
+    window.parent.focus()
     UIComponentServer.postMessage {name: "pasteResponse", data}
 
 UIComponentServer.registerHandler ({data}) -> handlers[data.name ? data]? data
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -340,8 +340,8 @@ commandDescriptions =
   toggleViewSource: ["View page source", { noRepeat: true }]
 
   copyCurrentUrl: ["Copy the current URL to the clipboard", { noRepeat: true }]
-  openCopiedUrlInCurrentTab: ["Open the clipboard's URL in the current tab", { background: true, noRepeat: true }]
-  openCopiedUrlInNewTab: ["Open the clipboard's URL in a new tab", { background: true, repeatLimit: 20 }]
+  openCopiedUrlInCurrentTab: ["Open the clipboard's URL in the current tab", { noRepeat: true }]
+  openCopiedUrlInNewTab: ["Open the clipboard's URL in a new tab", { repeatLimit: 20 }]
 
   enterInsertMode: ["Enter insert mode", { noRepeat: true }]
   passNextKey: ["Pass the next key to the page"]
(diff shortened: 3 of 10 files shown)


gdh1995:
What if you add the "clipboardRead" permission into the permissions array in manifest.json?

Just an idea needing checks - I have no laptop at hand.

mrmr1993:
> What if you add the "clipboardRead" permission into the permissions array in manifest.json?

We already have it, I'm afraid.

mrmr1993:
With 1548561 (implied necessary by https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Interact_with_the_clipboard), `document.execCommand("Paste")` returns `true` (rather than `false` before), but still nothing is pasted.

mrmr1993:
The 'working' [example](https://bugzilla.mozilla.org/show_bug.cgi?id=1312260#c42) from the issue in their issue tracker doesn't work on FF54 or today's nightly. I'm giving up on reading the clipboard for now.

mrmr1993:
Pushed some more commits and reordered. This probably wants merging up to c50c172479487dec1dd44a166897b3ceac4a5aa7. (There's no point merging 6659501 and d5db97ef08329eb440292fa8d6d0a74edb7638f6 since they don't actually fix pasting).

mrmr1993:
I've done all I want to with this now; IMO it's ready to go.

For posterity, several issues combined to stop pasting from working for me earlier:
* Firefox won't paste in frames that are `display: none`
* Firefox only pastes in `<textarea contentEditable=true>` elements
* Sometimes (specifically, when the active element is the document `<body>`), the HUD frame gets selected as `document.activeElement`, and we don't receive any more key events until we `blur` it from the document
* Firefox (almost?) always pastes nothing/whitespace when the clipboard contains rich text
  - this includes text selected from webpages, which was pretty confusing while testing
  - text from textboxes, the URL bar, etc. are fine

The last of these is still an issue for us, but there's no way of working around it.

gdh1995:
According to https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Interact_with_the_clipboard, an web-extension is able to do pasting on its background page using any content-editable elment. And I've succeed in pasting on background page, using `<body contenteditable="true">` or  `<textarea contenteditable="true">`.

But there're still some problems if using `<textarea c~e~=true>`:
* copied multiline text are converted into single-line text and all of `\n` are removed
* only pure text data (seems to be copied as `text/plain`) can be pasted
* `document.execCommand('Paste')` wouldn't work (although still return true) any more, if only executing `textarea.value = ""`, even if I created and used new `<textarea>`s - I don't know why but it was

So I suggest using `html > body > div[contenteditable=true]` on the background page to do pasting.

mrmr1993:
Chrome are [implementing](https://bugs.chromium.org/p/chromium/issues/detail?id=677564) an [async clipboard API](https://github.com/garykac/clipboard/blob/master/clipboard.md). It's been [suggested](https://bugzilla.mozilla.org/show_bug.cgi?id=1344410#c9) on (one of the many) Firefox issue(s). We should keep an eye to see if this comes to anything, since it would make all the pain and weirdness go away.

smblott-github:
@mrmr1993 What is the status of this?  Is it *good-to-go*, from your perspective?

mrmr1993:
Yes. I'd like for there to be a better solution, but there doesn't seem to be for the time being. I'll keep an eye on the Firefox issues so that we can use something better when it comes up.

On 15 Sep 2017 10:15 am, "Stephen Blott" <notifications@github.com<mailto:notifications@github.com>> wrote:

@mrmr1993<https://github.com/mrmr1993> What is the status of this? Is it good-to-go, from your perspective?

ahstro:
Any chance of merging this soon? 😇

smblott-github:
> Any chance of merging this soon?

Maybe.  Maybe not.  My understanding is that it's not as reliable as we'd like.  And something which works only sometimes is a bad UX.

majutsushi:
Would it be possible to at least get the yanking/copying *to* the clipboard part merged in the not-too-distant future? Judging by the multiple duplicates here and also my own experience that is the much more common operation, and the one that apparently works already. Especially if pasting from the clipboard might have to wait for the Firefox bug mentioned above, which would surely take a few months to actually be available in release versions. Thanks!

smblott-github:
@mrmr1993... @majutsushi makes a good point, no?

If the yanking is reliable then we should merge that and (perhaps) disable all of the paste commands on FF.

mrmr1993:
Sounds good. I'll rebase now.

mrmr1993:
Rebased. ([Old version here](https://github.com/mrmr1993/vimium/tree/ff-copy-paste-backup))

eyaleizenberg:
Any updates on this?

ahstro:
I had a dream this was merged. Now, I'm not saying I possess the gift of precognition, but I'm hoping for an early christmas miracle 🎄 ✨

philc:
@smblott-github I think this has reached a good state; it LGTM. Let's merge if there are no further objections.

mrmr1993:
Can do, but I preferred not to. If there are any issues, I thought it would be better if we could just point the commands back to the old code on Chrome.

mrmr1993:
In latest Firefox, we can paste into contenteditable elements that aren't `<textarea>`. This gives us the ability to paste rich text, fixing the last of the issues [I mentioned above](https://github.com/philc/vimium/pull/2601#issuecomment-322597060). I've pushed a commit to take advantage of this.

This PR now gives Firefox the same full clipboard support as Chrome.

gdh1995:
@mrmr1993 As I mentioned before, div[ce=true] is not safe enough, because Chrome will paste HTML content including images (in my tests, ``s). This means unexpected images might be loaded, if a user copy a line containing image icons from other web pages.

gdh1995:
So I'm going to test the different behavior on pasting between Chrome and Firefox, and try using browser-specified APIs like `onpaste` to solve them safely.

content_scripts/mode_normal.coffee
```diff
@@ -91,10 +91,19 @@ NormalModeCommands =
...
   copyCurrentUrl: ->
     chrome.runtime.sendMessage { handler: "getCurrentTabUrl" }, (url) ->
-      chrome.runtime.sendMessage { handler: "copyToClipboard", data: url }
+      HUD.copyToClipboard url
       url = url[0..25] + "...." if 28 < url.length
       HUD.showForDuration("Yanked #{url}", 2000)
 
+  openCopiedUrlInNewTab: (count) ->
+    HUD.pasteFromClipboard (url) ->
+      for i in [0...count] by 1
```
smblott-github: I might change this to use `mkRepeatCommand` in `main.coffee` (to keep the `count` logic in as few places as possible).

pages/hud.coffee
```diff
@@ -95,5 +95,19 @@ handlers =
       " (No matches)"
     countElement.textContent = if showMatchText then countText else ""
 
+  copyToClipboard: (data) ->
+    focusedElement = document.activeElement
+    Clipboard.copy data
+    focusedElement?.focus()
+    window.parent.focus()
+    UIComponentServer.postMessage {name: "unfocusIfFocused", data}
```
smblott-github: `unfocusIfFocused`, above, does not take `data`, does it? I'll remove `data` here; let me know if that's incorrect, @mrmr1993.

smblott-github:
@mrmr1993 I see that the new permission here causes Chrome to disable Vimium until the user accepts the new permission.

Might you be able to look into doing this dynamically: [https://developer.chrome.com/apps/permissions](https://developer.chrome.com/apps/permissions)?

gdh1995:
The clipboardWrite permission is safe to remove from manifest on Chrome.

The only effects I can see are:
* Chrome will show an alert dialog when I add it.
* The extension's detail will become "reading and changing your data...", while the old is only "reading"

But I think it's a good change to notice users that Vimium is allowed to modify the system clipboard.

mrmr1993:
> Might you be able to look into doing this dynamically: https://developer.chrome.com/apps/permissions?

@smblott-github I had a look into it ([branch here](https://github.com/mrmr1993/vimium/tree/experiments/optional-permission-clipboard)), but to no success. To quote [Mozilla's docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/permissions/request):

> The request can only be made inside the handler for a user action, such as
>
> * clicking the extension's browser action or page action
> * selecting its context menu item
> * activating a keyboard shortcut defined by the extension
> * clicking a button in a page bundled with the extension.

None of those things happens when a user tries to paste a URL using our commands. It looks like optional permissions are a no-go for us on FF.

(Sorry for the delay with this, I've had a few busy days.)

smblott-github:
OK.  Thank you.

PR #2602, opened 2017-08-15
Firefox: Simulate default action for clicking links with link hints

mrmr1993:
This makes the `LinkHints.activateModeToOpenInNew{,Foreground}Tab` commands work as expected on Firefox.

Currently the flag (`simulateClickDefaultAction`) is hardcoded on, which opens tabs twice in Chrome for the affected cases. It could be an option, use UA sniffing, or both (default set by UA sniffing), so I've left it for now.

@smblott-github this is a big usability win (currently FF can only open links in the current page).

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -117,6 +117,16 @@ TabOperations =
       catch
         callback.apply this, arguments
 
+  # Opens request.url in new window and switches to it.
+  openUrlInNewWindow: (request, callback = (->)) ->
+    winConfig =
+      url: Utils.convertToUrl request.url
+      active: true
+    winConfig.active = request.active if request.active?
+    # Firefox does not support "about:newtab" in chrome.tabs.create.
+    delete tabConfig["url"] if tabConfig["url"] == Settings.defaults.newTabUrl
+    chrome.windows.create winConfig, callback
+
 toggleMuteTab = do ->
   muteTab = (tab) -> chrome.tabs.update tab.id, {muted: !tab.mutedInfo.muted}
 
@@ -416,6 +426,7 @@ sendRequestHandlers =
   # with Chrome-specific URLs like "view-source:http:..".
   getCurrentTabUrl: ({tab}) -> tab.url
   openUrlInNewTab: (request) -> TabOperations.openUrlInNewTab request
+  openUrlInNewWindow: (request) -> TabOperations.openUrlInNewWindow request
   openUrlInIncognito: (request) -> chrome.windows.create incognito: true, url: Utils.convertToUrl request.url
   openUrlInCurrentTab: TabOperations.openUrlInCurrentTab
   openOptionsPageInNewTab: (request) ->
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -106,6 +106,7 @@ TabOperations =
       index: request.tab.index + 1
       active: true
       windowId: request.tab.windowId
+    tabConfig.active = request.active if request.active?
     # Firefox does not support "about:newtab" in chrome.tabs.create.
     delete tabConfig["url"] if tabConfig["url"] == Settings.defaults.newTabUrl
     chrome.tabs.create tabConfig, (tab) ->
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -11,6 +11,7 @@
 # The "name" property below is a short-form name to appear in the link-hints mode's name.  It's for debug only.
 #
 isMac = KeyboardUtils.platform == "Mac"
+simulateClickDefaultAction = true
 OPEN_IN_CURRENT_TAB =
   name: "curr-tab"
   indicator: "Open link in current tab"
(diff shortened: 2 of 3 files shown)


philc:
Any reason to delay merging this?

mrmr1993:
`simulateClickDefaultAction` needs to be set differently on the FF side vs. the Chrome side, so a decision needs to be made on how to do that. I'll push a UA sniffing commit here shortly.

AFAIC, this is otherwise mergeable. I don't like that we have to do it or how it's done, but it does work.

philc:
Sorry, I didn't read the PR description carefully enough. Thanks for clarifying. Yes, we need browser-branching before merging =)

mrmr1993:
Pushed UA sniffing for Firefox (as `Utils.isFirefox`). This should now be ready to go 🎉 

I added a small change that gets rid of the lag when reloading the extension on FF; I can extract it out if need be. See comment on the commit for more.

background_scripts/main.coffee
```diff
@@ -5,6 +5,7 @@ root = exports ? window
 chrome.runtime.onInstalled.addListener ({ reason }) ->
   # See https://developer.chrome.com/extensions/runtime#event-onInstalled
   return if reason in [ "chrome_update", "shared_module_update" ]
+  return if Utils.isFirefox()
```
gdh1995: Why to do nothing on `runtime.onInstalled` if the browser is Firefox?

mrmr1993: Re-injecting our content scripts doesn't work properly (ie. doesn't activate Vimium in tabs before they've been reloaded) on Firefox, and it makes the browser lag considerably while it's doing it. It was annoying me while testing, so I disabled it.

gdh1995: Oh, it does fail. Thanks! Then I want to test the update notification if you haven't, in order to ensure nothing is working on Firefox.

mrmr1993: > test the update notification if you haven't

I haven't. Go for it.

mrmr1993:
@smblott-github @philc can this be merged? I've been using it (on Chrome and Firefox) for nearly 2 weeks without any problems.

smblott-github:
What about like [this](https://github.com/smblott-github/vimium/commit/879a65537261bd3b090691d3f49299e1324249f7), @mrmr1993?  The test of the modifier keys looked too confusing to me.

gdh1995:
I think there may be some websites which don't call `preventDefault` for a mouse event on `<a href="#">function...</a>`, and neither do those on `<a href="javascript:;">...</a>`, so it's better to check `link.getAttribute("href") not in ["", "#"]` and `"://" in link.href`.

`href="#"` and `herf="javascript:;"` are often used to provide a underline for the link text and make an `<a>` tab-focusable.

mrmr1993:
Good call! I wondered about these, but assumed they would spawn a new window. I've certainly never come across a spec for it not.

ivanruvalcaba:
Hi all,

I don't know if the following information is relevant but I have noticed that this functionality doesn't work properly if you change the user-agent (tested on my Firefox Nightly 57.0a1, Linux 64bits host). I am currently using the [AgentX web extension](https://addons.mozilla.org/en-US/firefox/addon/agentx/?src=ss) for that. This happens when I replace the user-agent with Google Chrome on MS Windows, so if I have to use this functionality I must restore the user-agent to a default value.

Best regards.

smblott-github:
> this functionality doesn't work properly if you change the user-agent

That's a serious problem.  Could we figure out the browser on the background page and then inform the front end when each frame registers?

gdh1995:
Why not check `typeof browser === "object" && !!browser.runtime`?

mrmr1993:
> Could we figure out the browser on the background page and then inform the front end when each frame registers?

Done. I've used a method which should be fairly future-proof and shouldn't break if/when Edge and Chrome adopt Mozilla's extension API additions.

mrmr1993:
> I think there may be some websites which don't call `preventDefault` for a mouse event on `<a href="#">function...</a>`, and neither do those on `<a href="javascript:;">...</a>`, so it's better to check `link.getAttribute("href") not in ["", "#"]` and `"://" in link.href`.
>
> href="#" and herf="javascript:;" are often used to provide a underline for the link text and make an `<a>` tab-focusable.

@gdh1995 I've just done some testing on this. The current behaviour matches what clicking does with the relevant modifier keys (except for `javascript:` URLs, where Firefox won't let us open a tab). I don't think these are changes we should make.

gdh1995:
> The current behaviour matches what clicking does with the relevant modifier keys

OKay.

And the new API `runtime.getBrowserInfo` does works. Good job!

Byron-TW:
@mrmr1993 thanks for this fix! Thanks to it I can complete my switch from Chrome to Firefox without missing anything.
@philc Do you think that after fixing the merge-conflicts, this PR can be merged?

content_scripts/link_hints.coffee
```diff
@@ -384,7 +384,17 @@ class LinkHintsMode
             window.focus()
             DomUtils.simulateSelect clickEl
           else
-            clickActivator = (modifiers) -> (link) -> DomUtils.simulateClick link, modifiers
+            clickActivator = (modifiers) -> (link) ->
+              defaultActionsTriggered = DomUtils.simulateClick link, modifiers
+              simulateClickDefaultAction = Utils.isFirefox()
+              if simulateClickDefaultAction and
```
smblott-github: @mrmr1993.  We can merge this.  But I'd really appreciate it if you could find a way of avoiding repeating the modifier/new-tab logic here.  That logic rightly belongs -- in my opinion -- with the various modes at the top of this file.

mrmr1993: What do you think about moving it into `DomUtils.simulateClick`? Since it's a workaround for that rather than specifically for link hints, there makes more sense to me.

I'll rebase and make the change -- whichever you choose -- in a little bit.

mrmr1993:
I've rebased this, moved the fix code into `DomUtils`, (hopefully) made the code a bit clearer, and added an equivalent workaround for the `shift` key modifier.

Issue #2532, opened 2017-06-06
disable by html meta tag ?


zhengkai:
Hi, I'm a contributor of a html5 game which called "Slay One" https://slay.one/

I know how to exclude it in vimium, but for incoming players(and use vimium), I don't want each of them to do it yourself, Is there any way to automatically disable it? 

for example, noindex meta tag in html page:

    <META NAME="ROBOTS" CONTENT="NOINDEX, NOFOLLOW">

Maybe we need a novimium tag ?

Thanks

mrmr1993:
Duplicate of #2399.

Issue #2399, opened 2017-01-19
Is there a way to disable Vimium on my page as a site owner?


vincentwoo:
Pardon me if this is answered somewhere, but I couldn't find it in the FAQ or README.

I run a website that many programmers use. Some of the use Vimium, and it appears the extension tries to upgrade editing on certain elements on my page. This has been causing some tricky JS errors that are very hard to track down as well as a lot of user pain because they don't realize the failures are due to the conflict between our code and Vimium's.

Is there a way to tag elements with "if you are running Vimium, please do not enhance this element"? Thanks for your reply in advance.

smblott-github:
@vincentwoo...  Not currently.  Such a feature would require some careful thought.

vincentwoo:
Are you open to considering this feature? If so, can you say what some of your major concerns are? I obviously don't want to step on any toes, but I do want to offer a better user experience for my users who have Vimium installed (which is a surprising chunk of people).

leeny:
+1 This would be a really useful addition.

gdh1995:
You may dispatch a fake `unload` message on DOM ready / window load - if only Vimium has inited.

Here's my example: https://jsfiddle.net/2L36ypys/, and you'll see you can not use `f` to activate hints.

vincentwoo:
I'm not a vimium user myself, so I'll install it tomorrow and try to have a more concrete bug report for you. Thanks for the snippet.

gdh1995:
In fact, according to tests in https://jsfiddle.net/2L36ypys/1/, we can dispatch such a message from the "host" environment when the page is `"loading"`, because Vimium installs its `unload` event listener before all other page scripts.

But my [Vimium++](https://github.com/gdh1995/vimium-plus) does not use this listener and there's no easy method to destroy it by page scripts.

So I'm expecting your searches. A test page where you can reproduce errors is just OKay. Thanks a lot.

pimlottc:
Related to this, there is a 2-year old project to detect whether the user is using Vimium, might be useful here: https://github.com/EvanHahn/Detect-Vimium

gdh1995:
@pimlottc The extension has not worked for 1 or 2 years, because Vimium uses 'shadowDOM' to hide its nodes.

vincentwoo:
I could not exactly reproduce the problem that vimium users have on my site (probably because I don't use Vim personally). Is it still on the table to add some kind of vimium-specific CSS class to elements to ask the extension not to augment it?

[...]

PR #2604, opened 2017-08-18
Work better with Google Docs

### LINKED ISSUE #2075: docs.google.com completely steals the focus
The focus is stolen to the text area and I can't use any vimium bindings and have to use chromium default keybindings to switch to another tab or the like.  This is a cause of constrant frustration for me when iterating through tabs and suddenly I drop into this 'hole' of google docs tab and need to use a different key to 'escape'.

mrmr1993:
This makes `<esc>` blur out of frames whose body is an editable element.

Fixes #2075.

Notes:
* The focused frame for Google Docs is __not__ what it appears to be: a hidden iframe collects the typing events, but the content is rendered in the main frame.
* I can't find an obvious element that clicking should be enabled for to return focus to the editor, so you have to use the mouse to continue editing.
* **Any bindings that use modifier keys will drop focus back into the editor:** Docs captures the events for them and uses them to refocus it.

### DIFF
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -25,7 +25,7 @@ class InsertMode extends Mode
         if window.frameElement?
           # We can't get a |frameId| from a |window| in a content script, so the only way of communicating
           # with a specific window is a custom event.
-          vimiumBlurEvent = new CustomEvent "vimiumBlurEvent"
+          vimiumBlurEvent = new CustomEvent "vimiumBlurEvent", {detail: JSON.stringify event}
           window.frameElement.dispatchEvent vimiumBlurEvent
           return DomUtils.consumeKeyup event
         return @passEventToPage # We can't use the <esc>: assume the user intended it for the page.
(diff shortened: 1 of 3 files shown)


mrmr1993:
Updated.

gdh1995:
The new PR breaks my origin purpose of preventing the official Vimium from grabbing Esc events on my custom Vimium - my custom Vimium use `about:blank` as its FindMode iframe. Fortunately, I've dispatched a fake `unload` event to iframes to force Vimium to exit itself ^_^ , so up to now my custom Vimium is still safe.

mrmr1993:
> I've dispatched a fake `unload` event to iframes to force Vimium to exit itself

I've been meaning to add a check for `event.isTrusted` for some time, but I hadn't made a move on it because I couldn't think of a way it would leak data or give privileged access to the page. I'd forgotten that this was a possibility, and this is definitely something pages shouldn't be able to do, as much as the people in #2399 and #2532 might want to.

Thanks for the heads up, I'll throw together a PR now.

mrmr1993:
> I'll throw together a PR now.

Done: #2605.

philc:
@mrmr1993 this is well-written and well-documented. It solves a valid use case, but adds non-trivial complexity. I'm going to close this due to age and lack of interest.

Issue #2618, opened 2017-08-27
Non-Latin keyboard layouts do not initiate actions


giogziro95:
When I switch to a non-Latin keyboard layout (e.g. Georgian, Russian, etc.), pressing the keys doesn't do anything. Is it possible to use something like [this](http://keycode.info/) instead?

- Chrome version: 60.0.3112.90 (Stable channel)
- Vimium version: 1.59

giogziro95:
Woah, sorry. Just searched and it has already been reported several times.

smblott-github:
From [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode):

![snapshot](https://user-images.githubusercontent.com/2641335/29752051-8c8e99f6-8b4e-11e7-87da-4637b29b4784.png)

giogziro95:
@smblott-github Yes, I saw that. Is there a substitute for it?

giogziro95:
The [example on the page](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#Example) suggests a workable solution with a fallback.

smblott-github:
It's not exactly satisfactory, but you could do something like [this](https://github.com/philc/vimium/wiki/Key-Mappings) for whatever you language is.  And you can leave the Latin bindings in place too.

Vimium is using `event.key`.

> The example on the page suggests a workable solution with a fallback.

No.  We're already using `event.key` (the default, on that page), and the other two are deprecated.    In fact, the other two would never fire because Chrome defines `event.key`.  Same for Firefox.

mrmr1993:
The ideal solution would be to allow mapping keys by code ([`event.code`](https://developer.mozilla.org/en/docs/Web/API/KeyboardEvent/code), but this will complicate the key handling code quite a bit, and I don't think anybody's attempted it so far.

giogziro95:
> We're already using `event.key` (the default, on that page), and the other two are deprecated. In fact, the other two would never fire because Chrome defines `event.key`. Same for Firefox.

That's really inconvenient for the purposes of Vimium. Almost nobody would want to fire different actions based on the layout they're using. Is there a more consistent method across the different layouts?

> It's not exactly satisfactory, but you could do something like this for whatever you language is. And you can leave the Latin bindings in place too.

Yeah, I know that. Unfortunately, for many users, that's a bit of hassle.

giogziro95:
[...]

PR #2626, opened 2017-09-01
Ignore keyboard layout

smblott-github:
This adds a new (advanced) option `ignoreKeyboardLayout` (as per @lydell's suggestion in #2618) to allow users of non-Latin keyboard layout's to avoid switching layout.

### DIFF
--- a/lib/keyboard_utils.coffee
+++ b/lib/keyboard_utils.coffee
@@ -23,7 +23,6 @@ KeyboardUtils =
       key = event.code
       # Strip some standard prefixes.
       key = key[3..] if key[...3] == "Key"
-      key = key[5..] if key[...5] == "Digit"
       key = key[6..] if key[...6] == "Numpad"
       # Translate some special keys to event.key-like strings.
       if @enUsTranslations[key]
@@ -83,6 +82,16 @@ KeyboardUtils =
     "Period":        [".", ">"]
     "Slash":         ["/", "?"]
     "Space":         [" ", " "]
+    "Digit1":        ["1", "!"]
+    "Digit2":        ["2", "@"]
+    "Digit3":        ["3", "#"]
+    "Digit4":        ["4", "$"]
+    "Digit5":        ["5", "%"]
+    "Digit6":        ["6", "^"]
+    "Digit7":        ["7", "&"]
+    "Digit8":        ["8", "*"]
+    "Digit9":        ["9", "("]
+    "Digit0":        ["0", ")"]
 
 KeyboardUtils.init()
 
--- a/lib/keyboard_utils.coffee
+++ b/lib/keyboard_utils.coffee
@@ -19,8 +19,15 @@ KeyboardUtils =
     unless Settings.get "ignoreKeyboardLayout"
       key = event.key
     else
+      # The logic here is from the vim-like-key-notation project (https://github.com/lydell/vim-like-key-notation).
       key = event.code
+      # Strip some standard prefixes.
       key = key[3..] if key[...3] == "Key"
+      key = key[5..] if key[...5] == "Digit"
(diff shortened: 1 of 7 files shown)


lydell:
Nice! A couple of notes:

VimFx clarifies "Pretend that the en-US QWERTY layout is always used" below its "Ignore keyboard layout" checkbox. ¯\\\_(ツ)\_/¯

You probably want to translate some `event.code` values to more `event.key`-like ones: https://github.com/lydell/vim-like-key-notation/blob/3b35a1e3068acb73ff41b6c8ebf0a506d74a4982/index.js#L25-L47

You might want to handle the numpad specially. From vim-like-key-notation's readme:

> The numpad is an exception. `event.key` is used there even if the current layout is set to be ignored. That’s because it is not possible to correctly emulate it.
>
> Just as with capslock, we don’t know the state of numlock. This means that we cannot know if `Numpad1` should result in `1` or `<End>`, for example. Therefore, it is much more useful to the user if we actually rely on the current layout in this case, in order not to break the numpad. This is especially important on laptops, where `<Home>`, `<End>`, `<PageUp>` and `<PageDown>` usually only exist there. The keypad usually doesn’t change between layouts anyway—at least not enough to warrant drastic changes to it.

lydell:
Starting to look good!

- You might still want to _always_ use `event.key` for the numpad, though (as mentioned in my previous comment).
- The `enUsTranslations` mapping seems to be missing the the characters for "shift+digit", i.e. the right-most ones here:

```js
  "Digit2":       ["2",  "@"],
  "Digit3":       ["3",  "#"],
  "Digit4":       ["4",  "$"],
  "Digit5":       ["5",  "%"],
  "Digit6":       ["6",  "^"],
  "Digit7":       ["7",  "&"],
  "Digit8":       ["8",  "*"],
  "Digit9":       ["9",  "("],
  "Digit0": ["0", ")"],
```

lib/keyboard_utils.coffee
```diff
@@ -16,17 +16,31 @@ KeyboardUtils =
...
+      # We cannot correctly emulate the numpad, so fall back to event.key; see #2626.
+      key = event.key
+    else
+      # The logic here is from the vim-like-key-notation project (https://github.com/lydell/vim-like-key-notation).
+      key = event.code
+      key = key[3..] if key[...3] == "Key"
+      # Translate some special keys to event.key-like strings.
+      if @enUsTranslations[key]
+        key = if event.shift then @enUsTranslations[key][1] else @enUsTranslations[key][0]
+      key = key.toLowerCase() unless event.shiftKey
```
lydell: Hmm ... won't this result in inconsistent results for non-char-keys?

- <kbd>Enter</kbd> → `enter`
- <kbd>Shift</kbd><kbd>Enter</kbd> → `Enter`

lydell:
Reading through the code, all looks good to me now! 👍

Just for completeness sake, in vim-like-key-notation’s readme it is mentioned that we cannot know the state of Caps Lock and Num Lock. Well, today I learned it actually _is_ possible, through [`event.getModifierState`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState). Chrome seems to support it when I tested it!

- If you feel like it, you could do `shift = event.shiftKey != event.getModifierState('CapsLock')` and use `if shift` instead of `if event.shiftKey` to also handle Caps Lock.
- While it is possible to detect Num Lock using `event.getModifierState('NumLock')`, I’m not sure if it is worth the trouble. No VimFx users have complained about that their keyboard layout isn’t ignored for the numpad for two years at least. (I don’t have a keyboard with a numpad available right now, so I can’t test what the logic _would_ be either. It might also differ between OS:es.)

PR #2629, opened 2017-09-06
Only handle trusted events.

### LINKED ISSUE #2453: Key handling issue
From the Chrome store...

snapshot

smblott-github:
... otherwise, the page can generate events which trigger Vimium commands.

We should possibly/probably do this on all event handlers.  However, most of the others seem safe.

Fixes #2453.
Specifically, see https://github.com/philc/vimium/issues/2453#issuecomment-327289851 from @derrotebaron.


mrmr1993:
Is there a reason to use this over #2605?

smblott-github:
> Is there a reason to use this over #2605?

Nope.  That looks better.

PR #2636, opened 2017-09-12
Consider devicePixelRatio when calculating viewport

### LINKED ISSUE #2620: Labels fall out of sync on infinite scroll websites
Chrome: 62.0.3192.0 (Official Build) dev (64-bit)
Vimium: 1.59

An example of this can be seen when using Reddit Enhancement Suite and scrolling down, if RES lazy loads the next page into the stream it makes the positioning of tags and labels erroneous

### LINKED ISSUE #2633: Links rendered when display is scaled are not rendered in correct place
When scaling the display in my OS it causes the links to be rendered incorrectly. I tested this on both Windows 10 and Ubuntu 16.04.

Chrome Version: **Version 61.0.3163.79 (Official Build) (64-bit)**

See this GIF for an example:

Chrome example on Ubuntu when scaling display up

### LINKED ISSUE #2635: Wrong link hint positions on HiDPI screens
Explanation:
Just setup chromium/vimium on my new notebook (2560x1440 screen).
I use the "--force-device-scale-factor=2" cmdline option to scale the browser.
However i get wrong positions of the link hints (when pressing 'f').
Link hint posistions are initially right where they should be, but "drift up" when the page is scrolled down.

How to reproduce:
Start chromium with "--force-device-scale-factor=2".
Install vimium (if not already).
Open some site.
Scroll down a bit.
Press 'f'.
See the wrong link hint positions.
(See attached images for expanation, further down pictures represent further "scrolled-down" states).

Further explanation:
Use either "--force-device-scale-factor=2"
OR
set Xft.dpi: 192
in ~/.Xresources
(this is how i discovered it, nice option to make GUIs scale and chromium reads this value)
(both have the same effect, but for reproduction the bug, option number 1 is simpler)

My setup:
Arch x86_64 newest.
Chromium from official repos, newest (Version 61.0.3163.79 (dev-build) (64-Bit))

mrmr1993:
This should fix #2635, fix #2633 and fix #2620.

/cc @smblott-github @gdh1995

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -293,7 +293,8 @@ DomUtils =
     style = getComputedStyle box
     if style.position == "static" and not /content|paint|strict/.test(style.contain or "")
       zoom = +style.zoom || 1
-      top: Math.ceil(window.scrollY / zoom), left: Math.ceil(window.scrollX / zoom)
+      ratio = window.devicePixelRatio ? 1
+      top: Math.ceil(window.scrollY * ratio / zoom), left: Math.ceil(window.scrollX * ratio / zoom)
     else
       rect = box.getBoundingClientRect()
       top: -rect.top - box.clientTop, left: -rect.left - box.clientLeft


smblott-github:
I should just merge this without thinking too much, yes?

mrmr1993:
> merge this without thinking too much

I don't know much about HiDPI browsers or `devicePixelRatio` (or this part of the code), so it's probably worth a quick sanity check, just in case.

gdh1995:
Sorry again for my errors, but this PR is not enough yet - it causes error offsets in such a test:
* on a screen with a `devicePixelRatio` of 1
* use Chrome stable x64, 60.0.3112.113
* zoom the page into 150% using `ctrl` and `+`, and then `devicePixelRatio` is 1.5
* scroll down a page (such as this PR)
* all hints has wrong offsets.

PR #2649, opened 2017-09-14
Always use getBoundingClientRect for getViewportTopLeft

mrmr1993:
With this PR, we end up working with absolute positions and pixel offsets for everything in link hints, so nothing needs to be scaled.

This should fix all of the link hint zoom/scaling issues. It passes all the testcases I could think of.

Ping @gdh1995 @smblott-github.

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -292,13 +292,18 @@ DomUtils =
     box = document.documentElement
     style = getComputedStyle box
     rect = box.getBoundingClientRect()
-    {clientTop, clientLeft} = box
     if style.position == "static" and not /content|paint|strict/.test(style.contain or "")
       # The margin is included in the client rect, so we need to subtract it back out.
       marginTop = parseInt style.marginTop
       marginLeft = parseInt style.marginLeft
       top: -rect.top + marginTop, left: -rect.left + marginLeft
     else
+      if Utils.isFirefox()
+        # These are always 0 for documentElement on Firefox, so we derive them from CSS border.
+        clientTop = parseInt style.borderTopWidth
+        clientLeft = parseInt style.borderLeftWidth
+      else
+        {clientTop, clientLeft} = box
       top: -rect.top - clientTop, left: -rect.left - clientLeft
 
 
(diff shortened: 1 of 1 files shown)


gdh1995:
Good job! In my tests it also works well with various `documentElement`\'s styles including margin, zoom, and border.

smblott-github:
Thanks, @mrmr1993.

I'll try whatever tests I can think of later this morning, and push as 1.60.3, if all seems well.

smblott-github:
Looks great to me.  Thanks again, @mrmr1993.

I pushed 1.60.3.

Fingers crossed.

There some other issues from 1.60 which we need to look at, but nothing as systematic as this hint-positioning thing.

wchargin:
Hi there,

Just want to note that, like the original fix to link hint positioning, this _breaks_ link-hint positioning on Chrome <61. That is, every recent version of Vimium has had link-hint positioning working in exactly one of Chrome {60,61}.

With that said, please consider keeping this patch in. If you can find a way for this to work in both versions of Chrome, fantastic—but I've already changed Chrome versions from 60 to 61 back to 60 and now to 61 again, just to get link hints rendering correctly in Vimium. This is pretty annoying, and it'd be great to not have to downgrade once again if this is rolled back.

Thank you! <3

gdh1995:
@wchargin Could you provide a URL of those pages where Vimium's hints are broken on Chrome 60? I've tested a complexed page on Chrome 60.0.3112.113 (stable, x64) on Win 10 x64, but no errors were found.

wchargin:
It was broken on any page on which you zoomed in. For instance, https://news.ycombinator.com/.

However, after downgrading again to the same version of Chrome on which it was previously broken (60.0.3112.90 (Official Build) (64-bit) for Linux x64), and with the same version of Vimium (1.60.3), I'm not able to reproduce this. I don't understand this: I'm absolutely certain that these two versions were previously broken at the time that I wrote the comment, so there must be some factor in play that I'm missing. Hacker News (link above) is the page on which I noticed the issue and is what triggered me to write the comment. The only other remotely related system upgrade that I installed in the meantime was adobe-flashplugin from 1:20170808.1-0ubuntu0.16.04.1 to 1:20170912.1-0ubuntu0.16.04.1, but I assume that that shouldn't have an impact as Vimium doesn't use Flash.

PR #2653, opened 2017-09-15
Use .textContent instead of .innerHTML, where appropriate.

smblott-github:
From a FF extension reviewer (gatekeeper) on Mozilla add-ons:

  Comments:
  Thank you for your contribution.

  Please note the following for the next update:
  1) For inserting text, textContent (or JQuery text) or createTextNode() should be used instead of innerHTML.
  eg: $(&quot;saveOptions&quot;).innerHTML = &quot;Save Changes&quot;;

So, here, `.innerHTML` is replaced with `.textContent` for elements which only ever have text content.


mrmr1993:
Nice 👍

Issue #2459, opened 2017-03-25
Feature request: Vomnibar add tab searching & priority


rtgoodwin:
Vomnibar is fantastic, and already covers URLs, history, and bookmarks. But "T" is still required to search your open tabs. It would be great if:

1. tabs could be searched in Vomnibar
2. you could assign a priority to Vomnibar search. Example, "tabs, bookmarks, history" vs "history, bookmarks, tabs". This may not be needed if "tabs" was an option to add, if it had priority over history. But, as with all things, giving the user choice is often best. 

Quick Tabs has idea 1 and I find it reduces the mental load when I can't remember if I have something open in my 90 tabs or if it's off in a history or bookmark (which may already be open...now I have 91 tabs :) ). I use Great Suspender to suspend tabs and reduce Chrome memory, so it's quite often that I'll have a tab open already and have forgotten about it. 

It seems probably trivial, but I couldn't quite grasp the source enough to make a PR.  Thanks to everyone who contributes to this amazing extension!

mklinga:
I would love to see priority option on the vomnibar. Often I just press `O` instead of `B` and all I can find is history results from a page that I have also bookmarked, so I need to `<Esc>` out and press `B`.

zakvdm:
I think the single place where you can search all things, be it currently open tabs or bookmarks/history is very powerful from a usability perspective.

philc:
With #2656, we'll now show tab search results in regular Vomnibar, just like Chrome's Omnibar. That will go out in the next release of Vimium. Thanks for filing the idea!

PR #2656, opened 2017-09-16
Show tabs in the results when using the 'o' vomnibar

M-frankied:
Related to issue #2459.

Addresses a workflow like the following:
- Lots of tabs with e.g. issue tracker tickets, or admin web apps for lots of sites.
- I want to work on ticket x or site y, but I don't want x or y open in multiple tabs
- press 't', search for tab, not found
- close vomnibar
- press 'O', search for ticket or site

Getting access to the tab (if any) or otherwise the history item, bookmark, or search result from one vomnibar interaction is smoother here.

Changed the tab completer to have a quick weight falloff for subsequent tab entries to avoid crowding out other completion sources. This should have no influence on the workings of 't'.

### DIFF
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -364,19 +364,22 @@ class TabCompleter
     chrome.tabs.query {}, (tabs) =>
       results = tabs.filter (tab) -> RankingUtils.matches(queryTerms, tab.url, tab.title)
       suggestions = results.map (tab) =>
-        new Suggestion
+        suggestion = new Suggestion
           queryTerms: queryTerms
           type: "tab"
           url: tab.url
           title: tab.title
-          relevancyFunction: @computeRelevancy
           tabId: tab.id
           deDuplicate: false
+        suggestion.relevancy = @computeRelevancy suggestion
+        suggestion
+      .sort (a,b) -> b.relevancy - a.relevancy
+      suggestions.forEach( (sugg,i) -> sugg.relevancy /= ( i / 4 + 1 ) )
       onComplete suggestions
 
   computeRelevancy: (suggestion) ->
     if suggestion.queryTerms.length
-      RankingUtils.wordRelevancy(suggestion.queryTerms, suggestion.url, suggestion.title)
+      RankingUtils.wordRelevancy(suggestion.queryTerms, suggestion.url, suggestion.title) * 8.0
     else
       BgUtils.tabRecency.recencyScore(suggestion.tabId)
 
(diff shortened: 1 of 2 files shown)


quintrino:
If I'm reading this correctly and it means that I can access both my open tabs and history/bookmarks from a single bar I'm very excited.

deepredsky:
This is exactly what i have been looking for

background_scripts/completion.coffee
```diff
@@ -364,19 +364,22 @@ class TabCompleter
...
           type: "tab"
           url: tab.url
           title: tab.title
-          relevancyFunction: @computeRelevancy
           tabId: tab.id
           deDuplicate: false
+        suggestion.relevancy = @computeRelevancy suggestion
+        suggestion
+      .sort (a,b) -> b.relevancy - a.relevancy
+      suggestions.forEach( (sugg,i) -> sugg.relevancy /= ( i / 4 + 1 ) )
```
smblott-github: This computation requires comment explaining what and why.

Separately, elsewhere we use `suggestion` (not `sugg`) for an individual suggestion.

background_scripts/completion.coffee
```diff
@@ -364,19 +364,22 @@ class TabCompleter
...
       onComplete suggestions
 
   computeRelevancy: (suggestion) ->
     if suggestion.queryTerms.length
-      RankingUtils.wordRelevancy(suggestion.queryTerms, suggestion.url, suggestion.title)
+      RankingUtils.wordRelevancy(suggestion.queryTerms, suggestion.url, suggestion.title) * 8.0
```
smblott-github: The `* 8.0` here requires a comment/explanation.  Is this just a heuristic?

j-martin:
@smblott-github is there are any plan on merging this?

philc:
I just noticed that Chrome added support for this at some point, so that provides some additional rationale for mirroring their direction.

image

I'll test this out. Note that #2878 was just merged and is related.

philc:
Here's what the current UX looks like on this PR. The new tab favicons look OK to me. The favicons help the tab results stand out more, which doesn't seem particularly negative. CC @gdh1995, @smblott-github for any additional thoughts.

image

My main concern is whether the tabs often overwhelm the rankings. I'll use this for a few days to get a feel for it.

gdh1995:
To make it look better, we may give a same margin-left to URLs of history/bookmark items.

And now I have another idea to move icons before titles - currently my Vimium C looks like this:
image

Qquanwei:
I like this feature so much . @philc Can we release this feature as soon as possible ?

j-martin:
@philc thanks for looking into this. Worst come to worst, I think it could be option to show the tabs or not.

M-frankied:
thanks @philc, and everyone contributing to this thread

PR #2664, opened 2017-09-23
Search completion; add support for Amazon.co.jp completion

curipha:


smblott-github:
Thanks for this and the other one, @curipha.

PR #2688, opened 2017-10-01
Use Mocha instead of Shoulda.js for tests

mrmr1993:
This makes the bare minimum amount of changes to move from @philc's [shoulda.js](https://github.com/philc/shoulda.js) to [Mocha](https://mochajs.org) for tests, as part of my master plan to move to in-browser testing. This gives us a few advantages:
* Mocha supports asynchronous tests out-of-the-box, so we can remove the hacks we've been using to make DOM tests synchronous
* we can move to asynchronous testing methods
  - this lets us use selenium to test the extension in real Chrome and Firefox, on real pages
  - this would also let us create testcases for issues far more easily, and even let us test for browser/version specific issues
* we can use anything that would be useful from the Mocha ecosystem
  - it supports a large number of assertion libraries
  - we have choice over [test result output](https://mochajs.org/#reporters), and it's easy to write our own if we need it
  - there's plenty of tooling for it (e.g. `mocha-phantomjs`)

This PR itself has some disadvantages:
* it depends on `mocha-phantomjs` to run the DOM tests, rather than normal `phantomjs`.
  - I plan for this to get rid of this when I've rewritten the DOM tests using selenium
* it depends on `assertion` to make assertions in the DOM tests.
  - this effectively reimplements the [Node.js assert library](https://nodejs.org/api/assert.html) without being Node specific
  - we need this because we can't use Node's own libraries in-browser
  - if we move to a different assertion library, we won't need this any more
* the DOM tests require a file from the unit tests folder, since it sets up the polyfill for the shoulda.js calls
  - this is easily fixed, but I would prefer to convert the tests to use Node's assert (or another library) directly
* ~~a couple of (unit) tests are failing that weren't under shoulda.js.~~ **Fixed.**
  - ~~I haven't investigated this; it's possible that something wasn't running properly before, or that I've broken them in the conversion~~

@philc @smblott-github is this a direction worth pursuing?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "tests/shoulda.js"]
-	path = tests/shoulda.js
-	url = https://github.com/philc/shoulda.js.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ before_install:
   - "npm install -g coffee-script"
   - "npm install path@0.11"
   - "npm install util"
+  - "npm install mocha"
   - "npm install -g phantomjs@1.9.20"
   - "cake build"
 script: "cake test"
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ before_install:
   - "npm install path@0.11"
   - "npm install util"
   - "npm install mocha"
+  - "npm install assertion"
   - "npm install -g phantomjs@1.9.20"
   - "npm install mocha-phantomjs"
   - "cake build"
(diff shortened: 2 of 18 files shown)


smblott-github:
> is this a direction worth pursuing?

Very definitely yes!  I'll look at this soon.

Thanks for taking the initiative on this.

philc:
Thanks for the substantial proof of concept @mrmr1993. Regarding selenium tests, at the moment, I don't think the tradeoff of more realistic testing environments from in-browser testing vs.  the additional complexity is worth it.

As an aside, shoulda.js now uses a similar syntax as Mocha.

PR #2701, opened 2017-10-04
Allow using capital letters as link hint characters

### LINKED ISSUE #2414: Feature request: Use the link's name and capital letters for link-hint filtering
This is a request for a feature similar to *"Use the link's name and numbers for link-hint filtering"* with a difference that it would be possible to use capital letters instead of numbers to follow hints. I didn't find any existing setting for that and setting only capital letters as characters used for link hints didn't give expected results.

How this feature would work is that typing in lower case would follow links by text matching and use of capital letters would follow links by hints. It is the exact behavior that I am using in [VimFx](https://github.com/akhodakivskiy/VimFx) extension for Firefox.

ahstro:
I'm very new to the code base, so please double check so that things make sense. Also, I didn't write any tests because they didn't pass when I cloned `master`, and there's a pretty big PR related to them. I'd be willing to (try and) write some now or later if desired.

Closes #2414

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -239,8 +239,11 @@ class LinkHintsMode
     # NOTE(smblott) As of 1.54, the Ctrl modifier doesn't work for filtered link hints; therefore we only
     # offer the control modifier for alphabet hints.  It is not clear whether we should fix this.  As of
     # 16-03-28, nobody has complained.
-    modifiers = ["Shift"]
-    modifiers.push "Control" unless Settings.get "filterLinkHints"
+    modifiers =
+      if Settings.get "filterLinkHints"
+        []
+      else
+        ["Control", "Shift"]
 
     if event.key in modifiers and
       @mode in [ OPEN_IN_CURRENT_TAB, OPEN_WITH_QUEUE, OPEN_IN_NEW_BG_TAB, OPEN_IN_NEW_FG_TAB ]
@@ -288,8 +291,13 @@ class LinkHintsMode
 
     else
       @tabCount = previousTabCount if event.ctrlKey or event.metaKey or event.altKey
+      keyChar =
+        if Settings.get "filterLinkHints"
+          KeyboardUtils.getKeyChar(event)
+        else
+          KeyboardUtils.getKeyChar(event).toLowerCase()
       unless event.repeat
-        if keyChar = KeyboardUtils.getKeyChar(event).toLowerCase()
+        if keyChar
           keyChar = " " if keyChar == "space"
           if keyChar.length == 1
             @markerMatcher.pushKeyChar keyChar
@@ -473,7 +481,7 @@ class AlphabetHints
   # For alphabet hints, <Space> always rotates the hints, regardless of modifiers.
   shouldRotateHints: -> true
 
-# Use numbers (usually) for hints, and also filter links by their text.
+# Use characters for hints, and also filter links by their text.
 class FilterHints
   constructor: ->
     @linkHintNumbers = Settings.get "linkHintNumbers"
--- a/pages/options.html
+++ b/pages/options.html
@@ -115,11 +115,11 @@ b: http://b.com/?q=%s description
             </td>
           </tr>
           <tr id="linkHintNumbersContainer">
-            <td class="caption">Numbers used<br/> for link hints</td>
+            <td class="caption">Characters used<br/> for link hints</td>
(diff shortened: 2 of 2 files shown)


ahstro:
Ah, yes, I see your point. I thought it reasonable to mergw them, since it would no longer be restricted to numbers.

Do you suggest keeping `linkHintNumbers` but allowing non-numeric numbers in there?

mrmr1993:
Yes, I think so.

smblott-github:
> Do you suggest keeping linkHintNumbers but allowing non-numeric numbers in there?

Yes.

ahstro:
Well, that simplified the PR quite a bit 😅

mrmr1993:
You could tweak the comment by the filter hints option to document this feature, if you wanted to bulk up the PR.

ahstro:
There, one more line 😉

mrmr1993:
I meant [here](https://github.com/philc/vimium/blob/9da17bfd35f5e97bc4e6949debbaae073bc367e5/pages/options.html#L122), but the other tweak is nice to have too.

content_scripts/link_hints.coffee
```diff
@@ -289,7 +292,7 @@ class LinkHintsMode
     else
       @tabCount = previousTabCount if event.ctrlKey or event.metaKey or event.altKey
       unless event.repeat
-        if keyChar = KeyboardUtils.getKeyChar(event).toLowerCase()
+        if keyChar = KeyboardUtils.getKeyChar(event)
```
mrmr1993: We still want `toLowerCase` applied when `Settings.get "filterLinkHints"` is falsey, otherwise it breaks <kbd>shift</kbd> for 'alphabet hints'.

content_scripts/link_hints.coffee
```diff
@@ -288,8 +291,13 @@ class LinkHintsMode
 
     else
       @tabCount = previousTabCount if event.ctrlKey or event.metaKey or event.altKey
+      keyChar =
+        if Settings.get "filterLinkHints"
+          KeyboardUtils.getKeyChar(event)
+        else
+          KeyboardUtils.getKeyChar(event).toLowerCase()
```
ahstro: Not sure if duplicating `KeyboardUtils.getKeyChar(event)` was the most elegant solution, but it felt better than adding another variable just for keeping that value.

mc10: What about
```
keyChar = KeyboardUtils.getKeyChar event
unless Settings.get "filterLinkHints"
  keyChar = keyChar.toLowerCase()
```

ahstro: Not a huge fan of mutation, but if you think that's better, I'll change it?

smblott-github:
Do we not need a `toUpperCase()` somewhere for the hint character?  Otherwise, if the user sets lower-case hint characters, then some characters match link text and some are hints.  And the user cannot just type text for the link.

Also, setting upper-case hint characters on the options page, then it's weird that some shifted characters match hints which others match the link text.

smblott-github:
(Sorry.  Should have thought longer before posting.)

Would it be useful to have HUD feedback of the special meaning when `<Shift>` is depressed?

ahstro:
I did have some `toUpperCase()` logic back when I replaced `linkHintNumbers`; I could bring that back if you'd like? The way I did it was upper-cased `linkHintNumbers` on `filterLinkHints` toggle and options save. I left it out now because I didn't mind if people want to have some letters be filters and some be hints.

I'm not sure what you mean with the HUD feedback nor why that would help? One thing that could be beneficial would be that the hints reflected the case of the hints in settings, so that, if you're using non-filtering hints, the hits would be lower case?

That felt like I rambled a bit, hope you understand what I'm getting at. 🙂

smblott-github:
OK.

I'll take this a tweak it.

smblott-github:
I propose doing this [like this](https://github.com/smblott-github/vimium/tree/ahstro/feature/upper-case-link-hints); specifically, this [commit](bc8df497670658fdafa26660ee09987cf9d969ae).

From that commit:

1. Simplify the logic for calculating `modifiers`.
2. Revert the ordering of the `event.repeat` test and `keyChar` assignment.
3. Link hint characters *only* match in upper case.
4. If there are link-hint characters (non numbers), then shifted characters are only ever hints.
5. Otherwise, revert to doing all hint text comparison in lower case.

smblott-github:
Merged.  Closing.

PR #2713, opened 2017-10-08
Make link hints all equal length

eyqs:
In link-hint mode, I'm used to links opening immediately after I enter the last number. When there are more than ten links on the page, and I type "1", it always takes me a few seconds to realize that I then need to press enter.

This pull request fixes this by making all link hints the same length with leading 0s. The only downside I can see is that on a page with more than a hundred (!) links, typing "002" takes one extra keystroke compared to typing "2<Enter>". I think the frustration of remembering to press enter far outweighs this.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -17,6 +17,7 @@ Contributors:
   ConradIrwin
   Daniel MacDougall <dmacdougall@gmail.com> (github: dmacdougall)
   drizzd
+  Eugene Y. Q. Shen <eugene@eyqs.ca> (github: eyqs)
   gpurkins
   hogelog
   int3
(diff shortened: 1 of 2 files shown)


smblott-github:
I don't think that this is a great idea, @eyqs.  It's a pretty big change to the UI, with not that much evidence that the status quo is problematic.

eyqs:
I propose we leave this pull request open for now. The change is just an extra leading zero when you have more than 10 link hints, or two extra leading zeroes when you have ore than 100, which seems very minor to me.

philc:
Closing due to inactivity, and I agree with @smblott-github.

PR #2717, opened 2017-10-11
[POC] Selenium tests

mrmr1993:
This is a follow-on from #2688. Included are:
* basic infrastructure for running the tests via `cake`
* instructions for running the tests in `CONTRIBUTING.md`
* the first few link hints tests from the DOM tests converted for Selenium
* tests running in Chrome and Firefox
* simple option to use Xvfb, so that we can support Travis

Weird things:
* we can't get the extension ID when the browsers start, so we need the extension to open a page from `chrome-extension://` or `moz-extension://` for us.
  - we only get access to Chrome extension APIs via these pages using Selenium, so we want at least one
* Firefox can't open `file://` URLs, so we have to open them from Selenium, once the extension has opened its page(s) (so we don't race)
  - these pages have their content scripts injected as usual, so running tests on these is more authentic (as opposed to `chrome-`/`moz-extension://` pages)
* Firefox tests fail (timeout in initialisation) on Travis.
* there's a link hints-specific harness page that I accidentally included here.
* the diff here includes things from #2688, so it may be more helpful to look at [this diff](https://github.com/mrmr1993/vimium/compare/mocha-tests...mrmr1993:mocha-selenium-tests?expand=1#files_bucket) instead.

This isn't ready for merge, but it's a direction that I'd like us to move in.


mrmr1993:
Updated the title/description. @smblott-github what do you think?

philc:
Cool POC. I don't plan to leverage Selenium -- we now have Puppeteer and I plan to either modernize shoulda.js or use mocha, which should hit at least 80% of our testing use cases.

PR #2729, opened 2017-10-16
Updated Readme

dkrprogramming:


smblott-github:
That's branding, @dkrprogramming.  It's repeated in lots of other places.  So best not change it in just one of those places.

(Thanks anyway!)

PR #2733, opened 2017-10-19
Minimal Firefox Android support

### LINKED ISSUE #2662: Support Firefox on Android
It would be very nice to use vimium on my android tablet with a bluetooth keyboard.  Using firefox nightly, vimium loads fine and the configuration options are available, but it doesn't respond to any key presses.  Is there a simple solution to getting it working or will it require a significant effort?

mrmr1993:
`chrome.windows` and `chrome.browserAction.setIcon` [aren't supported by Firefox on android](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Differences_between_desktop_and_Android). By adding guards to these in the code, Vimium then works (or, at least, doesn't crash) on Firefox for Android.

I've only added guards where the code runs as expected on android without the APIs. Where code *needs* to access `chrome.windows` I've left it unguarded, so that we get errors that might be useful for debugging later on. E.g. opening an incognito window will still throw an error, since we're unable to call `chrome.windows.create`. None of these are in places where throwing will crash the extension.

This fixes #2662, in the loosest sense of the word.

### DIFF
--- a/background_scripts/bg_utils.coffee
+++ b/background_scripts/bg_utils.coffee
@@ -18,7 +18,7 @@ class TabRecency
       @deregister removedTabId
       @register addedTabId
 
-    chrome.windows.onFocusChanged.addListener (wnd) =>
+    chrome.windows?.onFocusChanged.addListener (wnd) =>
       if wnd != chrome.windows.WINDOW_ID_NONE
         chrome.tabs.query {windowId: wnd, active: true}, (tabs) =>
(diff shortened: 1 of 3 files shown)


smblott-github:
@mrmr1993.  I'll merge this because it is minimally disruptive.  However, I'm nervous about claiming FF on Android support.  Supporting desktop FF is enough hassle, but worth it for the *big* user base.  I'm not sure FF on Android is a priority, though.

PR #2740, opened 2017-10-24
Setting the funtion to hard reload mapped to letter R

krlooss:


smblott-github:
Thanks, @krlooss, but we actually got this already with #2738.

PR #2748, opened 2017-10-27
Move normal mode and its commands out of vimium_frontend.coffee

mrmr1993:
`vimium_frontend.coffee` has become quite large and disorganised, with a lot of things that could be separated out. This PR moves the `NormalMode` constructor and its commands into their own file, shortening `vimium_frontend.coffee` by over 200 LOC.

Note:
* All normal mode commands now live in the global object `NormalModeCommands`.
  - ~~We could use this to get rid of `Utils.invokeCommandString`, and call the commands from the object directly.~~ **DONE**
  - This would also free us up to rename/refactor the code without worrying about breaking the command names. (E.g. Link hints could benefit from this.)
* Some obviously find mode-related functions have been moved into `mode_find.coffee`
* There are now tests to check that all of the front-end commands listed in `background_scripts/commands.coffee` are implemented in `NormalModeCommands`.
  - Extending this to background commands should be simple enough, but I didn't want to rush into including `background_scripts/main.coffee` (and all the necessary stubs) in the command tests.
* The `focusInput` command's set-up code and its mode have been separated out and left in `vimium_frontend.coffee` (as `lastFocusedInput` and `FocusSelector`).

### DIFF
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -438,9 +438,6 @@ selectFoundInputElement = ->
       DomUtils.isDOMDescendant(findModeAnchorNode, document.activeElement))
     DomUtils.simulateSelect(document.activeElement)
 
-performFind = (count) -> FindMode.findNext false for [0...count] by 1
-performBackwardsFind = (count) -> FindMode.findNext true for [0...count] by 1
-
 getLinkFromSelection = ->
   node = window.getSelection().anchorNode
   while (node && node != document.body)
@@ -543,11 +540,6 @@ root.goNext = ->
   nextStrings = nextPatterns.split(",").filter( (s) -> s.trim().length )
   findAndFollowRel("next") || findAndFollowLink(nextStrings)
 
-# Enters find mode.  Returns the new find-mode instance.
-enterFindMode = ->
-  Marks.setPreviousPosition()
-  new FindMode()
-
 root.showHelp = (sourceFrameId) ->
   HelpDialog.toggle {sourceFrameId, showAllCommandDetails: false}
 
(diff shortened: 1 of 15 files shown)


smblott-github:
@mrmr1993... We should do this.

> We could use this to get rid of `Utils.invokeCommandString`, and call the commands from the object directly.

And we should do this too.

I just looked the code over once.  It's big, so I'll have to look it over again.

Is it good to go from your perspective?

mrmr1993:
> > We could use this to get rid of `Utils.invokeCommandString`, and call the commands from the object directly.

> And we should do this too.

I've pushed some commits to do this.

> I just looked the code over once. It's big, so I'll have to look it over again.

If you haven't been already, I would recommend going commit by commit. I tried to keep each commit fairly straightforward, so anything that's been copied wrong etc. is easy to catch.

> Is it good to go from your perspective?

Yes. I've been running this for the last few days and everything seems to work as normal.

content_scripts/vimium_frontend.coffee
```diff
@@ -705,8 +484,8 @@ root.Frame = Frame
 root.windowIsFocused = windowIsFocused
 root.bgLog = bgLog
 # These are exported for find mode and link-hints mode.
-extend root, {handleEscapeForFindMode, handleEnterForFindMode, performFind, performBackwardsFind,
-  enterFindMode, focusThisFrame}
+extend root, {focusFoundLink, selectFoundInputElement, focusThisFrame, FocusSelector,
```
smblott-github: @mrmr1993 Might it not be better to move these to `mode_normal.coffee` too?

They only exist to serve normal mode, do they not?

mrmr1993: `focusFoundLink` and `selectFoundInputElement` are now only used in `mode_find.coffee`, so they could go there.

`focusThisFrame` is also used by link hints. `FocusSelector` does a similar job to other focus trackers, so I was planning to try and merge them into a kind of `FocusOracle` that other code calls out to.

mrmr1993:
@smblott-github I've moved all of those exports into the relevant places. Last remaining holdout is `lastFocusedInput`, because it adds an event listener at definition time and frontend feels like the place that should happen.

smblott-github:
Just stop with whatever you think is best.

mrmr1993:
Tests now pass again. This PR is good to go from my end!

smblott-github:
OK.  This is very big, so it's hard to check that every line ended up in the right place.  Hopefully we'll pick up any mistakes pretty quickly, if there are any.

Thanks.

PR #2750, opened 2017-10-28
Add backup/restore for Vimium options.

### LINKED ISSUE #1440: [Feature request] Ability to import/export settings
I know Vimium can sync its settings, but that requires users to login in Chrome, leaving people who don’t do it without a way to easily keep custom configurations. I currently have three text files with my customisations — excluded URLs, CSS for styling, and key mappings, which have to be saved and reapplied manually (plus a few others that rely on checkboxes).

A way to export and import all settings, even if it just means saving everything to a plain accessible file (so at least we could automate moving it to the correct location) would be very welcome.

smblott-github:
See the *very* bottom of the options page (below advanced settings).

Clicking "Backup" creates a JSON file.

Selecting a backup populates the options inputs, the user then clicks *Save Changes* to confirm.

Looks like this:

snapshot

@mrmr1993 I've always been reticent about doing this in the past, mainly concerned that some combination of backup restore and a change to `Settings` could goof things up.

What do you think?  Also, should we add `settingsVersion` (or anything else) to the backup?

Fixes #1440.

### DIFF
--- a/lib/settings.coffee
+++ b/lib/settings.coffee
@@ -202,7 +202,7 @@ Settings.init()
 
 # Perform migration from old settings versions, if this is the background page.
 if Utils.isBackgroundPage()
-  Settings.onLoaded ->
+  Settings.applyMigrations = ->
     unless Settings.get "settingsVersion"
       # This is a new install.  For some settings, we retain a legacy default behaviour for existing users but
       # use a non-default behaviour for new users.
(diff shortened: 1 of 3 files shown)


mrmr1993:
> What do you think?

Code looks good, and works nicely on Chrome.

The create backup button is broken in Firefox: clicking it navigates to `about:blank`. The best "workaround" I can come up with is 09f2d27d1c29d55ad34d3aa0b2a25296b37ec2fd. Restoring from backup works as expected.

> Also, should we add `settingsVersion` (or anything else to the backup?

Having `settingsVersion` would be nice; I can't think of anything else that would be very useful.

PR #2751, opened 2017-10-29
Fix backup link for FF and Chrome.

smblott-github:
This makes the UI for generating an options backup the same for FF and Chrome.

We populate the *Backup* link on `mousedown`.

@mrmr1993 This seems to make the UI as similar as possible between Chrome and Firefox.

### DIFF
--- a/pages/options.coffee
+++ b/pages/options.coffee
@@ -333,8 +333,7 @@ document.addEventListener "DOMContentLoaded", ->
 #
 # Backup and restore. "?" is for the tests."
 DomUtils?.documentReady ->
-  $("backupButton").addEventListener "click", ->
-    document.activeElement?.blur()
+  populateBackupLinkUrl = ->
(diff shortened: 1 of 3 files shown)


mrmr1993:
I don't think this will work for Firefox, since it doesn't seem to respect the `download` attribute. I already tried something similar and had to reject it.

smblott-github:
I understand this to be the same as b1f6da21a882f84965de73eea175475f5d8f7b37, only without the need for the extra step of clicking a button to generate the link.

b1f6da21a882f84965de73eea175475f5d8f7b37 and this on FF both dump you in a page that looks like this:

snapshot

On Chrome, the download happens immediately.

mrmr1993:
Can you right click and save link as?

smblott-github:
Yes.

mrmr1993:
Okay, sorry for the noise then.

PR #2753, opened 2017-10-29
Always suppress keyups for keydowns that we handle, enforced in handlerStack

mrmr1993:
This PR tweaks the key handling so that we always suppress `keyup`s for `keydown`s we handle, and also don't suppress `keyup`s for `keydown`s we *didn't* handle. This led to bugs like #1522 in the past, and `keyup`s from number keys handled by `KeyHandlerMode` still leak in current `master`.

I've also removed some dead code (a `blur` listener that resets an unused variable) from `KeyHandlerMode`, and `passInitialKeyupEvents`, which we no longer need.

For this to work properly going forward, `keyup` handlers should avoid return values that skip the remaining handlers in the stack. Since we only really need to monitor `keyup` rather than handle it, hopefully this shouldn't be a problem.

### DIFF
--- a/lib/handler_stack.coffee
+++ b/lib/handler_stack.coffee
@@ -65,7 +65,11 @@ class HandlerStack
           true # Do nothing, but continue bubbling.
         else
           # result is @suppressEvent or falsy.
-          DomUtils.suppressEvent event if @isChromeEvent event
+          if @isChromeEvent event
+            if type == "keydown"
+              DomUtils.consumeKeyup event
+            else
+              DomUtils.suppressEvent event
           return false
 
     # None of our handlers care about this event, so pass it to the page.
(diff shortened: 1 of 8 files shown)


lib/handler_stack.coffee
```diff
@@ -65,7 +65,11 @@ class HandlerStack
           true # Do nothing, but continue bubbling.
         else
           # result is @suppressEvent or falsy.
-          DomUtils.suppressEvent event if @isChromeEvent event
+          if @isChromeEvent event
+            if type == "keydown"
+              DomUtils.consumeKeyup event
+            else
+              DomUtils.suppressEvent event
```
smblott-github: Help me understand this.

Here, lines 68 to 72, is the key idea of this PR.  We take the responsibility of deciding what to do with `keyup` out of higher-level modes, and into the handler stack.  If a `keydown` event was handled by Vimium, then the corresponding `keyup` event is blocked.  Otherwise it isn't.

That's it, yes?

mrmr1993:
That's exactly right, yes. In fact, this PR could have been just this and a couple lines that ignore the return value for `type =="keyup"`, to the same effect.

Is it add this (and warning when we do) to stop

PR #2762, opened 2017-11-02
Add some roles used by GMail to link hints

mrmr1993:
This PR also reduces the number of comparisons needed for (the vast majority of) elements which have no `role` or `contentEditable` attribute.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -658,9 +658,12 @@ LocalHints =
     isClickable ||= @checkForAngularJs element
 
     # Check for attributes that make an element clickable regardless of its tagName.
-    if (element.hasAttribute("onclick") or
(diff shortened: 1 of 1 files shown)


gdh1995:
This code of testing contentEditable contains a long term bug: `[contenteditable]` does not accept a value of `"contentEditable"`.

A test page is https://jsfiddle.net/erLq3a5u/ , and its document can be found in https://html.spec.whatwg.org/multipage/interaction.html#attr-contenteditable.

BTW, `[contenteditable].toLowerCase()` will never be `contentEditable` because it includes a char `E`.

mrmr1993:
Nice! Sorry I missed this (and wrote it in the first place). @smblott-github could you push a fix straight to master?

smblott-github:
I pushed this: e3ca00a9f6402730b21a059a751a1b6f1d9d19ea, but I don't really understand the rest of the comment.

PR #2763, opened 2017-11-02
FF: Don't use Selection.getRangeAt(0) when there are no ranges

### LINKED ISSUE #2757: [Firefox] / sometime opens firefox quicksearch.
Build identifier: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:56.0) Gecko/20100101 Firefox/56.0
Vimium 1.61.1

/ sometime opens firefox quicksearch. Happens quite randomly, after a few tries I always get to the vimium search balloon.

mrmr1993:
This fixes #2757.

### DIFF
--- a/content_scripts/mode_find.coffee
+++ b/content_scripts/mode_find.coffee
@@ -82,6 +82,7 @@ class FindMode extends Mode
     FindMode.handleEscape() if event
 
   restoreSelection: ->
+    return unless @initialRange
     range = @initialRange
     selection = getSelection()
     selection.removeAllRanges()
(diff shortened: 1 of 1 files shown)


gdh1995:
What's the most early version of Firefox does Vimium support? According to MDN, Firefox supports `Selection::type` since ver 57.

Though Vimium has got a polyfill for it (`DomUtils.getSelectionType` in https://github.com/philc/vimium/blob/73b1bd8057ef9b71f12bca5e81dae0103281800b/content_scripts/mode_find.coffee#L236), I find it's using `selection.type` directly in `mode_find.coffee` (https://github.com/philc/vimium/blob/master/content_scripts/mode_find.coffee#L9) and https://github.com/philc/vimium/blob/master/lib/dom_utils.coffee#L222 .

gdh1995:
Then `window.getSelection().type` is still `undefined` (Firefox 56.0.2 stable, Chinese version), so we may need to fix these misuses.

Issue #2769, opened 2017-11-04
Bind shortcut with enter key.


schvabodka-man:
How do i bind shortcut with enter key? For example i want to bind <a-enter> as new tab. I was trying <a-return>, <a-ret>, <a-Enter>, <a-Return>, but nothing worked. I can't find anything useful in documentation about ret key. How to do that?

smblott-github:
This is not currently supported.  (Although I see no reason why it shouldn't be.)

schvabodka-man:
> Although I see no reason why it shouldn't be.

That's very edgy and radical.

schvabodka-man:
But anyway, thx for response

PR #2770, opened 2017-11-04
Add more named keys

smblott-github:
This adds several additional named keys:

    tab
    enter
    delete
    insert
    home
    end

This is partially in response to #2769.

The rationale is: why not?  Just let the user decide what they want.


smblott-github:
> Is there a reason not to do this by default, casting any key without event.key to lower case and then remapping arrows and space as currently?

So, just accept anything `event.key` gives us, except the legacy names?  That sounds good.

mrmr1993:
> accept anything `event.key` gives us

I meant [`event.code`](https://w3c.github.io/uievents-code/#code-value-tables). Only real issue I can see is that it makes modifiers mappable. Is that a problem?

PR #2772, opened 2017-11-05
Add more named keys (version 2)

### LINKED ISSUE #2769: Bind shortcut with enter key.

smblott-github:
This allows any special key to mapped, simply by using the `event.key`
key name. E.g.

    map <enter> scrollDown

@mrmr1993... What would be the benefit of using `.code` over `.key`?

Replaces #2770.
Fixes #2769.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,9 +141,10 @@ The following special keys are available for mapping:
 
 - `<c-*>`, `<a-*>`, `<m-*>` for ctrl, alt, and meta (command on Mac) respectively with any key. Replace `*`
   with the key of choice.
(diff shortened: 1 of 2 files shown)


mrmr1993:
> What would be the benefit of using `.code` over `.key`?

For non-printing keys, I thought it would be better if they're identified the same regardless of keyboard layout. In the general case, they will be anyway, so I suppose it doesn't matter.

PR #2785, opened 2017-11-09
Special-case modifiers now they are resolved as keys

mrmr1993:
This fixes number prefixes for normal and visual modes, which were broken by #2772.

### DIFF
--- a/content_scripts/mode_key_handler.coffee
+++ b/content_scripts/mode_key_handler.coffee
@@ -59,7 +59,7 @@ class KeyHandlerMode extends Mode
       @reset if @keyState.length == 1 then @countPrefix * 10 + digit else digit
       @suppressEvent
     else
-      @reset() if keyChar
+      @reset() unless event.key in ["Control", "Shift", "Alt", "OS", "AltGraph"]
       @continueBubbling
 
   # This tests whether there is a mapping of keyChar in the current key state (and accounts for pass keys).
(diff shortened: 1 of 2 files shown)


mrmr1993:
Added some code which let us disregard modifier keys. This lets us trigger all the mappings as expected where:
* a mapping involving a modifier exists (e.g. `map a<shift>a LinkHints.activateMode`), and
* either
  - a mapping where a modifier needs to be pressed exists (e.g. `map F LinkHints.activateModeToOpenInNewForegroundTab`);
  - a mapping not involving the modifier but with a shared prefix exists (e.g. `map ab LinkHints.activateModeToOpenInNewTab`);
  - or both (e.g. `map aF LinkHints.activateModeWithQueue`).

It also gives priority to a mapping including a modifier over a mapping not including it. So:
* `a<shift>a` gets priority over `aa`
* `a<shift>a` gets priority over `aab`

This is obviously an edge case, but hopefully it is predictable and reasonable behaviour.

smblott-github:
This is a good catch, @mrmr1993, but I'm more inclined to revert to the old behaviour of modifiers in `keyboard_utils.coffee`.  I can see allowing modifiers through coming back to bite us again.

mrmr1993:
I prefer this to the alternative (e.g. #2788), and I find it more honest and flexible to handle them all the same way until we want them to do something different. This also lets us experiment with some niceties like having mode modifiers for alphabet link hints be user configurable (via the options in the mapping).

Obviously the final decision is yours.

smblott-github:
Merged #2788 before I saw your comment here, @mrmr1993.

I think the idea to admit modifiers was kind of accidental, and is most unusual.  We quickly tripped up, and my suspicion is that it will come back and bite us again and again.  The benefit is so marginal that I prefer to #2788.

PR #2791, opened 2017-11-11
Remove WaitForEnter and TypingProtector modes from link hints code

mrmr1993:
Update: Part of this PR has been merged as #2823.

The remaining changes here are:
* `WaitForEnter` mode logic is moved into `LinkHintsMode` main logic: the mode doesn't try to exit until enter is pressed
  - the leaves the HUD showing the link hints contents rather than `Hit <enter> to proceed`
  - the user can 'back out' of the current selection by typing <kbd>backspace</kbd> and select a different link, since they're still in the full link hints mode
* `TypingProtector` mode logic is also moved into `LinkHintsMode` main logic: the mode waits until it seems like the user has stopped typing before trying to exit
  - this means that the link is no longer highlighted as soon as there is only 1 match left, but instead when it is clicked. This is an easy one-line change if we want to keep the old behaviour.
  - again, the user can 'back out' of the current selection (if they are fast enough)

Old description:

------------------------

This PR does some tidying up of link hints. The commit messages give a good summary of everything that this changes.

Most notably, this kills `WaitForEnter` mode. Instead, this PR makes link hints mode stay active until enter is pressed. The only user-visible UI changes are:
* in filter hints with `waitForEnterForFilteredHints` enabled:
  - the HUD still shows the link hints contents rather than `Hit <enter> to proceed`
  - the user can 'back out' of the current selection by typing <kbd>backspace</kbd> and select a different link, since they're still in the full link hints mode
  - the border highlight isn't shown before pressing <kbd>enter</kbd> (which was a bit strange anyway)
* in alphabet hints, the border highlight for the selected link now covers all parts of the link, as it already did for filter hints.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -230,6 +230,10 @@ class LinkHintsMode
 
   # Handles all keyboard events.
   onKeyDownInMode: (event) ->
+    if @exitTimer? # There is only one hint left, but the user is still typing, so don't exit yet.
+      clearTimeout @exitTimer
+      @exitTimer = null
+
     return if event.repeat
 
     # NOTE(smblott) The modifier behaviour here applies only to alphabet hints.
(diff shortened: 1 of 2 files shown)


mrmr1993:
@smblott-github any interest in this? IMO it makes link hints exiting behaviour less strange, and the code easier to follow.

mrmr1993:
Rebased onto #2823.

mrmr1993:
Rebased onto master and updated description/title now #2823 has been merged.

philc:
Closing due to age.

PR #2792, opened 2017-11-11
Simplify tab count handling in link hints

mrmr1993:
Currently, tab count handling is.. weird. This PR tries to make it less so.

This changes the behaviour of <kbd>enter</kbd> and <kbd>space</kbd>.
* It should make no difference for <kbd>enter</kbd>, as it exits the mode.
* It makes no sense for rotating the layering of the hints to clear the active hint (when the next key is pressed), so I suspect it was an oversight.
  - This also only applies in the current frame, so all other frames still have the old `tabCount`
  - Inadvertently changing frame straight after rotating hints therefore means that tab may change its behaviour. This is unlikely, but still.. weird.

Also, reduces the number of lines in link hints slightly!

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -232,13 +232,9 @@ class LinkHintsMode
   onKeyDownInMode: (event) ->
     return if event.repeat
 
-    previousTabCount = @tabCount
-    @tabCount = 0
-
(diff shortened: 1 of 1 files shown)


smblott-github:
@mrmr1993 See f93410695cf5b62014cc9e37c0978c0e992b058e. Is that (in addition to this) correct?

mrmr1993:
Yup, looks right to me. Thanks for merging!

PR #2806, opened 2017-11-15
Allowed binding of < and > with modifier keys.

jeeger:
By escaping the > and <, both can be bound in the modifier keys
format. This enables binding shortcuts like <c-\>> and <c-\<>.


mrmr1993:
Why do we need this? We already accept `<c-<>` and`<c->>` as keys for mappings, and the key handling code will never emit these (so the bindings will never fire).

jeeger:
Ah, so the browser will never emit `<c->>` anyway? That's a pity. I'll close this then.

mrmr1993:
I misspoke there. However, the old regex will always prefer to interpret `<c-<>` and `<c->>` as expected, rather than as `<<>c-<<><>>`, so Ican't see what benefit the escaping brings.

jeeger:
Oh, I can actually bind those keys. 

I tried it before, and it didn't work, so I assumed it was due to escaping shenanigans. Thanks for your help!

PR #2827, opened 2017-11-21
[RFC] Add a (painfully) minimal commands listing

mrmr1993:
This adds a command listing, complete with the current mappings, as its own dedicated page. This is the absolute minimum start on my idea from #1280, as [requested a few years ago](https://github.com/philc/vimium/issues/1269#issuecomment-68362665). All the other Vimium-like extensions I've tested have some sort of page like this.

In its current state, this is essentially a duplicate of the help dialog, but as a large, unstyled table. If we run with this, we should:
1. style it. (I will need help with this.)
1. add more fleshed-out descriptions to the commands
1. document the options for each command in the descriptions
1. have collapsible descriptions (so that the options don't show by default and overwhelm the command list)
1. document the quirks/issues for each command in the descriptions
1. make the help dialog show only the subset of commands that it does by default, and point to this page for the rest
1. close or re-evaluate #1269
1. make the list searchable

**WARNING:** This is really unattractive and underwhelming. This PR is just to get *something* put together. Currently looks like this:

screenshot from 2017-11-21 13-48-57

@smblott-github is this a direction we could move in?

### DIFF
--- a/pages/commands.coffee
+++ b/pages/commands.coffee
@@ -9,33 +9,10 @@ compareKeys = (a,b) ->
   if a < b then -1 else if b < a then 1 else 0
 
 window.showTable = (helpPageData) ->
-  article = document.getElementsByTagName("article")[0]
-  article.innerHTML = ""
   for own group, commands of helpPageData
-    groupSection = document.createElement "section"
-    groupSection.className = "group-#{group}"
-
-    groupHeader = document.createElement "h2"
-    groupHeader.appendChild document.createTextNode group
-    groupSection.appendChild groupHeader
-
     for command in commands
-      commandSection = document.createElement "section"
-      commandSection.className = "command-#{command.command}"
-      commandHeader = document.createElement "h3"
-      keysSpan = document.createElement "span"
-      keysSpan.className = "keys"
-      descriptionParagraph = document.createElement "p"
-      descriptionParagraph.className = "description"
+      commandSection = document.getElementsByClassName("command-#{command.command}")[0]
+      keysSpan = commandSection.getElementsByClassName("keys")[0]
 
-      commandHeader.appendChild document.createTextNode command.command + " "
+      keysSpan.innerHTML = ""
       keysSpan.appendChild document.createTextNode (command.keys.join ", ")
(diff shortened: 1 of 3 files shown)


smblott-github:
@mrmr1993 I support doing this.  I've wondered about doing this myself.  With this approach, we have more space to describe what commands do, and -- in particular -- weird stuff like command options.

(Yes, the current format is v. ugly.)

philc:
+1. Happy to help style this if no one else jumps in.

mrmr1993:
I've pushed a few commits that hardcode the commands, and adds some basic styling (so that it doesn't hurt to look at). Looks like this now:

screenshot from 2017-11-21 23-28-56

@philc I'm terrible with styling, please make any changes you think would benefit.

philc:
I plan to merge this or a modified version of it for the next Vimium release.

philc:
I'm going to delay merging this until after the es6 port. I played with it and here are my thoughts. This requires a little more thought before we merge.

* I don't like that command names and descriptions from background_scripts/commands.coffee aren't used.
* I don't like having to keep two places up to date when we add a command (commands.coffee which drives the help dialog, as well as this commands.html page).
* We do need a place to append verbose descriptions (quirks, options) to our existing terse command descriptions. That's probably why this PR has the documentation in the commands.html page be self-contained and not generated from the command names and existing strings in background_scripts/commands.coffee -- it easily affords us the capability to write elaborate documentation.

The best idea I could come up with is to make the data structures describing each command in background_scripts/commands.coffee more elaborate, e.g. objects with properties commandName, shortDescription, documentation, defaultBinding. Then we can generate the help dialog and the commands page from that single data structure. I think this would be an improvement over what's in this PR.

philc:
FYI I've begun work on a modern version of this.

8061041-art:
nkvgvgkvyv yvy8uf7tugh

8061041-art:
k hjnnnnnnnnnnnnnnnnnnnnnnnnnnnmuobjlo nkjjinhjj98hubgyvtcrxe68s97d05f86-rgt97=8y0t6

PR #2846, opened 2017-11-30
Rework insert mode (check status dynamically).

### LINKED ISSUE #2841: Insert mode not activated on Github login
Logging in to the user interface on github, when I type "p" in the username text box, Chrome closes the current link and opens the Google link.

smblott-github:
This removes @insertModeLock and instead always checks whether we are in insert mode dynamically (as we previously did only for contentEditable).

@mrmr1993... Would you be able to check this, please?  It is intended to be mostly a no-op.

Fixes #2841.

### DIFF
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -10,23 +10,23 @@ class InsertMode extends Mode
 
     handleKeyEvent = (event) =>
       return @continueBubbling unless @isActive event
-      return @passEventToPage if @insertModeLock is document.body
+
+      # See comment here: https://github.com/philc/vimium/commit/48c169bd5a61685bb4e67b1e76c939dbf360a658.
+      activeElement = @getActiveElement()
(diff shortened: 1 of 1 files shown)


content_scripts/mode_insert.coffee
```diff
@@ -36,67 +36,16 @@ class InsertMode extends Mode
...
-    if (target and target == @insertModeLock) or @global or target == undefined
-      @log "#{@id}: deactivating (permanent)" if @debug and @permanent and @insertModeLock
-      @insertModeLock = null
-      # Exit, but only if this isn't the permanently-installed instance.
-      super() unless @permanent
+    return true if @global
+    DomUtils.isFocusable @getActiveElement()
+
+  getActiveElement: ->
+    document.activeElement?.shadowRoot?.activeElement ? document.activeElement
```
mrmr1993: Shadow roots can be nested, so the `.shadowRoot?.activeElement?` test should loop here to get the innermost active element.

PR #2851, opened 2017-12-02
Recurse an arbitrary depth into shadow DOMs for insert mode element

mrmr1993:
This addresses [my comment](https://github.com/philc/vimium/pull/2846#discussion_r154135427) from PR #2846.

/cc @smblott-github


smblott-github:
Thank you.

PR #2852, opened 2017-12-02
No clipboardWrite permission for Chrome Store version.

smblott-github:
Following on from #2601, it would be better if we do not unnecessarily require *all* Chrome Store users to reactivate Vimium, when that really isn't required.

@mrmr1993 reported elsewhere that requesting permissions dynamically was unlikely to work.

So, here, we just hack the manifest for the Chrome Store version.

### DIFF
--- a/Cakefile
+++ b/Cakefile
@@ -75,6 +75,7 @@ task "package", "Builds a zip file for submission to the Chrome store. The outpu
     blacklist.map((item) -> ["--exclude", "#{item}"]))
 
   spawn "rsync", rsyncOptions, false, true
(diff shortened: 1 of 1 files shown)


Cakefile
```diff
@@ -75,6 +75,7 @@ task "package", "Builds a zip file for submission to the Chrome store. The outpu
     blacklist.map((item) -> ["--exclude", "#{item}"]))
 
   spawn "rsync", rsyncOptions, false, true
+  spawn "sed", "-i /clipboardWrite/d dist/vimium/manifest.json".split /\s+/
```
mrmr1993: Won't doing this here (before packaging for either FF or Chrome) mean removing the permission for both..?

Issue #2670, opened 2017-09-26
t for 'open in new tab' does not respect Firefox new tab location preferences


vaibhavsagar:
I have `browser.tabs.insertRelatedAfterCurrent` set to `false` in `about:config` but Vimium does not respect this preference when opening a new tab and opens it after the current one instead.

mrmr1993:
We can't read preferences from `about:config` ([here for some vaguely-related discussion](https://discourse.mozilla.org/t/webextension-read-write-access-to-about-config/12268)), and our behaviour was set up to mirror Chrome's default, which is non-configurable.

I don't get to make these decisions, but I can't imagine this is high-value enough to warrant its own option in our settings.

vaibhavsagar:
It's high-value for me :smiley:. Would you be open to a PR? I think it might fit in 'Advanced Options'.

reinerh:
That behaviour is also slightly annoying to me. I'm used to creating new tabs at the end of the list, which is currently unfortunately not possible with vimium.
I would also appreciate an advanced setting for this (or a new command that creates tabs at the end).

benhartley:
Firefox is not behaving in the same way as Chromium for me. In Chromium, new tabs are opened in context groups, but the newest tab in the group is opened in the furthest right position. In Firefox, the new tab is always opened in the position immediately to the right of the current tab.

This is sub-optimal, since when opening several tabs from one page, a likely use case is to then review those tabs in the order they were opened (left to right), giving the benefit that the next oldest opened link will most likely have already loaded by the time you navigate to it, and so on as you move right through the tabs.

I'd be happy enough if Firefox behaved the same way as Chromium here so tabs were opened in the furthest right position within a context group. Failing that it does seem like something that should be configurable.

CyanBlob:
I'd like to add that how Vimium opens new tabs on Firefox can be a huge pain when using TreeStyleTabs, which is a popular Firefox addon. Right now, every new tab is opened as a child of the current tab, which causes some annoyingly deep trees. I would prefer for createTab to create the new tab at the root of the tree, not as a child of the current tab

PhilippWendler:
I agree, but I would like to add that `Vomnibar.activateEditUrlInNewTab` should still open tabs next to the current tab, because such a new tab will be related to the current tab. Only `Vomnibar.activateInNewTab` should open a new tab at the end of the tab bar.

philc:
This is partially addressed with a new option to the createTab command in #2895, and will go out with the next Vimium version.

PR #2863, opened 2017-12-12
Do not specify tab.index for new tab.

ris58h:
Seems like it's a fast fix for #2670.


smblott-github:
Thanks, @ris58h.  However, this fix applies to everywhere that `openUrlInNewTab` is used.  Perhaps we need to be more selective.  For example, when called from visual mode, the status quo behaviour seems best.  We'd have to look at the other cases one by one.

philc:
This is addressed as best we can with a new option to the createTab command in #2895, and will go out with the next Vimium version.

Issue #2865, opened 2017-12-12
Feature Request: Open the bookmark from the Bookmarks Bar


ris58h:

ris58h:
In which command group should I place this command?

PR #2866, opened 2017-12-12
Command for 'Open the bookmark from Bookmarks Bar'.

ris58h:
#2865

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -231,7 +231,8 @@ Commands =
     "closeOtherTabs",
     "enterVisualLineMode",
     "toggleViewSource",
-    "passNextKey"]
+    "passNextKey",
+    "openBookmarkFromBar"]
 
 defaultKeyMappings =
   "?": "showHelp"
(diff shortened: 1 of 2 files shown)


smblott-github:
Thanks @ris58h, but I'm not sure that this is of sufficiently general interest to warrant a whole new command.  It's a bit niche.

You can achieve similar effects using global marks and/or [createTab](https://github.com/philc/vimium/wiki/Tips-and-Tricks#creating-tabs-with-urls-and-windows).

ris58h:
@smblott-github Thank you for your feedback. Marks & createTab look like what I'm looking for, but it seems like they both create a new tab instead of change current one. Did I miss something?

smblott-github:
With global marks, create the mark *before* scrolling at all.  Then, an existing  tab will be reused.

ris58h:
@smblott-github Thank you for the response, but I still can't make it work. My workflow is:
1. Go to https://github.com (there is the only one tab in the browser's window)
2. Set a global mark with mG (no scrolling before as you said)
3. Go to any other site (google.com for example) in the same tab
4. Go to the global mark `G
It opens marked site (github in my example) in a new tab. What did I do wrong?
Thank you for your time.

smblott-github:
> 3. Go to any other site (google.com for example) in the same tab

Go to any other site in *another* tab, then `G takes you back to the original tab.  If no matching tab is open, then a new tab is created.

philc:
Thanks for exploring this @ris58h. I'm going to close this because it's too niche to be in the main repo. I haven't seen much demand for it.

PR #2871, opened 2017-12-14
Detect click listeners for link hints.

smblott-github:
This is a simpler version of #1167.  It detects clickable elements with listeners added with `addEventListener()`.

It includes some of @mrmr1993's ideas from #1167 (in fact, it's mostly those ideas tweaked into a slightly different form).

### DIFF
--- a/content_scripts/injected.coffee
+++ b/content_scripts/injected.coffee
@@ -1,0 +1,20 @@
+# The code in `injectedCode()`, below, is injected into the page's own execution context.
+#
+# This is based on method 2b here: http://stackoverflow.com/a/9517879, and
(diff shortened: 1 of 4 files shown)


gdh1995:
Um, this method (to add a new attr) seems to ugly for me, and in my personal view frontend developers won't be willing to see tons of `<a _vimium-has-onclick-listener ></a>`. So why Vimium chose such a method instead of #1859 ? If there's no reasons, then I beg for a reversion.

smblott-github:
#1859 is, in my view, extremely complicated and likely to be very difficult to maintain.

Let's just try this approach, and see what happens.  We can always back it out if an issue arises.

Issue #2847, opened 2017-11-30
Improve tab-switching bar UI


romgrk:
It would be nice if the tab-switching bar could have tab icons displayed for easier navigation.
Another nice improvement could be an option to display a shortcut key (similar to link-hints) next to each tab, and be able to press a single letter to switch to that tab.

Would a PR for this be welcomed? If so I might be interested to contribute.

smblott-github:
1. Icons.  We tried this some years ago and it turned out to be a bad idea because the availability of favicons was patchy, and it really did not look good at all.

2. Link-hints to select tabs.  There's no API to know the position of a tab's tab (I think). So this probably can't be done.

gdh1995:
Well, I think it's not patchy if we use 'chrome://favicon/16/<url>', and I'v implement this feature on my custom Vimium.

PR #2878, opened 2017-12-16
show favIcons for browser tabs on Vomnibar

gdh1995:
This make Vomnibar show favicons of current browser tabs. Implements #2847 .

### DIFF
--- a/background_scripts/bg_utils.coffee
+++ b/background_scripts/bg_utils.coffee
@@ -85,6 +85,9 @@ BgUtils =
       continue if line[0] in '#"'
       line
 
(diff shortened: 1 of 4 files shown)


smblott-github:
This looks quite nice, actually.  I was a little surprised...

image

However, we need a decision on #2656.  If we go with that proposal, then we'd need to make sure the two look ok together.

seniorivn:
for urls in #2656 we can use google favicon resolver, so there will be consistent view(with a fallback image for unresolved favicons)
http://codingclues.eu/2009/retrieve-the-favicon-for-any-url-thanks-to-google/

mmikeww:
bump

philc:
Looks great. Merged in cea42f3bd370d44531f0d76b7923f5bf7e67c19b. Thanks!

PR #2882, opened 2017-12-19
allow mapkey to 'enter' (v. isEnterMap, <c-]> by default)

rbtwrrr:
Hi!

This allows something such as `mapkey <a-q> <c-]>` to accept input in the vomnibar, or to submit a page's form in insert mode.

**Motive:**

In pentadactyl, for whatever reason, I liked being able to press `qq` to accept input and `jj` to blur focus. In https://github.com/philc/vimium/pull/2306 you made it possible to map something like `alt-j` to blur, which is easier to get used to than pressing escape. For some reason. However, unless I'm mistaken, it's still not possible to map some chord to accept input (`keymap <a-q> <enter>` does nothing).

My maps largely make use of the top left of the keyboard (qwerasdf) for common functions, and without the ability to double-tap q (or similar - `alt-q` in this case) I find myself sometimes reaching over the entire keyboard with my left hand to press enter. I can often get what I want with a few characters and a tap of tab. Let's say, Reddit - `re<Tab>qq`. I made a (personal) fork to implement `qq` in the vomnibar, but on discovering #2306 I thought I'd try to implement it in a more Vimium-y way. So here you go.

I expect you might have issue with:
- My choice of default bind (`C-]`). I just liked that it was adjacent to escape (`C-[`). There's no more thought to it than that; it exists to allow me to remap it with mapkey (and it's out of the way).
- The way I handle it when focus is on a page's input. Obviously a page can handle `Enter` in a variety of ways and this is potentially problematic. As long as users know it's basically a shortcut for "submit" and not actually "enter", I think this works.
- You might want to add a `useVimLikeEscape` analogue for this?
- I might be the only person who does this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ dist
 jscoverage.json
 tags
 .cake_task_cache
+pack.ps1
+.nvimrc
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@ dist
 jscoverage.json
 tags
 .cake_task_cache
-pack.ps1
-.nvimrc
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -116,6 +116,8 @@ Commands =
     # Inform `KeyboardUtils.isEscape()` whether `<c-[>` should be interpreted as `Escape` (which it is by
     # default).
     chrome.storage.local.set useVimLikeEscape: "<c-[>" not of keyStateMapping
+    # Same for `KeyboardUtils.isEnter()` of `<c-]>`.
+    chrome.storage.local.set useVimAdjacentEnter: "<c-]>" not of keyStateMapping
 
   # Build the "helpPageData" data structure which the help page needs and place it in Chrome storage.
   prepareHelpPageData: ->
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -261,7 +261,7 @@ class LinkHintsMode
         # knows not to restart hints mode.
         @hintMode.exit event
 
-    else if event.key == "Enter"
+    else if event.key == "Enter" || KeyboardUtils.isEnterMap event
       # Activate the active hint, if there is one.  Only FilterHints uses an active hint.
       HintCoordinator.sendMessage "activateActiveHintMarker" if @markerMatcher.activeHintMarker
 
@@ -878,7 +878,7 @@ class WaitForEnter extends Mode
 
     @push
       keydown: (event) =>
-        if event.key == "Enter"
+        if event.key == "Enter" || KeyboardUtils.isEnterMap event
           @exit()
           callback true # true -> isSuccess.
         else if KeyboardUtils.isEscape event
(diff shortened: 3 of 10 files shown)


smblott-github:
If we were to do this, it would have to be done uniformly for filtered link hints and find mode, too.  (Are there any other places?)

The function in `KeyboardUtils` should be `isEnter(event)` and should handle all of the cases (just like `isEscape()` handles all cases, and `isBackspace()`).

rbtwrrr:
I didn't think it'd be a good idea to intercept all actual presses of enter, since pages can do a lot of different things with it; an input might not have a form but it still handles enter; it slides a new form element into view.. etc. Whereas escape is pretty much always unfocus, and that's what the user wants with a keyboard-centric browser regardless of what the page wants. I didn't have luck sending a fake enter event in another userscript to get around that, either. It's significantly easier to keep it internal to vimium. But if that doesn't work for you, I can look at the options.

I didn't think about the other modes, I'll add it to those where relevant.

rbtwrrr:
I'm afraid I'm not sure how to get this to yank in visual mode - particularly because pressing enter doesn't seem to yank when I have a breakpoint set. But it does otherwise. Schrodinger's keypress.

It does work with filtered hints and in find mode mode. There's the problem of this not acting as "enter" to click a link upon focusing one with find mode, however. I'm not sure how to handle that. As far as I can tell you're letting the browser handle that - you focus the link and leave find mode on the first enter press, and the browser handles the next. So.. that would take some extra logic.

rbtwrrr:
Above commit makes it work for yanking in visual mode. Yay. It uses a fat arrow though, which.. seems a bit out of place?

rbtwrrr:
And the above commit is placeholder for a possible to way to simulate real enter in the page. I'm not sure if that's the right place. Also, placeholder since I need to charge my laptop.

rbtwrrr:
asd
asd
asd

rbtwrrr:
A few things, sorry if I'm spamming your inbox:
- Renamed the function, working around prior mentioned issues by checking for true enter in the relevant places.
- You'll notice in https://github.com/philc/vimium/pull/2882/commits/7738a6dfd7e33e79d1b93a0e31343305515779bf I moved the openInNewTab assignment - it wasn't used anywhere else in that function. 
- I'm relatively sure this is finished, unless you have problems with the way I've done something in particular (understandably).

philc:
@rbtwrrr, thanks for exploring this. I'm going to close this PR because I think this is niche and want to keep the codebase as simple as possible for future maintainers.

rbtwrrr:
That's fine. I understand the difficulties in maintaining public repositories, trust me. Thanks!

gdh1995:
Here's an AD: my customized version, named [Vimium C](https://github.com/gdh1995/vimium-c), supports `mapkey` in all modes, including FindMode and Vomnibar, and it also supports per-mode `mapkey` (e.g. `mapkey <f:o> <enter>` for Vomnibar), to use different keys in different modes. The detailed usage is in https://github.com/gdh1995/vimium-c/issues/110#issuecomment-580088981 .

Issue #2893, opened 2017-12-28
[FF] Open new tab after last tab


wladmis:
Current behavior is: when `t' key is pressed, Firefox opens new tab after current tab. I found it less convenient than if new tab was opened after last tab in the tab list.

Please, change the behavior to open new tab after the last tab if you find it reasonable, or make some way to able to change it by user preferences.

drwilly:
Firefox controls this via `browser.tabs.insertRelatedAfterCurrent`.
However opening a new tab through Vimium seems to ignore this setting.

reinerh:
Duplicate of #2670.

wladmis:
> Firefox controls this via browser.tabs.insertRelatedAfterCurrent

No, it's not what I want. I want that Firefox opens related tab after current tab but a new tab after last tab.

It's no duplicate of #2670.

philc:
An option for this was implemented in #2895 and will go out with the next Vimium release in late Jan. Thanks for the request @wladmis.

gdh1995:
@philc Sorry for my mistake but now I find the `position=end` may not work on Firefox: if `chrome.tabs.create {position: null}`, Firefox will open a tab next to a current tab...

Maybe I just tested the PR on Chrome. I'll put a new PR to fix it. Sorry again.

philc:
position:null opens a new tab after the current tab on Chrome also; it's not Firefox specific. That's how it's always been. Note that when I press cmd+T in Chrome, the new tab is inserted at the end. I'm not sure why Vimium's T and Chrome's Cmd+T behave differently. Maybe historical legacy?

The code in master preserves the current tab behavior for both Firefox and Chrome, and just adds some new options without changing the default behavior. I'm not sure we should change the default behavior at this point to reconcile Vimium to Chrome and Firefox's defaults. I prefer "new tab after existing tab" and have gotten used to it; if we change it now, many users will notice and ask about it.

gdh1995:
Oh you're right, then I'll split `position=end` and `position=undefined`.

gdh1995:
Oh, I find the root cause which failed my tests:
on Firefox, `chrome.tabs.create({url:'http://localhost/', index: null, openerTabId: 3})` will create a tab just next to the tab of `id=3`. And if no `openerTabId`, its default action is to open a tab at the end. So I think the new PR is still useful since `tabIndex=9999` is enough.

m6z:
Hello,
This works well in Firefox: `map t createTab position=end`
[...]

PR #2895, opened 2017-12-29
add an option "position=end/..." for command createTab

gdh1995:
This is a simple solution for #2893.

`map t createTab position=end about:blank` will open a blank page at the end of the browser tab list.


mrmr1993:
`position=end` (with potential for `start`, `before`, `after`, etc.) would be more in line with the other options.

philc:
Nice. I merged this in 2acf8dd51d6c3378e994bcbc34e21f6f15ee55f2. Thanks @gdh1995 !

PR #2912, opened 2018-01-13
WIP: Better visibility calculation for frames in link hints

smblott-github:
This addresses two related problems:

1. When detecting elements for link hints, all elements within a frame are selected (even if just part of the frame is visible).

2. Elements are even selected for frames which are wholly outside of the viewport.

Both of these happen because link detection happens only w.r.t. the frame's own window, without regard to where that window and viewport is within the overall viewport.

This PR attempts to address this by propagating information about frame visibility from parent frames to their children frames.  There's a test page [here](https://f.smblott.org/frames).

Issues:

1. I'm not sure that the viewport calculation is correct in all cases.
2. *Help required!* The mechanism for propagating visibility from parent to child here is trigered by scrolling.  This is not a good approach, in general.  What would be a better mechanism?  I'd prefer to maintain the information automatically all the time (rather than requesting it synchronously when the user hits `f`).


lib/dom_utils.coffee
```diff
@@ -131,21 +130,69 @@ DomUtils =
...
+            rect = do ->
+              rect = element.getBoundingClientRect()
+              left = Math.max 0, -rect.left
+              top = Math.max 0, -rect.top
+              right = Math.max 0, Math.min rect.width, window.innerWidth - rect.left
+              bottom = Math.max 0, Math.min rect.height, window.innerHeight - rect.top
+              right = left if right < left
+              bottom = top if bottom < top
+              Rect.create left, top, right, bottom
+            element.contentWindow.postMessage {name: "vimiumRegisterFrameViewport", rect, secret}, "*"
```
mrmr1993: * Don't send `vimiumSecret` to a frame that has non-Vimium code running in it. Otherwise, a malicious sub-frame can open and control any of the `UIComponent` pages.
* Don't use `window.postMessage`:
  - pages with `message` listeners may break when we send messages
  - pages who want to mess with Vimium can send themselves/child frames a message misreporting their size

As an alternative, you could try
```coffee
event = new CustomEvent "vimiumRegisterFrameViewport-#{secret}", { detail: { rect } }
element.contentWindow.dispatchEvent event
```
and attach the corresponding listener on each page.

lib/dom_utils.coffee
```diff
@@ -131,21 +130,69 @@ DomUtils =
...
+
+    window.addEventListener "load", onScroll
+    window.addEventListener "scroll", onScroll
```
mrmr1993: These should probably use `forTrusted`.

lib/dom_utils.coffee
smblott-github: > Don't send vimiumSecret to a frame that has non-Vimium code running in it. 

Doh.  Yes.  Thanks.

> As an alternative, you could try .... (a custom event)

I had tried that and just double checked.  Unfortunately it triggers a cross-origin frame security error.

image

> pages who want to mess with Vimium can send themselves/child frames a message misreporting their size

I'd be less worried about that (there's nothing for such a site to gain).  For example, a site can already mess with Vimium by removing the Vomnibar frame.

But...

> pages with message listeners may break when we send messages

How much of a concern is that really?  A page is pretty broken if sending it an arbitrary message breaks it.  I'd put this in the *let's try it out and see* class of issues.

philc:
@smblott-github do you still think this is a good idea?

Also, how important? Can you motivate with some real world pages that have bothered you?

mcint:
https://github.com/philc/vimium/issues/3515
https://github.com/philc/vimium/issues/3745
Spotify iframe embed in Medium: https://alinautrata.substack.com/p/can-you-see-your-computer-and-can The `gf`/`gF` trick even doesn't change link hinting to parent page.

Issue #2464, opened 2017-04-05
Make vomnibar suggestion URLs editable in the vomnibar input.


suptimal:
i want to use the highligthed entry. for example i type "git" -> [TAB] -> [CTRL+SPACE]:
![108d7d0f-25f2-45d4-8582-20be6ebc96ef](https://cloud.githubusercontent.com/assets/4957651/24695027/7afd7080-19e4-11e7-816c-17e70ac707dd.png)
and get 
![c154ff97-00ea-428a-959b-6af7824193cb](https://cloud.githubusercontent.com/assets/4957651/24695509/68c74b8c-19e6-11e7-81c0-d6271d1ea042.png)

smblott-github:
I don't understand the problem.

> i want to use the highligthed entry

So hit `<Enter>`?

blendi-93:
enter opens the url. I want a update of the input field, to modify the url.
for example.
after visiting github.com/baskerville/bspwm i have github.com/baskerville/bspwm in my history, if i want to visit github.com/baskerville/xdo for the first time i have to retype the whole url.

but i want to "reuse" to old entry and modify it by searching for bspwm select github.com/baskerville/bspwm hit [ctrl+space] and have github.com/baskerville/bspwm in the textfield so i can remove bspwm and replace it with xdo.

smblott-github:
Ah.  That makes more sense.

blendi-93:
thx for fixing my title ;)

i dont think its a widely used feature but i would like something like that (used ff and vimperator before)

gdh1995:
+1

BTW, my [Vimium++](https://github.com/gdh1995/vimium-plus) has implemented this feature.

Gu1nness:
Can I take this one ?

My expected behaviour for this would be that as soon as tab is used, the input in the vomnibar is modified accordingly with the url (the way I used to do when using VimFX in firefox).

smblott-github:
@mrmr1993 What do you think about this?

- For searches, populate the input with the previous search terms (as it does currently).
- For URLs (bookmarks and history), populate the input with the URL.

In both cases, the user can continue editing.

mrmr1993:
@smblott-github sounds good to me!

smblott-github:
@Gu1nness Have a look, if you like.  Make sure and look how things work at the moment for search suggestions.  Try to use the existing infrastructure.  (As I recall, that bit of the code is pretty tangled.)

Gu1nness:
[...]

PR #2914, opened 2018-01-15
Populates vomnibar input with url ( Issue #2464)

Gu1nness:
When <Tab> is pressed, if the suggestion is a bookmark or history,
populates the input with the corresponding url.

Refers and solve  https://github.com/philc/vimium/issues/2464

### DIFF
--- a/pages/vomnibar.coffee
+++ b/pages/vomnibar.coffee
@@ -114,6 +114,8 @@ class VomnibarUI
     else if (key == "down" ||
         (event.ctrlKey && (key == "j" || key == "n")))
       return "down"
+    else if (event.ctrlKey && event.key == "Enter")
+      return "ctrlenter"
     else if (event.key == "Enter")
       return "enter"
     else if KeyboardUtils.isBackspace event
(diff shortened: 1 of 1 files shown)


smblott-github:
Thanks, @Gu1nness.

This is quite a big change to the usability of the Vomnibar.  There are use cases where this is the right thing to do, and probably others where it is not.

A (less discover-able) alternative might be to use `<Control-Enter>` to copy the currently-selected URL to the input.

Bottom line... I'm not sure.

<hr>

Separately...

1. If the URL inserted into the input is too long for the input, then we probably should put the focus at the end.
2. What does it mean to do this in tabs mode (`T`)?
3. There's something very broken in tabs mode and in bookmarks mode (`b`).  Selecting an entry and editing the input, you'll see.

Gu1nness:
Indeed, I had missed these points (actually I didn't even know that this exists).
I'll look into it so this can be patched.
I agree for the `<Control-Enter>` this way it won't break the workflow the users.

Gu1nness:
Hi @smblott-github ,
I eventually pushed a new version, with <Ctrl-Enter> .

Gu1nness:
I have corrected the small typo that I had mistakenly inserted.

philc:
I wasn't able to merge this PR directly due to conflicts, but pushed your change in 2268e498637bc3190070e3d32146f981a33be726. I hadn't realized I've been missing this all these years. Thank you @Gu1nness !

philc:
I wish this was easier to discover. It would be cool to somehow have an opt-in cheat sheet built into Vomnibar, in the same spirit as typing "?" in Vimium. For now, I've documented this feature on [this wiki page](https://github.com/philc/vimium/wiki/Tips-and-Tricks).

Issue #2884, opened 2017-12-19
Vimium destroys some functionality of Angular application


vbrandl:
The following Angular application doesn't work with vimium activated: https://keyserver.ntzwrk.org/preview

I don't know what exactly causes the problems but here are the steps to reproduce:
Problem 1: Header goes missing
1. Go to https://keyserver.ntzwrk.org/preview
1. Navigate to any other subpage, e.g. `Submit` and navigate back to the search page
1. The header above the search bar is missing

Problem 2: Statistics and graphs are not loading:
1. Go to https://keyserver.ntzwrk.org/preview
1. Navigate to the `Stats` or `Graphs` subpage
1. Statistics and graphs are not loading. When visiting https://keyserver.ntzwrk.org/preview/#/stats or https://keyserver.ntzwrk.org/preview/#/graphs directly, both work fine but when navigating there, they don't

Currently the source code is not available on GitHub but it should be in a few days.

smblott-github:
Cannot reproduce Problem 1, but there's definitely something weird happening for Problem 2.  The console shows various CSP errors, eg...

![image](https://user-images.githubusercontent.com/2641335/34195597-596c5772-e556-11e7-8724-19bca3ee686a.png)

vbrandl:
Strange. I just tried again using a standalone version of Firefox 57.0.1 with only vimium installed and I could reproduce the issue.
Also with vimium deactivated, the graphs load as intended despite the CPS error.

JoshuaRichards:
Vimium-ff seems to break angular change detection. I've found an example at https://jsfiddle.net/yurzui/1hk7knwq/

Disabling vimium and refreshing the page makes the button text change when clicked as expected

vinaychandra:
A simpler example where Vimium is failing Angular applications is on (Angular Material)[https://material.angular.io/components/input/overview]. Clicking on the side tab does not work as intended. Click does not navigate to the correct page. Disabling the addon restores the functionality immediately.

aleph-naught2tog:
With Vimium activated, I can reproduce Issue 1 reliably on FF. However, disabling Vimium does *not* fix it in FF; I haven't tried removing it to see if it does.

In Chrome, I had no error on the site until installing Vimium, then the `reflected-xss` error appeared as well--*but* the header above the search bar stays regardless of that error and Vimium being enabled.

rfvoltolini:
Just to add, I could reproduce it with FireFox 57.0.4 (64-bit) and with

[...]

PR #2933, opened 2018-01-29
Revert "Detect click listeners for link hints."

smblott-github:
This reverts commit f2d7e185c704e3cf8f6ca0c9d3d4da2498610475.

Reason: see #2884.  This seems to break Angular applications, which is quite serious.

I propose merging this and then pushing a bug-fix release 1.62.9.

We could consider a variant of #2410 by way of a partial replacement.


gdh1995:
We can hook 'Function.prototype.toString' to fix crashes with libraries using 'lodash.isNative', just like what angular/zone.js has done in 2017/03. I think the new hook is surely enough because zone.js has no more issues about it.

smblott-github:
I don't completely understand your proposal, @gdh1995, but feel free to submit a PR.

Issue #2915, opened 2018-01-15
Cannot input successive Japanese/Chinese/Korean words on MacOS


regmarmcem:
When I input some texts written in Japanese/Chinese/Korean, I cannot input any characters following a Japanese/Chinese/Korean word before searching.  These languages need to type `<enter>` key to convert characters to a word that makes sense, but I cannot search two successive words because of this problem.

This problem arises both in `find mode` and in search window.

Environment:
Mac OS: High Sierra 10.13.1
Google Chrome: 63.0.3239.132 (Official Build)(64 bit)
Vimium: 1.62.4

I put an example below. 

![output](https://user-images.githubusercontent.com/27683817/34937128-023dddaa-fa27-11e7-8b51-18e2e0108173.gif)

smblott-github:
> These languages need to type `<enter>` key to convert characters to a word that makes sense,

Is it just `<Enter>` on its own, or is there also a modifier?

regmarmcem:
Just `<Enter>` on its own.
In Japanese, only `<Enter>` key decides what Japanese words I type.
According to my trying to search in other languages, this rule appears to be applicable to Chinese/Korean.

smblott-github:
How do other sites differentiate between `<Enter>` meaning *done*, and meaning *convert characters*?

E.g., [Whatsapp](https://web.whatsapp.com/), where `<Enter>` normally means *send*.

regmarmcem:
Same problem occurs in sites like that.
(Actually I tried only in [Slack][1])

[1]: https://slack.com/

innaterebel:
You mean when you entering CJK phonetics in a Vimium input and press <kbd>Enter</kbd> intending to confirm the words conversion, Vimium always immediately catches the <kbd>Enter</kbd> as `send`, right?

I cannot repro on Windows, tried Chinese and Japanese IME.

regmarmcem:
@innaterebel 
Thank you for perfect explanation! That's right.

gdh1995:
I'm using the Chinese Pinyin IME of Windows 10, and `<space>` is used to select a word, so I haven't seen such a problem. And when the IME is enabled, `<Enter>` only types letters to a Vimium input (not commits them).

regmarmcem:
I tried on Windows10, but I cannot repro this problem. This problem seems to be unique to MacOS.

PR #2934, opened 2018-01-31
Enable to convert CJK characters on MacOS(Issue#2915)

regmarmcem:
on isuue #2915 , I enabled to convert CJK characters on MacOS.
I tried on Windows10, Ubuntu16.04, and MacOS(High Sierra). But I could reproduce this problem only on MacOS.

I fixed `pages/vomnibar.coffee` and `pages/hud.coffee`.

This change is based on the feature that typing `Enter` key contains three event, `keydown 13`, `keypress 13`, and `keyup 13`. 
On the other hand, converting CJK words contains `keydown <keycode>` and `keyup 13`, not containing `key press 13`.

This change makes us type successive CJK words on MacOS.

before
before

after
after

### DIFF
--- a/pages/hud.coffee
+++ b/pages/hud.coffee
@@ -15,16 +15,30 @@ setTextInInputElement = (inputElement, text) ->
 document.addEventListener "DOMContentLoaded", ->
   DomUtils.injectUserCss() # Manually inject custom user styles.
 
+document.addEventListener "keypress", (event) ->
+  inputElement = document.getElementById "hud-find-input"
+  return unless inputElement?
+
+  if event.key == "Enter"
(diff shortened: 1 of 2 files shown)


smblott-github:
Thanks for this, @regmarmcem.

What about 61951f222c5588ea662993f8078cbe071eb2ece8 as a simpler alternative to this.  Would that have the same effect?

regmarmcem:
Thank you for your change. Your change also works well. 
That is  a simpler alternative to mine.

smblott-github:
Closing.  Fixed via 61951f222c5588ea662993f8078cbe071eb2ece8.

PR #2951, opened 2018-02-13
Make the reload command accept a count.

### LINKED ISSUE #674: Reload all tabs
It would be nice to have a 'reload all tabs' functionality like in Firefox. Chrome doesn't have that built-in and I don't feel like installing extensions for that.

smblott-github:
Move the `reload` command to the background page and make it accept a count prefix.

Fixes #674.

Although the demand for this is low and the use cases limited, it is a very vim-ish extension to the existing command.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -336,7 +336,7 @@ commandDescriptions =
   scrollFullPageDown: ["Scroll a full page down"]
   scrollFullPageUp: ["Scroll a full page up"]
 
-  reload: ["Reload the page", { noRepeat: true }]
+  reload: ["Reload the page", { background: true }]
   toggleViewSource: ["View page source", { noRepeat: true }]
 
   copyCurrentUrl: ["Copy the current URL to the clipboard", { noRepeat: true }]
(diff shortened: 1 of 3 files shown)


gdh1995:
Sorry, but I think this makes some unexpected difference:
* the old code allows someone to only reload the current iframe
* the new version can only reload the whole tab.

So, maybe a new command `reloadTab` is better.

smblott-github:
(I merged this at the exact moment you commented.)

> the old code allows someone to only reload the current iframe

Yes.  I realised that.  However, that is arguably also a bug.  For example, if the focus is in an iframe, then the user might be confused as to why `r` doesn't reload the tab.

gdh1995:
Um, then let's wait more feedbacks.

politicus:
From the changelog 1.63 (2018-02-16), "The reload command now accepts a count prefix; so 999r reloads all tabs (in the current window). Tested it and confirms that it is not working, like already said [here](https://github.com/philc/vimium/issues/674#issuecomment-365108651).

Maybe updating the changelog so no other Vimium user tries to make this work is a good idea.

PR #2960, opened 2018-02-18
Fix Firefox link hints.

### LINKED ISSUE #2958: Vimium-FF 1.63 'f' button triggers pop-ups blocker and 'F' button do not work
Version: Vimium-FF 1.63 from Firefox addons store (most recent)
Firefox: 58.0.2  (64 bit)

Current Behavior
Using `f` button triggers firefox pop-ups blocker 
After using `F` button, no new tab show up
rolling back to Vimium-FF version 1.62.8  fix these problems

Expected Behavior
Using `f` button should open the link 
Using `F` button should open the link in new tab

smblott-github:
Fixes #2958 (probably).

This appears to be correct for target "_blank" (or not) and modifiers (or not).  That's four cases.

It will be *incorrect* if there is a click listener on a target "_blank" link.  Some conditions are tested to try to prevent this from happening.

This only affects Firefox.  Chrome is unaffected.

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -247,8 +247,6 @@ DomUtils =
             element.setSelectionRange element.value.length, element.value.length
 
   simulateClick: (element, modifiers = {}) ->
-    console.log modifiers
-    console.log Object.keys(modifiers).length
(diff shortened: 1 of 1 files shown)


troglotit:
Tried it locally, `f` and `F` work as expected!

PR #2968, opened 2018-02-24
Escape [ and ] in hostname in popup.

### LINKED ISSUE #2967: Rules by default do not work for IPv6 hosts
When connected to e.g. http://[fe80::cce4:1680:e720:eacb]:8080/ when opening the small extension menu it will by default create this pattern: `https?://[fe80::cce4:1680:e720:eacb]:8080/*`

This pattern does, however, not work unless '[' and ']' is escaped (and this should be done by default).

Chrome: 65.0.3325.73
Vimium: 1.63

Thanks, great project :)

smblott-github:
From URL:

    http://[fe80::cce4:1680:e720:eacb]:8080/index.html

generate:

    https?://\[fe80::cce4:1680:e720:eacb\]:8080/*

in the popup (note that `[` and `]` are now escaped).

Fixes #2967.

### DIFF
--- a/pages/options.coffee
+++ b/pages/options.coffee
@@ -181,7 +181,9 @@ class ExclusionRulesOnPopupOption extends ExclusionRulesOption
     if /^https?:\/\/./.test @url
       # The common use case is to disable Vimium at the domain level.
       # Generate "https?://www.example.com/*" from "http://www.example.com/path/to/page.html".
-      "https?:/" + @url.split("/",3)[1..].join("/") + "/*"
+      # Note: IPV6 host addresses will contain "[" and "]" (which must be escaped).
+      hostname = @url.split("/",3)[1..].join("/").replace("[", "\\[").replace "]", "\\]"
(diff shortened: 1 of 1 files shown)


smblott-github:
Note... We have a separate longstanding bug whereby `.` is not escaped in exclusion rules, and therefore matches *any* character.  I looked at escaping all Regexp characters here, including `.`, but it's just too noisy.  And, since nobody has ever complained, let's just leave that be.

gdh1995:
@smblott-github Then how can I write a pattern to match sites by `new RegExp( "https?://video[0-9]+\.example\.com/.*" )` ?

smblott-github:
> Then how can I write a pattern to match sites by ...

The change only affects the case where we are generating a *new* rule with the *default* pattern.  If an existing rule exists, such as the one you propose, then the new code is not touched at all.  Nothing is changed.

gdh1995:
Oh, yes. Sorry for my misunderstanding.

PR #2971, opened 2018-02-26
Use count prefix for closeTabs___ commands.

### LINKED ISSUE #2970: Close tabs on the left/right command not very useful?
Hello, I'v been use Vimium many years. IMO the advanced command **Close tabs on the left/right** are not very useful. I think the command are same as **Close all other tabs** in many situation. Do you think change the command to **Close tab on the left/right**(not tabs) will be more suitable for daily use. Thanks and best wishes.

smblott-github:
E.g.:

    map cl closeTabsOnRight

    # Close just two tabs on the right.
    2cl

    # Always close all tabs on the right
    map cl closeTabsOnRight count=999999

This is a change to the established behaviour.  However, it is more vim-like, and arguably more elegant.

Feedback on "do"/"don't do" this would be appreciated.

Fixes #2970.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -389,9 +389,9 @@ commandDescriptions =
   togglePinTab: ["Pin or unpin current tab", { background: true, noRepeat: true }]
   toggleMuteTab: ["Mute or unmute current tab", { background: true, noRepeat: true }]
 
-  closeTabsOnLeft: ["Close tabs on the left", {background: true, noRepeat: true}]
-  closeTabsOnRight: ["Close tabs on the right", {background: true, noRepeat: true}]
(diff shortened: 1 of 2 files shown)


philc:
@smblott-github responding to your request for feedback, this seems better, but not that important, and thus probably not worth changing the behavior at this time.

gdh1995:
It's not needed to change the old behavior: we can check whether `count` option exists or not, and:
* if no `count` option, close all
* if `count=1`, then `2cl` close 2 tabs
* if `count=2`, then `2cl` close 4 tabs

philc:
@gdh1995's proposal is better that changing the default behavior. But I'm still going to close due to not being that important right now. Reopen if you feel strongly otherwise.

philc:
This is now revived in #4296!

PR #2975, opened 2018-02-27
Add options to control selection for focusInput.

### LINKED ISSUE #2969: Feature Request: "gi" should highlight already existing input text
In pages that dynamically generate content, many times we use the same input field over and over again. Typing "gi" to get in the input field and then having to press also ctrl+A in order to overwrite the existing text would become faster if "gi" automatically selected the input text.

smblott-github:
This implements the suggestion in [this comment](https://github.com/philc/vimium/issues/2969#issuecomment-368491430).

Examples:

    map X focusInput select=all
    map X focusInput select=start
    map X focusInput select=end

Fixes #2969.


philc:
I think this is too niche. I've started discussion in #2969.

PR #2978, opened 2018-03-06
Add zoom command.

### LINKED ISSUE #1866: Feature Request - Adjust zoom/fontsize via keypress
Being able to adjust font size via key press is super handy. There is currently a [vim script](https://github.com/drmikehenry/vim-fontsize) that has this behavior, and the script gives a good precedent.  The script uses the following keybindings:

```
\\+ => Zoom Bigger
\\++ => Zoom 2x Bigger 
\\+++ => Zoom 3x Bigger 
(etc)
\\- => Zoom Smaller
\\-- => Zoom 2x Smaller
\\---- => Zoom 3x Smaller 
(etc)
```

I suspect many users would find this a helpful functionality.

smblott-github:
Examples:

    map = zoom in factor=1.1
    map - zoom in factor=1.1

    map = zoom factor=1.1
    map - zoom factor=0.90909

Fixes #1866.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -172,7 +172,8 @@ Commands =
       "nextFrame",
       "mainFrame",
       "Marks.activateCreateMode",
-      "Marks.activateGotoMode"]
+      "Marks.activateGotoMode",
+      "zoom"]
     vomnibarCommands:
       ["Vomnibar.activate",
       "Vomnibar.activateInNewTab",
@@ -231,7 +232,8 @@ Commands =
     "closeOtherTabs",
     "enterVisualLineMode",
     "toggleViewSource",
-    "passNextKey"]
+    "passNextKey",
+    "zoom"]
 
 defaultKeyMappings =
   "?": "showHelp"
@@ -409,6 +411,8 @@ commandDescriptions =
 
   "Marks.activateCreateMode": ["Create a new mark", { noRepeat: true }]
   "Marks.activateGotoMode": ["Go to a mark", { noRepeat: true }]
+
+  zoom: ["Zoom in or out"]
 
 Commands.init()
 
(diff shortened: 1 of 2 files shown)


glumanda99:
Please merge :heart_eyes_cat:

NightMachinary:
Please merge.

maxim-danilov:
It's a very useful feature. I use zooming on every site. Please add support:
```
map q zoomIn
map w zoomOut
```
Or just merge the commit.

alfozan:
+1

philc:
@smblott-github we used to have a zoom command in the early days of Vimium. Here's the initial [commit)](d367e58a555606b7ad232fa880484cad8a2f7844). We even saved the preference per domain, and had a configurable default zoom level [(commit)](3b8c0f1060640efd0a9383707a1d892d102e414f) in Vimium.

I [removed](3b8c0f1060640efd0a9383707a1d892d102e414f) this command because setting document.body.style.zoom doesn't work the same as whatever Chrome is doing when you use Chrome's native zoom. In particular, text would start unexpectedly wrapping on some pages based on the zoom level, which never happens with Chrome's native zoom.

See `git log --grep=zoom` for all related commits and for some of the complexity this feature caused.

If we rework this to use [Chrome's native zoom API](https://developer.chrome.com/extensions/tabs#method-setZoomSettings) rather than document.body.style.zoom, that should solve the zoom robustness piece.

To make this feature more robust, we could then add:

1. Remembering the setting for the given domain as Chrome does (assuming use of the zoom settings API doesn't already do that by default; requires further investigation).
2. Implement a default zoom level setting, as you can with Chrome. That way, we start zooming or unzooming from that default, rather than from 100%.

I would argue that this command is not acceptable without adding #1, because our zoom is inferior to Chrome's native zoom in a really annoying way (i.e. it has a hidden sharp edge). I could see people navigating to reddit.com, zooming in, and the next time they navigate there, zooming in again because their settings weren't saved, and doing that every time they load the page. With Chrome's native zoom functionality, they would only have to zoom once.

j-cr:
Saka key has this feature:
https://github.com/lusakasa/saka-key/blob/3ec7984c1a1d630f6c1eb5b9a6932678cf6a21a3/src/modes/command/background/commands/tab.js#L161

Seems straightforward?

(1) and (2) from the comment above already work as expected in saka key.

philc:
Closing this PR because it's outdated, but I've reopened #1866 to track the feature request.

Issue #2734, opened 2017-10-19
[Firefox] Vimium breaks Vim-mode in CodeMirror


xuhdev:
[CodeMirror](http://codemirror.net/demo/vim.html) provides a Vim-mode. However, if Vimium is installed, I cannot come back to normal mode once I enter insert mode in the codemirror editing box.

mrmr1993:
We handle <kbd>esc</kbd>, and so the CodeMirror input never gets it. To workaround this, you'll want to map some key to `passNextKey` (in the options page), and then you can press that before <kbd>esc</kbd> to pass it.

mrmr1993:
In current master (due to commit 66b11f98f480bbd83cb2cf720e219d0df8203f80), if you add `map <c-[> passNextKey` to your mappings, then we won't handle <kbd>ctrl</kbd>+<kbd>[</kbd> and the CodeMirror input will interpret it as <kbd>esc</kbd>. There isn't any workaround for the versions in the Chrome/Firefox stores that I can see.

@smblott-github I like the sound of something like Vimperator's `<c-esc>` that toggles all Vimium keys ([noted in the CodeMirror issues](https://github.com/codemirror/CodeMirror/issues/4391#issuecomment-289170542)). What do you think?

jez:
I've opened a pull request that special-cases CodeMirror into Vimium so that the two can co-exist peacefully: #2993.

I've been using it locally for a while, and it seems to fit my needs; feel free to try it out.

PR #2993, opened 2018-03-19
Add CodeMirror as special case in InsertMode

jez:
CodeMirror (a code editor component that can be embedded in web pages)
implements a simple Vim emulation mode. It supports navigating the text
with Vim normal-mode operators, and also editing text with insert-mode
operators. In particular, CodeMirror supports pressing ESC to go from
Insert mode to Normal mode.

To Vimium (and the DOM), CodeMirror appears as an editable, focusable
element. This means that when you press ESC in a CodeMirror editor,
Vimium will blur CodeMirror's focus. This prevents CodeMirror users from
using Vim mode.

This change allows CodeMirror's Vim keymap and Vimium to co-exist on the
same page. Without this, Vimium will never let a CodeMirror's editor
return to Normal mode. If we're inside a CodeMirror editor, instead of
trapping the ESC, we pass it through to the editor.

This has the slight drawback that you'll need to use the mouse to take
focus away from CodeMirror. I consider this preferable, because it
makes CodeMirror feel as close to Vim while it does have your focus.

For more context:

- <https://github.com/philc/vimium/issues/2734>

To reproduce and play around with the existing CodeMirror behavior:

- <https://codemirror.net/demo/vim.html>

Also, I'm happy to pursue alternative workarounds to this! Looking
forward to hearing your thoughts.

### DIFF
--- a/content_scripts/mode_insert.coffee
+++ b/content_scripts/mode_insert.coffee
@@ -20,8 +20,13 @@ class InsertMode extends Mode
         new PassNextKeyMode
 
       else if event.type == 'keydown' and KeyboardUtils.isEscape(event)
-        activeElement.blur() if DomUtils.isFocusable activeElement
-        @exit() unless @permanent
+        # Special-case so that CodeMirror's Vim keyMap and Vimium can co-exist on one page.
+        # Without this, Vimium would never let a CodeMirror's editor return to Normal mode.
+        if DomUtils.isCodeMirror activeElement
+          return @passEventToPage
+        else
+          activeElement.blur() if DomUtils.isFocusable activeElement
+          @exit() unless @permanent
 
       else
         return @passEventToPage
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -205,6 +205,12 @@ DomUtils =
   isFocusable: (element) ->
     element and (@isEditable(element) or @isEmbed element)
 
+  isCodeMirror: (element) ->
+    if element.classList.contains('CodeMirror')
+      return true
+    else
+      return element.parentNode && DomUtils.isCodeMirror element.parentNode
+
   isDOMDescendant: (parent, child) ->
     node = child
     while (node != null)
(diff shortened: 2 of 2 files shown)


smblott-github:
Thanks, @jez.

> This prevents CodeMirror users from using Vim mode.

Not entirely.

Using [pass-next-key](https://github.com/philc/vimium/wiki/Tips-and-Tricks#using-the-escape-key-in-inputs) should work, should it not?

Having Vimium's `<Escape>` wholly disabled in CodeMirror might confuse (and frustrate) users.

Separately, the tests are failing. It looks like `.classList.contains()` is not available in phantomjs.

jez:
Ah, I'll give `passNextKey` a try, and see how it feels. I was hesitant to try it at first, because I tend to press `ESC` over-eagerly to ensure that I'm in normal mode.

Using `passNextKey`, I'd have to be more mindful of always pressing `<C-]>` when editing a buffer with CodeMirror's Vim. My goal is for using CodeMirror with Vim mode to feel quite close to normal Vim. (For some context: I'm working on a small plugin that turns normal textareas into CodeMirror editors).

While I'm trying out `passNextKey`, could I get your thoughts on these two alternative implementations?

1. What if we make the key that blurs the active element configurable? Right now, we're hard-coding `isEscape`:

https://github.com/philc/vimium/blob/5745bb344959a61297fc16e00bb6de0a2a8744d9/content_scripts/mode_insert.coffee#L22-L23

2. What if we implement something like was discussed in https://github.com/philc/vimium/issues/2734#issuecomment-337988544: we add a keybinding to enter a "disabled mode" where all key events are passed through to the page (toggled with a keystroke)?

3. We register a *different* keystroke (something other than `ESC` to blur the active element when a CodeMirror editor is active, so that it's still possible to blur a CodeMirror editor using only the keyboard?

The third seems sort of like a hack in response to a hack, whereas I think the first two solutions are fairly general enough to not feel like a hack.

Also, thanks for responding so quickly!

Issue #3002, opened 2018-03-31
Vimium breaks "Remove" button on pay.google.com


ravron:
To reproduce this bug, you must have a Google account. Repro steps:

1. Ensure Vimium is installed and enabled.
1. Navigate to https://pay.google.com/send/paymentMethods?authuser=0.
1. If you have one or more payment methods already added, skip to step 3. Otherwise, add any valid payment method and return to https://pay.google.com/send/paymentMethods?authuser=0.
1. Choose one of your listed payment methods and click "Remove." 

Expected behavior: pay.google.com creates a pseudo-popup confirming you want to remove the payment method.
Observed behavior: nothing appears to happen. 

Disabling Vimium through its own settings — i.e. by adding an empty rule for play.google.com — doesn't fix the issue. Disabling it via chrome://extensions does, though.

- Chrome: 65.0.3325.181
- Vimium: 1.63

smblott-github:
This is probably a duplicate of #2997.

Assuming that that's correct, I'll close this and move the dicusssion over there.

ravron:
@smblott-github ah, yep, that looks right. Thanks for finding that dupe!

Issue #2997, opened 2018-03-27
Disable DOM injection on specific site


hanfeisun:
When I enable vimium on [google payment center](https://payments.google.com/payments/u/0/home#paymentMethods), some of their function will not work at all. The reason is that they use safeHTML check to avoid potential malicious DOM manipulation. And vimium will appended a tag  _vimium-has-onclick-listener to the DOM, which will break their safeHTML checking.

I tried to disable vimium on that site, but that doesn't work. Even though all the keybindings are disabled, the DOM injection still happens, so the safeHTML still breaks.

Is there a way to also disable DOM injection on specific site?

smblott-github:
Thanks, @hanfeisun.  I imagine that took some time to track down!

This sounds plausible.  We will need to find another mechanism for informing the Vimium Javascript context when a click listener is added to an element in the page Javascript context.

smblott-github:
I disabled the offending injected (5745bb344959a61297fc16e00bb6de0a2a8744d9) and push this as 1.63.3.  This is a temporary fix.

PR #3004, opened 2018-04-02
Use a Vimium class to detect click listeners.

smblott-github:
5745bb344959a61297fc16e00bb6de0a2a8744d9 disabled our detection of click listeners added with `addEventListener` (see #2997 and #3002).

This PR proposes using the CSS class list to inform Vimium that an element has a click listener (instead of a custom attribute).

### DIFF
--- a/content_scripts/injected.coffee
+++ b/content_scripts/injected.coffee
@@ -1,6 +1,3 @@
-# NOTE(smblott) Disabled pending resolution of #2997.
-return
-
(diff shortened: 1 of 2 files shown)


philc:
@smblott-github Can I close this due to inactivity?

The DOM injection strategy is currently disabled, and has always felt problematic to me.

philc:
Closing to tidy up the PR queue.

Issue #3028, opened 2018-05-10
Is Code Coverage broken?


ris58h:
I've tried to run Code Coverage (after installing dependencies of course) but I've got an error.

npm run coverage

> vimium@ coverage /path_to_vimium
> cake coverage

/path_to_vimium/lib/utils.js:8
  if (window.forTrusted == null) {
  ^

ReferenceError: window is not defined
    at Object.<anonymous> (/path_to_vimium/lib/utils.js:8:3)
    at Object.<anonymous> (/path_to_vimium/lib/utils.js:492:4)
    at Module._compile (module.js:662:30)
    at Object.Module._extensions..js (module.js:673:10)
    at Module.load (/path_to_vimium/node_modules/coffeescript/lib/coffee-script/register.js:45:36)
    at tryModuleLoad (module.js:515:12)
    at Function.Module._load (module.js:507:3)
    at Module.require (module.js:606:17)
    at require (internal/module.js:11:18)
    at Object.action (/path_to_vimium/Cakefile:202:13)
    at invoke (/path_to_vimium/node_modules/coffeescript/lib/coffee-script/cake.js:44:26)
    at Object.exports.run (/path_to_vimium/node_modules/coffeescript/lib/coffee-script/cake.js:70:20)
    at Object.<anonymous> (/path_to_vimium/node_modules/coffeescript/bin/cake:15:42)
    at Module._compile (module.js:662:30)
    at Object.Module._extensions..js (module.js:673:10)
    at Module.load (module.js:575:32)
    at tryModuleLoad (module.js:515:12)
    at Function.Module._load (module.js:507:3)
    at Function.Module.runMain (module.js:703:10)
    at startup (bootstrap_node.js:193:16)
    at bootstrap_node.js:665:3

smblott-github:
It's probably broken.  I've never run it.  Ever.

Issue #3023, opened 2018-05-07
build failure


yhuangatimvucom:
I am using these command to build it:
`npm install --save-dev coffeescript
cake build`
But fail with following output
`npm WARN saveError ENOENT: no such file or directory, open '/Users/yhuang/Dropbox/myproject/git/firefox/philc/vimium/package.json'
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN enoent ENOENT: no such file or directory, open '/Users/yhuang/Dropbox/myproject/git/firefox/philc/vimium/package.json'
npm WARN vimium No description
npm WARN vimium No repository field.
npm WARN vimium No README data
npm WARN vimium No license field.

+ coffeescript@2.3.0
updated 1 package in 0.568s
/Users/yhuang/Dropbox/myproject/git/firefox/philc/vimium/node_modules/coffeescript/lib/coffeescript/optparse.js:147
    shortFlag = shortFlag != null ? shortFlag.match(SHORT_FLAG)[1] : void 0;
                                                               ^

TypeError: Cannot read property '1' of null
    at buildRule (/Users/yhuang/Dropbox/myproject/git/firefox/philc/vimium/node_modules/coffeescript/lib/coffeescript/optparse.js:147:64)
    at /Users/yhuang/Dropbox/myproject/git/firefox/philc/vimium/node_modules/coffeescript/lib/coffeescript/optparse.js:119:22
    at buildRules (/Users/yhuang/Dropbox/myproject/git/firefox/philc/vimium/node_modules/coffeescript/lib/coffeescript/optparse.js:122:7)
    at new OptionParser (/Users/yhuang/Dropbox/myproject/git/firefox/philc/vimium/node_modules/coffeescript/lib/coffeescript/optparse.js:27:20)
    at Object.exports.run (/Users/yhuang/Dropbox/myproject/git/firefox/philc/vimium/node_modules/coffeescript/lib/coffeescript/cake.js:73:14)
    at Object.<anonymous> (/usr/local/lib/node_modules/coffeescript/bin/cake:22:42)
    at Module._compile (module.js:660:30)
    at Object.Module._extensions..js (module.js:671:10)
    at Module.load (module.js:573:32)
    at tryModuleLoad (module.js:513:12)
    at Function.Module._load (module.js:505:3)
    at Function.Module.runMain (module.js:701:10)
    at startup (bootstrap_node.js:193:16)
    at bootstrap_node.js:617:3

`

smblott-github:
[...]

PR #3029, opened 2018-05-10
Use package.json to manage dependecies.

ris58h:
CoffeeScript is a dependency now and there shouldn't be such issues as in https://github.com/philc/vimium/issues/3023

But there are some things to add:

- It would simplify build if we could get rid of shoulda.js and that git submodule and use any other test runner library/framework.

- I couldn't run Code Coverage https://github.com/philc/vimium/issues/3028 so this part remains unchanged. It would be nice to move jscoverage and temp to devDependencies too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,12 @@
 sudo: false
 language: node_js
 node_js: "8"
-before_install:
-  - npm install -g coffee-script
-  - npm install path@0.11
-  - npm install util
-  - npm install -g phantomjs@1.9.20
-  - cake build
-script: cake test
+script:
+  - npm run build
+  - npm run test
+cache:
+  directories:
+    - "node_modules"
 notifications:
   email: false
 branches:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,8 +26,7 @@ Please include the following when reporting an issue:
 Vimium is written in Coffeescript, which compiles to Javascript. To
 install Vimium from source:
 
- 1. Install [Coffeescript](http://coffeescript.org/#installation).
- 1. Run `cake build` from within your vimium directory. Any coffeescript files you change will now be automatically compiled to Javascript.
+ 1. Run `npm run build` from within your vimium directory. Any coffeescript files you change will now be automatically compiled to Javascript.
  1. Navigate to `chrome://extensions`
  1. Toggle into Developer Mode
  1. Click on "Load Unpacked Extension..."
@@ -35,7 +34,7 @@ install Vimium from source:
 
 ## Development tips
 
- 1. Run `cake autobuild` to watch for changes to coffee files, and have the .js files automatically
+ 1. Run `npm run autobuild` to watch for changes to coffee files, and have the .js files automatically
     regenerated
 
 ## Running the tests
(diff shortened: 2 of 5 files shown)


smblott-github:
Thanks, @ris58h.

I get this error...

    ❯ npm install 

    > phantomjs@1.9.20 install /home/blott/local/project/vimium/node_modules/phantomjs
    > node install.js
    
    Considering PhantomJS found at /usr/bin/phantomjs
    Looks like an `npm install -g`
    Linking to global install at /usr/lib/node_modules/phantomjs/lib/phantom/bin/phantomjs
    Writing location.js file
    npm WARN vimium@ No repository field.
    npm WARN vimium@ No license field.
    
    added 107 packages from 75 contributors in 1.433s
    npm ERR! Object.entries is not a function
    
    npm ERR! A complete log of this run can be found in:
    npm ERR!     /home/blott/.npm/_logs/2018-05-11T08_37_40_608Z-debug.log

(The complete error log is 3000 lines, so I'll not post it here.)

I tried with a fresh clone too.  Same thing.  npm version is 6.0.1.

Separately, this is going to use the global `cake` and `coffee` commands, is it not?

ris58h:
@smblott-github It's strange. Everything works fine on my machine and [travis](https://travis-ci.org/ris58h/vimium) seems ok too. Could you provide the log?

ris58h:
@smblott-github I believe that problem is related to node. As I can see in the log you have node v6.12.2. I use v9.6.1. Vimium's travis file contains node_js 8.

smblott-github:
> I believe that problem is related to node...

That's it, thanks. And...

> Separately, this is going to use the global cake and coffee commands, is it not?

It seems that npm does indeed arrange that `$PATH` is set correctly and uses the local copy of `cake` and `coffee`.

Could you...

- fill in the license and repo in `package.json` (to prevent unnecessary warnings), and
- note the required version of `node` in `CONTRIBUTING.md`, please?

Note -- something good about this... Legacy users with existing clones are unaffected, I think, right?

ris58h:
> not for this PR

I know. It's related to dependencies so I've just asked.
> I have no opinion on this.

OK. I would investigate that later.

One more question: I see that `cake` is used in pre-commit git hook, but when i try run that shell-script it fails with `mktemp: too few X's in template ‘vimium’` because the name of the template must contain at least 3 X according to `man mktemp`. Is that hook used at all?

smblott-github:
> Is that hook used at all?

I don't use it (because sometimes I want to commit code with the tests broken).

ris58h:
> fill in the license and repo in package.json (to prevent unnecessary warnings)

According to [npm docs](https://docs.npmjs.com/getting-started/using-a-package.json) only two fields are required: name & version. I would like to keep things simple and don't add unnecessary fields because Vimium is not going to be published in the npm registry anyway.

> note the required version of node in CONTRIBUTING.md, please?

I've done more research about that problem and it seems like it's just an [npm-related issue](https://github.com/npm/npm-audit-report/issues/9) (fixed).

ris58h:
@smblott-github Any feedback?

smblott-github:
I'll get to this soon.

philc:
Related: #3226. Thanks @ris58h. I may incorporate some of this work as part of [converting](https://github.com/philc/vimium/issues/3459) coffeescript to es6.

philc:
1. I've checked in a package.json file in 1f4f4e61f0eff22c951cd251b9206030a685a143 to fix an issue with Travis builds, but I don't think it's actually needed. Our only dependency now is a utils module for node, and puppetteer for dom testing. I may remove it down the line.

2. Coffeescript is no longer a dependency, now that everything has been ported to JS (#3459).

Issue #3054, opened 2018-06-28
Close vomnibar when not focused


Troyciv:
When I open the vomnibar with `o` and (by accident) loose focus of it (e.g. trying to click a link) I cannot close the vomnibar with `ESC`. I always have to click inside the vomnibar and then press `ESC`

Is there a way to change that?

ris58h:
@Troyciv I didn't dig deep into it, but found a temporary solution: just use `tab` to focus Vombinar's input back. After that you can close it with `Esc`.

innaterebel:
I found Vimium on Chrome already does the title? What's your system information?

ris58h:
@innaterebel just middle click under Vomnibar to lose focus. I've fixed this in https://github.com/philc/vimium/pull/3057

Troyciv:
@ris58h Thank you, `tab` is a working solution for now. How ever I'd still prefer automatic close when loose focus, or something similar.

smblott-github:
This should be fixed with #3057.

PR #3057, opened 2018-07-02
Issue #3054

ris58h:
Vombinar frame's body has zero height but the frame has almost full screen height, so `document.body.addEventListener "click"` is useless. We should use `document.addEventListener "click"` instead.

### DIFF
--- a/pages/vomnibar.coffee
+++ b/pages/vomnibar.coffee
@@ -262,7 +262,7 @@ class VomnibarUI
       @input.focus()
       event.stopImmediatePropagation()
     # A click anywhere else hides the vomnibar.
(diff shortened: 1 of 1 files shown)


smblott-github:
> ... so `document.body.addEventListener()` "click" is useless.

And yet, it seems to work.  (If you take that line out, then clicking under the Vomnibar no longer closes it.)

Your proposed version and the status quo both seem to work.

ris58h:
>And yet, it seems to work.
Actually, no. And that is the issue. For me and the reporter at least. Can you reproduce it?

smblott-github:
OK.

PR #3079, opened 2018-08-10
Fixed #3068

### LINKED ISSUE #3068: Frustrating IME leftover issue when press ESC in input
Hi there, I got an annoying issue when using this fancy plugin. After searched the whole internet, finally I realized that this issue was caused by vimium.

And here, I would like to re-explain the previous issue again: https://github.com/philc/vimium/issues/2768 , with more details.

OS: OS X 10.12
Chrome: Latest and most recent versions
IME: System built-in Chinese Pinyin (Simplified)
Details:

When we need to input some Chinese or Japanese words in form, we need to use corresponding IME, like Chinese Pingyin Input Method.

Usually, when we type a whole Chinese sentence, we key in many alphabet words, then they are converted to Chinese and show in the on-the-fly area of IME, then, after we press blank space,  the Chinese chars will be sent to input box.
Easy and simple, aha? Key in alphabet words, IME convert them to Chinese chars, press blankspace, IME send the buffered Chinese to input box. just as follow:

ezgif-4-5a9f6be5d5

(Please note that before Chinese were sent to inputbox, I pressed blank space)

If we want to give up(discard) the current buffered text(usually them are long), we just need to press ESC instead of blankspace, as follow:

ezgif-4-b6c7788434

Did you see that? when I want to discard whole buffered words, I just need to press ESC.

This is the good and normal behavior by default.

BUT, when I have vimium installed in Chrome, if I discard the buffered words by pressing ESC, the following annoying thing happened:

ezgif-4-035edda4c6

Did you see that? when I press ESC to discard the buffered text, the original alphabet was left over in the inputbox, which should be cleared or discarded. This is super annoying and frustrating when composing an email or something else, coz we need to use mouse to select those leftovers then delete, or delete them one by one.
------------

I hope I described the problem clearly, it took me 50 minutes to make GIFs and reproduce this issue. but since I'm a 6+ years vimium user, I don't think 50m is a waste of time.

More than that, then I writing this issue, I found another forked plugin which resolved this bug perfectly:
https://chrome.google.com/webstore/detail/vimium%2B%2B/hfjbmagddngcpeloejdejnfgbamkjaeg

Here is the screenshot when using vimium++:

ezgif-4-1806c329a6

But again, I'm still a vimium user and want to make it better, hope this helps.
Please take a look at this issue, thank you guys so much.

fangli:
Fixed frustrating IME leftover issue when press ESC in input

### DIFF
--- a/lib/keyboard_utils.coffee
+++ b/lib/keyboard_utils.coffee
@@ -66,7 +66,10 @@ KeyboardUtils =
 
     (event) ->
       # <c-[> is mapped to Escape in Vim by default.
-      event.key == "Escape" or (useVimLikeEscape and @getKeyCharString(event) == "<c-[>")
+      # Escape with a keyCode 229 means that this event comes from IME, and should not be treat as a directly/normal ESC press.
+      # In this case, IME will take care of the ESC press event, not vimium.
+      # See https://lists.w3.org/Archives/Public/www-dom/2010JulSep/att-0182/keyCode-spec.html
+      (event.key == "Escape" and event.keyCode != 229) or (useVimLikeEscape and @getKeyCharString(event) == "<c-[>")
 
   isBackspace: (event) ->
     event.key in ["Backspace", "Delete"]


smblott-github:
Excellent job tracking that down, @fangli.  Thank you.

I'll tweak the long lines and merge in a moment.

fangli:
So in the old version, we use keyCode to check if an event is ESC or not, which works pretty good, coz if a press comes from IME, the keyCode is 229 instead of 27.

But in the new version, we use keyNames, which are always "Escape", no matter where this event comes from.

fangli:
Yes please. since I'm not a native English speaker, just tweak the comments and lines.

Issue #3055, opened 2018-06-29
Feature Request: Focus Links (without clicking)


NoisomePossum:
This was a feature I got really used to in VimFX and it would be great to have it here.

**Description**
I'd like to be able to select links using their shortcuts just as when using 'f' or 'F' but, instead of actually clicking on the link, the plugin should simply put the keyboard focus on that element.

**Use-cases**
1) On some pages you can't scroll using either the browser's native scroll functions (arrow keys) or Vimium's shortcuts because the page puts keyboard focus somewhere that isn't very useful such as a menu button. The search function sometimes works to get around this but not always. Being able to focus a specific element inside the section you want to scroll is super helpful for solving this kind of problem.

2) Drop-downs or other unfolding menus are currently a bit hard to work with because they get clicked immediately when you access them with Vimium's shortcuts. This often results in the page being reloaded.

fturcheti:
**Additional use-case**

3. Sometimes it's wise to check the destination URL of a link element. To do that the user should be able to focus the link to see its `href` in the browser status bar.

potpath:
#2174
#2188

mmikeww:
issue should be closed, #3097 works, solution is added to wiki here:
https://github.com/philc/vimium/wiki/Tips-and-Tricks#hovering-over-links-using-linkhints

philc:
Great; thank you @mmikeww.

PR #3084, opened 2018-08-11
Add a new command to focus link.

### LINKED ISSUE #2188: Feature request: Focus hint
1. Press `gf`.
2. Hints are displayed (the same as when `f` is pressed).
3. Selecting a hint in the usual way, causes the url to be displayed in the status bar (in the same way as a mouseover). (e.g. [Keysnail](https://raw.githubusercontent.com/mooz/keysnail/master/plugins/hok.ks.js) calls `elem.focus` on the url.)

Related to: #2083

potpath:
This would fix #2188 and #3055.

### DIFF
--- a/background_scripts/commands.coffee
+++ b/background_scripts/commands.coffee
@@ -167,6 +167,7 @@ Commands =
       "LinkHints.activateModeToDownloadLink",
       "LinkHints.activateModeToOpenIncognito",
       "LinkHints.activateModeToCopyLinkUrl",
+      "LinkHints.activateModeToFocus",
       "goPrevious",
       "goNext",
       "nextFrame",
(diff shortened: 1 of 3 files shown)


smblott-github:
Thanks, @potpath.

I'm not sure we should do this in this way.

You care about focusing links.  Others care about hovering, or unhovering, or yanking link text, or whatever.

We would end up with too many new commands.  I'd rather consider doing this like this...

```
map A LinkHints.activateMode
map B LinkHints.activateMode action=focus
map C LinkHints.activateMode action=hover
map D LinkHints.activateMode action=unhover
map E LinkHints.activateMode action=yank-text
```

Then, we just document the variations on the Wiki, and don't pollute the help page.

PR #2881, opened 2017-12-19
Add command options for hints (hover, unhover, copy link text).

smblott-github:
Add...

```
map X LinkHints.activateMode action=hover
map X LinkHints.activateMode action=unhover
map X LinkHints.activateMode action=text
```

The latter yanks the link's text.

Issue #2829, opened 2017-11-22
Creating a way to use the F command to hover and not click.


andrewsavill:
Many website such as Amazon use menus that need to be hovered over instead of clicked to open up more items.

If we could add in some way of using the F command to "hover the mouse" instead of clicking.

cyberhck:
I'm not sure if it'll work on every site though, some might be listening to onMouseEnter (amazon does for their menu-aim thing), some websites might be using `:hover` psuedo class, not sure if it's possible, but definitely a good feature to add.

Maybe instead of F we can go for something else? since it's a little lesser used feature than link, right? maybe on H for hover?

smblott-github:
Perhaps we should add something like...

    map X LinkHints.activateMode hover
    map X LinkHints.activateMode copyLinkText
    map X LinkHints.activateMode openInNewWindow

(and so on).

mrmr1993:
IMO better would be

```
map X LinkHints.activateMode action=hover
map X LinkHints.activateMode action=copyText
map X LinkHints.activateMode action=copyUrl
map X LinkHints.activateMode action=openWindow
```

this also opens us up to do things like

```
map X LinkHints.activateMode action=click shift=true ctrl=true
```

I'm not a big fan of the naked options, since it could really limit our flexibility later, and opens us up to nonsense like

```
map X LinkHints.activateMode hover copyLinkText
```

But yes, we should do something like this.

smblott-github:
Agreed.

everyonesdesign:
Thought about the same thing today, it'd be awesome. Any updates on this?

Trollwut:
Also want to know if someone had an idea on how to solve this :)

Also wouldn't it suffice to just place the mouse cursor above the link?

alexanderfarr:
Is there a feature to hover with Vimium already or is there a plan until when it will be supported?

gdh1995:
#3097 should have fixed this. @alexanderfarr

Trollwut:
Can someone explain how I can hover something? I fear I'm too stupid to understand it in the linked MR.

gdh1995:
&nbsp;This MR adds a new option (aka.&nbsp; parameter) to the command of LinkHints.activateMode.&nbsp; So you just need to open the option page, map a key sequence to the command and append an option of "action=hover" to the end of the line.



---Original---

duckyb:
This worked for me:
+ `map e LinkHints.activateMode action=hover`

[...]

PR #3097, opened 2018-08-17
Link-hint mode command options (incl. hover, focus)

smblott-github:
This adds:

    map X LinkHints.activateMode action=hover
    map X LinkHints.activateMode action=focus
    map X LinkHints.activateMode action=copy-text

For "hover", `<Escape>` generates the `mouseout` event.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -45,9 +45,28 @@ DOWNLOAD_LINK_URL =
   name: "download"
   indicator: "Download link URL"
   clickModifiers: altKey: true, ctrlKey: false, metaKey: false
+COPY_LINK_TEXT =
+  name: "copy-link-text"
+  indicator: "Copy link text"
+  linkActivator: (link) ->
+    text = link.textContent
+    if 0 < text.length
+      HUD.copyToClipboard text
+      text = text[0..25] + "...." if 28 < text.length
+      HUD.showForDuration "Yanked #{text}", 2000
+    else
+      HUD.showForDuration "No text to yank.", 2000
+HOVER_LINK =
+  name: "hover"
+  indicator: "Hover link"
+  linkActivator: (link) -> new HoverMode link
+FOCUS_LINK =
+  name: "focus"
+  indicator: "Focus link"
+  linkActivator: (link) -> link.focus()
 
 availableModes = [OPEN_IN_CURRENT_TAB, OPEN_IN_NEW_BG_TAB, OPEN_IN_NEW_FG_TAB, OPEN_WITH_QUEUE, COPY_LINK_URL,
-  OPEN_INCOGNITO, DOWNLOAD_LINK_URL]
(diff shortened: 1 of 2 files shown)


gdh1995:
In order to make `mouseout` and `mouseover` work well, we need to set `event.relatedTarget` before dispatching them, because some pages use this to show/hide popup menus:
image

Doc: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/relatedTarget

smblott-github:
Thanks, @gdh1995.

So, setting `relatedEvent` to what?  The parent element?

Ideally, if you could give me a commit that I can cherry pick, then that would be perfect.

gdh1995:
`relatedEvent` should be the element moving "from" or "to".

I have no enough rest time today, and if you wait a commit from me, it will be in 8 hours.

BTW, according to https://developer.mozilla.org/en-US/docs/Web/Events/mouseout , `event.detail` should be `0`  for `mouseout` nad `mousemove` events - this line should need change: https://github.com/philc/vimium/blob/d1283709ec58b952a0941bb800bf7bbdfc8534e1/lib/dom_utils.coffee#L282

gdh1995:
By the way, I haven't understood changes about `lastHoveredElement` when a new `HoverMode`  replaces an old. I'll test it on my computer in a few hours - but not now (on China, it's work time~).

gdh1995:
I've tested on http://jsfiddle.net/wUNhU/81/ , and after I hovered `#c1` and then `#c2` at once, the result was `, mouseenter, mouseover, mouseout, mouseleave, mouseenter, mouseover, mouseout, mouseleave`:

image

Explain: The last `mouseout` (which is wrong) is dispatched when the previous `HoverMode` exited, but during the `mouseover` on `#c2` Vimium also dispatched a `mouseout`.

A plain idea: rework HoverMode into:
*  do nothing and exit if any key except `Escape`
* dispatch mouseout and exit if a first key is `Escape`

cadillion:
Are there any reasons why this PR hasn't been merged? I'm interested in using the feature set and would love to offer any fixes that may be required.

gdh1995:
Seems to replace #2881 and solve #2829.

horak:
Just bumping @cadillion's comment—is there anything preventing this from merging?

4fthawaiian:
any movement on this?

dacodas:
I would love this feature :)

philc:
Great work @smblott-github! Works great.

This functionality is pretty undiscoverable -- we need a better way to document command parameters (e.g. #2827).

For now I've documented it in [our wiki](https://github.com/philc/vimium/wiki/Tips-and-Tricks#hovering-over-links-using-linkhints), and will link to this wiki entry in the release notes when this hits the store.

eyalev:
I see the option to copy the link text with `action=copy-text`

Is there an option to copy the link URL?

gdh1995:
@eyalev There has been a command named `LinkHints.activateModeToCopyLinkUrl` for this.

PR #3119, opened 2018-08-31
Handle scrolling on Reddit redesign

marcotc:
This PR adds support for scrolling the through the new Reddit post popups, introduced during the ongoing Reddit redesign effort.

Before scrolling with j and k would scroll the background, instead of the foreground article:

before

With the changes in this PR, it will now scroll the popup post correctly:

after

## Implementation notes

Unfortunately `document.scrollingElement` will always return the document's body, meaning we never reach our fallback codes that could handle edge-cases like this.

The solution implemented in this PR is ad-hoc, which is subpar to say the least. It is based on a recent similar fix for Twitter.

If this solution is acceptable for now, I'm willing to refactor these two ad-hoc to reduce code duplication.
I could move them into an external configuration file instead, as the only variables are the URLs affected, and how to find the scrollable element.

### DIFF
--- a/content_scripts/scroller.coffee
+++ b/content_scripts/scroller.coffee
@@ -319,6 +319,15 @@ if DomUtils.isTopFrame() and window.location.host == "twitter.com"
         activatedElement = element ? getScrollingElement()
         func arguments...
 
+if DomUtils.isTopFrame() and window.location.host in ["reddit.com", "new.reddit.com"]
+  for method in ["scrollTo", "scrollBy"]
+    do ->
+      func = Scroller[method]
+      Scroller[method] = ->
+        element = document.getElementById "overlayScrollContainer"
+        activatedElement = element ? getScrollingElement()
+        func arguments...
+
 root = exports ? (window.root ?= {})
 root.Scroller = Scroller
 extend window, root unless exports?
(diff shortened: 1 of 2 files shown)


smblott-github:
This is more or less what I was going to try to get to.  Reddit is sufficiently important to warrant a site-specific hack.

> If this solution is acceptable for now, I'm willing to refactor these two ad-hoc to reduce code duplication.

Yes.  Please refactor to eliminate duplication.  Perhaps with configuration via a dictionary mapping hostnames to CSS.  No need to go so far as an external configuration file just yet.  It's enough to have the code isolated from the real scrolling logic at the bottom of `scroller.coffee`.

marcotc:
Thank you for the review @smblott-github, I'll make changes to implement your feedback.

smblott-github:
Thanks, @marcotc.

> It seems to me that any scrolling state should be reset when history changes...

That's an interesting observation.  You're right.

I made a couple of tweaks (72aaf054201755f837ba84f26a4d74846a81ffba).  Let me know if you spot a mistake.

Once again, thanks.

marcotc:
Hi @smblott-github, thank you for merging this.
Seems fine to me, thanks again for all the help!

marcotc:
Confirmed regression, almost certain Reddit has changed their layout.

tehp:
Bump, also experiencing this issue.

kitaev:
Doesn't work for me with Firefox 65.0.2 on Linux. Vimium version 1.64.3.

sooriravindra:
Doesn't work on Chrome Version 73.0.3683.103 (Official Build) (64-bit)

dseeni:
same here as well, using the 74.0.3729.108 build of chorme

Issue #3130, opened 2018-09-11
Escaping '/' for search engine completion


rperryng:
I find that using [searching engine completion](https://github.com/philc/vimium/wiki/Search-Completion) is a very convenient way of building shortcuts for long URLs, e.g. to visit a specific subreddit I could do

```
r: https://www.reddit.com/r/%s Reddit
```

and type `r games` in the omnibar to visit the games subreddit.

Sometimes, there are URLs in which I would like to use a `/` token, e.g.

```
gh: https://www.github.com/%s
```
so that I could do `g pulls` to visit my pulls, but also `g rperryng/dotfiles` to visit a repo within an org/user.  However when trying this the string is URIencoded into `rperryng%2Fpulls`.

It seems like this [functionality is built in](https://github.com/philc/vimium/blob/17fb86e02971948f90215934cbe5417674364a29/lib/utils.coffee#L133), in some way, or at least implemented internally but I couldn't understand from the wiki pages or code if this can be achieved via the options.  Any help appreciated, thanks!

smblott-github:
So, you're saying that we should only apply `encodeURIComponent` if the query contains non-URI characters?  That sounds reasonable.

gdh1995:
I suggest that we use `%S` (upper-case 'S') to represent a "raw" query and not encode it, and keep `%s` to be URI-encoded.

gdh1995:
Oops, I find an old issue #2580 which is also about this.

smblott-github:
Fixed in #3136.

Issue #2580, opened 2017-07-28
Multi-word custom search with + separator breaks some engines


dagerikhl:
### Issue

I've discovered the issue while using a custom search engine, [Prisjakt](https://www.prisjakt.no). The issue persist on the UK version [Pricespy](https://pricespy.co.uk). When replacing `%s` with the search terms provided, Vimium inserts a `+` (encoded: `%2B`) between all words. This is fine for many engines, but some, such as this, provides different results when using a space ` `/`%20` versus using a plus `+`/`%2B`.
And for this site, using spaces provides better search results than using plus-signs.

A similar issue has been raised previously in #1334, but the fix introduced in #1384 introduced the `+` that's creating this difference.

### Steps to reproduce

1. Add custom search engine Pricespy: `p: https://pricespy.co.uk/#rparams=ss=%s Pricespy`
2. Search with `p playstation 4` in the vomnibar
3. Observe that the resulting search-query becomes `playstation+4`.
4. Observe that changing it to `playstation 4` changes the results (for the better).

### Suggested solution 1

Use spaces (`%20`) as separators for search terms. Not sure how this will affect other search engines.

### Suggested solution 2

Allow users to choose between tow searchterm variables: `%s` and `%S`, where one inserts plus-signs and the other spaces as separators.

### Suggested solution 3

Allow users to specify a custom separator for their custom search engine, i.e. as `p: https://pricespy.co.uk/#rparams=ss=%s ( ) Pricespy`, where `( )` specifies the separator (here, spaces).

### System information

* **OS:** Windows 10 Pro 64bit v. `10.0.15063`
* **Browser:** Google Chrome 64bit v.  `59.0.3071.115`
* **Vimium:** v. `1.59`

smblott-github:
Suggested solution 1 is probably a non-starter; for sure it would break stuff for other users.

Solution 2 is an interesting idea.

Solution 3 seems most general.

Let's leave this for a bit and see if anybody else reports the same or similar issues.

gdh1995:
You may try my Vimium++ (https://chrome.google.com/webstore/detail/vimium%2B%2B/hfjbmagddngcpeloejdejnfgbamkjaeg), and it support `%s` or `%S` (aka `$s`, `$S`) to use `+` or `(space)` - `%S` means to concat URL using raw spaces.

Besides, it supports a simple "template" syntax:
* `$S{_}` means to concat query words using `_`, and so does `$S{/}`
[...]

PR #3136, opened 2018-09-14
Dont encode '%S'

rperryng:
Hello, this addresses https://github.com/philc/vimium/issues/3130 / https://github.com/philc/vimium/issues/2580

It adds support for using `%S` instead of / in addition to `%s` in search engine completion.
`%S` tokens will not be URI encoded.

Example use-case:

```
gh: https://www.github.com/%S Github
```

and using `philc/vimium`to resolve to `https://www.github.com/philc/vimium` instead of `https://www.github.com/philc%2Fvimium`


dagerikhl:
Awesome! Thanks for following through with this :D

smblott-github:
Merged.

Thanks for this.

Note f2ab8bee6e1fa8d9b34837c934e6b8ec0cb7abc9.

PR #3175, opened 2018-10-28
use Shadow DOM v1 if it exists

gdh1995:
Chrome 70 has deprecated Shadow DOM v0
    and displayed a warning when it's used.
And accroding to https://www.chromestatus.com/features/4507242028072960,
    Shadow DOM v0 will be removed since Chrome 73.

The support for Shadow DOM v1 has been there since Chrome 53 and
    is enabled by default on Firefox 63,
so it's time to use it.

Old support is still needed because Vimium's min_chrome_version is 51.

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -27,7 +27,7 @@ class UIComponent
         seamless: "seamless"
       shadowWrapper = DomUtils.createElement "div"
       # PhantomJS doesn't support createShadowRoot, so guard against its non-existance.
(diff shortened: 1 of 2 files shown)


smblott-github:
@gdh1995 Thanks for this.  I came looking for a fix for exactly this.

(So you and others know... I have been extraordinarily tied up with work commitments for the last few months, hence my almost total absence here.  I hope to catch up with Vimium needs in the next couple of weeks.  Sorry about that.)

Issue #3146, opened 2018-09-24
Firefox: 'f' no longer works in GMail


raboof:
Finding links with 'f' used to work fine for me in GMail, but now it has switched to the 'new layout' it no longer works for me.

Hitting 'f' I see a Vimium status message saying "Collecting hints...", then nothing.

Vimium version: v1.64
Firefox version: 61.0.1

raboof:
Ah, spoke too soon: it works on the 'inbox' page, but when viewing an email/thread.

raboof:
From the addons debugger console:
```
style is null dom_utils.js:369
getViewportTopLeft
moz-extension://8d5de73c-e1cb-42fd-a860-e1b730f78724/lib/dom_utils.js:369:1
getLocalHints
moz-extension://8d5de73c-e1cb-42fd-a860-e1b730f78724/content_scripts/link_hints.js:1141:14
getHintDescriptors/</<
moz-extension://8d5de73c-e1cb-42fd-a860-e1b730f78724/content_scripts/link_hints.js:124:88
onLoaded
moz-extension://8d5de73c-e1cb-42fd-a860-e1b730f78724/lib/settings.js:69:16
getHintDescriptors/<
moz-extension://8d5de73c-e1cb-42fd-a860-e1b730f78724/content_scripts/link_hints.js:121:18
DomUtils.documentReady
moz-extension://8d5de73c-e1cb-42fd-a860-e1b730f78724/lib/dom_utils.js:23:18
getHintDescriptors
moz-extension://8d5de73c-e1cb-42fd-a860-e1b730f78724/content_scripts/link_hints.js:119:14
linkHintsMessage
moz-extension://8d5de73c-e1cb-42fd-a860-e1b730f78724/content_scripts/vimium_frontend.js:279:14
init/<
moz-extension://8d5de73c-e1cb-42fd-a860-e1b730f78724/content_scripts/vimium_frontend.js:316:20
apply self-hosted:4613:5 applySafeWithoutClone
resource://gre/modules/ExtensionCommon.jsm:234:16
addListener/asyncWithoutClone/<
resource://gre/modules/ExtensionCommon.jsm:1918:20
```

gdh1995:
Confirmed on Firefox 61.0.1 x64 on Windows 10, but Firefox 62 has no such a problem.

[...]

PR #3185, opened 2018-11-15
Avoid crash when document style not found

raboof:
Fixes crash when using 'f' in gmail on Firefox, #3146.

The gmail inbox view would initially work, but after finishing loading crash
vimium. Debugging revealed while gmail worked, 'f' would trigger 2 calls of
getViewportTopLeft. After loading the page completely, 'f' would trigger 3
calls of getViewportTopLeft, and crash on the third because 'style' was null.

After handling the case of the 'style' being null (and assuming a viewport top
left location of '0,0' in that case), 'f' works again even when gmail is fully
loaded

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -328,19 +328,22 @@ DomUtils =
     box = document.documentElement
     style = getComputedStyle box
     rect = box.getBoundingClientRect()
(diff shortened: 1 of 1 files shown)


smblott-github:
Closing based on [this comment](https://github.com/philc/vimium/issues/3146#issuecomment-440659408).

PR #3188, opened 2018-11-19
Document how to debug under firefox

raboof:
Needs an application ID to use local storage, will fail otherwise.


smblott-github:
Thank you.  I'll take your word that this is correct.

PR #3111, opened 2018-08-27
use Element::scrollBy if it exists

gdh1995:
This is to fix #3034 , #3041 and #3107 .

Explanation: https://github.com/philc/vimium/issues/3041#issuecomment-415346708

PR #3201, opened 2018-12-16
scroller.coffee(performScroll): Fix `h` `l` horizontal scrolling

teddywing:
When scrolling horizontally, `axisName` has value `"scrollLeft"`. Since
this line checks against value `"x"` it would always fail, causing `h`
and `l` to scroll vertically up and down respectively.

Use the `direction` argument in the condition instead, which is set to
`"x"` when scrolling horizontally with `h` and `l`.

This fixes horizontal scrolling using `h` and `l`.

### DIFF
--- a/content_scripts/scroller.coffee
+++ b/content_scripts/scroller.coffee
@@ -58,7 +58,7 @@ performScroll = (element, direction, amount) ->
   before = element[axisName]
   if typeof element.scrollBy is "function"
     scrollArg = behavior: "instant"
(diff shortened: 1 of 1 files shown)


gdh1995:
Oops, thanks for your fix. This should solve my typo in #3111.

@smblott-github Sorry for disturbing you, but maybe Vimium needs to merge this PR and bumps into a new release.

teddywing:
Thanks for the note @gdh1995

PR #3206, opened 2018-12-24
Add 'Bookmarks blacklist' setting

cjbassi:
Allows specifying bookmark folders whose contents should not be included in the bookmark vomnibar search

### DIFF
--- a/background_scripts/completion.coffee
+++ b/background_scripts/completion.coffee
@@ -244,13 +244,14 @@ class BookmarkCompleter
 
   # Recursive helper for `traverseBookmarks`.
   traverseBookmarksRecursive: (bookmark, results, parent={pathAndTitle:""}) ->
-    bookmark.pathAndTitle =
-      if bookmark.title and not (parent.pathAndTitle == "" and @ignoreTopLevel[bookmark.title])
-        parent.pathAndTitle + @folderSeparator + bookmark.title
-      else
-        parent.pathAndTitle
-    results.push bookmark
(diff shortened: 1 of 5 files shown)


gdh1995:
An interesting idea. Then do we need a "history blacklist" to filter out some history items?

cjbassi:
That wouldn't be a bad idea either. It would be implemented slightly differently though since that would have to be some sort of pattern match.

smblott-github:
This seems quite niche.  What's the use case?

cjbassi:
Well blacklisting the porn folder for one, not gonna beat around the bush on that one :laughing:. But it's also good for blacklisting folders of random internet snippets/articles that can accumulate and get in the way of the standard websites people bookmark and frequently visit like youtube, reddit, etc.

gdh1995:
Will it be needed to re-enable searching for those text, when a user type "/path/to/banned/folder" ?

cjbassi:
@gdh1995 That's not a bad idea. If this is going to get merged then I could implement that if we want to. It's just that the logic would have to be changed a fair bit because that requires dynamic blacklisting during querying rather than during indexing of the bookmarks which is what is currently being done.

smblott-github:
I still think this is pretty niche.  Let's just leave it for the moment and see whether anybody comes up with a compelling use case.

cjbassi:
@smblott-github The main use case for this is more than likely going to be for nsfw folders, which I think is a compelling enough use case on its own imo.

philc:
Closing as too niche to be in the main Vimium distribution, IMO. Thanks for prototyping the idea!

PR #3209, opened 2018-12-26
Add option to enable Javascript URLs

### LINKED ISSUE #3178: Javascript bookmarks not working anymore
My javascript bookmark use to work on Chrome 69, but it doesn't on version 71.0.3578.30.

Its content is:
`javascript:(function()%7B(function () %7Bvar i%2C elements %3D document.querySelectorAll('body *')%3Bfor (i %3D 0%3B i < elements.length%3B i%2B%2B) %7Bif (getComputedStyle(elements%5Bi%5D).position %3D%3D%3D 'fixed'||getComputedStyle(elements%5Bi%5D).position %3D%3D%3D 'sticky') %7Belements%5Bi%5D.parentNode.removeChild(elements%5Bi%5D)%3B%7D%7D%7D)()%7D)()`

https://github.com/philc/vimium/pull/3167 might solve this.

smblott-github:
This is an idea for resolving #3178 (and includes #3167).

1. Wholly disabling Javascript would be unpopular.
2. Apparently as of Chrome 71, `chrome.tabs.update` can no longer be used to open Javascript URLs (hence #3178).
3. #3167 fixes this by using `chrome.tabs.executeScript` instead.
4. Unfortunately, with #3167 the Javascript now executes in Vimium's context, hence with Vimium's elevated permissions.  To enable #3167 behind users' backs would be irresponsible.

So this PR proposes:
- Take #3167 as is, however...
- Disable Javascript URLs, but add an option on the options page for users to re-enable them.

The option looks like this:

image

The default (which everybody will inherit) is `false`.

Fixes #3178.
Fixes #3167.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -97,7 +97,10 @@ do ->
 TabOperations =
   # Opens the url in the current tab.
   openUrlInCurrentTab: (request) ->
-    chrome.tabs.update request.tabId, url: Utils.convertToUrl request.url
+    if Utils.hasJavascriptPrefix request.url
+      chrome.tabs.executeScript request.tabId, {code: request.url}
+    else
+      chrome.tabs.update request.tabId, url: Utils.convertToUrl request.url
 
   # Opens request.url in new tab and switches to it.
   openUrlInNewTab: (request, callback = (->)) ->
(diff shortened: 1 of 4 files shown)


gdh1995:
We may send the URL to content scripts and then vimium_frontend injects a `<script>` to run target code. 
The injection is safe enough even on sandboxed pages, since Chrome 68 (actually, since https://github.com/chromium/chromium/commit/5a5267ab58dd0310fc2b427db30de60c0eea4457)

smblott-github:
> We may send the URL to content scripts and then vimium_frontend injects a

That sounds better.

PR #3222, opened 2019-01-09
findNext: check last FindMode query correctly

### LINKED ISSUE #3217: `n` command does not search for previous search term
1. Search for `abc` in a tab using `/` command.
2. Go to other tab and press `n`.
3. It shows `no query to find` error.
4. It used to search for the term a week or two earlier.

Env:
vimium: Version 1.64.3
chrome on macos: Version 71.0.3578.98 (Official Build) (64-bit)

gdh1995:
This is to fix #3217 and replace commit 7c77b9bef6720283b04f6253b5e3cec0c7612b15 - the commit checked `FindMode.query.rawQuery` directly and might mistakenly ignore find mode history.

### DIFF
--- a/content_scripts/mode_find.coffee
+++ b/content_scripts/mode_find.coffee
@@ -56,6 +56,7 @@ class PostFindMode extends SuppressPrintable
 class FindMode extends Mode
   @query:
     rawQuery: ""
+    parsedQuery: ""
     matchCount: 0
     hasResults: false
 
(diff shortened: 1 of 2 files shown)


smblott-github:
Thanks, @gdh1995.

Have you left `parsedQuery` in here by mistake?  (It seems not to be used.)

gdh1995:
No. Although it's still correct when `parsedQuery: ""` was not added, I want to use it to ensure `getQuery` always returns a string (but not `undefined`).

> Thanks, @gdh1995.
> 
> Have you left `parsedQuery` in here by mistake? (It seems not to be used.)

0xC0FFEE:
Is there anything I could do to help get this merged? I switched over from cVim and would really appreciate if this feature would work in Vimium too.

PR #3226, opened 2019-01-11
Use Yarn for dependency management and running tasks

dtinth:
This PR aims to modernize the build system a little bit, to make it easier for new contributors (me included) to hack on and contribute to this project. Goals:

- To reduce the amount of manual steps required to set up a development environment.
- …while keeping all the existing build commands (cake) intact.
- To reduce the amount of global npm dependencies.
- To have a better integration with VS Code.

What I did:

1. Instead of requiring users to install coffeescript globally (`npm install -g coffeescript`) they only have to run `yarn`. This will:

   - Install the necessary dependencies (except PhantomJS) locally.
   - Run `git submodule update` to pull in `shoulda.js` (via `postinstall` script).
   - Run `cake build` (via `prepare` script).

   After this the project is immediately ready to be installed as a Chrome extension. It’s also ready to be tested (using `yarn test`) if the user have PhantomJS installed.

   The `cake` command is available as `yarn cake`. (`yarn` behaves somewhat like `bundle exec` here, allowing developers to run locally-installed CLI apps.)

2. Since the project now follows npm script conventions the Travis CI configuration has been simplified.

3. Relevant documentation has been updated (in CONTRIBUTING.md).

   - Steps has been reduced.
   - Make clear in the docs that PhantomJS v1.9.20 is required (the DOM tests fail in latest version of PhantomJS, but this is out of scope of this PR.)

4. Added `.vscode/tasks.json` for better integration with VS Code. When opening the project in VS Code, pressing Cmd+Shift+B will launch the `autobuild` task.

   image

Related: #3029.


smblott-github:
I agree that the current build system is showing its age.  However, is this PR really the direction to go?

I'm on Debian, so -- on a fresh clone -- I'd like to just say...

```
npm install .
make build
make test
make autobuild
```

That would be pretty easy to do and would work on Mac too, but probably wouldn't be very Windows friendly.

dtinth:
I cannot say about whether this is "the" direction to go, but I would say it is a step forward :smile:. At the same time I try to make this PR fully backwards-compatible; it should have no impact for existing maintainers, but would improve the experience for new contributors.

This is the proposed workflow in this PR:

```
yarn install      # This also runs `git submodule update` and `cake build`
yarn test
yarn autobuild
```

There are 3 parts in this PR:

1. `package.json`
2. Yarn or npm (`yarn.lock` or `package-lock.json`)
3. `.vscode/tasks.json`

Here are my opinions:

1. `package.json` is definitely a way to go.

    - This allows all the dependencies used to be declared in a single file, which reduces having to install multiple packages separately from npm into a single command. (But still, I have trouble getting a PhantomJS@1.9.20 installed locally — the 3-year-old package doesn’t seem to support local installs, so I kept it global for now.)

    - [Locally installed packages is now more preferred than globally installed packages in the JavaScript community.](https://codeburst.io/maybe-dont-globally-install-that-node-js-package-f1ea20f94a00)

2. Yarn vs npm: I don’t have a strong opinion here, but I chose Yarn for this PR because:

    - The `yarn` command lets you easily run locally installed packages. Instead of `./node_modules/.bin/cake`, we can use `yarn cake`.

    - `yarn` is also [preferred in VS Code’s codebase](https://github.com/Microsoft/vscode/wiki/How-to-Contribute#build-and-run).

3. `.vscode/tasks.json` 

    - [VS Code is now the most popular development environment as of 2018, according to StackOverflow developer survey](https://insights.stackoverflow.com/survey/2018/#development-environments-and-tools).

It would be great to hear some more second opinions. 😃

smblott-github:
> It would be great to hear some more second opinions.

Indeed.

philc:
Thanks for proposing this. I'm keeping this in mind as we [modernize](https://github.com/philc/vimium/issues/3459) the codebase.

Soon the coffeescript dependency and associated cake tasks will be gone. Then, the only other packages required will be whatever we need for running tests. This might make yarn less added value. I'd love to retire cake for a Makefile, but I'm not sure that will happen. There's a decent hunk of logic in the cakefile related to test running and code coverage.

philc:
The cakefile is now gone, as is the transpilation and autobuilding step, so we now have very few dependencies and no build steps. As a result, we don't need to add something as capable as `yarn`.

PR #3244, opened 2019-02-13
Allow css customisation of blank.html

FelixWeis:
I like dark themes, so this could allow setting a custom dark background on vimium's NewTabPage.


gdh1995:
First, a raw `<script>...</script>` is not allowed in HTML files, and we have to put JS code in *.js file and refer it using `<script src="*.js"></script>".

Second, I think it's just enough if you write a black page and make it Vimium's New tab URL.

FelixWeis:
okay thanks

PR #3246, opened 2019-02-16
delay window.focus() when needed

### LINKED ISSUE #3242: Focus switches randanly after using Vomnibar.activateInNewTab command in chrome 72.
Issue appears on windows 10 64 bit chrome version 72.0.3626.96.
Previous version 71.0.3578.80 is OK.
After openning new tab through vomnibar, focus mostly stays on current tab, but also switch to the newly opened tab sometimes like previous version always do.

gdh1995:
On Chrome, if an extension's content script call window.focus()
  during Port.onMessage handlers, then Chrome will activate the tab.
Therefore there's a race condition between:
* `tabs.update(tabId, {active:true}` in `main.coffee`
* and `focusThisFrame -> window.focus()` in `vimium_frontend.coffee`

In my tests, a `setTimeout` is just enough to avoid this activating, so this PR should fix #3242.

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -93,7 +93,7 @@ class UIComponent
                 handler: "sendMessageToFrames",
                 message: name: "focusFrame", frameId: @options.sourceFrameId, forceFocusThisFrame: true
             else
-              Utils.setTimeout 0, -> window.focus()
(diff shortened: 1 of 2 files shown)


smblott-github:
Thanks, @gdh1995.

PR #3264, opened 2019-03-12
translate (w/ google) selected text with "T"

iddoeldor:


philc:
Cool idea and minimal implementation. At the moment, I think this is outside the scope of this extension.

PR #3282, opened 2019-03-31
add 'dblclick' event to simulateClick

### LINKED ISSUE #3281: Feature Request: doubleclick
Add functionality for doubleclicking. 

The only place I can find where this is useful is spotify web player, but presumably there are other places where this would be useful. The problem I have on spotify is that you need to doubleclick tracks to get them to play. If I do 2f[keys], it doesn't have the desired effect. I'm assuming because the sequence
`["mouseover", "mousedown", "mouseup", "click"]`
is repeated, and `"dblclick"` doesn't happen.

I've hacked about a bit in the source code, and this code does what I want:

```    mouseEvent = new MouseEvent('dblclick', {
        'view': window,
        'bubbles': true,
        'cancelable': true
    });
    element.dispatchEvent(mouseEvent)```

Maybe this could be implemented as 2f[keys] or f2[keys]?

lukebryant:
Would close #3281.

The rationale behind this is that most places aren't listening for
'dblclick', and the ones that are (https://open.spotify.com is the only
one I can find) don't respond to single clicks.

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -247,7 +247,7 @@ DomUtils =
             element.setSelectionRange element.value.length, element.value.length
 
   simulateClick: (element, modifiers = {}) ->
-    eventSequence = ["mouseover", "mousedown", "mouseup", "click"]
+    eventSequence = ["mouseover", "mousedown", "mouseup", "click", "dblclick"]
     for event in eventSequence
       defaultActionShouldTrigger =
         if Utils.isFirefox() and Object.keys(modifiers).length == 0 and event == "click" and


gdh1995:
Well, the "dblclick" event should not be dispatched by default.

We need to care about the events and make them more like the native process.

If there's a command option to toggle (enable) it at runtime, it would be better.

(BTW, as for the necessary of this feature, I'm still doubting.)

lukebryant:
Yeah, I agree this PR doesn't make much sense in its current state. I just found this quick fix that works for my purposes, so I thought I'd put it here to illustrate the issue.

philc:
Thanks for the commentary. Closing, as this isn't actionable in its current state.

PR #3288, opened 2019-04-05
Scroller: listen click events on Firefox

### LINKED ISSUE #3287: fishshell.com partial functionality
Hello,
I've been visiting http://fishshell.com/docs/current/index.html and vimium scrolling doesn't seem to be working with Firefox. The site seems to have two frames, a side menu and a main one. By default only side menu is chosen (which scrolling works) but any attempt to switch to main with `gf`or mouse tapping does not seem to be working.

gdh1995:
on Firefox, there're no DOMActivate events, so click is needed

This should fix #3287


philc:
LGTM, thanks!

PR #3343, opened 2019-06-04
Moved Release Notes

JonathanGrocott:
Working on a class project for inclusivity issues. Limiting the amount of information on the front page README.md will help others with different information processing styles not be overwhelmed.


ght:
The most commonly used name seems to be `CHANGELOG.md`. The [Keep a Changelog](https://keepachangelog.com/) site collected some related best practices.

On a sidenote, the way you first added an edited version of the changelog (with all the leading `-` of the bullet points removed), then removed the lines from the Readme, and finally changed the `RELEASENOTES` to the original form in three separate commits breaks gits "moved/copied lines" detection, which helps keeping `git blame` useful for files that have been renamed or lines that have been moved/copied between files.

If this change is accepted, it would be cleaner to rewrite your commits (adding a commit to rename the new file, then squashing them all into one should be enough). You can check by doing `git blame -C CHANGELOG.md` (check `man git-blame` for details), and it should blame the lines in the new file on the commits that originally added them to `README.md`. It's not really important for documentation, but why not do it properly while you're already at it.

philc:
This was added in ab3c49d187478b1dddb11e55cd3e248bb2263cc1. Thanks for the idea!

PR #3346, opened 2019-06-05
Disabled Custom CSS for Vimium UI option

bergsm:
This commit disables the Custom CSS for link hints feature from the
advanced options menu, but leaves the code in place.

I ran the tests mentioned in CONTRIBUTING.md and received results identical to the current version. 

This should resolve issue #636. It's an old issue, so I understand if it is no longer relevant.

Thanks for your consideration and let me know if I need to make any changes!

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -33,6 +33,7 @@ Contributors:
   Murph (github: pandeiro)
   Niklas Baumstark <niklas.baumstark@gmail.com> (github: niklasb)
   rodimius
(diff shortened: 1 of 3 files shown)


philc:
Thanks for putting your eyes on a "help-wanted" issue. I'm going to close this for now, because I don't want to deal with removing something existing people might need, and I think hiding it for new users isn't a great option (because we still have the code complexity -- we're just reducing the UI complexity).

I'll reference this PR from the issue.

Issue #3380, opened 2019-08-21
Uncaught TypeError


Beasta:
Uncaught TypeError: Cannot read property 'length' of null

Context
_generated_background_page.html

Stack Trace
lib/utils.js:422 (anonymous function)

Details:
crashed when I used onetab to collapse all my open tabs

Browser and os:
OS: OSX
Browser: Brave Version 0.67.125 Chromium: 76.0.3809.100 (Official Build) (64-bit)

![Screen Shot 2019-08-21 at 12 33 45 PM](https://user-images.githubusercontent.com/6305836/63462577-16a5db80-c410-11e9-9406-c6e3df510e35.png)

philc:
We think this is fixed in master, which will be released to the store in 1-2 days.

PR #3381, opened 2019-08-22
fix a crash on network errors

gdh1995:
I think #3380 is because some code might call `fetch's callback` in `AsyncDataFetcher` twice, and after reviewing code I find such a bug may be to blame.

Although this fix may be not enough, it does no harm and can be merged directly.

### DIFF
--- a/background_scripts/completion_search.coffee
+++ b/background_scripts/completion_search.coffee
@@ -48,18 +48,12 @@ CompletionSearch =
     xhr = new XMLHttpRequest()
     xhr.open "GET", url, true
     xhr.timeout = 2500
-    finished = 0
-    # Note(gdh1995): Maybe the 2 events below are not needed, but I just leave it unchanged for robustness
-    xhr.ontimeout = xhr.onerror = ->
-      finished++
-      if finished == 1
-        callback null
-
+    # According to https://xhr.spec.whatwg.org/#request-error-steps,
+    # readystatechange always gets called whether a request succeeds or not,
+    # and the `readyState == 4` means an associated `state` is "done", which is true even if any error happens
     xhr.onreadystatechange = ->
       if xhr.readyState == 4
-        finished++
-        if finished == 1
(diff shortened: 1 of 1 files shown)


background_scripts/completion_search.coffee
```diff
@@ -48,12 +48,19 @@ CompletionSearch =
     xhr = new XMLHttpRequest()
     xhr.open "GET", url, true
     xhr.timeout = 2500
-    xhr.ontimeout = xhr.onerror = -> callback null
-    xhr.send()
+    finished = 0
```
philc: On inspection, it seems to me that callback should never be called twice. The only way it can be called twice is if the definition of the ontimeout and onerror events overlap (meaning onerror fires when a timeout occurs).

If that's true, then the simpler fix to this problem is to just remove the ontimeout handler, assuming onerror is the more generic event.

background_scripts/completion_search.coffee
```diff
@@ -48,12 +48,19 @@ CompletionSearch =
...
+    # Note(gdh1995): Maybe the 2 events below are not needed, but I just leave it unchanged for robustness
+    xhr.ontimeout = xhr.onerror = ->
+      finished++
+      if finished == 1
+        callback null
 
     xhr.onreadystatechange = ->
       if xhr.readyState == 4
-        callback if xhr.status == 200 then xhr else null
+        finished++
```
philc: Debatable, but I think this bookkeeping is clearer as a bool, because it explicitly communicates that we care only about calling the callback once. `finished` indicates that we care about how many times it was called, until you carefully read how it's used.

So something like this:

```
callbackCalled = false

xhr.ontimeout = ->
  if !callbackCalled
    callback null
  callbackCalled = true
```

gdh1995:
I've added a comment to refer the specification, and then `onerror`/`ontimeout` seem useless so I remov them.

Issue #3157, opened 2018-10-07
vomnibar color 


ralphee:
how can i change the colour of the vomnibar in firefox?

shibumi:
You need to write CSS for it.. (at least thats how It works in chromium)

gdh1995:
Steps: on Vimium's options page, click "show advanced options" at the page bottom, and then write CSS in the textarea of "CSS for Vimium UI".

shibumi:
Here is my current configuration:
```
/* Arc Dark theme */

/* Link hint boxes */
div > .vimiumHintMarker {
  background: #383c4a;
  font-size: 12px;
  border: 0.25em solid #383c4a;
  border-radius: 0.68em;
  box-shadow: 0em 0.1em 0.6em 0.1em rgba(0, 0, 0, 0.4);
}

/* Link hint text */
div > .vimiumHintMarker span {
  color: #e2e3e7;
  font-size: inherit;
  text-shadow: none;
}

/* Link hint matching characters */
div > .vimiumHintMarker > .matchingCharacter {
  color: #5294e2;
}

input#vomnibarInput.vimiumReset {
	color: #FFF !important;
  background-color: #111 !important;
}

#vomnibar {
   border-color: black;
   background-color: black;
}

#vomnibar .vomnibarSearchArea {
   background-color: #222;
   border-bottom: #3f3f3f;
}

#vomnibar input {
   border-color: #393939;
}

#vomnibar ul {
    background-color: #222;
}

#vomnibar li {
   border-bottom: #3f3f3f;
}

#vomnibar li .vomnibarUrl {
   color: #5294e2;
}

#vomnibar li .vomnibarMatch {
   color: white;
}

#vomnibar li.vomnibarSelected {
    background-color: #181D24;
}

#vomnibar li em .vomnibarMatch, #vomnibar li .vomnibarTitle .vomnibarMatch {
   color: white;
}

#vomnibar li em, #vomnibar li .vomnibarTitle {
   color: #ddd;
}
```

shibumi:
Maybe we should open a `hub` for color schemes in the vimium wiki

nadya-p:
Alternatively, [Dark Reader](https://chrome.google.com/webstore/detail/dark-reader/eimadpbcbfnmbkopoojfekhnkhdbieeh?hl=en) chrome extension darkens everything including the vomnibar.

shibumi:
@nadya-p last time When I've tried this it didn't work..

[...]

PR #3402, opened 2019-10-05
Add dark mode option

mwanchap:
I use Vimium pretty heavily, but it's quite bright and I love dark modes, so I thought I'd try adding one to fit in Chrome's dark mode. This is my first time working on a Chrome extension or with coffeescript, so would very much appreciate any feedback. Anyway, this PR applies darker colours to the Vomnibar, help, options and blank tab pages, if an option is checked in Vimium's options menu
 - Dark mode checkbox in options
 - Put all the style overrides in /pages/darkMode.css
 - Inject darkMode.css into Vomnibar, help, options and blank tab pages if dark mode option is checked, similar to how userDefinedLinkHintCss is injected

Ideally, I'd much rather use Chrome's theme colours, and then Vimium would fit in with _any_ custom themes the user has installed. But I couldn't figure out how to do that...


gdh1995:
Good job.

BTW, Chrome's theme colors are not available.

mwanchap:
Ah, that explains why I couldn't figure out how to do it.  The closest thing I could find was
the `chrome.devtools.panels.themeName` property, but that's just for Chrome's dev tools panels, and doesn't seem to be accessible anyway

mwanchap:
Err, I somehow completely forgot to add dark mode to the new tab page...

mwanchap:
@gdh1995 any idea what's up with the Travis builds? It looks like a call to path.join in phantom_runner is failing because util.isString is undefined, even though it's definitely defined in the package source. I had a look into it but I didn't get very far, all I could figure out was the below.
```coffeescript
path = require 'path'
util = require 'util'
console.log 'path.join is ' + path.join  // prints "path.join is function () {...."
console.log 'util is ' + util  // prints "util is [object Object]"
console.log 'util.isstring is' + util.isString('asdf') // prints "TypeError: 'undefined' is not a function (evaluating 'util.isString('asdf')')"
```

philc:
If we implement dark mode with an invert filter, we'll need to exclude favicons from that filter (or apply an inverse filter) now that favicons can show in the vomnibar for currently open tabs (#2878).

mwanchap:
Hey Phil, thanks for all the feedback. I was thinking the same thing about having to maintain all the extra CSS. Love the idea of simplifying it all away with an invert filter, I'll experiment with that and see if I can exclude favicons.

mwanchap:
Not quite finished yet, but how does that look?
TODO:
- blank page
- merge upstream repo

leosolid:
Please add it to the Open URL command box too, if you haven't.

mwanchap:
@leosolid thanks for reminding me. Done!
@philc adding another invert filter to the icons gets them back pretty close to what they were originally, though I'm not sure what would be the inverse of a sepia filter?

mwanchap:
Aww dammit, just realised I had made all these commits using the wrong email address so they're not counted as being mine :( Haven't had to rewrite commits in a GitHub PR before... if I do, is it going to erase your comments and such?

philc:
Don't worry about my comments. Do what you have to do to get the code ready so we can merge =)

mwanchap:
Too hard basket, gave up and added my work email to my GitHub account instead.

I just tried testing this PR in Firefox and noticed that when a Vimium UI element is first displayed, this message gets logged to the console: "WARNING: Settings have not loaded yet; using the default value for darkMode.", so the setting is false and the darkMode class isn't added until the element is hidden and shown again. I'll see if I can figure out why (but please let me know if there's something really stupidly obvious that I've totally missed 😅)

mwanchap:
Bummer, I'm not going to have time to figure this out today and won't have access to a computer for a couple of weeks. If you're okay with at least Chrome users getting this feature, go ahead and merge; I'll have another go at the FF bug when I get back.

gdh1995:
@mwanchap `maintainDarkMode` should wait for `Settings.onLoaded`, like:
image

mwanchap:
Yeah that was the first thing I tried. It prevents the Settings.get call from happening and prevents the warning being logged, but it doesn't cause the setting to be loaded any earlier, so the buggy behaviour remains. Is there somewhere we're supposed to pre-load/initialise settings before using them?

gdh1995:
Seems that hud page lacks initiation of settings? You may look at how vimium_frontend.coffee triggers loading and add the same code to the bottom of pages/hud.coffee .

mwanchap:
Really appreciate any help! I'd do it myself but my wife and I are about to hop on a plane and go far away from computers for a little while...

gdh1995:
I've tested your master branch on my computer, and it works well if I just add a settings.onload to the "maintain" function.

Firefox 72.0.1 On windows 10

mwanchap:
Done on my phone, no access to PCs at the mo. Hopefully I didn't miss anything...

gdh1995:
@mwanchap Sorry but it should be `=>` after `Settings.onLoaded `, or we can replace `@setDarkModeState` with `DomUtils.setDarkModeState`.

mwanchap:
Finally back home, nice to actually be able to type properly and test what I'm doing...
@gdh1995 thanks so much for the tip about `=>`, would have taken me ages to figure that out
@philc all done, ready to merge

philc:
Awesome! I'll take this updated PR for a spin in the next week.

philc:
This looks good based on the diff. However, it doesn't merge cleanly, and I'd like it to be one commit since it's one feature. Can you rebase this on top of master, resolve the conflicts, and get rid of the merge commit? It might be faster to just manually recreate the changes in this PR's diff in one new commit and overwrite this branch.

mwanchap:
Good point, we usually squash PRs at my work too. Done!

philc:
Cool; now this is tight and easy to merge. I'll likely merge this before porting the codebase to coffeescript, although I may disable the UI for it until we can resolve this debate and do the extra work to remove the Setting:

After reflecting, I don't think we should have an option for this at all. If the browser's theme is dark mode, Vimium should be dark, just like the rest of Chrome's UI. The user shouldn't have to keep Vimium in sync with their browser. In fact, in OSX, Chrome doesn't have a dark mode option -- it just follows whatever the system wide setting is. OSX Catalina even has an "Auto" setting, where all apps use either light/dark based on the time of day (I have a hard time imagine that working smoothly, but it's there). Ideally, Vimium's UI should follow this setting, assuming Chrome's UI does.

We can remove the need for this setting by just directly detecting if dark mode is used. There are at least a few ways to [detect](https://stackoverflow.com/questions/50840168/how-to-detect-if-the-os-is-in-dark-mode-in-browsers) if the browser is in dark mode. From the looks of it, the javascript method can be used in both content scripts and background pages, so we may not even need to propagate information from background pages to content scripts.

mwanchap:
Yeah Chrome doesn't have a dark mode in Windows either, you can just set the OS' theme to dark and most things (including Chrome) update their UI based on that.

No debate at all, I totally agree with you.  I was actually hoping to do something like this in the first place to keep the feature as simple as possible, although I went in the wrong direction, trying to access the individual theme colours rather than detecting the light/dark setting. But I'm glad that my original intention might actually be possible, even if it was a roundabout way to get there! I'll have a look into it as soon as I can.

mwanchap:
This is so much simpler! All the CSS is in once place, no Javascript is needed at all, and Vimium's colours update immediately when the OS colour scheme changes. At least they do on Windows - could any of you guys test on Mac/Linux?

I've also noticed a weird behaviour in Firefox - enabling any CSS filters causes the HUD to disappear, except for its box-shadow. Seems like it's something to do with calculating its vertical position, because setting `position: inherit` or `bottom:unset` makes it display normally again. I wonder if the reason this wasn't happening before is because the dark mode CSS was applied a split second _after_ the element was rendered in place, so its position was already calculated? A related thing seems to happen in Chrome where the HUD has moved a few px from where it was before.  Anyway, I've fixed it in Firefox by adding this to the Vimium CSS, although it feels a bit hacky?

```
  @-moz-document url-prefix() {
    div.vimiumHUD {
      position: inherit;
    }
  }
```

mwanchap:
I'm not sure how I missed this, but in addition to styling Vimium UI elements, the dark mode CSS gets applied to the body tag of every web page.  Hilarious!

Bug...or feature? 😅

philc:
I tweaked this a bit and fixed the bug where all pages became dark. Merged in 4efaf1437c544524ab1eaaf575bab7e4c9fa123e. Thanks @mwanchap!

Future work is to make the dark mode styles nicer (i.e. not use the sepia filter, at least for the help dialog). That will be much cleaner if we can remove the use of vimiumReset from the help dialog, HUD, and vomnibar. I think there's no reason to have vimiumReset anymore since they're rendered in iframes, but I could be mistaken.

philc:
I'm wondering if we might want this configurable via an option (with the default being to use the system's theme colors). If you use the Dark Reader extension as mentioned in #3157 (which I do), it applies a page-level invert filter, and so now vomnibar in dark mode gets inverted again, such that it's white:

image

Any ideas?

mwanchap:
Yeah good idea! I use [this ](https://chrome.google.com/webstore/detail/dark-mode/dmghijelimhndkbmpgbldicpogfkceaj) dark mode extension, which already seems to exclude all of Vimium's UI elements

philc:
I reverted this in 41cbc1f9aec69c19596e23baa0c3229770e0b17d, but plan to revisit it soon. Two major TODOs:

1. See if we can make this play nice with Dark Reader. I bet there's a lot of overlap between Vimium and Dark Reader users.
2. Make the styles nicer (i.e. use specific dark styles, rather than a Sepia filter).

juanandresgs:
Bump

leosolid:
I really want this feature to be built in, so I’ll take a stab at it! From skimming the previous discussion, the requirements for the dark mode feature are: 

- It is compatible with dark reader
- It is dependent on the users system preference. (There’s a CSS selector for this)
- It uses a specific color scheme. (Seems natural to match Chrome’s default dark theme)
- It won’t need to be updated for future UI additions.

The last one seems to be the biggest blocker. I‘ll list out some potential solutions tomorrow.

philc:
Leo, IMO the fourth item is a nice-to-have. But if you can get the first three working, this feature is worth having to maintain two sets of CSS as we update the UI in the future. Hopefully the additional dark mode CSS is minimal.

#1 is the highest risk item; I don't know how to do it. We might need to collaborate with the dark mode extension authors.

mwanchap:
Sorry guys, I'd totally forgotten about this... Anyway, other than the dark reader compatibility, the stuff from my previous commits hit the other requirements already, so you're welcome to start from there if you want to make some changes.

Is the dark reader stuff an absolute show-stopper for this feature? If so, do we also need to think about compatibility with the various other popular "dark mode" extensions? Perhaps instead we could recommend a specific dark mode extension that _is_ compatible with Vimium, such as [Dark Mode](https://chrome.google.com/webstore/detail/dark-mode/dmghijelimhndkbmpgbldicpogfkceaj) which doesn't interfere with Vimium's dark mode at all, since it changes colours rather than inverting.

leosolid:
Dark Reader seems by far the most used dark mode extension. I'm sure there is a large intersection of users who would appreciate it being compatible.

Anyhow, I started where you left off in [mwanchap/vimium](https://github.com/mwanchap/vimium)  and I added minimal styling is compatible with Dark Reader in my environment. I made a PR in your fork mwanchap/vimium/pull/2, So, if you merge that we can review it right here.

mwanchap:
Nice! Confirmed that it works for me too, both in Chrome and Firefox, albeit only with Dark Reader's "Static" and "Dynamic" modes - Filter and Filter+ both use inversion so they cause the same issue mentioned previously. Dynamic mode seems to be the default (and the best), so I'd consider this compatible. @philc , what do you think?

I haven't used Dark Reader for ages since it was just "inversion on/off" and I wanted more flexibility, but it looks like they've greatly improved it since then.  Thanks for re-introducing me, it seems to be way better than my current Dark Mode extension now.

leosolid:
Ah, that's why it seemed to just work. Is dynamic mode the default? It's been installed on my browsers a while so I'm not sure. If so, then I think it's good to go as well.

Either way I can make a PR soon on dark reader to ignore Vimium for filter and filter+. They merge fixes fairly quickly.

philc:
Great progress everyone. I'm very interested in merging this as I use DarkReader. I had mine set to Filter mode; I'm not sure why. I'm a long-time user. If the default is Dynamic, I must've set it to Filter at one point with some intention, likely because I saw enough break cases with Dynamic to change it.

I've just switched it over to Dynamic mode and will see how it compares in my daily usage.

I'm on the fence as to whether this is OK to merge if it works with Dynamic mode but not Filter mode. It depends on what percent of users have set their mode to Filter. As a DarkReader user, I wouldn't have even thought to try other filtering modes when if I saw the inversion weirdness with Vimium's UI.

Note that making it work with all modes would likely significantly cut down on user annoyance and questions on our issue trackers. So I'm very interested @leosolid to see if you can get a working PR sent to dark reader, and what the response is.

mwanchap:
I've just tried installing it from scratch on my wife's Google account (she's never used it before) and "Dynamic" is definitely the default mode.

mwanchap:
Another alternative idea if the Dark Reader PR doesn't work out: Vimium can already "read and change all your data on websites that you visit" - we could directly check if there's an invert filter being applied to the page and update Vimium accordingly. When Dark Reader is using an invert filter, it adds a style element to the head tag containing the following:

`html {
  -webkit-filter: invert(100%) hue-rotate(180deg) !important;
  filter: invert(100%) hue-rotate(180deg) !important;
}`

And Vimium can either remove the dark mode CSS or apply another invert filter to un-invert things.

leosolid:
Got it working with the filter and filter+ modes! And just remembered to add styles to the help dialog.

Annotation 2020-05-29 220935

I actually didn't make a PR on Dark Reader. It was not possible to add a simple CSS exception for Vimium due to the fact that the vomnibar—and a couple other components—are added through iframes and shadow DOM. I could've tried to parse for the iframe, but there was no easy way to identify Vimium and so it would've probably been too messy to merge.

The actual approach is basically what @mwanchap suggested. For Filter and Filter+ modes, Dark Reader adds `<style id="dark-reader-style" type="text/css">` inside the `<head>` tag, but not for Dynamic mode. So, I added a mutation observer to the head, such that whenever a child is added or removed from `<head>` the observer checks if `#dark-reader-syle` exists. If it does, then a class is added to the iframe that reverses the inversion. 

For robustness, we could take it a step further and check if the `<html>` actually has an invert filter applied, but I didn't - in favor of simplicity. And I've read that checking for style changes (with `getComputedStyles`) is a bit slow.

This logic was added to `ui_component.coffee` in order to apply to the help dialog as well.

@mwanchap I'll PR this to your branch once again

philc:
Hey guys, note that the whole codebase has been converted to plain javascript since this PR was created, so there's no longer any coffeescript files to modify. This PR should be rebased onto master (or create a fresh PR). I recommend doing that before writing any code.

leosolid:
Woops... Well now I know some coffeescript, lol!

I’ll undo the change and rebase first

mwanchap:
Sigh, all those days spent mastering Coffeescript, _wasted_ 😜
Leo, I've rebased my master branch and pushed it over to [darkmode-rebase](https://github.com/mwanchap/vimium/tree/darkmode-rebase), I hesitate to use `push --force` to rewrite master since you've forked it, although I can if you want.  Then you can PR your rewritten JS to that, and I'll open a fresh PR for Phil to check out. Right? (Or am I overcomplicating this?)

leosolid:
Yeah, I think that's fine. I'll make a new PR to that branch later today. 👍

leosolid:
@mwanchap here ya go! mwanchap/vimium/pull/4

Also, I updated it to not invert static mode. I forgot to check that last time. Although, I'd be surprised if anyone is using static mode, as it doesn't seem to work too well...

mwanchap:
Oh thanks, I'll really appreciate that as I've started using static mode occasionally for certain sites where Dynamic mode seems to slam my CPU for some reason (e.g. Salesforce). It doesn't look great as you said, but at least my browser stops lagging. Maybe they have some weird DOM manipulating JS that causes Dark Reader to need to do a lot more work analysing things.
Anyway, I think @philc will want all these commits squashed into one, so I'll go do that first.

FranklinYu:
Hello, why was this PR closed? I thought the plan was to squash and have it reviewed?

mwanchap:
I'm going to open it as a new PR as discussed; I just haven't had a chance to do it as I've been away for the last few days. Thanks for your patience!

zbyna:
dsasdf

zbyna:
dfasdfa

PR #3403, opened 2019-10-06
setup listeners: use capture mode

### LINKED ISSUE #3398: Some keys do not work on https://www.nydailynews.com/ (both chrome and firefox)
link hints (f/F), open URL (o/O) keys don't work
some keys like scrolling, tab navigation etc., do work.

chrome: 77.0.3865.75
firefox: 69.0
Vimium: 1.64.6

Regards

gdh1995:
This should fix #3398 (the problematic page suppresses DOMContentLoaded events).

### DIFF
--- a/content_scripts/ui_component.coffee
+++ b/content_scripts/ui_component.coffee
@@ -62,10 +62,11 @@ class UIComponent
                       @hide false
                     false # We will not be calling sendResponse.
                   # If this frame receives the focus, then hide the UI component.
-                  window.addEventListener "focus", (event) =>
+                  window.addEventListener "focus", (forTrusted (event) =>
                     if event.target == window and @options?.focus
                       @hide false
                     true # Continue propagating the event.
+                  ), true
                   # Set the iframe's port, thereby rendering the UI component ready.
                   setIframePort port1
                 when "setIframeFrameId" then @iframeFrameId = event.data.iframeFrameId
--- a/content_scripts/vimium_frontend.coffee
+++ b/content_scripts/vimium_frontend.coffee
@@ -16,10 +16,12 @@ normalMode = null
 windowIsFocused = do ->
   windowHasFocus = null
   DomUtils.documentReady -> windowHasFocus = document.hasFocus()
-  window.addEventListener "focus", forTrusted (event) ->
+  window.addEventListener "focus", (forTrusted (event) ->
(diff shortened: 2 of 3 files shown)


content_scripts/ui_component.coffee
```diff
@@ -62,10 +62,11 @@ class UIComponent
                       @hide false
                     false # We will not be calling sendResponse.
                   # If this frame receives the focus, then hide the UI component.
-                  window.addEventListener "focus", (event) =>
+                  window.addEventListener "focus", ((event) =>
                     if event.target == window and @options?.focus
                       @hide false
-                    true # Continue propagating the event.
+                    true
+                  ), true # Continue propagating the event.
```
lydell: The comment belongs to the `true` on the line above, doesn’t it?

pages/vomnibar.html
```diff
@@ -15,7 +15,7 @@
   <body>
     <div id="vomnibar" class="vimiumReset">
       <div class="vimiumReset vomnibarSearchArea">
-        <input id="vomnibarInput" type="text" class="vimiumReset">
+        <input id="vomnibarInput" type="text" class="vimiumReset" autocomplete="off">
```
lydell: Looks unrelated to the rest of the PR?

gdh1995: yes, unrelated😂

content_scripts/ui_component.coffee
gdh1995: updated.

in fact, the first 'true' here does nothing - Browser doesn't use true/false to decide whether to continue on such listeners

philc:
It seems conceptually reasonable that we would listen to the capture phase of an event, and this should make Vimium more robust to individual pages suppressing important events.

It's a pretty core change, however. Have you foreseen any issues yet with this?

gdh1995:
Compared with bubble-phase listeners, the only obvious differences is, for `focus`, `blur`, `load`, and `unload` events:
* capture-phase listeners will be run if only a bound target or its descendant generates such a event
* but bubble-phase listeners will only accept events whose `target` is the bound `this`

Maybe there're also some other events following such a rule, but it will always be safe if we add enough checks.

My Vimium C has used caputure-phase listeners for 4 years and no unexpected breaks found.

PR #3404, opened 2019-10-08
Fixed failing build by rewriting path.join call in phantom_runner

mwanchap:
Builds have been failing for a while due to some issue inside the path/util modules. A better solution would be to figure out what the issue with util.isString was, but I couldn't figure that out and this still works for joining paths ;)


philc:
Thanks for the fix. I agree that your workaround is sufficient and not worth debugging further for now.

PR #3409, opened 2019-10-13
FindMode: clear focus of HUD iframe on exit

### LINKED ISSUE #3245: [Firefox] grey highlight in find mode, followed by locked commands in visual mode
**Environment:**
Firefox version: 65.0
Host OS: Win10
Vimium version: 1.64.3

**Addons:**
Smart HTTPS
uBlock Origin
Vimium

**Issue:**
After using find mode a grey highlight appears, which locks command functionality after entering visual mode.

**Solution after facing issue:** 
Use the mouse to focus back on the page, which removes the grey highlight, and allows for commands to operate again.

**Steps to replicate:**
1. Search for a word using find mode (ex. screenshot below, highlighting the word issue - the selected word has a grey highlight.)
image
2. Immediately go into visual mode.
3. Attempt to use a command.

**Workaround:**
    1. Use find mode to search for a word.
    2. While the word still has grey highlight press n or N to change focus back to orange highlighting.
    3. Visual mode should now be operational.

**Additional troubleshooting:**
Disabled addons
Replicated steps across multiple types of websites
Wasn't able to replicate in Chrome
Able to enter visual mode first > use mouse to highlight > commands work

*edit*
Cleaned up formatting

gdh1995:
On Firefox, if an `<iframe>` disappears when it's focused, then it will keep focused. This feature has caused the 2nd problem in #3245 .

This PR should fix #3245 , and it works well on my Firefox 69 (stable x64) + Win 10.

As for Firefox 67, there's still some incorrect results about colors of selection text after FindMode has exited, but I think it's because of inner bugs of Firefox - this PR works well enough on Firefox 69...

### DIFF
--- a/content_scripts/hud.coffee
+++ b/content_scripts/hud.coffee
@@ -5,7 +5,6 @@
 HUD =
   tween: null
   hudUI: null
-  _displayElement: null
   findMode: null
   abandon: -> @hudUI?.hide false
 
@@ -113,6 +112,11 @@ HUD =
 
   unfocusIfFocused: ->
     document.activeElement.blur() if document.activeElement == @hudUI?.iframeElement
+
+  clearFocus: ->
+    if @hudUI?.showing
+      @hudUI.iframeElement.blur()
+      window.focus()
 
 class Tween
   opacity: 0
--- a/content_scripts/hud.coffee
+++ b/content_scripts/hud.coffee
@@ -111,9 +111,6 @@ HUD =
     @pasteListener data
 
   unfocusIfFocused: ->
-    document.activeElement.blur() if document.activeElement == @hudUI?.iframeElement
-
-  giveUpFocus: ->
     # On Firefox, if an <iframe> disappears when it's focused, then it will keep "focused",
     # which means keyboard events will always be dispatched to the HUD iframe
     if @hudUI?.showing
--- a/content_scripts/hud.coffee
+++ b/content_scripts/hud.coffee
@@ -114,7 +114,9 @@ HUD =
     document.activeElement.blur() if document.activeElement == @hudUI?.iframeElement
 
   giveUpFocus: ->
-    if @hudUI.showing
+    # On Firefox, if an <iframe> disappears when it's focused, then it will keep "focused",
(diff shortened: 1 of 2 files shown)


content_scripts/mode_find.coffee
```diff
@@ -78,6 +78,7 @@ class FindMode extends Mode
     HUD.showFindMode this
 
   exit: (event) ->
+    HUD.giveUpFocus()
```
philc: Shouldn't this be `clearFocus`? Can you add a comment as to why focus needs to be cleared (reference #3409 if needed).

content_scripts/hud.coffee
```diff
@@ -114,6 +113,11 @@ HUD =
   unfocusIfFocused: ->
     document.activeElement.blur() if document.activeElement == @hudUI?.iframeElement
 
+  clearFocus: ->
+    if @hudUI?.showing
```
philc: I didn't look hard, but Is this `@hudUI?` conditional necessary? Is it possible to be exiting find mode without the @hudUI being initialized?

content_scripts/mode_find.coffee
gdh1995: Sorry `clearFocus` was just a typo. I've rebased the code and add comments.

content_scripts/hud.coffee
gdh1995: The `?` seems necessary on slow pages. When a page is loading slowly, press <kbd>/</kbd> will call `showFindMode` to init and show HUD iframe on `documentComplete`, so if a user click/press <kbd>Esc</kbd> before `documentComplete`, it will exit before `HUD.hudUI` gets ready.

content_scripts/hud.coffee
```diff
@@ -114,6 +113,13 @@ HUD =
   unfocusIfFocused: ->
```
philc: Can we delete this `unfocusIfFocused` method and replace it with `giveUpFocus`? I find it confusing that there's now two methods with very similar names and functionality.

gdh1995: Yes they're doing the same thing. I've updated the code.

PR #3414, opened 2019-10-21
Move RELEASE NOTES to its own Section

wendyliga:
as release notes will increase as this repo move forward, it will create very long `README` which will look intimidating and not necessarily on the front page of the repo ( as secondary info ).

I suggest moving release notes on its own page so it will not bloat the main repo page.


philc:
Thanks for motivating this idea. I've implemented this in ab3c49d187478b1dddb11e55cd3e248bb2263cc1 using a slightly different filename for the release notes.

Issue #3418, opened 2019-10-28
Custom CSS for Link Hints inside iFrames seems to not be working correctly again


HarvardYard:
I noticed this has been referenced before.
#2594
And fixed by:
#2683

However, it is still occurring for me. First started noticing it today [10.28.19].

Chrome Version: 78.0.3904.70 (Official Build) (64-bit) (cohort: 78_70_Win)
Vimium Version: 1.64.6

gnarendran:
+1
To be able to use with the new "chrome://flags/#enable-force-dark", I tried to change the colors for link hints in "CSS for Vimium UI", and then noticed this issue. Currently working around by modifying vimium.css itself.

gdh1995:
https://developer.chrome.com/extensions/content_scripts#document_start says:

> scripts marked "`document_start`" are injected after after files from css

But in tests on Chrome 53/61/74/78 x64 on Win 10, I found user-defined code which is injected into iframes by `chrome.tabs.insertCSS` may be overridden by those in vimium.css, and this means, I think, the document above is not always correct.

qubist:
Chiming in here: in Firefox, my custom link hint CSS doesn't apply in the iframes of embedded YouTube videos.

macOS 11.0.1 (20B29)
Firefox 89.0.2 (64-bit)

philc:
Closing this in favor of #3452, which I think is the same issue.

PR #3421, opened 2019-11-03
Fix css injection in iframes

gdh1995:
This is a work-around for #3418.


gdh1995:
Well, this PR may cause a page (which is still loading) shows link hints in wrong styles ... So this is not as useful as I once expected.

philc:
Thanks for taking a look at this. Given the caveat you describe, I think this is not a valid solution in its current form.

PR #3127, opened 2018-09-08
Fix Firefox focus switch for find mode

marcotc:
> _Fixes #3106_

Firefox won't allow us to change focus to our search HUD _unless_ nothing is currently focused in the current webpage document (`document.activeElement` on the main page).

If I had to guess, it seems like Firefox prioritizes the brower's focus to the main page first, and then, if it doesn't find anything to focus there, goes down the document chain, finding Vimium's DOM with a valid `document.activeElement` set next.

PR #3460, opened 2020-01-06
another solution to fix enterFindMode can not grab focus

### LINKED ISSUE #3106: focusing on certain elements makes search feature not functional.
Open this [link](https://www.google.com/search?q=blot&ie=utf-8&oe=utf-8&client=firefox-b-ab) after the page is loaded, press <kbd>Tab</kbd> then press <kbd>/</kbd> the search box will show up but it won't be focused so if you type nothing gets typed in it, if you press <kbd>Esc</kbd> and then <kbd>/</kbd> again, it won't work. if you press <kbd>/</kbd> one more time the firefox search will appear and it will have the focus you can try many things but I could never make that search box have the focus again except by clicking on it with mouse and then pressing <kbd>Tab</kbd>.

gdh1995:
This is to fix #3106 and replace #3127.

Compared with #3127, this PR is more acurate since the old one only grabs focus before HUD begins to load.

### DIFF
--- a/pages/hud.coffee
+++ b/pages/hud.coffee
@@ -91,7 +91,9 @@ handlers =
     countElement.id = "hud-match-count"
     countElement.style.float = "right"
     hud.appendChild countElement
-    Utils.setTimeout TIME_TO_WAIT_FOR_IPC_MESSAGES, -> inputElement.focus()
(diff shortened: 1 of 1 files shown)


philc:
Thanks @gdh1995. I expanded the comment for this in 9005144cd02fab59577c24f50aae279f4160ce06 and made window.focus happen only for firefox.

PR #3463, opened 2020-01-10
Better position for create tab

gdh1995:
On Chrome the default action of `chrome.tabs.createTab` when `position=null` is to open at the end of tabs, but on Firefox it isn't. So this PR fixes it.

@philc See https://github.com/philc/vimium/issues/2893#issuecomment-572889467. And this PR has been well-tested.

### DIFF
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -116,7 +116,7 @@ TabOperations =
     switch position
       when "start" then tabIndex = 0
       when "before" then tabIndex = request.tab.index
-      when "end" then tabIndex = null
+      when "end" then tabIndex = (if Utils.isFirefox() then 9999 else null)
       # "after" is the default case when there are no options.
       else tabIndex = request.tab.index + 1
     tabConfig.index = tabIndex
--- a/background_scripts/main.coffee
+++ b/background_scripts/main.coffee
@@ -116,6 +116,8 @@ TabOperations =
     switch position
       when "start" then tabIndex = 0
       when "before" then tabIndex = request.tab.index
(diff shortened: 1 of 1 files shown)


background_scripts/main.coffee
```diff
@@ -130,7 +130,7 @@ TabOperations =
     tabConfig.openerTabId = request.tab.id if canUseOpenerTabId
 
     chrome.tabs.create tabConfig, (tab) ->
-      callback extend request, {tab, tabId: tab.id}
+      callback extend request, {tab, tabId: tab.id, position: "", active: false}
```
philc: Why are these extra arguments needed? For Chrome they must be no-ops, assuming Chrome's current behavior isn't changed. So if these are important for Firefox, document why.

background_scripts/main.coffee
```diff
@@ -116,7 +116,7 @@ TabOperations =
     switch position
       when "start" then tabIndex = 0
       when "before" then tabIndex = request.tab.index
-      when "end" then tabIndex = null
+      when "end" then tabIndex = (if Utils.isFirefox() then 9999 else null)
```
philc: Please add a comment explaining why firefox needs this logic and chrome doesn't -- i.e. explain why Firefox has a special case. I'm noticing that the code base has started accumulating firefox-specific logic which works, but isn't well documented as to why we need it, and so it's not going to survive big refactors or changes by relatively novice contributors.

background_scripts/main.coffee
gdh1995: OKay, I've added two comments in code.

PR #3467, opened 2020-01-12
Fix broken documentComplete

gdh1995:
This fixes a typo bug in #3403, which has made the HUD and help dialog disappear.

According to the specifications, a page's final "load" event will always be dispatched onto `document`.


philc:
Great, thank you. I had it on my TODO list to chase this down before release =)

PR #3474, opened 2020-01-18
Add improved elementFromPoint util

poacher2k:
Now works properly with Shadow DOM, ref. discussion at https://github.com/philc/vimium/commit/caf13c2dbfa9b3c5f35aae76dfcffa4cfea14bac#r36859231

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -749,7 +749,7 @@ LocalHints =
   getElementFromPoint: (x, y, root = document, stack = []) ->
     element = root.elementFromPoint(x, y)
 
-    if stack.includes(element)
+    if stack.indexOf(element) != -1
       return element;
     
     stack.push(element);
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -747,7 +747,7 @@ LocalHints =
   # until we hit an actual element.
   #
   getElementFromPoint: (x, y, root = document, stack = []) ->
-    element = root.elementsFromPoint(x, y)[0]
+    element = if root.elementsFromPoint then root.elementsFromPoint(x, y)[0] else root.elementFromPoint(x, y)
 
     if stack.indexOf(element) != -1
       return element
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -747,17 +747,17 @@ LocalHints =
   # until we hit an actual element.
   #
   getElementFromPoint: (x, y, root = document, stack = []) ->
-    element = root.elementFromPoint(x, y)
+    element = root.elementsFromPoint(x, y)[0]
 
     if stack.indexOf(element) != -1
-      return element;
+      return element
     
-    stack.push(element);
+    stack.push(element)
     
     if element and element.shadowRoot 
       return LocalHints.getElementFromPoint(x, y, element.shadowRoot, stack)
     
-    return element;
+    return element
   
   #
   # Returns all clickable elements that are not hidden and are in the current viewport, along with rectangles
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -742,6 +742,19 @@ LocalHints =
     visibleElements
 
   #
+  # Returns element at a given (x,y) with an optional root element.
+  # If the returned element is a shadow root, call the function use that recursively 
+  # until we hit an actual element.
+  #
+  getElementFromPoint: (x, y, root = document) ->
+    element = root.elementFromPoint(x, y)
+    
+    if element and element.shadowRoot 
+      return LocalHints.getElementFromPoint(x, y, element.shadowRoot)
+    
+    return element;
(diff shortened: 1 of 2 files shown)


poacher2k:
With this util the result is the following:
image

gdh1995:
BTW, I recommend to check `element?.getRootNode().elementsFromPoint().includes(element)` firstly, and `getRootNode` exists since Chrome 54 and Firefox 53.

poacher2k:
@gdh1995 : the tests failed when I tried to use `Array.includes`, as well as when I used `elementsFromPoint`.

At what place do you recommend doing the check you write?

gdh1995:
Firstly, fix my typo: it should be `element.getRootNode?().elementsFromPoint?().includes(element)`

We may update the test,  since:
* `Node::getRootNode` exists since Chrome 54 and Firefox 53
* `DocumentOrShadowRoot::elementsFromPoint` exists since Chrome 43 and Firefox 63
* `Array::includes` always exists since Chrome 49 and Firefox 43

So we only need to check the first two functions.

All the version code of Chrome has been well-tested on my Win10, while Firefox versions are referred from MDN (like https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/elementsFromPoint)

poacher2k:
Updating the test then sounds like a good idea, as it saves some unnecessary checks.

`element.getRootNode?().elementsFromPoint?().includes(element)` is, as far as I can see, a bit too broad, as we're only interested in the first and topmost, element. this is why `elementFromPoint` would be ideal, but given that with the "Code"-button @ [the test case page](https://bugs.chromium.org/p/chromium/issues/detail?id=1036742) isn't returned by `elementFromPoint` when I tested, but *is* first in the returned array by `elementsFromPoint` given the exact same x and y, the latter seems to be a more consistent way to get the topmost element in a cross-browser setting.

poacher2k:
See this screenshot for reference:
image

content_scripts/link_hints.coffee
```diff
@@ -805,7 +824,7 @@ LocalHints =
       element = visibleElement.element
 
       # Check middle of element first, as this is perhaps most likely to return true.
-      elementFromMiddlePoint = document.elementFromPoint(rect.left + (rect.width * 0.5), rect.top + (rect.height * 0.5))
+      elementFromMiddlePoint = LocalHints.getElementFromPoint(rect.left + (rect.width * 0.5), rect.top + (rect.height * 0.5))
       if elementFromMiddlePoint && (element.contains(elementFromMiddlePoint) or elementFromMiddlePoint.contains(element))
```
gdh1995: Why check `elementFromMiddlePoint.contains(element)`? Is it for those with a CSS style of `pointer-events: none;` ?

poacher2k: I seem to recall this was because sometimes `elementFromMiddlePoint` would be a parent to the clickable element, and I would get more consistent results when including this check. When testing out now, though, I can't seem to find any difference with or without it, so I guess it was some strange edge cases that might no longer be around.

gdh1995: I find an opposite case: on https://material.angular.io/components/button-toggle/overview , the sidebar has overflow:hidden, so if I remove this call, those invisible list items will be not hinted.

philc:
Great job spotting this issue and posting a fix. Just to check, is this ready to go? I'd like to do a release soon.

poacher2k:
Yep, as far as I know. @gdh1995 ?

gdh1995:
Here's a small warning: the detection will be "totally broken" if only an ancestor element is zoomed by `element.style.zoom = 2/3/...`. For example, if `document.documentElement.style.zoom = 1.2`, then most small elements will excluded by this detction.

Note: `.style.zoom` is only on Chrome.

lydell:
@gdh1995 Interesting! Didn’t know about `zoom`. Do you know of any site that uses it?

gdh1995:
Um, no I don't have. QQMail once used `documentElement.style.zoom` to implement global font-size adjusting, but now it turns to use `document.body.style.transform` (like `scale(1.5)`).

gnarendran:
I have been using poacher2k:master for the last week or so. I occasionally find a few a few missing link hints. For example, attached a screenshot of https://news.google.com/?hl=en-IN&gl=IN&ceid=IN:en, where the mouse underlines the link but there is no link hint - there are other links in the screenshot with the same issue.
missingLinkHints

philc:
Oops, I meant to merge this in as part of 1.65. Oh well; I'll make it a point release if people continue complaining about shadow dom links not being covered by link hints.

gdh1995:
@gnarendran The page marks its links `[aria-hidden="true"]`, so Vimium ignore them. I think it's a misuse of `aria-hidden`...

gnarendran:
@gdh1995 Oh ok, thanks for the clarification. Also wanted to note that when that problem page was scrolled, some of the missings link hints started appearing (not always though). Regards.

PR #3482, opened 2020-02-07
fix broken isFirefox on pages/*

gdh1995:
The commit 9005144cd02fab59577c24f50aae279f4160ce06 which modified #3460 may not work, because `Utils.isFirefox` was always `false` unless in Vimium's background / content_scripts page. So this PR passes the result of `isFirefox` from content scripts to `pages/*.html`, including the HUD.

### DIFF
--- a/content_scripts/link_hints.coffee
+++ b/content_scripts/link_hints.coffee
@@ -372,7 +372,7 @@ class LinkHintsMode
             Utils.nextTick -> focusThisFrame highlight: true
           else if localHintDescriptor.reason == "Scroll."
             # Tell the scroller that this is the activated element.
-            handlerStack.bubbleEvent "DOMActivate", target: clickEl
+            handlerStack.bubbleEvent (if Utils.isFirefox() then "click" else "DOMActivate"), target: clickEl
           else if localHintDescriptor.reason == "Open."
             clickEl.open = !clickEl.open
           else if DomUtils.isSelectable clickEl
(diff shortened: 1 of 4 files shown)


philc:
Thanks for catching! We can drop the trailing `if` statement on my commit if that makes this easier.

However, I'm confused by the complexity of the `isFirefox` detection mechanism and this PR. Why can't we just use the [userAgent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent), so it doesn't require information passing from background pages?

gdh1995:
A web developer may override useragent through The Developer Tools, so it's not robust enough.


This PR is depended on by another of my PRs (the "ime in findmode")



---Original---

content_scripts/link_hints.coffee
```diff
@@ -372,7 +372,7 @@ class LinkHintsMode
             Utils.nextTick -> focusThisFrame highlight: true
           else if localHintDescriptor.reason == "Scroll."
             # Tell the scroller that this is the activated element.
-            handlerStack.bubbleEvent "DOMActivate", target: clickEl
+            handlerStack.bubbleEvent (if Utils.isFirefox() then "click" else "DOMActivate"), target: clickEl
```
philc: Where did this change come from? Seems unrelated to this PR.

content_scripts/ui_component.coffee
```diff
@@ -72,6 +72,8 @@ class UIComponent
                 when "setIframeFrameId" then @iframeFrameId = event.data.iframeFrameId
                 when "hide" then @hide()
                 else @handleMessage event
+      if Utils.isFirefox()
```
philc: Add a newline before this line, to separate it from the very long block before it.

content_scripts/link_hints.coffee
gdh1995: Unrelated. This is just to fix another issue that on Firefox `LinkHints` can not notify `Scroller` what to scroll.

PR #3499, opened 2020-02-14
make message hud non-clickable

### LINKED ISSUE #3496: Hint box about command executing in vimium  with new version 1.65.2 , hide some link hints
I changed the css to hide this new box, but continues to hide link hints, in some buttons(in this screen the button next ) and some links happens. 
I attached the example screen ,
chrome_5BgUgNd09Z

gdh1995:
This fixes #3496 by giving message HUD `pointer-events: none` to remove it from the result list of `document.elementsFromPoint`.

Also collect code abouting HUD's visibility changes, just for clear logic. I think this should have no impact on Vomnibar and tests has passed on Chrome 80.0.3987.106 and 74.0.3729.169 (x64 version) on my Win 10.

### DIFF
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -330,7 +330,7 @@ iframe.vimiumHUDFrame {
   right: 20px;
   margin: 0 0 0 -40%;
   border: none;
-  z-index: 2149999998; /* Two less than the reference value. */
+  z-index: 2139999998; /* Two less than the reference value. */
   opacity: 0;
 }
 
@@ -407,7 +407,7 @@ iframe.vomnibarFrame {
   margin: 0 0 0 -40%;
   border: none;
   font-family: sans-serif;
-  z-index: 2149999998; /* Two less than the reference value. */
+  z-index: 2139999998; /* Two less than the reference value. */
 }
 
 div.vimiumFlash {
@@ -429,4 +429,8 @@ iframe.vimiumUIComponentVisible {
 
 iframe.vimiumUIComponentReactivated {
   border: 5px solid yellow;
+}
+
+iframe.vimiumNonClickable {
+  pointer-events: none;
 }
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -413,8 +413,9 @@ DomUtils =
     if not node?
       return null
     if node == sel.anchorNode and sel.focusOffset == sel.anchorOffset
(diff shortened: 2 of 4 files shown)


content_scripts/vimium.css
```diff
@@ -330,7 +330,7 @@ iframe.vimiumHUDFrame {
   right: 20px;
   margin: 0 0 0 -40%;
   border: none;
-  z-index: 2149999998; /* Two less than the reference value. */
+  z-index: 2139999998; /* Two less than the reference value. */
```
philc: Why this change?

lib/dom_utils.coffee
```diff
@@ -413,8 +413,9 @@ DomUtils =
     if not node?
       return null
     if node == sel.anchorNode and sel.focusOffset == sel.anchorOffset
-      # The selection either *is* an element, or is inside an opaque element (eg. <input>).
-      node = node.childNodes[sel.focusOffset]
+      # If the selection is not a caret inside a `#text`, which has no child nodes,
+      # then it either *is* an element, or is inside an opaque element (eg. <input>).
+      node = node.childNodes[sel.focusOffset] || node
```
philc: Why is this change needed?

gdh1995:
the 214... is a typo, and make HUD hide (cover) some hint markers. While the related issue wants to see hints for elements which are at bottom-right corner and covered by HUD, so I think it's needed to show hint markers above HUD.



---Original---

gdh1995:
I found that a first Escape on HUD's input box can not exit FindMode on Chrome 74, if only the input box is empty, while it just blurs input. Only a second Escape exits it and make HUD hide.
So I add this check in the second commit of this PR to fix the crash behind this problem.



---Original---

content_scripts/vimium.css
gdh1995: The 214... is an old typo, and make HUD hide (cover) some hint markers. While the related issue wants to see hints for elements which are at bottom-right corner and covered by HUD, so I think it's needed to show hint markers above HUD.

lib/dom_utils.coffee
gdh1995: I found that a first <kbd>Escape</kbd> on HUD's input box can not exit FindMode on Chrome 74, if only the input box is empty, while it just blurs input. Only a second Escape exits it and make HUD hide.
So I add this check in the second commit of this PR to fix the crash behind this problem.

PR #3500, opened 2020-02-15
Migrate from phantomjs to puppeteer

philc:
Migrate our dom tests to a better (and maintained) headless chrome implementation. Allows us to delete some code.


philc:
@gdh1995 is the shadow dom change I made in a929fda safe?

gdh1995:
Um, `Element::attachShadow` is since Chrome 53, but currently Vimium only requires Chrome 51+ :
https://github.com/philc/vimium/blob/ebc6d66e42fb9a60ac127168ea90f2ddf9ba6d45/manifest.json#L9

philc:
Great, thanks for the research @gdh1995. I'll leave the conditional checks, and I've updated the comments.

PR #3509, opened 2020-02-28
Send mouse events that bubble across shadow DOM

tbodt:
The latest version of Gerrit has been redesigned so that the link elements don't actually handle mouse events. They're instead handled by other elements a few layers of shadow DOM up. Events created by initMouseEvent don't have `composed` set, so they won't bubble up shadow DOM. This sets `composed` using the MouseEvent constructor.

Also initMouseEvent is deprecated according to MDN.

Link: https://fuchsia-review.googlesource.com/

### DIFF
--- a/lib/dom_utils.coffee
+++ b/lib/dom_utils.coffee
@@ -282,9 +282,17 @@ DomUtils =
         @simulateMouseEvent "mouseout", undefined, modifiers
         lastHoveredElement = element
 
-      mouseEvent = document.createEvent("MouseEvents")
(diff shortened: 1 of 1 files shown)


gdh1995:
LGTM.

BTW, although https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/MouseEvent says this constructor only exists on Chrome 47+, in my tests Chrome 31 has supported it.

philc:
Nice fix, and readability improvement. Thanks!

PR #3540, opened 2020-04-18
Export Vomnibar.activate

onvej:
This PR exports a function that launches the vomnibar to page script's scope. It makes possible to create a homepage that automatically launches the vomnibar:
```
<html>
  <body onload="activateVomnibar()">
  </body>
</html>
```

### DIFF
--- a/content_scripts/api.coffee
+++ b/content_scripts/api.coffee
@@ -1,0 +1,6 @@
+activateVomnibar = () ->
+  root = exports ? (window.root ?= {})
+  root.Vomnibar.activate(root.sourceFrameId, {options : {}})
+  return
+
+exportFunction(activateVomnibar, window, {defineAs:'activateVomnibar'})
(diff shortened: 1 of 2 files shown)


gdh1995:
1. In this way a page can not define command options, so I doubt whether it's so useful
2. `exportFunction` only exists on Firefox, not on Chrome

philc:
1. Since this only works for Firefox, I don't think it's viable, given most of our users are on Chrome.
2. I don't think we should generally give pages the ability to invoke the vomnibar; it might lead to an exploit.

gdh1995:
um, if not limited in the exportfunction way, we may use window.postmessage to pass commands, and the second point can be solved by adding an allow list of such URLs.

philc:
Closing due to age, and I don't think this approach is particularly wise (allowing pages to invoke Vimium UI functions).

Issue #3450, opened 2019-12-23
Feature request: LinkHints.activateModeToOpenInNewWindow


parogen:
Similar to LinkHints.activateModeToOpenInNewTab, allow the user to instead open links in a new window.

itaj-sherman:
That would be great.
Even more so if it would keep the maximized status of the original window.

PR #3542, opened 2020-04-19
Add LinkHints.activateModeToOpenInNewWindow command - open a link in a new window

AndreasDM:
This adds a new mode that can be mapped: "LinkHints.activateModeToOpenInNewWindow". When mapped, this opens a link in a new window with focus. Works in both latest chrome and firefox.
Removed seemingly irrelevant code from the existing `openUrlInNewWindow` function.


philc:
Related: #3450.

Closing this PR due to age. Sorry we never got discussion on this at the time of publishing.

This prompted me to think about how to add more link hints options like this without creating yet more commands: #4634.

Issue #160, opened 2010-06-09
like 'f' but for saving images


jamescole:
when the user would press the designated key for this feature, all the images on the page would be given yellow labels (like the links currently are when the user presses 'f' or 'F').  if the user then typed the label for a particular image, a save image dialog would appear for that image.

PR #3543, opened 2020-04-20
Fix LinkHints.activateModeToDownloadLink

### LINKED ISSUE #2804: Download link url (Firefox)
The ```LinkHints.activateModeToDownloadLink``` command doesn't seem to be working on Firefox (it either does nothing or does the same as ```LinkHints.activateMode```). Is this a case of a missing API, like the clipboard issue with FF?

Frederick888:
The first commit 4fd37d7 replaces the <kbd>Alt</kbd>+Click simulation with `downloads` API, which is available in both [Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download) and [Chrome](https://developer.chrome.com/extensions/downloads#method-download) atm. It requires an additional `downloads` permission but should work more stably than the simulation I reckon.

While the first commit alone should be sufficient to fix #2804, the following two further allows users to add `save_as` and `conflict_action` options to their mappings. For instance, I found it pretty handy to have:
```
# browser.download.useDownloadDir = false in about:config
# so by default it always pops up a dialog

# suppress the dialog
map gd LinkHints.activateModeToDownloadLink save_as=false
# use the default behaviour
map gD LinkHints.activateModeToDownloadLink
```

I tested these commits on both master and v1.66 under Firefox and they seemed to be ok.

However, to make `registryEntry.options` available in `linkActivator` context, the second commit 
39dd153 actually impacts not only `DOWNLOAD_LINK_URL` but all `availableModes`. I went through the contribution guide but since it's my first time to program in CoffeeScript, it may potentially cause some bugs. Please let me know if you spot any issue in this PR, thanks!

PS: With the `downloads` API, #160 then seems to be simply a matter of hinting `` elements and passing `src` to `downloadUrl`?


content_scripts/link_hints.coffee
```diff
@@ -152,6 +152,7 @@ LinkHints =
       when "hover" then mode = HOVER_LINK
       when "focus" then mode = FOCUS_LINK
     #
+    mode.options = registryEntry?.options
```
Frederick888, line 155: Out of curiosity, why don't I need to make a clone of `mode` here? I thought it would modify the original `DOWNLOAD_LINK_URL` hence resulted in side-effects to subsequent calls with unassigned options but this worked just alright lol...

Frederick888: Ah, nvm, dumb question... Just realised that I was looking at the wrong line in debugger. It does change the original mode, but `mode.options` also gets overwritten every time the function is called.

politicus:
Experiencing "Download link url (LinkHints.activateModeToDownloadLink)" not working.

twio142:
I converted your codes to javascript and merged it into the latest version, it actually works... Thx a lot 👯

philc:
Using this API seems like the right fix to this problem. Thanks @Frederick888!

Given its age this doesn't merge into the codebase, so I'm closing this PR.

I'm considering deprecating this command. See the discussion in #2804.

PR #3595, opened 2020-06-08
Add dark mode option, take 2, plus Dark Reader support thanks to leosolid

mwanchap:
Adding a dark mode to Vimium, take 2 of PR #3402, now with [Dark Reader](https://chrome.google.com/webstore/detail/eimadpbcbfnmbkopoojfekhnkhdbieeh) compatibility provided by @leosolid!

Includes Leo's PRs mwanchap/vimium#3 and mwanchap/vimium#4

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -50,5 +50,7 @@ Contributors:
   tobimensch
   Ramiro Araujo <rama.araujo@gmail.com> (github: ramiroaraujo)
   Daniel Skogly <daniel@wishy.gift> (github: poacher2k)
+  Matt Wanchap <matt@wanchap.com> (github: mwanchap)
+  Leo Solidum <leo.g.solidum@gmail.com> (github: leosolid)
 
 Feel free to add real names in addition to GitHub usernames.
--- a/content_scripts/ui_component.js
+++ b/content_scripts/ui_component.js
@@ -37,6 +37,7 @@ class UIComponent {
 
       this.shadowDOM.appendChild(styleSheet);
       this.shadowDOM.appendChild(this.iframeElement);
+      this.handleDarkReaderFilter();
       this.toggleIframeElementClasses("vimiumUIComponentVisible", "vimiumUIComponentHidden");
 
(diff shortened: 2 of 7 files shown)


leosolid:
Woo, excited for this! Although, the changes from https://github.com/mwanchap/vimium/pull/4 seem to be missing from this PR

mwanchap:
Whoa, I'm so glad you picked that up! Somehow I must have missed those commits? I gave up trying to figure out where my mistake was and just cherry-picked them in from Leo's repo, and then squashed it all again, so it looks like they're included now.  Btw, if you have any idea of what I did wrong (or how I could prevent this in the future), I'd love some feedback...

BachoSeven:
I'd love to see this merged; I was trying out Dark Reader recently but the way it inverted the Vomnibar's colors was horrible. 

Btw, would there be a way to disable the color inversion for just the Vomnibar, in that case? I have some nice dark colors setup that I would want to keep even with Dark Reader active.

mwanchap:
@BachoSeven it's not possible to exclude a particular element from a CSS inversion that has been applied to the html element (which is how Dark Reader's filter modes work), the only option is to add another inversion to do the opposite - which is what @leosolid has done in our PR with the reverseDarkReaderFilter class:
https://github.com/philc/vimium/blob/1e7507228b14fbb566dd3dfbdd904fc0e9983584/content_scripts/vimium.css#L440-L443

Also if you haven't tried this already, I'd suggest experimenting with the other theme generation modes in Dark Reader, you might find they work better with your preferred colors.

philc:
It's awesome and it works great; I love it. Great job @mwanchap and @leosolid!

mwanchap:
Woohoo! Thanks!

leosolid:
Thanks! Now I can take off these sunglasses at night hehe 😎

j39m:
Is it possible to enable dark mode across Vimium without installing the Dark Reader add-on?

mwanchap:
Yeah, that was kind of the main idea of this PR 😉 Although the Vimium extension hasn't been updated in the Chrome Web Store since March, so it doesn't include this feature yet.

PR #3600, opened 2020-06-14
Fix some issues of global mark

### LINKED ISSUE #3599: Jump to global mark should focus window containing target tab
Jumping to a global mark does not also focus the window within which the target tab is located (but it *does* activate the marked tab within the target window as expected). This forces the user to switch windows via window manager in order to focus the target window.

It is unclear whether this is the desired behaviour, but it makes marks less useful for navigation when multiple shared-session browser windows are in use (which may be fairly common on multi-monitor setups).

**To Reproduce**
1. Create a window A.
2. Mark default tab A in window A via `mA`.
3. Create a tab B in window A, and focus tab B.
4. Create a new window B, and focus window B.
5. Attempt to jump to window A, tab A via `` `A``.
6. Notice that window B remains focused, but when changing windows via other means (e.g. the window manager), then in window A tab A is now focused.

**To Fix**
From a cursory glance, a fix could be akin to adding something similar to `chrome.windows.update(windowIdOfTab, {focused: true});` at `background_scripts/marks.js:75-76`, where `windowIdOfTab` would need to be obtained from the target `tabId`.

Firefox version 77.0.1
Vimium-ff version 1.66

gdh1995:
* fix a bug that `m` could not create global marks when <kbd>Shift</kbd> is held on
* fix that `Marks.activateGotoMode` did not focus her inactive windows

This will fix #3599

### DIFF
--- a/background_scripts/marks.js
+++ b/background_scripts/marks.js
@@ -71,8 +71,11 @@ const Marks = {
   },
 
   // Focus an existing tab and scroll to the given position within it.
-  gotoPositionInTab({ tabId, scrollX, scrollY, markName }) {
-    chrome.tabs.update(tabId, { active: true }, () => chrome.tabs.sendMessage(tabId, {name: "setScrollPosition", scrollX, scrollY}));
+  gotoPositionInTab({ tabId, scrollX, scrollY }) {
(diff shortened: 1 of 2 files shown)


philc:
Excellent. Thank you @gdh1995 !

PR #3624, opened 2020-07-10
fix usage of :read-only on Firefox

### LINKED ISSUE #3623: Advanced options on firefox: Cannot restore from backup
The "Backup and Restore" section's `Restore` button for browsing files is broken/disabled or something and thus does not allow to restore from backup.

### LINKED ISSUE #3634: Vimium-FF – Some buttons in options screen do not work
When I use vimium-FF on Firefox (78.0.2), in the options screen I can't click on the delete button (x) in the "Excluded URLs
and keys" section and I can't click on the "Browse..."-button in the "Backup and Restore"-section. So I'm not able to import any settings. Chrome is working totally fine.

### LINKED ISSUE #3653: firefox 80 won't let you hit the browse button in options to restore settings from a backup file
firefox 80 won't let you hit the browse button in options to restore settings from a backup file

### LINKED ISSUE #3678: Vimium-FF: Button for removing an excluded URL rule is unclickable
**Describe the bug**
Currently I use regular Firefox, but this has also been the case for Firefox Developer Edition: When I go to Vimium-FF options and try to remove a rule from the first section "Excluded URLs and keys", clicking on the "X" button does nothing.

**To Reproduce**
1. Open Firefox.
1. Go to Vimium-FF options.
2. Go to first section "Excluded URLs and keys".
3. If there are no excluded URLs, click on "Add Rule" to add one.
4. Try to remove a rule, by clicking the "X" button on the far right of the rule.

**Browser, OS and Vimium version**
Firefox 81.0
Windows_NT 10.0 19041
Vimium-FF 1.66

### LINKED ISSUE #3725: Button in vimium options doesn't work on Firefox
Whenever I want to use vimium advanced options to restore my backup file, the button simply doesn't work, it doesn't even act like a button. I notice that the button **"Add Rule"** changes color on hover, the button "Browse" does not and it does absolutely nothing when clicked. I noticed that this issue only happens with **Firefox** and **Firefox-developer-edition** because it worked on **chromium**.

**Steps to reproduce the behavior:**
1. Go to vimium options 
2. Click on "Show advanced options"
3. "Click Browse"

Firefox-developer-edition
Arch Linux
Mozilla Firefox 84.0b8
Vimium 1.66 version

I am attaching a picture with the button I'm referring to.

image

gdh1995:
According to https://developer.mozilla.org/en-US/docs/Web/CSS/:read-only, the `:read-only` means all non-editable DOM elements, including `<p>` and event those `<input>` whose `isSelectable` are false.

Spec: https://html.spec.whatwg.org/multipage/semantics-other.html#selector-read-only .

This may fix #3623, fix #3634, fix #3653, fix #3678 and fix #3725.

### Behavior Differences

Firefox 78 begins to support `:read-only` besides `:-moz-read-only`, so the misusage of `:read-only` only causes some differences since it.

Chrome seems to have different opinions, and it doesn't match those `<input>` with `isSelectable()=false` into `:read-only`.

Here's a test page from web-platform-tests: http://wpt.live/html/semantics/selectors/pseudo-classes/readwrite-readonly.html , and you can see Firefox 78 passes all tests while Chrome 83 fails 15 tests.

There're also some related discussions that may help:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1532968#c11
* https://stackoverflow.com/q/62571297/5789722

### DIFF
--- a/pages/options.css
+++ b/pages/options.css
@@ -143,7 +143,7 @@ input#searchUrl {
   right: -320px;
   width: 320px;
 }
-input:read-only {
+input[type=text]:read-only, input[type=number]:read-only, textarea:read-only {
   background-color: #eee;
   color: #666;
   pointer-events: none;


Dabsen:
Thanks for this. First: I've been using it, and it works, including with the current master.

Second: This is a neat one-liner fix that solves a very annoying bug and seems unlikely to cause further problems. I'm suspecting (though I don't know) that the reason it's still pending might be that the description is not very clear on what the fix does. So here's my take.

Since FF 78, all buttons in the vimium interface have stopped working (issues #3623 #3634 #3678 #3653), including, annoyingly, the `Browse…` button to restore settings. 
The reason, it seems, is that FF changed its interpretation of the CSS pseudo-class `:read-only`. This used not to select buttons; now it does. As a result, one selector in vimium's `options.css` stylesheet now mistakenly applies to buttons—which makes them unusable! 
@gdh1995's fix solves this by adding further restrictions to the selector in question, so as to exclude buttons. I don't see any particular reason why these restrictive clauses should break anything, either on the FF side or on the Chrome side (whether Chrome changes behavior eventually or not…). Admittedly, I did not test this on Chrome. I can confirm it works with FF 79.

gdh1995:
Um, this is just because Vimium development is slow again - nothing about potential disadvantages of the changed line.

gdh1995:
@philc Hello, this PR seems very safe and can be merged.

wcpines:
I think I've run into into this issue (#3653) --I'm unable to restore a vimium options JSON file after migrating from Brave to FF.  It would be super rad to see this merged if it's safe to do so :D

philc:
Thanks @gdh1995, and @Dabsen for the nice writeup. This will go out with the next version of Vimium, to be released soon.

PR #3672, opened 2020-09-23
Remove unnecessary style at <input> for selecting backup file

rhysd:
## Description

With Japanese locale, `<input>` for selecting backup file is a bit broken. The message is squashed by '...' since its width is not sufficient. This is because `width: 200px` is set on the element.

スクリーンショット 2020-09-23 19 39 17

However, the style seems not necessary. I confirmed removing the style fixed this issue and did not break any other UI.

スクリーンショット 2020-09-23 19 44 16


philc:
Nice find, thanks!

PR #3679, opened 2020-10-01
Add a timeout for multiple-character actions

jdpopkin:
Implements #2170.

## Description
I saw that #2170 was marked as help-wanted, so I implemented this. I've been surprised by the current behavior before (I think I've accidentally lost my place in my Twitter timeline because I hit `g0` after having pressed `g` earlier). This PR adds a 30 second timeout, which seems extremely conservative. Let me know if I should make this configurable, though, and I'll dig a little deeper to try to figure that out.


philc:
Thanks for the PR. I have questions about whether this is worth adding, which I've added to #2170.

PR #3684, opened 2020-10-12
extend goNext & goPrev button with go,proceed,cancel,discard

Arshiamidos:
## Description
it's straight forward to go forward or backward via ']]' or '[[' without Shift+F,A,B(4key press). specially in modal forms


gdh1995:
we may make `goNext` accept an option named `patterns`, instead of adding words into the default list.

PR #3693, opened 2020-10-18
Move default search engine option the basic options

### LINKED ISSUE #3692: Move default search engine option
Currently the option to change the default search engine is located in the advanced options.
Most users would probably expect this options to be directly visible when opening the options menu since it is quite an important setting. For me personally it took a few minutes to figure out where to change the search engine used by vimium.

thomas-zahner:
Fix #3692

### DIFF
--- a/pages/options.html
+++ b/pages/options.html
@@ -88,6 +88,18 @@ b: http://b.com/?q=%s description
               <textarea id="searchEngines"></textarea>
           </td>
         </tr>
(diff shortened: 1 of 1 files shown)


philc:
As part of v2.0, all options are visible by default on the options page. I've also moved the default search engine higher on the page in f8d56273ed602a40c6f9b5722111c74a32c37c3d.

Issue #3641, opened 2020-08-17
VIMIUM-FF Smooth scrolling is very choppy when "privacy.resistFingerprinting" is set to true in about:config


clonex10100:
I'm not sure if this is Firefox's fault or the extensions fault, but I thought I'd post it here just incase someone else has choppy scrolling and can't figure out why.

**To Reproduce**
Ensure smooth scrolling is on in Vimium-FF and firefox preferences
Go to about:config and set "privacy.resistFingerprinting to true"
Scroll

**Browser and Vimium version**
Firefox Version: 79.0
Os: Arch
Vimium-FF Version: 1.66

gdh1995:
This is because when this configuration is true, Firefox gives Vimium fake timestamps (something like 0, 0, 0.1sec, 0.1sec, 0.1sec, 0.2sec, 0.2sec, 0.3sec, ...), and Vimium fails in computing scrolling distance if a second timestamp is equal with the one before.

kamkudla:
Are there any solutions to this? I can confirm that `window.scrollBy` works around the issue with timing:
```
window.scrollBy({
          top: 100,
          left: 0,
          behavior: 'smooth'
        });
```

gdh1995:
Vimium can work with `behavior: 'smooth'` - only when the "Use smooth scrolling" is disabled and you don't hold on the key of scroll commands.

kamkudla:
How do I implement this? I toggled off "Use smooth scrolling" and I am running v1.66.

PR #3701, opened 2020-10-27
added option to fix choppy firefox scroll

kamkudla:
**Problem:** Fake timestamps are provided to Vimium when `privacy.resistFingerprinting` is set `true` in Firefox which results in choppy and unpredictable scrolling. 
**Solution:** Added option for alternative smooth scroll using native `scrollBy` toggling between options `smooth` and `instant`

Fixes https://github.com/philc/vimium/issues/3641

### DIFF
--- a/content_scripts/scroller.js
+++ b/content_scripts/scroller.js
@@ -61,7 +61,7 @@ const performScroll = function(element, direction, amount) {
   const axisName = scrollProperties[direction].axisName;
   const before = element[axisName];
   if (element.scrollBy) {
-    const scrollArg = {behavior: "instant"};
+    const scrollArg = {behavior: Settings.get("scrollBehaviorSmooth") ? "smooth" : "instant"};
     scrollArg[direction === "x" ? "left" : "top"] = amount;
     element.scrollBy(scrollArg);
   } else {
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -177,6 +177,7 @@ const Settings = {
   defaults: {
     scrollStepSize: 60,
     smoothScroll: true,
+    scrollBehaviorSmooth: false,
     keyMappings: "# Insert your preferred key mappings here.",
     linkHintCharacters: "sadfjklewcmpgh",
     linkHintNumbers: "0123456789",
--- a/pages/options.html
+++ b/pages/options.html
@@ -141,6 +141,20 @@ b: http://b.com/?q=%s description
             <td verticalAlign="top" class="booleanOption">
               <div class="help">
                 <div class="example">
+                  Fixes choppy scroll in Firefox when privacy.resistFingerprinting is set true
+                </div>
+              </div>
+              <label>
+                <input id="scrollBehaviorSmooth" type="checkbox"/>
+                Use alternative smooth scroll (Firefox fix)
+              </label>
+            </td>
+          </tr>
+          <tr>
+            <td class="caption"></td>
+            <td verticalAlign="top" class="booleanOption">
+              <div class="help">
+                <div class="example">
                   In link-hint mode, this option lets you select a link by typing its text.
                 </div>
               </div>
--- a/pages/options.js
+++ b/pages/options.js
@@ -252,6 +252,7 @@ const Options = {
   ignoreKeyboardLayout: CheckBoxOption,
   scrollStepSize: NumberOption,
   smoothScroll: CheckBoxOption,
+  scrollBehaviorSmooth: CheckBoxOption,
   grabBackFocus: CheckBoxOption,
   searchEngines: TextOption,
   searchUrl: NonEmptyTextOption,


kabeersvohra:
Yep ran into this issue too, could this please be merged? :)

kabeersvohra:
@kamkudla I compiled your change in an unpacked version of the latest version of vimium and I am still getting choppy scrolling after selecting the `Use alternative smooth scroll (Firefox fix)` option that you added, I tried this with and without the normal `Use smooth scrolling` option selected, do you know why this may be?

kamkudla:
@kabeersvohra It's probably due to the new KV pair not getting set to storage unless you refresh 1) `Use smooth scroll` must be off 2) Make sure to press "Save Changes" in the Vimium options UI 3) Restart all previous tabs

Try `Settings.get('scrollBehaviorSmooth')` in console in the Vimium options page, does it return `true`?

kabeersvohra:
@kamkudla yes sorry about that, this was my bad that it was clashing with the normal vimium that I had installed already. However when I try and hold down j now it moves down for a small amount, stops for half a second and then scrolls but still in a slightly jittery manner, it is much smoother than the normal setting but still not the same as the original smooth scrolling. Perhaps this is a limitation of the 'scrollBehaviourSmooth' setting?

kamkudla:
@kabeersvohra Change the scroll step size in the Vimium options to something greater like 100px or to your preference and save. It could also be your system is resource constrained, close some tabs and check again. It's scrolling smoothly on my end with the key pressed.

kabeersvohra:
Thanks for the help, will have another go tomorrow and Let you know

kabeersvohra:
Sorry didn't get back to you, I am still having the same issue after changing the pixel size and trying on 3 different devices, are you getting exactly the same behaviour as the normal smooth scrolling with fingerprinting on?

kamkudla:
Go to the Vimium options page and open up the console - what does `Settings.get('scrollBehaviorSmooth')` return?

kabeersvohra:
It returns `true`, I do notice that the scrolling is smoother than with the option off, however when I hold it, it moves for a bit, stops, and then continues, here is the output:

ezgif-2-529b19adce94

gdh1995:
Does your version require `Use smooth scrolling` to be disabled, when `Use alternative smooth scroll (Firefox fix)` is enabled?

gdh1995:
During my tests on Firefox 80 x64, Win 10, with system animation enabled, it works only when I disabled Use smooth scrolling and enabled "alternative smooth scroll", and there's still a difference: the Vimium Options page scrolled much slower in such a case, compared with "the original smooth scrolling + not resist fingerprinting".

BTW, my working around for this issue is to assume a small time period of 17ms, when two timestamps from requestAnimationFrame are same:

https://github.com/gdh1995/vimium-c/blob/7c1b595d09aef528d445cb7f25dd0261f94c0448/content/scroller.ts#L76-L80

kamkudla:
@kabeersvohra Sounds like it could be a system issue. Behavior on my end is normal. I am on MacOS running Firefox Developer Edition 83.0b7. It could be the calling of `Settings.get` on every `performScroll` call - in order to isolate you can hardcode the value and bypass the setting. Change line 64 of `content_script/scroller.js` and rebuild and reinstall https://github.com/philc/vimium/blob/00228e0323af1f5aaf515f148e5596aed22a9571/content_scripts/scroller.js#L64

From `    const scrollArg = {behavior: Settings.get("scrollBehaviorSmooth") ? "smooth" : "instant"};` 
To `    const scrollArg = {behavior: "smooth"};`

`Use smooth scroll` in the options must be off, `Alternative smooth scroll` doesn't matter because we're bypassing it

kamkudla:
@gdh1995 Yes, `Use smooth scroll` has to be disabled for the alternative smooth scroll to work as there's a call that checks this setting and if disabled goes to 'jump scrolling' https://github.com/philc/vimium/blob/00228e0323af1f5aaf515f148e5596aed22a9571/content_scripts/scroller.js#L224-L229

Which then performs a native JS `scrollBy` bypassing the timing. Originally the `scrollBy` was hardcoded with `{behavior: "instant"}`. My fix is just a checkbox to change this setting from `"instant"` to `"smooth"`. Although the slowness on your end and @kabeersvohra could be a system issue as it works fine on my end. Hardcoding `{behavior: "smooth"}` on line 64 of scroller.js could help isolate the issue. 


Thanks for sharing your solution. I'll have to study the sequence a bit further. Is your scroll logic in your fork fundamentally different than this one?

gdh1995:
My animation logic is forked from this Vimium, and the scrolling step algorithm has no changes.

On my win10 laptop/PC, both Chrome and Firefox will make scrolling very slow, if `{behavior: "smooth"}`, and it's because smooth scrolling causes `element.scrollTop` change step by step, so in a middle time (like +17ms) point Vimium only gets a scroll offset of 1~2px, then a next `element.scrollTo(...)` in a next animation frame will stop the earlier scrolling action, and use the 2px as its new base to execute further animation - as a result, the `scrollTop` will be `0px, 2px, 3px, 5px, ...`, so the final result will be scrolling slows down.

quentincaffeino:
Thanks for the PR.

I think this is a good PR but it lacks in being generic. Scroll behavior is also supported by other browsers and should generally have better performance than custom. Wouldn't it be better to call it just `Native smooth scroll` and use it by default when it is available, ie:

```js
'scrollBehavior' in document.documentElement.style
```

If not - fallback to old custom smooth scrolling.

philc:
Rather than adding another option for Firefox which interacts with Vimium's existing smooth scrolling option, it would be nice if we could always use the browser's smooth scrolling behavior. I've opened #4306 for discussion.

Issue #3695, opened 2020-10-22
Problem installing in Firefox 


pwlodawi:
I'm trying to install in Firefox 78.3.1(32bit). After I accept permissions I get the following error
![vimium_fail](https://user-images.githubusercontent.com/49796436/96924981-c31b9b80-14b3-11eb-89cc-15080ed59888.png)
The message says: Add-on "Vimium" can not be installed because Firefox is not able to modify required file.

When I do Load Temporary Add-on it  loads and seems to work fine.
Any clues what could be the cause of the problem or how to troobleshoot further ?

kamkudla:
The problem arises because the Extension ID is not explicitly set in the `manifest.json` for the Firefox build in the `browser_specific_settings` key and Firefox requires this. The `application` key as used for the temporary debug version is ignored when trying to install an `.xpi` or `.zip`. I'm submitting a PR on this.

philc:
I believe this is fixed in 8e3c529af78423fe37e94c404286313a7a8220a5 and f5a14083fe45049e0d5b06be5b4dcff9feacca39 as part of v2.0. We can now generate a firefox-specific manifest during development with `./make.js write-firefox-manifest`

PR #3702, opened 2020-10-27
fixed corrupted firefox build by adding ID to manifest at build

kamkudla:
The Firefox build requires an explicit ID set in its manifest.json under the `browser_specific_settings` key as referenced by the docs [here](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings) and [here](https://extensionworkshop.com/documentation/develop/extensions-and-the-add-on-id/#when-do-you-need-an-add-on-id). `id` must be set since we're using the `storage` API. Without this an error is thrown that the extension is corrupted. Also, the `application` key is ignored. Here is the browser console logs: 

```
21:27:47.553 1603834067553	addons.xpi	WARN	Ignoring applications property in manifest
21:27:47.556 1603834067555	addons.xpi	WARN	Invalid XPI: Error: Cannot find id for addon /Users/vimium/dist/firefox/vimium-firefox-1.66.zip(resource://gre/modules/addons/XPIInstall.jsm:1539:19) JS Stack trace: loadManifest@XPIInstall.jsm:1539:19
async*init@XPIInstall.jsm:2039:18
createLocalInstall@XPIInstall.jsm:3007:20
getInstallForFile@XPIInstall.jsm:4315:25
XPIProvider[meth]@XPIProvider.jsm:3200:28
promiseCallProvider@AddonManager.jsm:253:31
getInstallForFile/<@AddonManager.jsm:1848:29
getInstallForFile@AddonManager.jsm:1862:7
getInstallForFile@AddonManager.jsm:4056:33
installAddonsFromFilePicker/</<@aboutaddonsCommon.js:274:42
```

This PR addresses this and fixes https://github.com/philc/vimium/issues/3695

### DIFF
--- a/make.js
+++ b/make.js
@@ -62,6 +62,7 @@ function buildStorePackage() {
   // Chrome considers this key invalid in manifest.json, so we add it during the build phase.
   firefoxManifest["browser_specific_settings"] = {
     gecko: {
+      "id": "vimium@vimium.github.io",
       "strict_min_version": "62.0"
     }
   };


philc:
I believe this is fixed in 8e3c529af78423fe37e94c404286313a7a8220a5 and f5a14083fe45049e0d5b06be5b4dcff9feacca39. We can now generate a firefox-specific manifest during development with `./make.js write-firefox-manifest`

PR #3718, opened 2020-12-03
add a chinese version README

pys078:
As the title


philc:
Thank you @pys078. I think for now, the benefit of having this localized readme is not quite worth the cost of maintaining it. I can't read this, so from here on out, others would need to modify this readme as we make changes to the main readme. Related: #3776.

PR #3724, opened 2020-12-12
Make FOCUS_LINK LinkHint mode available to Vimium via sl keybinding.

### LINKED ISSUE #1: Restoring of tabs should work for tabs closed with CMD+W
We may need to snapshot the tab during the chrome.tabs.onRemoved event instead of the D keystroke, although I'm not sure if at that point we can capture the page's URL, scroll position, and window ID.

stephenwithav:
Defaulted to `sl` shortcut since `f` is taken.

`sl` stands for "Spotlight link".  (Yes, I used thesaurus.com to find
synonyms for focus.)

## Description

This PR makes more sites accessible to users who can't use their hands, like me.  I recently rejoined Slack for a potential job and I prefer to browse the global "All Unread" section in the morning to catch up.  Navigating is a PITA, however, as my Pixelbook isn't as hackable as Linux to make it more user friendly; I use `xkbset` to map F7 to toggle MouseKeys mode on my Linux desktop.

This will also help me navigate more sites, like Manning and PacktPub.  If an area of a page requires focus to scroll up/down with j/k, its unusable to me.  (Thank you HTML5.)

There are several other sites this will help with.  Those three were the first to come to mind who may be able to sponsor Vimium.

### DIFF
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -315,6 +315,7 @@ const defaultKeyMappings = {
   "F": "LinkHints.activateModeToOpenInNewTab",
   "<a-f>": "LinkHints.activateModeWithQueue",
   "yf": "LinkHints.activateModeToCopyLinkUrl",
+  "sl": "LinkHints.activateModeToFocusLinkUrl",
 
   // Using find
   "/": "enterFindMode",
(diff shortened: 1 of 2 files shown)


gdh1995:
This PR is not so necessary: currently `LinkHints.activateMode action=focus` uses `FOCUS_LINK`, so you just need:

```
map sl LinkHints.activateMode action=focus
```

stephenwithav:
Thank you.  Actually reading the [wiki](https://github.com/philc/vimium/wiki/Tips-and-Tricks) was very useful for additional tricks. :)

philc:
Thank you for the PR Steven and thanks Dahan for the quick tip!

Issue #2903, opened 2018-01-04
Support other languages


jorgesumle:
Vimium design goals says:

> A secondary goal is to make Vimium approachable, or in other words, to minimize the barriers which will prevent a new user from feeling awesome.

Now, Vimium is only approachable to English-speaking people. I think that translation support should be added. Just a pot file is necessary for us to submit translations. But if you prefer a GUI webpage to translate, I suggest the free software website [Weblate](https://weblate.org). I can work on the Spanish translation and the Esperanto translation if the support is added.

philc:
While I certainly want people to have the extension in their native language, I don't believe it's worth the overhead of instrumenting the codebase with localization and maintaining the strings files, based on my current knowledge. I unfortunately don't have the bandwidth to maintain translated strings and documentation.

PR #3776, opened 2021-03-02
internationalization

joshmuente:
Approach to support translations.
https://developer.chrome.com/docs/extensions/reference/i18n/

## Description
Added some German translations mostly to test things.
My idea was to use `content_scripts/localize.js` on every page and use translate methods for "populated" html.
But before I continue working on it i would like to get your opinion first.

### DIFF
--- a/content_scripts/localize.js
+++ b/content_scripts/localize.js
@@ -1,0 +1,8 @@
+window.addEventListener('load', () => {
+    document.querySelectorAll('[data-locale_text]').forEach(elem => {
+        elem.innerText = chrome.i18n.getMessage(elem.dataset.locale_text)
+    });
+    document.querySelectorAll('[data-locale_placeholder]').forEach(elem => {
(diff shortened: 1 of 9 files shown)


content_scripts/localize.js
```diff
@@ -0,0 +1,8 @@
+window.addEventListener('load', () => {
```
gdh1995, line 1: This seems only for those html files in `pages/`, so we may move it to `pages/localize.js`.

philc:
While I certainly want people to have the extension in their native language, I don't believe it's worth the overhead of instrumenting the codebase with localization and maintaining the strings files, based on my current knowledge. I haven't heard much demand for localization to date (outside of #2903), and I suspect most of the userbase is able to use Vimium in its current form (given all of the other resources -- like the chrome store page and the github README -- are necessarily in English).

I'll leave this open to invite discussion. What do folks think?

PR #3779, opened 2021-03-05
Fix word movements in visual mode

jdmnd:
## Description
Performing word movements (pressing `e`) in visual mode would sometimes cause infinite mutual recursion between the `runMovement` and `runMovements` function. 

See the first commit – it's a one-line fix. The bug was probably introduced in the conversion from coffeescript to js. 

While I was figuring out how it worked, I decided to simplify a little. It turns out these recursive calls were unnecessary, and there was an alternative mechanism for moving the cursor and checking that progress was made.


philc:
I tested this and I couldn't see any obvious regressions. Thanks for the fix and nice simplification!

Issue #3367, opened 2019-07-30
Vimium tab navigation stops working with Google Docs / Gmail


geoffnewell:
When using 'J' and 'K' or gt / gT to navigate between tabs, this stops working when either Google Docs or GMail tab is hit.

In a Google Doc,  the 'J' and 'K' keys start typing in the doc.  
In GMail, the 'J' and 'K' keys don't work.

Is there a key modifier / hotkey for Vimium that could be invoked to take control when hitting these websites?  As a heavy G-Suite user, vimium is pretty limited for tab navigation.

gdh1995:
The situation means Vimium C was in "Insert" mode when a Google Docs tab gets active, so it wouldn't trigger commands on those keys.

Chrome and Firefox have provided an API of `chrome.commands` to support global shortcuts, which will be available whatever page you're using. But up to now Vimium doesn't use it.

So here's an AD: I have a customized version of Vimium https://github.com/gdh1995/vimium-c, and it supports this feature. The latest published version of Vimium C is v1.76.1 (v1.76.2 is waiting for reviewing and publishing), and it allows much more shortcuts. Here's also a related issue about shortcut usages: https://github.com/gdh1995/vimium-c/issues/55

geoffnewell:
Same problem with Google Sheets as well.  Suggestion from gdh1995 above sounds good for master branch.

pickle-and-pork:
@philc do you think a review of commit in a fork with attempt to understand if it's need to be incorporated into original vimum worth it? 

the commit in a fork version with the fix for this issue seems to be this one: https://github.com/gdh1995/vimium-c/commit/ad87e66f04ab2a2198a888aaf6b03904a5e38527


@gdh1995 can you please advise if this one part of the major refactoring you did and won't be possible to pick up here easily? Is it even okay to try to apply the solution you did in your fork to the original repo?

pickle-and-pork:
I think i see similar behaviour on youtube btw, default youtube shortcuts override vimium behaviour Vomnibar doesn't work.


However this one has a workaround if i press F or Shift + F it seems to renders the links and picks up the control over the page.

Unfortenatelly same workaround is not applicable for google sheets.

Issue #3279, opened 2019-03-25
Strategy Vimium & Google


Fandekasp:
There seem to be a lot of vimium issues related to google products.
Myself regularly using Gmail, Calendar, Docs, Spreadsheet & Presentations, it is rather bothersome.

Example: 
* Want to go to Gmail (first tab) ?  `g0
* Want to go back to previous tab ? `^` (oops, no, you disactivated vimium on gmail)
* Want to go to that 3rd tab on the left ? `3J` (oops, you forgot the 3, and entered a spreadsheet which took focus, and won't relinquish it with Escape)

Can Vimium experts share their tips and tricks on how to efficiently deal with google-heavy chrome sessions ? I would really appreciate it to boost my performance, and I'm assuming this could reduce the number of github issues created here.

smblott-github:
For Google docs, just disable Vimium.  I can't find any useful way of making the two work together.

For GMail, I just add the GMail bindings I prefer as [pass keys](https://github.com/philc/vimium/wiki/Disabling-Vimium#partially-disable-vimium-on-a-site).

gdh1995:
A Chrome extension is allowed to declare 4 global shortcuts, and if a user assign a key combination to a shortcut, them he/she can use the key if only a Chrome window is focused. Therefore I suggest that Vimium's previousTab/nexTab should be two configurable global shortcuts.

elquimista:
@Fandekasp Have you tried `i` (insert mode) ?

Fandekasp:
@elquimista as far as I know, the `i` vimium binding has nothing to do with this discussion. 
Its purpose is to disable vimium bindings in a page until Escape is pressed. 
What I'm looking for is to be able to force vimium shortcuts on all pages, whether google ones or not (or learn about tips on how other users manage a mouse-free experience with plenty of Google services tabs open).
@gdh1995 suggestion looks great, and @smblott-github tip seems quite useful, I'll spend some time trying to tune my vimium configuration later on, thank you!

gdh1995:
BTW, here's another idea about switching Vimium status quickly: https://github.com/philc/vimium/issues/3272#issuecomment-475296695 - it makes the exclusion rules not so "absolute" and give you more choices.

philc:
Closing due to age.

Issue #3254, opened 2019-02-24
Cannot disable for Google Slides


everyonesdesign:
I prefer to disable Vimium for Google Products because they have decent keyboard navigation by themselves; but I fail to do so when I edit Google Slides.

When I try to disable it, it gives current URL as "about:blank" and it still works after I try to block it.

![image](https://user-images.githubusercontent.com/3520704/53297947-7be48f80-3826-11e9-85a5-976fff97ba7c.png)

It seems that it switches to some iframe, because there're a few iframes with such a URL.

Is it a known issue? How can I overcome this?

gdh1995:
Currently, you can add another rule with a pattern of "https://docs.google.com" to also disable Vimium on the main (topest) frame.

### As far as the "about:blank"

Um, this is a *new* issue that I didn't expect. I think Vimium needs to re-think about its exclusion logic: it should also be disabled on such iframes, obeying what it is on the parent/top frame.

victortrac:
This has been driving me crazy for a few days.

gdh1995:
I've released a new version of modified Vimium (I call it Vimium C), and it applies rules from a top frame to such about:blank frames. You may have a try.

Link to download: https://chrome.google.com/webstore/detail/vimium-c/hfjbmagddngcpeloejdejnfgbamkjaeg or https://github.com/gdh1995/vimium-c/releases .

srsudar:
@gdh1995 , do you have any pointers to how you got rules from a top frame applying to `about:blank` frames? I looked through your changelog and tried searching the repo for `about:blank` but didn't see anything immediately obvious.

gdh1995:
Vimium C checks URLs which don't look like `*://*` and turns to apply passkeys of their top frames in:
https://github.com/gdh1995/vimium-c/blob/7e4672cd80ab59ba130e127edde116a6c136640e/background/exclusions.ts#L62-L67

binlaniua:
any update?

Issue #3190, opened 2018-11-21
White list of enabled keys by URL


borislt:
The topmost feature configuration in Vimium's settings is: 'Excluded URLs and keys'. 
Which is basically a blacklist.
But to enable only a couple of keys (e.g. scroll, tab switch) - you have to go through and disable all other keys. It's tedious.
Would be nice to have also a White list of enabled keys by URL.

davidgenesiscruz:
+1
I use keyboard shortcuts, for example, on video streaming sites to control playback. Vimium shortcuts would sometimes interfere with them so I disable Vimium on those URLs, but it would be nice if I could just keep the bare essentials like opening tabs.

afilp:
I also need this a lot. 

`Youtube` is one example. The `F` key there is "Full Screen". I had to exlcude "vimium" all together for Youtube, because I really need the "Full Screen" shortcut also.

smblott-github:
> Would be nice to have also a White list of enabled keys by URL.

Yes.  But it would also have to be obvious for new users.  And when you bundle it together with the feature request that you can also exclude (or include) modified keys, it all becomes quite complicated.

That's why I've never taken it on.  I just list the keys I want to use from the page.

@davidgenesiscruz  and @afilp ... You might be looking for insert mode (`i`), or perhaps the `passNextKey` command.

afilp:
@smblott-github Thanks! How about an interface similar to this?

![image](https://user-images.githubusercontent.com/7850073/51046146-fe462680-15cd-11e9-987d-1ff36ea6c266.png)

Notes:
- The `Keys` input box will be **disabled** when `All Keys` value is selected. This will have (internally) the same representation as using `Specific Keys` with no specific value in `Keys` (so same as today, but with a new explicit option too).
- The `All Keys Except` option allows you to add keys that "survive" any exclusion from other rules (i.e. "white list").

You may also use colors in the background of `Keys` (red/green) to show their purpose:

![image](https://user-images.githubusercontent.com/7850073/51046489-dc00d880-15ce-11e9-848f-530bce3dfd5a.png)

And this is the styling with the "All Keys" showing the `Keys` input as disabled.

off-by-one:
I second this - my current workaround is to list all keys I remember and then remove the keys I want to use. This is especially true on websites where I use vim controls internally (cocalc, overleaf), which mean I need very restrictive whitelists to keep vimium from interfering with the site.

gdh1995:
[...]

PR #3785, opened 2021-03-14
Expose global commands at chrome://extensions/shortcuts

srsudar:
## Allow Global Access to Some Commands

This PR adds the ability to set up hotkeys that are passed directly to the extension, without needing any of the hot key interceptions in a normal page. Visiting `chrome://extensions/shortcuts`, you see something like the attached image.

You need to define these in `manifest.json`, so they are static. I only bothered adding ones that I was interested in for now. If there is interest in adding something like this, I can try to be more complete.

This CL also adds a tab jump list (see `:h jumplist` in vim), which lets you move back and then forward through previously visited tabs. That isn't really related to this PR, but separating them cleanly before knowing if there's interest was more work than I wanted to do.

Some related issues are #3367, #3279, #3254, and #3190.

@gdh1995 has had some comments on this and might know more, including [here](https://github.com/philc/vimium/issues/3367#issuecomment-516357312).

Screen Shot 2021-03-14 at 1 47 06 PM


package-lock.json
```diff
@@ -1,8 +1,416 @@
 {
   "name": "vimium",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
fearphage, line 4: I'd revert this file since you aren't making any `package.json` changes.

background_scripts/main.js
```diff
@@ -350,6 +360,58 @@ const BackgroundCommands = {
   }
 };
 
+chrome.commands.onCommand.addListener(function(command) {
+
+  const sendCommandToCurrentTab = function(requestName) {
```
fearphage, line 365: A better name for this variable would be `command`.

background_scripts/main.js
```diff
@@ -350,6 +360,58 @@ const BackgroundCommands = {
...
+    case "switch-to-previous-tab":
+      chrome.tabs.query({ active: true, currentWindow: true }, (tabs) => {
+        if (tabs[0]) {
+          const tabId = tabs[0].id;
+          BackgroundCommands.visitPreviousTab({count: 1, tab: { id: tabId} });
+        }
+      });
+      break;
+    default:
+      console.error('unrecognized command: ', command);
```
fearphage, line 411: You don't need to add a space after the string. The logger joins multiple things together with spaces by default.

manifest.json
```diff
@@ -7,6 +7,32 @@
               "48": "icons/icon48.png",
              "128": "icons/icon128.png" },
   "minimum_chrome_version": "69.0",
+  "commands": {
+    "jump-back-tab": {
+      "description": "Jump back in the tab stack."
```
fearphage, line 12: Everywhere else you call it a tab list (instead of a stack). These should be consistent.

background_scripts/bg_utils.js
```diff
@@ -69,6 +104,84 @@ class TabRecency {
...
+  constructor(tabs) {
+    this.tabs = tabs;
+    this.activeIdx = tabs.length - 1;
+    // Tabs can be deleted after a TabJumpList has flattened the tabs into an
+    // array. Rather than O(N) look through the tabs, we'll just maintain
+    // deleted IDs and skip them.
+    this.deletedTabs = new Set();
+  }
+
+  isCoherent(currentTabId) {
```
fearphage, line 124: Did you mean `isCurrent` here?

background_scripts/bg_utils.js
```diff
@@ -69,6 +104,84 @@ class TabRecency {
...
+// the list is all open tabs--the current (i.e. most recently visited) tab is
+// the last element. The tab visited the longest in the past is the 0th tab.
+// Jumping backwards moves through the open tabs. The index is maintained,
+// allowing jumping forward to move again back to newer tabs. A manual
+// navigation through a mechanism other than a jump resets the jump list.
+class TabJumpList {
+
+  constructor(tabs) {
+    this.tabs = tabs;
+    this.activeIdx = tabs.length - 1;
```
fearphage, line 117: This name could be more descriptive/explicit. Is that `activeTabId`?

fearphage (changes requested):
I'm not a maintainer, but I had a few suggestions.

manifest.json
```diff
@@ -7,6 +7,32 @@
...
+    "open-vomnibox-bookmark": {
+      "description": "Open the vomnibox with history."
+    },
+    "open-vomnibox-bookmark-new-tab": {
+      "description": "Open the vomnibox with history."
+    },
+    "switch-to-previous-tab": {
+      "description": "Switch to the previously used tab."
+    }
+  },
```
gdh1995, line 35: Just a personal opinion:

Maybe we can use command names of Vimium directly, like what I've done in https://github.com/gdh1995/vimium-c/blob/dc5e026083db67fc6f471c907e5ca37b7b5a5316/manifest.json#L38-L63 .

Then we may add new syntaxes to key mappings to assign options to such global "shortcuts" (like https://github.com/gdh1995/vimium-c/wiki/Trigger-commands-in-an-input-box#shortcuts)

gdh1995 (changes requested):
Um, I suggest you use `git rebase` to modify commits and remove `package-lock.json` from this PR.

fredfortier:
Is some version of this implemented? Are global shortcut events dispatched outside of a frame as well (like in the address bar)? If so, I would welcome a universal "normal mode leader" kind of shortcut that sends the following keys to the active frame in normal mode if available. I certainly prefer this over special bindings to switch tabs, which isn't much different from using the built-in tab switcher.

philc:
Cool; thanks for the PR @srsudar. I don't see a strong need for global shortcuts now that the vomnibar-related use cases are handled in the latest new tab page support (#4795). The one exception is maybe tab switching, but that can be handled outside of this PR.

For the jump list, I don't have use for that command, but if you think it's more broadly useful, then the next step is to open an issue to discuss when it's useful.

srsudar:
Wow, I totally forgot about this! Happy to leave it closed. =)

Issue #3448, opened 2019-12-19
[Question] Cycling between input boxes


ahmetb:
On https://www.reddit.com/subreddits page I see 3 input boxes. Two of them are yellow, one is red after hitting `gi`. This makes me think I can cycle between the input boxes using `gi`, but the `?` documentation doesn't say so.

![image](https://user-images.githubusercontent.com/159209/71133991-77240a00-21b1-11ea-8863-7dba432a4419.png)

Is there actually a way to do this?

(Also FYI https://vimium.github.io/ seems to be out of date with `?` pane.)

innaterebel:
Forgot where I knew it, but it is <kbd>Tab</kbd>.
The `gi` also accepts count prefix, meaning you can directly <kbd>3</kbd>`gi` to focus the 3rd input in the viewport.

ahmetb:
Damn I have been trying gi2 gi3 etc.

ahmetb:
Let's keep open for adding documentation around this.

PR #3816, opened 2021-05-09
Updated command descriptions in README

aaronshenhao:
## Description

Updated description for "insert mode", "gi" and "g0" in README. For insert mode, added <c-[> as an alternative to Esc (for consistency). For gi, added detail on how to use <tab> to cycle through input boxes (see issue #3448). Also added detail on how ng0 will go to nth tab which is not intuitive.

The HUD should probably be updated too.


philc:
Thanks for this @aaronshenhao. The fact that `<c-[> ` is an alias for Esc is mentioned under the command listing in the readme, so I didn't include that.

PR #3840, opened 2021-06-24
Disable smooth scrolling in visual mode

webkadiz:
This PR disables smooth scrolling in visual mode because there is a problem in visual mode:
when the scroll key is held down (for example J), scrolling breaks and nothing is scrolled,
scrolling will work if J is pressed constantly, you have to press J, then press J again, and so on,
but it's long, you can't hold J, scrolling doesn't work that way.

I don't think it is possible to fix smooth scrolling to make it work in visual mode,
because I tried the native implementation of scrolling in the browser `el.scrollIntoView({ behavior: "smooth" })`
and it had the same problems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-2.0.0 (2023-09-28 -- partially released to the Chrome store)
+2.0.1 (2023-10-04)
+
+- Fix exception when migrating some pre-v2.0 settings. ([#4323](https://github.com/philc/vimium/issues/4323))
+
+2.0.0 (2023-09-28 -- partially rolled out to users on the Chrome store)
 
 - Support manifest v3, as now required by Chrome. This involved a partial rewrite and many changes.
   Please report any new issues [here](https://github.com/philc/vimium/issues).
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 3,
   "name": "Vimium",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "The Hacker's Browser. Vimium provides keyboard shortcuts for navigation and control in the spirit of Vim.",
   "icons": { "16": "icons/icon16.png", "48": "icons/icon48.png", "128": "icons/icon128.png" },
   "minimum_chrome_version": "102.0",


gdh1995:
On my Chrome 91 on Win 10, Vimium's Visual mode is able to make selection area scroll into view smoothly, when any key is pressed and its command gets executed.

What did you mean by `if J is pressed constantly` ? `j` should move down selection caret by one line, but not scroll a page.

webkadiz:
> j should move down selection caret by one line in Visual mode, but not scroll a page

When a user selects text that spans on to several screens they will obviously want to be able to see the selection.

> What did you mean by if J is pressed constantly

The only way that scrolling works for me is by pressing J repeatedly. Holding the key does not work. That's rather incovenient.

philc:
I've verified the poor UX that @webkadiz describes: when in visual (or visual line) mode, holding `j` will create a selection that eventually extends off screen, and once it does, the page does not scroll to keep the bottom of the selection in view, unless you type `j` repeatedly.

The fix in this PR works, but it has two issues:

1. We shouldn't be modifying state (especially `Settings`) around function calls. We can instead add a second parameter to `scrollIntoView` called `smoothScroll`, and in `scrollIntoView`, when that argument is null, initialize it with the value from `Settings.get("smoothScroll")`.
2. I don't actually understand why smoothScroll doesn't work in visual mode based on a scan of mode_visual.js. Before we implement some version of this workaround, can anyone expand on what is the actual underlying issue?

webkadiz:
I tried smooth scrolling with browser native function `element.scrollIntoView({ block: "nearest", behavior: "smooth" })`. This solving work, but it have problem: smooth scrolling starts work after keyup `J`, that is, I don't get feedback when I keep `J`.

I think scrolling in visual mode does not need smooth, it may not be possible to achieve the desired effect due to the fact that the visual mode jumps over elements of different heights and positions, for smoothness it may be necessary to predict the position of these elements in advance. In addition, the smooth scrolling that I added on my knee using the browser function is not a convenience

I suggest that we focus on the scrolling that my PR gives. But really need to take into account the remark about code @philc. I don't think I will be able to do this in the near future, maybe at the end of the month

tasnuva1:
Still want this feature, it's really necessary to being able to see what I'm selecting. I hope in the future it will be fixed.

PR #3841, opened 2021-06-25
Add special scrolling element for Telegram

webkadiz:
This PR provide special scrolling element for Telegram.
Vimium can't find the necessary element for scrolling after loading page, because Telegram is a SPA and its scrolling element is nested in the DOM structure


philc:
Great, thanks!

(I don't have a telegram account and so haven't actually verified this fix).

PR #3847, opened 2021-07-02
Fix frozen search mode

### LINKED ISSUE #3557: In the search mode, the screen does not move when querying the next one

RagnarGrootKoerkamp:
Fix #3557.

See the next post for the up-to-date fix. Below the line is the description of the previous fix.

---

During the coffeescript -> javascript migration, some lines that disable the PostFindMode early on were (unintentionally?) swapped in https://github.com/philc/vimium/commit/01895d8233766772a8b896c3b948359e786c4fdd.
Because of this, PostFindMode was entered after pressing enter, blocking all keypresses until a mouseclick in the window happens, or the browser window is unfocussed and focussed again.

This reverses them back to their order in the original coffeescript.

My testing shows that searching now works again.

On 2nd thoughts, this fix does introduce javascript errors, and those errors are likely why the broken behavior isn't triggered.

### DIFF
--- a/content_scripts/hud.js
+++ b/content_scripts/hud.js
@@ -121,7 +121,7 @@ const HUD = {
     if (exitEventIsEnter) {
       FindMode.handleEnter();
       if (FindMode.query.hasResults)
-        postExit = () => new PostFindMode;
+        postExit = () => newPostFindMode();
     } else if (exitEventIsEscape) {
       // We don't want FindMode to handle the click events that FindMode.handleEscape can generate, so we
       // wait until the mode is closed before running it.
--- a/content_scripts/mode_find.js
+++ b/content_scripts/mode_find.js
@@ -32,6 +32,13 @@ class SuppressPrintable extends Mode {
 //      inheriting from SuppressPrintable.
 //   3. If the very-next keystroke is Escape, then drop immediately into insert mode.
 //
+var newPostFindMode = function() {
+  if (!document.activeElement || !DomUtils.isEditable(document.activeElement))
+    return;
+
+  return new PostFindMode();
+}
+
 class PostFindMode extends SuppressPrintable {
   constructor() {
     const element = document.activeElement;
@@ -46,9 +53,6 @@ class PostFindMode extends SuppressPrintable {
       keypress(event) { return InsertMode.suppressEvent(event); },
       keyup(event) { return InsertMode.suppressEvent(event); }
     });
-
-    if (!document.activeElement || !DomUtils.isEditable(document.activeElement))
-      return;
 
     // If the very-next keydown is Escape, then exit immediately, thereby passing subsequent keys to the
     // underlying insert-mode instance.
@@ -284,7 +288,7 @@ class FindMode extends Mode {
 
     if (FindMode.query.hasResults) {
       focusFoundLink();
-      return new PostFindMode();
+      return newPostFindMode();
     } else {
       return HUD.showForDuration(`No matches for '${FindMode.query.rawQuery}'`, 1000);
     }
--- a/content_scripts/mode_insert.js
+++ b/content_scripts/mode_insert.js
@@ -67,7 +67,7 @@ class InsertMode extends Mode {
     return activeElement;
   }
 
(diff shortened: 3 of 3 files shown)


RagnarGrootKoerkamp:
Note that even with this PR, the state is still broken/frozen when the search result is a textbox. For example, refreshing GitHub and searching `a`, followed by pressing Enter gives a frozen state.
This is already broken on master, but working fine in the last release.

RagnarGrootKoerkamp:
Hmm, things may actually work as intended when the search result is a textbox.

Current behavior in 1.66:
- reload GitHub, search a, press enter
- textbox is focussed
- j/k scroll
- esc unfocusses the textbox
- i enters insert mode in the textbox

Behavior with this PR:
- reload GitHub, search a, press enter
- textbox is focussed
- all keys but escape are ignored
- escape enters insert mode in the textbox
- escape again enters normal mode

The documented behavior actually seems to be somewhat in the middle:
https://github.com/philc/vimium/blob/master/content_scripts/mode_find.js#L27-L34

RagnarGrootKoerkamp:
Ok, even though this PR does 'fix' the behavior, this is because it actually introduces a JS error (calling `return` before calling the super class constructor).

I still don't understand how all the keypress event handling code works, but at least this could indicate the way to a proper fix.

RagnarGrootKoerkamp:
Nevermind all previous findings. It turns out the fix is much simpler :)

In the migration, an accidental void return statement was added to the line below in https://github.com/philc/vimium/commit/e66c10805c0da33dd5a09da68666afbc367af052#:
https://github.com/philc/vimium/blob/cfc1b0d95be4b82b2a2afc6b3d90c6958ffe918c/content_scripts/mode_insert.coffee#L56
creating this line:
https://github.com/philc/vimium/blob/e66c10805c0da33dd5a09da68666afbc367af052/content_scripts/mode_insert.coffee#L61
which was then converted to
https://github.com/philc/vimium/blob/b4aa6adb6f79dafdc7bb2803e4a730abcea57975/content_scripts/mode_insert.js#L70
which isn't equivalent to the original anymore.

Side note: I only realised this very late one, but the debugging was complicated by the fact that I'm using ctrl as esc as well, and this first sends ctrl-key-down and ctrl-key-up events before sending the esc-down event, breaking the behavior when esc is the very first key to be pressed. But that's OK - just using `i` to go to insert mode is more natural to me anyway.

philc:
Great catch! This is why I hate converting source from coffeeScript. There are a few painful bugs like this one that have been introduced into master that still need to be tracked down.

Adding `return event` to `suppressEvent` improves the behavior -- you can now hit `n` to proceed to the next match. However, entering find mode for a second time on a page that has already had find mode triggered once locks the Vimium UI again. I haven't looked deeply; LMK if you can spot the issue.

RagnarGrootKoerkamp:
Phew - this was much easier to fix now. Turns out that the original fix in this PR (moving the if statement in the PostFindMode constructor before calling `super()`) was still needed as well. I've extracted it to a wrapper function and now call that one in both places that previously called the constructor directly.

gdh1995:
Then the last commit is a replacement of #3801 . What will `./make.js test` report with it?

RagnarGrootKoerkamp:
@gdh1995 `./make.js test` on the last commit in this PR (https://github.com/philc/vimium/pull/3847/commits/ccc2e9499aead106740a344e8bb7a5c2762f2934) runs passes all 169+109 tests.

RagnarGrootKoerkamp:
@gdh1995 FYI: replacing my wrapper function by your `this.exit()` breaks the tests again, so apparently we really need the wrapper function.

RagnarGrootKoerkamp:
@gdh1995 I think your commit using `this.exit()` also passes the tests if you wrap the `this.exit()` and `return` in some braces ;) Currently it just always `returns` right afterwards.

gdh1995:
Oh... You're right. I just made a stupid typo.

philc:
Hooray! Thanks @RagnarGrootKoerkamp

Issue #3832, opened 2021-06-07
UXSS inside Vimium vomnibar feature


barakolo:
**Describe the bug**
Attacker can use vimium omnibar feature to execute arbitrary JS code over any site.
This can be caused by combination of low entropy PRNG used as vimium "secretToken" as well as invalid messages sent between different content-scripts in page.

**Full Technical Description:**
https://docs.google.com/document/d/1BdsJ3IRV05dodVfgWHLDRgM1FTkgswiMPxGHl_ufNho/edit

**Bug Video:**
https://youtu.be/bklyYCtECF0

**To Reproduce**
Steps to reproduce the behaviour:
1. Checkout this GitHub:
https://github.com/barakolo/VimiumUXSS
2. Find out current "VimiumSecret" (This can also be extracted by simple brute-force because of bad PRNG as I explained in the sharable docs):
To do that - 
** open Chrome DevTools, and change the Javascript context to “Vimium extension” (Under the Console tab -> click on the arrow sign below it).
b. Now Run this line of code (we are now running it in zotero content-scripts context):
chrome.storage.local.get('vimiumSecret', e=>console.log(e['vimiumSecret']))
c. Save this code aside (this is a small number with a max 10-digits which is being used across all sites).
3. Change Line 27 inside abc.js to the current "vimiumSecret" extracted.
4. Now, open SimpleHTTPServer over this github repo:
python2 -m SimpleHTTPServer 8080
5. Surf to this site, and click enter once or two.
6. Observe that alert() is executed over "example.com" now.

**Browser and Vimium version**
Vimium Version: 1.66 (March 2, 2020)
Chrome version: 91.0.4472.77
OS version: Windows 10 Home 19042.985

barakolo:
In short - vulnerable lines are:
*The first vulnerable line that generates small secret is here:
https://github.com/philc/vimium/blob/bdf654aebe6f570f427c5f7bc9592cad86e642b5/background_scripts/main.js#L37

*The second vulnerable line is actually the one that gets executeScript message from the background-script and executing this JS in the incorrect frame:
* The background scripts triggers this vulnerable-line inside of the content-scripts:
https://github.com/philc/vimium/blob/bdf654aebe6f570f427c5f7bc9592cad86e642b5/content_scripts/vimium_frontend.js#L191
This line Doesn't verify request.sender (there might be in there frameId we can use to validate and such).
But probably it's best just to keep unique UUID in each content-script to use when sending/getting repsonses to avoid confusions.

[...]

PR #3850, opened 2021-07-03
Added crypto random string. verified.

barakolo:
## Description
A fixed security issue - the randomness wasn't enough as an attacker could have tried breaking this random over time, without any limitations.
Please look at the following issue: https://github.com/philc/vimium/issues/3832


philc:
Great work @barakolo on the bug and this PR. I've adapted the contents of this PR a bit in 1d80e66aa20f7f1d17d7f16ee046b1253368aa16, and included a stub so that the tests pass. Thanks!

PR #3851, opened 2021-07-04
Added Brave Search as an option for search suggestions.

lemonnuggets:
## Description
Just personally wanted to use the new Brave Search suggestions and thought that adding it as an option might be helpful for others as well.


matthewblott:
Oh wow, I just found the details to add this and was about to submit a pull request when I found it's already here. This was submitted a long time ago any reason for not merging?

philc:
Thanks @lemonnuggets. I pushed an updated implementation of this in 9e09b270.

PR #3853, opened 2021-07-10
Add feature: close one tab on the left/right

dyckia:
Add the feature of closing one tab on the left/right.


dyckia:
Add feature: close N tabs on the left or right of the current tab. 

If no count number is specified, it will close the most adjacent neighbor tab.

philc:
Hey @dyckia, thanks for building this out. I think we can implement this without having to introduce another command. See #4296; I'll close this in favor of that.

PR #3854, opened 2021-07-10
fix: ensure find query history in case of always private windows

### LINKED ISSUE #3614: n and N do nothing (Firefox, Ubuntu)
**Describe the bug**
After searching for text with `/<some text><ENTER>` `n` and `N` ought to jump to respectively the next and previous occurrences of `<some text>`. Instead, however, they do nothing.

**To Reproduce**
1. Go to URL `example.com`
2. Type `/am<ENTER>`
3. Type `n`
What ought to happen: jump to the next occurrence of `am`
What actually happens: nothing

The screenshot shows before and after step 3 (webpage is dark because of Dark Reader extension).
Vimium-Firefox-bug-n-N-do-nothing

about:support
Name: Firefox
Version: 78.0.1
Build ID: 20200630195452
Distribution ID: canonical
User Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:78.0) Gecko/20100101 Firefox/78.0
OS: Linux 5.4.0-40-generic
Application Binary: /usr/lib/firefox/firefox

about:addons
Version: 1.66
Last Updated: 3 July 2020

sid-maddy:
## Description
Fixes #3614.

In my observation, `this.rawQueryList` always remains `undefined`, if a user always browses in private/incognito windows.
As per my understanding, this is because of the way it is populated from local storage.

### Changes
Added a fallback when setting it from local storage changes to ensure `this.rawQueryList` is always defined.

### DIFF
--- a/lib/find_mode_history.js
+++ b/lib/find_mode_history.js
@@ -18,7 +18,7 @@ const FindModeHistory = {
       this.storage.get(this.key, items => {
         if (!chrome.runtime.lastError) {
           if (items[this.key]) { this.rawQueryList = items[this.key]; }
-          if (this.isIncognitoMode && !items[this.key]) {
+          else if (this.isIncognitoMode) {
             // This is the first incognito tab, so we need to initialize the incognito-mode query history.
             this.storage.get("findModeRawQueryList", items => {
               if (!chrome.runtime.lastError) {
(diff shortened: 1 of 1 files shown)


philc:
Thanks for tracking this down @sid-maddy !

PR #3860, opened 2021-07-14
Clean up and simplification

fearphage:
* switched `spawn` to use current default values and switched the calls to spawn to make use of the default values
* simplified `runUnitTests`
* cleaned up `buildStorePackage` to dynamically create manifests and de-duplicated some commands
* Changed `Process.exit` to exit with the number of tests that failed

I've run both functions locally to verify they are still functional.


philc:
Nice work, thank you!

PR #3870, opened 2021-07-21
Add option to exit insert mode on activating a tab

psxvoid:
## Description
Hi,

This PR adds an advanced option that allows exit insert mode after focusing on a tab.

### Why is this important?

Previously, when a user enters an insert mode and switches a tab using a mouse, then after using Vimium's `J/K` keyboard shortcuts for switching tabs it will NOT exit the insert mode, effectively not allowing a user to switch to the next tab (from a one with an active insert mode), until a user manually exits the insert mode (e.g. by pressing `Escape`). It was always counterintuitive for me and prevented me from using those hotkeys because sometimes they weren't working and I wasn't curious enough to ask why and eager enough to find a solution to avoid it.

P.S.: fixing unit tests is still in progress~

### DIFF
--- a/content_scripts/mode.js
+++ b/content_scripts/mode.js
@@ -75,6 +75,7 @@ class Mode {
       keydown: this.options.keydown || null,
       keypress: this.options.keypress || null,
       keyup: this.options.keyup || null,
+      focus: this.options.focus || null,
       indicator: () => {
         // Update the mode indicator.  Setting @options.indicator to a string shows a mode indicator in the
         // HUD.  Setting @options.indicator to 'false' forces no mode indicator.  If @options.indicator is
--- a/content_scripts/mode_insert.js
+++ b/content_scripts/mode_insert.js
@@ -25,11 +25,7 @@ class InsertMode extends Mode {
       if (Settings.get("passNextKeyKeys").includes(keyString)) {
         new PassNextKeyMode();
       } else if ((event.type === 'keydown') && KeyboardUtils.isEscape(event)) {
-        if (DomUtils.isFocusable(activeElement))
-          activeElement.blur();
-
-        if (!this.permanent)
-          this.exit();
+        this.removeFocusAndExit();
 
       } else {
         return this.passEventToPage;
@@ -42,7 +38,12 @@ class InsertMode extends Mode {
       name: "insert",
       indicator: !this.permanent && !Settings.get("hideHud")  ? "Insert mode" : null,
       keypress: handleKeyEvent,
-      keydown: handleKeyEvent
+      keydown: handleKeyEvent,
+      focus: (event) => {
+        if (Settings.get("removeFocusAfterSwitchingTabs") && event && event.target === window) {            
+          this.removeFocusAndExit();
+        }
+      }
     };
 
     super.init(Object.assign(defaults, options));
@@ -65,6 +66,19 @@ class InsertMode extends Mode {
     while (activeElement && activeElement.shadowRoot && activeElement.shadowRoot.activeElement)
       activeElement = activeElement.shadowRoot.activeElement;
     return activeElement;
+  }
+
+  removeFocusAndExit(element) {
+    element = element || this.getActiveElement();
+
+    if (element == null)
+      return;
+
+    if (DomUtils.isFocusable(element))
+      element.blur();
+
+    if (!this.permanent)
+      this.exit();
   }
 
   static suppressEvent(event) { return this.suppressedEvent = event; }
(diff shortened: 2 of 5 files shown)


gdh1995:
&nbsp;Well, in fact, we may add global (browser-level) shortcuts to trigger commands in background_scripts, by listening chrome.commands.onCommand



---Original---
From: "Pavel ***@***.***&gt;
Date: Wed, Jul 21, 2021 20:18 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;
Subject: [philc/vimium] Add option to exit insert mode on activating a tab (#3870)



Description
 
Hi,
 
This PR adds and advanced option that allows exit insert mode after focusing on a tab.
 
Why is this important?
 
Previously, when a user enters an insert mode and switches a tab using a mouse, then after using Vimium's J/K keyboard shortcuts for switching tabs it will NOT exit the insert mode, effectively not allowing a user to switch to the next tab (from a one with an active insert mode), until a user manually exits the insert mode (e.g. by pressing Escape). It was always counterintuitive for me and prevented from using those hotkeys, because sometimes they wasn't working and I wasn't curious enough to ask why and eager enough to find a solution to avoid it.
 
P.S.: fixing unit tests is still in progress~
 
 
You can view, comment on, or merge this pull request online at:
 
&nbsp; https://github.com/philc/vimium/pull/3870
 
Commit Summary
    
Add option to exit insert mode on activating a tab 
 
File Changes
    
     M     content_scripts/mode.js     (1)      

     M     content_scripts/mode_insert.js     (26)      

     M     lib/settings.js     (3)      

     M     pages/options.html     (15)      

     M     pages/options.js     (1)    
 
Patch Links:
    
https://github.com/philc/vimium/pull/3870.patch   

https://github.com/philc/vimium/pull/3870.diff

psxvoid:
@gdh1995 Could you please expand your thought?

From my perspective, global commands aren't solving the same issue.

I was also adding global commands into another chrome extension that has vim-like navigation inside the popup, and it wasn't flexible enough in terms of configuration. So, I decided to use key-down events instead and removed global commands.

gdh1995:
If a page breaks, or is `chrome://*`, then Vimium doesn't work on it, so the method using keydown events will fail. On the contrary, `chrome.commands` will always work.

As for configuration, we may extend syntax of key mappings. For example, I've added `shortcut <shortcutName> command=<commandName> [...<options>]` in my [Vimium C](https://github.com/gdh1995/vimium-c).

BTW, here's some ways to trigger keys in an input box in my Vimium C: https://github.com/gdh1995/vimium-c/wiki/Trigger-commands-in-an-input-box . I think the easiest way is "global shortcuts".

psxvoid:
@gdh1995 Any chances it can be merged? I'm still using locally build version of Vimium with this change. If there is a chance it can be merged, I would continue with fixing unit tests for this PR.

gdh1995:
Um, as far as I know, Vimium is in a very inactive (and slow) developing status.

BTW, recently I add some interesting features to my customized version, named [Vimium C](https://github.com/gdh1995/vimium-c), and now it can run a "`esc`" action once a command is triggered, and the usage is:
```
map <f7> runKey keys="nextTab+2dispatchEvent" o.esc
```

Explanations:
* `runKey` and its `keys`:  https://github.com/gdh1995/vimium-c/wiki/Auto-run-a-tree-of-commands
* `dispatchEvent count=2 esc=true`: means to force Vimium C to blur a currently active element.

psxvoid:
@gdh1995 Thanks a lot, I didn't realize there is "Vimium C" extension on the market as well! It seems it's working great and I'm switching to "it!

I remapped <kbd>K</kbd>,<kbd>J</kbd> to `runKey keys="nextTab+2dispatchEvent" o.esc` and `runKey keys="previousTab+2dispatchEvent" o.esc` respectively and they working as expected.

From a user perspective, I don't think anyone would guess that he can do such a trick (using the `runKey` to allow tab-switching work even when an element is focused). I still believe that adding an option for this feature is much better.

gdh1995:
Um, here's a tip only for Vimium C users: by default mappings of `K` and `J` won't work in insert mode, but `<f1>` - `<f12>` will. If someone do want to use `K`, `mapKey <K:i> <f1>` is necessary to notify Vimium C's Insert Mode to treat `K` as `<f1>` - see https://github.com/gdh1995/vimium-c/wiki/Trigger-commands-in-an-input-box .

gdh1995:
Um, here's a tip only for Vimium C users: by default mappings of `K` and `J` won't work in insert mode, but `<f1>` - `<f12>` will. If someone do want to use `K`, `mapKey <K:i> <f1>` is necessary to notify Vimium C's Insert Mode to treat `K` as `<f1>` - see https://github.com/gdh1995/vimium-c/wiki/Trigger-commands-in-an-input-box .

philc:
Thanks for the submission @psxvoid. Closing due to lack of interest.

Issue #3880, opened 2021-08-07
Feature Request: Normal and Visual mode in vomnibar


kiuKisas:
Hey, first of all, thank's for this plugin ! It's really pleasant to have a vim-like experience in the browser :)
I was thinking if it can be consider to add other mode in the `vomnibar`, `normal mode like` at least and `visual mode like` can be a cool addition.

The most obvious use-case is when editing the current URL, if we can replace some part of the URL quickly it cool be great.

As I saw it, vomnibar will open in `Insert mode` by default, then hiting `Esc` will switch to normal mode. A second `Esc` hit will quit vomnibar. As this behavior will not match all usage, I think it should also be an option that we can enable/disable in the vimium options/

gdh1995:
`Vomnibar` should have exited on the first `Esc`. Maybe there was something wrong with your Vimium, if it did enter normal mode in Vomnibar iframe.

kiuKisas:
No no, it's a feature request and I described the behaviour I would like if
vnibar had normal/visual mode.
I apologize if it was not clear

henwid:
I would like add a vote for something like this. In a perfect world it would work identically to the bash setting 'set -o vi' which enables vim-style command line history searching and editing.

francisjunr:
Here's another upvote for this feature request. If you think its not possible for some reason, adding redline short cuts, like in vim insert mode where you can use ctrl-e to move to the end of the line, ctrl-a to move to the begining, ctrl-h to backspace etc, would also do the trick to an extent. And hopefully add it to insert mode everywhere and not just vomnibar

PR #3881, opened 2021-08-07
Vomnibar: fix a bug preventing ctrl-enter from working

gdh1995:
Chrome may assign `"\n"` as the `key` property of an `Enter` key,
which made `Ctrl+Enter` not work as expected.

This helps commit 2268e498637bc3190070e3d32146f981a33be726 (see #2914) work well on Windows, and may help #3880


philc:
I didn't test this on Windows, but it doesn't break OSX. Thanks @gdh1995

gdh1995:
Added: Here's my test on Chrome 98 and Win11:

image

Issue #3610, opened 2020-06-28
feature request: Safari support on macOS Big Sur with new Web Extension API


zhigang1992:
https://developer.apple.com/documentation/safariservices/safari_web_extensions

with *Safari Web Extensions* we should be able to support Safari using the same codebase.

This would be very great since current alternatives on Safari doesn't have feature parity with Vimium.

gdh1995:
It requires an Apple Developer account, which means 99 US dollars every year...

neilyio:
Is there anyway to donate to the Vimium effort to help cover the cost of something like this? The Safari Web Extensions API was probably the WWDC announcement that excited me most, particularly because of the potential of seeing Vimium work.

zhigang1992:
We have couple of options here:

1. We can have donations to cover the cost. (GitHub Sponsor etc)
1. We can sale the app in AppStore to cover to cost. Like (Dark reader)
1. Deploy with some one's personal account (I'm happy to contribute mine since we need the enroll in Apple Developer Program anyway)

nbelzer:
I'm also interested in this. Currently I'm working a bit on Vimari, one of the Safari alternatives out there on [my own fork](https://github.com/nbelzer/vimari). As I'm running the macOS Big Sur beta I was able to successfully run the Safari Web Extension converter on the Vimium project with the following result:

```
$ xcrun safari-web-extension-converter ./vimium --project-location ./vimium-xcode
Xcode Project Location: ./vimium-xcode
App Name: Vimium
App Bundle Identifier: com.yourCompany.Vimium
Language: Swift
Is this correct? [yes]: yes
Warning: The following keys in your manifest.json are not supported by your current version of Safari. If these are critical to your extension, you should review your code to see if you need to make changes to support Safari:
        history
        clipboardRead
        match_about_blank
        bookmarks
        sessions
        notifications
```

Initially the extension could not be run as there was an error in the manifest `content_scripts` on line `78` where the `matches` key is defined for `content_scripts/file_urls.css`.
```
"matches": ["file:///", "file:///*/"],
```
I'm not familiar with the codebase for Vimium but I figured that removing a css file from the manifest wasn't going to break the application, if it even worked at all, so I removed this section of the manifest. After this I was able to build and activate the extension in Safari 14. 

I did encounter an issue where upon activation Safari crashes due to it not being able to convert something to grayscale (presumably the logo, as all extension logos are grayscale). The icon of Vimium also doesn't show up in Safari after a restart.

[...]

PR #3885, opened 2021-08-12
Safari web extension [WIP]

cheif:
# Description
This creates a Safari web extension-wrapper around the current code, with the idea of being able to run Vimium natively in Safari (See #3610). I've basically just taken the steps from other users in that issue and wrapped it up in a PR for now, with the hope that we can collaborate on this here.

# Status
Right now this is runnable, but not everything is working, as in https://github.com/philc/vimium/issues/3610#issuecomment-884343691 some commands doesn't work (see below).

## How to test
1. Get the branch for this PR.
2. Open `Vimium/Vimium.xcodeproj` in XCode (You'll need XCode, I'm running XCode 12.5)
3. Run the project, then you'll need to make sure that you have enabled running unsigned extensions in safari: https://developer.apple.com/documentation/safariservices/safari_web_extensions/running_your_safari_web_extension
4. On further launches you'll just need to build in XCode, and it should reload properly.
5. For debugging, open `Develop>Web Extension Background Pages>Vimium` in Safari.

## Things to test
- [ ] Test through all known features, to make sure that they work as expected.
- [ ] Test on iPad/iOS

## Known bugs:
- [ ] Commands requiring the clipboard (`yy`, visual mode copy, `P`, etc.)
- [ ] `B` doesn't have access to bookmarks ([`bookmarks`-API isn't supported in safari](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_support_for_JavaScript_APIs#bookmarks))
- [ ] `F` loads a URL in the currently selected tab rather than a new one
- [ ] `X` doesn't seem to restore the last closed tab ([`sessions`-API isn't supported in safari](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_support_for_JavaScript_APIs#sessions))

# Questions/considerations
- [ ] Right now this PR adds a `Vimium` folder with an `.xcproj`, this is what was created by running `xcrun safari-web-extension-converter`, but might not be the desired folder structure
- [ ] Some code in `main.js` is running conditionally on whether or not we're on `chrome`, is this safe, or should we find a way to trigger it in similar conditions in safari?
- [ ] The `manifest.json` was changed to have a broader `matches` for one `.css` file, is this safe or do we need to do something smarter here?
- [ ] In #3610 there were some discussions on using other artwork, since the current logo has the Chrome logo in the background.


cheif:
Looking into the API availability a bit, and it seems like the `sessions` API isn't available in safari (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Browser_support_for_JavaScript_APIs#sessions), which is the reason for `X` not working.

gdh1995:
`content_scripts/file_urls.css` is only to make some text like file names selectable on `file:///*`, so if Safari has no such problem then it can be removed when packing as a Safari extension.

gdh1995:
BTW, if I only write extensions and always use them by `loading unsigned extensions`, do I need an Apple Developer account (which means US $99 per year) yet?

cheif:
> `content_scripts/file_urls.css` is only to make some text like file names selectable on `file:///*`, so if Safari has no such problem then it can be removed when packing as a Safari extension.

Hmm, I tried just creating a local `test.html` and opening it, but it seems like the safari extension can't run at all on `file://` url:s, so it's probably safe to strip it for safari.

cheif:
> BTW, if I only write extensions and always use them by `loading unsigned extensions`, do I need an Apple Developer account (which means US $99 per year) yet?

Yeah, I think that should be possible, although I guess the "load unsigned" stuff might toggle off after restarting etc. as well, haven't looked into it much myself though.

cheif:
I've managed to fix the `F` issue by directly communicating with the bg-page (instead of simulating a CMD-click), this seems to be done in a lot of other cases, so I think the change should be pretty safe.

gdh1995:
No. You change will prevent pages using JavaScript listeners and doing whatever they want. Please consider about update `lib/dom_utils.js#simulateClick` to make Vimium work on Safari in a similar way as on Firefox.

cheif:
> No. Your change will prevent pages using JavaScript listeners and doing whatever they want. Please consider about update `lib/dom_utils.js#simulateClick` to make Vimium work on Safari in a similar way as on Firefox.

Hmm, I think I understand the problem, although I'm not able to get it working the way I thought.
The `MouseEvent` that's created/dispatched in `simulateMouseEvent` is handled by safari, but for some reason the `metaKey` is ignored. So if I just change `simulateClick` to have something like `Utils.isFirefox() || Utils.isSafari()` then a link will be opened in both the current tab + a new tab.

I guess we could just tailor the behaviour specifically for `F` on Safari, or do you see a better way of doing it? (that could be done with something like:
```javascript
if Utils.isSafari() && modifiers.metaKey { 
  simulateClickDefaultAction(element, modifiers)
}
```

cheif:
I've also looked into the yanking/pasting a bit, and I managed to get the yanking working by using the [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API), although I can't get pasting working through that. Perhaps because it needs to happen in a user-interaction chain, some more info [here](https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/).

Re: the clipboard-API, would it be safe to switch to using that, instead of using the `execCommand` stuff currently in use?

gdh1995:
Well, an ideal way is to hook the process of event propagation, and append `event.preventDefault()` if only no user code calls it.

But the code will be too complicated ... Seems we have to skip normal event dispatching directly.

gdh1995:
I'm not sure. I suggest we only call `navigator.clipboard` on Safari, since it has too strict limits.

gdh1995:
As far as I know, there will not exist any reliable method to read or paste from an extension, in Manifest V3 (https://developer.chrome.com/docs/extensions/mv3/intro/mv3-overview/), which is very annoying, but it seems nothing can be done. `navigator.clipboard` adds a much stricter limits for such shortcut kinds of extensions.

SamirHafez:
Has anything changed with regards to Web Extensions on Safari this past year?
IMO, there is still good appetite for having full fledged Vimium on Safari

Issue #3661, opened 2020-09-08
Vomnibar custom search delay in Firefox


SylEleuth:
Problem:
After opening vomnibar (o, shift+o) and entering a letter (e.g. "w" for wikipedia) and then space, I have to wait around 4-5 seconds for it to switching to chosen search engine.

Running Firefox with only Vimium enabled gives the same effect.
Removing and installing Vimium doesn't change anything, settings are the same. Is it possible to make a clean start? I can't find where settings for addons are stored.

This might be a problem that I could easily resolve just resetting everything in Firefox, refreshing profile or just starting a new one and slowly migrating there but I'll do it only if nothing else would work, that is why I wanted to ask people smarter than myself.

philc:
This might be resolved since #3887 was merged, and the Vomnibar has been written since this bug was filed. Please reopen if it's still slow.

PR #3887, opened 2021-08-13
completer: fix a typo causing too many refreshing

### LINKED ISSUE #3886: Feature Request: Allow Vomnibar.activate only with custom serach engine completion
To currently use a custom search engine, I have to open the Vomnibar, enter my keyword and then wait until completions for tabs, bookmarks and history are finished before the custom search is activated.

This takes several seconds, if one has many bookmarks, tabs and/or a long history.

Several issues have already talked about the slow performance:
- https://github.com/philc/vimium/issues/3661
- https://github.com/philc/vimium/issues/2131
- https://github.com/philc/vimium/issues/1802
- https://github.com/philc/vimium/issues/1790

It would be nice to use custom search immediately without having to wait for redundant completions.

gdh1995:
I find a typo making Vomnibar slow down when there're too many bookmarks, which was imported when migrating from CoffeeScript to ES6. And it's very easy to fix.

This typo was included in v1.67, so #3661 seems not related with this, but I think this PR can fix #3886 .

### DIFF
--- a/background_scripts/completion.js
+++ b/background_scripts/completion.js
@@ -702,8 +702,8 @@ class MultiCompleter {
 
   cancel(port) {
     for (let c of this.completers)
-      if (c.refresh)
-        c.refresh(port);
+      if (c.cancel)
+        c.cancel(port);
   }
 
   filter(request, onComplete) {


philc:
image

PR #3890, opened 2021-08-18
Make vimium generate invariant hints for RPA applications

wswag:
This PR modifies the generation of hints in non-link-mode to...

- ... provide invariant element hints which won't change on site updates, scrolling area, etc. - this is done by hashing either the element.id if present, or the outerHTML string, and taking the first 3 base36 chars of this hash.
- ... force an exact match to follow a hint, so all letters have to be typed

the background of this is to make vimium applicable for stable RPA applications where content/context dependent element hints are a severe problem. It is not intended to be merged without providing an option to activate this behaviour but to provide a baseline for making this a feature in the future.

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -447,12 +447,13 @@ class LinkHintsMode {
 
   updateKeyState({hintKeystrokeQueue, linkTextKeystrokeQueue, tabCount}) {
     Object.assign(this.markerMatcher, {hintKeystrokeQueue, linkTextKeystrokeQueue});
-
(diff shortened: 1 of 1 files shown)


philc:
Thanks @wswag for the prototype. I get the use case.

I haven't found a need to use this myself, and I think it's too niche of a use case to merge into the main repo. There's more info in [this section](https://github.com/philc/vimium/blob/master/CONTRIBUTING.md#which-pull-requests-get-merged) of the contributing guide for why I'm trying to keep main repo focused on the core features.

PR #3910, opened 2021-09-15
Chore: small refactor for user agent detection

onpaws:
As a Vimium user who spends time in Chrome DevTools I've been finding it mildly irksome to repeatedly experience the following warning in Chrome v96.0.4642.0 (and likely newer).
image

## Description
It seems the 'classic' way of accessing user agent info has been deprecated. Fortunately the revised [user agent API](https://blog.chromium.org/2021/05/update-on-user-agent-string-reduction.html) still contains the platform info that `keyboard_utils.js` apparently needs, so this PR is an attempt at using the new approach and thus avoiding the warning.

### Note 
I have not tested this thoroughly. While I believe it should provide the minimum necessary changes to squelch the warning, please test in whatever ways are appropriate.

### DIFF
--- a/lib/keyboard_utils.js
+++ b/lib/keyboard_utils.js
@@ -9,19 +9,38 @@ const KeyboardUtils = {
   },
 
   init() {
-    // TODO(philc): Remove this guard clause once Deno has a userAgent.
-    // https://github.com/denoland/deno/issues/14362
-    // As of 2022-04-30, Deno does not have userAgent defined on navigator.
-    if (navigator.userAgent == null) {
-      this.platform = "Unknown";
-      return;
-    }
-    if (navigator.userAgent.indexOf("Mac") !== -1) {
-      this.platform = "Mac";
-    } else if (navigator.userAgent.indexOf("Linux") !== -1) {
-      this.platform = "Linux";
+    if (navigator.userAgentData) {
+      /**
+       * navigator.userAgentData is supported by Chrome + Edge
+       * @see https://caniuse.com/mdn-api_navigator_useragentdata
+       */
+      const { userAgentData: { platform } } = navigator;
+      if (platform.indexOf("macOS") !== -1) {
+        this.platform = "Mac";
+      } else if (platform.indexOf("Linux") !== -1) {
+        this.platform = "Linux";
+      } else {
+        this.platform = "Windows";
+      }
     } else {
-      this.platform = "Windows";
+      // TODO(philc): Remove this guard clause once Deno has a userAgent.
+      // https://github.com/denoland/deno/issues/14362
+      // As of 2022-04-30, Deno does not have userAgent defined on navigator.
+      if (navigator.userAgent == null) {
+        this.platform = "Unknown";
+        return;
+      }
+      
+      /**
+       * For older browsers, use legacy navigator.userAgent
+       */
+      if (navigator.userAgent.indexOf("Mac") !== -1) {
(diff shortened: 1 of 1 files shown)


onpaws:
I think I'm noticing that the codebase is ES5 where as this PR is ES6.
Do we need definitely need to support ES5 only versions of Chrome? If so please let me know and I can adjust accordingly.

gdh1995:
The code base is ES6, since it uses lots of arrow functions.

onpaws:
Thanks for the heads up.
BTW, it looks like Chrome's [timeline](https://blog.chromium.org/2021/09/user-agent-reduction-origin-trial-and-dates.html) for the rollout of this

gdh1995:
Um, `userAgentData` is wholly a mess... Anyway, extensions have to eat this \*\*\*\*.

philc:
@onpaws thanks for knocking out this chore. I'm happy to merge it -- can you modify it so that it merges cleanly?

gdh1995:
@philc Sorry, but I forgot to address that this PR requires `navigator.userAgentData` must exist, which is not true when a browser is Firefox or before Chrome 83.

onpaws:
Thanks for reviewing my little PR @philc

Today I rebased on master to merge cleanly, and did a small refactor based on the concerns @gdh1995 raised + the Deno guard I noticed. It's the same in principle as before, but safer.

- `navigator.userAgentData` is gated behind a feature detect - we only use when available.
  That should squelch the warning for Chrome+Edge. (I confirmed on Chrome v104.0.5071.0 + macOS)
- If `navigator.userAgentData` is unavailable, which sometimes [may be the case](https://caniuse.com/?search=navigator.useragent), we keep the previous platform detection as-is.
- Regarding the Deno guard clause, while I don't have context on the specifics for that, I noticed the link in the comment there. There may be some recent [developments](https://github.com/denoland/deno/issues/14362) that could affect whether we still need it there. Just wanted to let you know.

NB: I only tested this on macOS b/c I don't have easy access to a Linux/Windows machine. 

FWIW `./make.js test` passed:

```shell
Check file:///Users/paws/Developer/vimium/make.js
run: started
test: started
Pass (169/169)



Pass (108/108)
test: finished (1371ms)
run: finished (1373ms)

```

philc:
Checking in here -- it looks like Chrome has completed its navigator.userAgent [migration](https://www.chromium.org/updates/ua-reduction/#proposed-timeline) and the information we need to determine the platform (Mac, Linux, or Windows) is still present in the user agent string. I also don't see a warning being issue in Chrome's dev Console.

Since userAgentData is not supported by Firefox, and navigator.userAgent has what we need, we can keep the code as-is, correct?

onpaws:
Thanks for checking in on this @philc !  
I also no longer observe the warning in the Chrome console on v117.0.5898.0 either, so considering that was the original issue this PR sought to address, I'll close it.

Not sure if Chrome's "Issues" tab warning for `navigator.userAgent` usage was spurious in the first place, or if the Chrome folks' plans changed, or what exactly. I suppose my takeaway is to regard future Chrome JS console warnings with considerably more skepticism, unless I'm mistaken it seems like much ado about nothing in the end. 🤔

gdh1995:
Um, I can still reproduce the issue report on `Issues` panel of DevTools:

image

onpaws:
On this version of Chrome Canary
image

`Issues` tab is empty for me
image

If you're on the stable channel, you may have to wait longer.

PR #3925, opened 2021-09-28
Add feature move current tab to an existing window

dyckia:
## Add feature move current tab to an existing window

Add a new command `Vomnibar.moveTabToWindow` which moves the current tab to an existing window.

Press key w which brings up the omnibar that displays a list of existing windows.

Once a window is selected, the current tab will be moved to the end position in the selected window, and focus is automatically switched to that tab.

- Suggestions only display the windows that have the same state (normal/incognito).
- Suggestions will automatically hide the window the tab is currently in.


developomp:
can this PR be merged?

philc:
Hey @dyckia, sorry for the delay.

This seems like a reasonable feature and the implementation looks good. But for completeness, can you describe some of your use cases where you need this command?

Second, the code base has changed quite a bit recently and this will need to be reworked a bit to work again.

Third, there are some unrelated changes (the addition of closeOneTabOnLeft and other commands, scrollPageUp magnitude) which should be removed.

IdoKendo:
Hi @philc @dyckia I am also interested in this feature quite a bit. If this PR is not panning out I would be willing to try and implement it myself, if it's fine by you.

philc:
Hey @IdoKendo, I think now is a good time to consider this. Would take a stab?

gliptak:
alternative implementation at https://github.com/philc/vimium/pull/4168

IdoKendo:
> Hey @IdoKendo, I think now is a good time to consider this. Would take a stab?

Hey @philc please see my implementation at #4417

philc:
Ok great. I’m traveling this week and will review when I’m back.

philc:
I think the UX for this feature needs more discussion, so I've opened #4640.

dyckia:
> I think the UX for this feature needs more discussion, so I've opened #4640.

Hi @philc, do we have a finalized UX for this feature yet? If so, I’m happy to update this PR and make it compatible with the latest codebase.

PR #3948, opened 2021-11-06
Add support for escape map, unmap, ifurlmatches, endif.

kzaher:
**ifurlmatches and endif only function for escape map/unmap.**

It also logs unparsable lines from config file instead of ignoring them.

This is a proof of concept so it's possible to write something like

```
ifUrlMatches https://colab.research.google.com/.*
  unmap <escape>
  map <s-escape> <escape>
endif
```

This implements https://github.com/philc/vimium/issues/1188.

It doesn't implement full blown support for ifUrlMatches, but only for escape key. This for me gives the biggest benefit.

Something similar could be trivially implemented for other commands if command file is parsed per web page and not globally.

Because this is not implemented I had to add another level of indirection in configuration and store per region mappings. This could be removed in case command file is parsed per web page.

If somebody could move command parsing to be per web page in that case this could be trivially implemented for all commands.

I'm making this PR primarily so I don't lose these code changes since Colab is useless without them with vimium, and vimium is useful for quickly jumping to different headers.

I'm aware of `map <c-]> passNextKey` hack, but this is not a realistic solution for heavily used sites. It is better to just turn off vimium in that case.

If somebody is frustrated like me with vimium interaction with rich web sites and finds this useful, cheers.

If there is a desire to merge this, I can add unit tests, but can't spend much time on this otherwise.

### DIFF
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -19,17 +19,38 @@ const Commands = {
     this.mapKeyRegistry = {};
 
     const configLines = Object.keys(defaultKeyMappings).map((key) => `map ${key} ${defaultKeyMappings[key]}`);
-    configLines.push(...BgUtils.parseLines(customKeyMappings));
+    configLines.push(...BgUtils.parseLines(customKeyMappings).reverse());
 
     const seen = {};
     let unmapAll = false;
+    let escapes = { '': {"<escape>": true} };
+    let urlStack = [''];
+
+    function getCurrentEscapes() {
+      url = urlStack[urlStack.length - 1];
+      if (!(url in escapes)) {
+        escapes[url] = { "<escape>": true };
+      }
+      return escapes[url];
+    }
+
     for (let line of configLines.reverse()) {
-      const tokens = line.split(/\s+/);
+      const tokens = line.trim().split(/\s+/);
       switch (tokens[0].toLowerCase()) {
+      case "ifurlmatches":
+        urlStack.push(tokens[1]);
+        continue;
+      case "endif":
+        urlStack.pop();
+        continue;
       case "map":
         if ((3 <= tokens.length) && !unmapAll) {
           var _, optionList, registryEntry;
           [_, key, command, ...optionList] = tokens;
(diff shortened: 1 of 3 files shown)


gdh1995:
It seems this PR deletes the feature of using `<c-[>` as `Escape`, which I think is not what we want.

BTW, I have a simpler way to pass `<escape>`: just update exclusion passKeys to support composed keys, like `<escape>` and `<c-[>` (currentlly it only supports single characters) . And I have implemented this feature in my `Vimium C` (https://github.com/gdh1995/vimium-c).

kzaher:
Hi @gdh1995,

Thanks for answering. I think making `<c-[>` work would be trivial, but since there is a pretty negligible chance of this being merged, I didn't bother.

> And I have implemented this feature in my Vimium C (https://github.com/gdh1995/vimium-c).

I unfortunately can't use Vimium C. Vimium has gone through a security review by my employer while Vimium C has not.

There were even objections to some of the methods in this repo from security perspective. Some functionality has been disabled, like
```
diff --git a/content_scripts/vimium_frontend.js b/content_scripts/vimium_frontend.js
index dd803ed1..35b559b4 100644
--- a/content_scripts/vimium_frontend.js
+++ b/content_scripts/vimium_frontend.js
@@ -188,7 +188,7 @@ const initializePreDomReady = function() {
     },
     linkHintsMessage(request) { return HintCoordinator[request.messageType](request); },
     showMessage(request) { return HUD.showForDuration(request.message, 2000); },
-    executeScript(request) { return DomUtils.injectUserScript(request.script); }
+    executeScript(request) { return console.log("This should be disabled"); }
   };
 
   chrome.runtime.onMessage.addListener(function(request, sender, sendResponse) {
```
I agree that even after https://github.com/philc/vimium/issues/3832 it's safer to just disallow this.

philc:
@kzaher, thanks for prototyping this.

I think full-blown support for `ifurlmatches` is probably too niche to live in the main repo.

I think the remaining discussion about escape is valuable. Can this be closed in favor of #4284?

PR #3981, opened 2022-01-08
Add Etymonline completion

stanoarn:
## Description
This PR adds search completion for [Etymonline](https://www.etymonline.com), the online etymology dictionary.


philc:
Thanks for putting this together. I think this is too niche to be in by default. We don't even have a dictionary-website completion engine, and presumably that is a much more common use case. I'm not sure how UpToDate and Qwant made it into the current code base.

PR #3993, opened 2022-01-24
Add command to open current tag in an incognito window

ehdr:
## Description
It is often useful to open a page in incognito mode, e.g. if you want to see what a page looks like when not logged in. Instead of having to yank the URL of the current tab, open a new incognito window, and paste the URL, this new command combines those actions into one.

The new `openTabInIncognitoWindow` command is bound to `yI` by default (analogous to `yt` for `duplicateTab`), and listed under "Advanced commands".

### DIFF
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -238,6 +238,7 @@ const Commands = {
       "firstTab",
       "lastTab",
       "duplicateTab",
+      "openTabInIncognitoWindow",
       "togglePinTab",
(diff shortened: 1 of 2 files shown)


philc:
The use case makes sense, but I think it's too niche to introduce a new command for it.

Note also that there is no similar command in Chrome's UI that I'm aware of -- one must copy the current URL, open a new incognito tab/window, and paste, as you say -- which I think is reasonable given the infrequency of this action.

PR #3994, opened 2022-01-24
Add command to show link URL without opening it

ehdr:
## Description
You may want to see the URL of a link before you open it with `f`/`F`, e.g. to make sure it's not malicious. This change introduces a new `LinkHints.activateModeToShowLinkUrl` mode, bound to `sf` by default, that lets you choose a link on the page, and will show the corresponding URL in the HUD.

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -48,6 +48,20 @@ const COPY_LINK_URL = {
     }
   }
 };
+const SHOW_LINK_URL = {
+  name: "link",
+  indicator: "Show link URL",
+  linkActivator(link) {
+    if (link.href != null) {
+      let url = link.href;
+      if (url.slice(0, 7) === "mailto:") { url = url.slice(7); }
+      if (28 < url.length) { url = url.slice(0, 23) + "...."; }
+      HUD.showForDuration(`Link URL: ${url}`, 5000);
+    } else {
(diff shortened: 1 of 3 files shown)


content_scripts/link_hints.js
```diff
@@ -48,6 +48,20 @@ const COPY_LINK_URL = {
...
   }
 };
+const SHOW_LINK_URL = {
+  name: "link",
+  indicator: "Show link URL",
+  linkActivator(link) {
+    if (link.href != null) {
+      let url = link.href;
+      if (url.slice(0, 7) === "mailto:") { url = url.slice(7); }
+      if (28 < url.length) { url = url.slice(0, 23) + "...."; }
```
ehdr, line 58: It's unfortunate that it seems the HUD can't display long messages (see screenshot), and thus must truncate long URLs, which partly defeats the purpose of this feature (which is to review the link URL before you follow it). Is there a way around this?

image

sfmqrb:
Why this is not merged yet?
useful feature

LGTM

content_scripts/link_hints.js
Malnasser: Neovim typically handles this with a double keybind pattern - press Ctrl+k to open the popup, then Ctrl+k again to focus inside it and enable scrolling. Could we improve this PR to implement similar functionality?

Issue #3989, opened 2022-01-20
Firefox open link in new tab opens two duplicate tabs


wchao:
**Describe the bug**
When I press F to open link in new tab in Firefox, Vimium opens two of the same tab. This only happens in Firefox. I also use Chrome, and Vimium on Chrome does not exhibit the same behavior.

**To Reproduce**
1. While viewing a web page, press F and choose a link.
2. You'll see that two of the same tab will open in Firefox.

**Browser and Vimium version**

{
  "application": {
    "name": "Firefox",
    "osVersion": "Windows_NT 10.0 22000",
    "version": "96.0.1",
    "buildID": "20220113185450",
    "distributionID": "",
    "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:96.0) Gecko/20100101 Firefox/96.0",
    "safeMode": false,
    "updateChannel": "release",
    "supportURL": "https://support.mozilla.org/1/firefox/96.0.1/WINNT/en-US/",
    "osTheme": "",
    "numTotalWindows": 1,
    "numFissionWindows": 1,
    "numRemoteWindows": 1,
    "launcherProcessState": 0,
    "fissionAutoStart": true,
    "fissionDecisionStatus": "enabledByRollout",
    "remoteAutoStart": true,
    "policiesStatus": 0,
    "keyLocationServiceGoogleFound": true,
    "keySafebrowsingGoogleFound": true,
    "keyMozillaFound": true
  },
  "securitySoftware": {
    "registeredAntiVirus": "Microsoft Defender Antivirus",
    "registeredAntiSpyware": "",
    "registeredFirewall": "Microsoft Firewall"
  },
  "processes": {
    "remoteTypes": {
      "privilegedabout": 1,
      "extension": 1,
      "webIsolated": 9,
      "prealloc": 3,
      "gpu": 1
    },
    "maxWebContentProcesses": 8
  },
  "environmentVariables": {
    "MOZ_CRASHREPORTER_DATA_DIRECTORY": "C:\\Users\\wchao\\AppData\\Roaming\\Mozilla\\Firefox\\Crash Reports",
    "MOZ_CRASHREPORTER_EVENTS_DIRECTORY": "C:\\Users\\wchao\\AppData\\Roaming\\Mozilla\\Firefox\\Crash Reports\\events",
[...]

PR #3985, opened 2022-01-13
not prevent click events since Firefox 96

gdh1995:
It seems Firefox 96 allows extensions to trigger native click behaviors.
See https://github.com/philc/vimium/issues/3964 for reports.

Although this is not listed in https://www.mozilla.org/en-US/firefox/96.0/releasenotes/,
I think it's safe enough to skip preventing since Firefox 96.

This will fix #3964, fix #3979, fix #3986 and fix https://github.com/philc/vimium/issues/3579#issuecomment-981237048

Issue #3964, opened 2021-11-30
[Bug] using F opens links twice on Firefox Nightly


s-marios:
Hi,

When I'm using shift + f to open a link in a new tab, instead of one tab two tabs with the same content appear. 

I'm using Firefox Nightly and this started happening 1-2 days ago. Here's at least one more user experiencing the same problem https://github.com/philc/vimium/issues/3579#issuecomment-981237048 .  

Anybody else affected?

Artturin:
im on nightly too

Artturin:
https://github.com/gdh1995/vimium-c does not have this problem though a weird thing is that when i disable vimium-ff and then install vimium-c, the vimium-c will have the same problem but it is fixed after i restart firefox

gdh1995:
@Artturin Vimium C uses a trick which is not safe on a loaded page, so it only enables this trick if it runs when a page begins loading. If you want to test Vimium C again, please refresh tabs after you enable it and before tests.

evenlee:
I am using firefox 96.0b2 dev edition, have same problem

JoshMock:
Also experiencing this on 96.0b2 developer edition.

sandeshghanta:
Also experiencing this issue in Firefox developer edition 96.0b4 version.

somers-all-the-time:
Experiencing the issue on Firefox Nightly 97.0a1

rvighne:
This has also been happening in beta for a long time. Is a fix possible / planned as of now? FF 96 will hit the release channel very soon on 1/11 which will affect a lot more users.

QHayward:
I installed Firefox today on a new computer, and can confirm that I am having this issue

keepiru:
I'm experiencing this after upgrading to FF 96.0 today.

mcauley-penney:
Just upgraded to 96.0 and also experiencing

Drk01:
Same error here, i just updated to FF 96. Only happens in Firefox

philc:
I haven't looked at this, but it's clearly important and manifesting for many users. PRs welcome!

gdh1995:
@philc I've filed #3985 . And I think #3867 is also worth merging.

R2robot:
Happening on the general public releases now.  96.0.1

cupcakearmy:
Same issue here on 96.0.1

sarrietav-dev:
Same issue here on 96.0.1. I hope this gets fixed soon.

[...]

PR #4000, opened 2022-01-30
fix opening links twice when `privacy.resistFingerprinting` is enabled or on Firefox 91.6+

### LINKED ISSUE #4021: Open a link with uppercase caused two same tabs opened
**Describe the bug**

1. Press `f` to generate hints for all links visible

image

2. Press uppercase letter like `L` in above picture to open link in new tab. However, it opens the same tab twice

image

- Firefox version: 91.6.0esr (64-bit)
- Vimium version: 1.67.1
- OS: Arch Linux x86_64
- Kernel: 5.16.11-arch1-1

### LINKED ISSUE #4032: Using F to open a link in a new tab does so twice
### Issue:

- The problem: Using F (shift+f) to open any link in a new tab opens two new tabs each loaded with the specified link.
- Expected behavior: Doing the described action should result in just one new tab opened with the specified link loaded.

### Technical specifications:

- OS: Windows 10 64-bit
- Browser: Firefox 97.0.2 (64-bit) (privacy-oriented hardened about:config)
- Vimium: Vimium-FF 1.67.1 from the Firefox addons store

gdh1995:
As said in https://github.com/philc/vimium/issues/3989#issuecomment-1019377405, Firefox will report a version of "91.0" in `navigator.userAgent`, if only `privacy.resistFingerprinting` on `about:config` is enabled.

This bug is reported in https://github.com/philc/vimium/issues/3989#issuecomment-1019370376 and https://github.com/philc/vimium/issues/3964#issuecomment-1024960014.

And https://github.com/philc/vimium/pull/3985#issuecomment-1043402172 reported Firefox 91.6.0 ESR has updated its blocking policy and then Vimium opens 2 tabs on it again. The latest commit adds support for this special case.

This also fixes #4021 and fixes #4032.

### DIFF
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -161,7 +161,8 @@ const TabOperations = {
       delete tabConfig["url"];
 
     // Firefox <57 throws an error when openerTabId is used (issue 1238314).
-    const canUseOpenerTabId = !(Utils.isFirefox() && (Utils.compareVersions(Utils.firefoxVersion(), "57") < 0));
+    const canUseOpenerTabId = !Utils.isFirefox() || Utils.firefoxVersion() instanceof Promise
+        || (Utils.compareVersions(Utils.firefoxVersion(), "57") >= 0);
     if (canUseOpenerTabId)
       tabConfig.openerTabId = request.tab.id;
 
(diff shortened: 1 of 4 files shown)


Drk01:
Sad this isn't already merge :cry:

woky:
I've just tested this, and it fixes the problem with opening new tab twice in current Firefox ESR (`91.7.0esr`). Thank you!

There's still one minor difference when `privacy.resistFingerprinting` is enabled compared to when it's disabled (in Firefox ESR). When it's disabled, you can press Shift when either when opening hits, that is Shift-F, your you can press Shift at last letter, i.e. when link letters are `SSA`, the following two key press sequences open the link in a new tab:
1. `Shift+F`, `S`, `S`, `A`
2. `F`, `S`, `S`, `Shift+A`
When `privacy.resistFingerprinting` is enabled, only the first one works though. I'm not sure if the 2nd case is supposed to work or is just a side effect, but I got quite used to it. It's no big deal though.

gdh1995:
@woky It's another impact of `resistFingerprinting` - it blocks `Ctrl/Shift/Alt` key events and then LinkHints's won't learn you have pressed `Shift+S` or `Shift+A`. In Normal mode, Vimium uses another way to check `Shift`, so `Shift+F` still works.

My extension, Vimium C (https://github.com/gdh1995/vimium-c) is also affected, and I've implemented a working around for this issue in the master branch (though not released yet). You may take a try or just wait for a few weeks to try Vimium C v1.97+ .

I have no plans to file new PRs to add the working around into Vimium, because: 1) there're no other people reporting this, so this bug seems insignificant; 2) I suspect there're still other impacts of `resistFingerprinting`

philc:
Tricky and complex, unfortunately =( Thanks for adding this @gdh1995

gdh1995:
Hello, as I've updated in https://github.com/philc/vimium/pull/3985#issue-1101757110 , I find a new failure on Firefox 91.6+ ESR - if only `Settings -> Privacy & Security -> Block pop-up windows` is enabled, "a plain click on `<a target=_blank href=...>`" will still be blocked; while Firefox 96+ doesn't have such a problem.

However, Firefox 102.2 ESR is arriving in the end of this August, and then Mozilla's support for Firefox 91 ESR will end. So we may just ignore this minor issue.

PR #4005, opened 2022-02-05
Fixed a minor handler-id assignment bug

gurjeet:
Before the fix, the anonymous name, generated for assigning to an
otherwise unnamed handler, had a different value in the name than the id
assigned to the handler.

This bugfix makes sure that the id used in the name of an anonymous
handler matches the id assigned to the handler.


philc:
@gurjeet I think this logic fix makes sense. Does the existing bug have any impact in practice?

gdh1995:
Seems no impact on UX behavior - `_name` is only used by `console.log` in `HandlerStack`'s `logResult` and `show`.

https://github.com/philc/vimium/blob/7a9fc8d631d7ac5abc1a377194d6a46796c27898/lib/handler_stack.js#L127-L148

philc:
I couldn't merge this cleanly through Github's UI for some reason, so I ported these changes over in 4c0b91fd30960a9ef47035882e81d8eabaff63c8. Thanks @gurjeet

Issue #3469, opened 2020-01-14
Incremental search and highlights


Vozf:
Are there any plans to adopt commands for incremental search and highlighting such as
```
set incsearch
set hlsearch
```
?

gdh1995:
Browsers don't allow extensions to call their inner functionality of highlighting all matched text, while it's too complex and difficult to simulate such a highlighting style by Vimium itself.

Vozf:
There is an extension which works in a similar way.
If it helps - https://chrome.google.com/webstore/detail/selection-highlighter/nepmkgohgoagfgcoegjaggacodcpdibj

gdh1995:
The extension is hosted on https://github.com/neaumusic/selection-highlighter, and it creates new DOM elements to add styles to matched text. This behavior may break some pages - this may be why it has a `blacklistedHosts`.

Well, this issue is somehow a duplicate of #591, and there seems no perfect API to do so.

Vozf:
Can it be done as an optional feature with the same blacklist as vimium already has?

Xiao6:
Add usecase why this is helpful: I need to view log file from logging apps, the log is usually large. Chrome default search can highlight all search appearances, and also has outline in the right scroll bar, it makes is easy to navigate the log. I don't quite use vimium search function because of lacking this feature.

sepsol:
I absolutely love this extension. I think it'd be helpful if the other matched searched results were kept highlighted even when they're not the activated matched result. I hope this finds its way into a future update.

Issue #3100, opened 2018-08-22
Anyway to highligh the search in find mode?


dilipm:

smblott-github:
If you mean highlighting *all* matches (instead of just the current match), then there's no API for that.

ryankhart:
> If you mean highlighting _all_ matches (instead of just the current match), then there's no API for that.

Could you clarify what you meant by "no API for that?" Does that mean that it's impossible to write a Chrome extension for that without the risk of a Chrome update breaking the extension? Or does that mean that Vimium itself doesn't have an API for it yet, but theoretically, someone could fork the project, develop that feature, and submit a pull request?

Issue #591, opened 2012-07-18
In find mode, highlight all matches on the page, not just the current


livelazily:
In chrome find mode,it can highlight all match word in the page
Can vimium also do this?

koolkalang:
Recently found out about vimium, can't believe I haven't installed this sooner.

+1

emperornorton:
+1

ihaku4:
+1

Matrinica:
+1

TripleFire:
+1
and I think this would be another killer feature

spillevink:
:+1: This would be great

ThePinkPanther:
+1

wbolster:
please STOP replying with useless +1 comments. it does not add anything to the discussion, and will waste a lot of people's time since they will get notifications of your useless "contribution". use the "watch" function to subscribe to a ticket.

spillevink:
+1

JamesValero:
+1

dragonxlwang:
+1, the default chrome behaves highlighting all matched hit. If implementing it again would be a performance issue, is it possible to directly call chrome's search?

mrmr1993:
> is it possible to directly call chrome's search?

No.

To implement something like this would want something like (an updated) #1081 so the browser doesn't hang on every search, which increases the development/maintainance burden substantially.

smblott-github:
Closing.
Popular as it is, we don't really have a way to implement this.

(Vimium doesn't actually do the highlighting at all, currently.  We just call `window.find()`.)

yech1990:
4 years...

ResolverJay:
+1 This is really a great feature if it can be implemented.

I did some research. This seems to do it http://james.padolsey.com/javascript/replacing-text-in-the-dom-solved/

But I think there's a further question - when and how to remove the highlight. For when, I think when starting the next search or when executing something like `:noh`; for how, I assume `execCommand('undo')` would do it.

Really look forward to the feature.

fohrums:
This is needed and not sure if it's suppose to be a vimium-specific feature, but other vim-like extensions do this (*like surfingkeys for example*). It's already been nearly 5 years and this feature isn't here yet. **What in the world is going on?**

mrmr1993:
> I did some research. This seems to do it http://stackoverflow.com/a/5887719/96100 (the solution includes IE, so it can be further simplified by removing the IE part)

[...]

PR #4010, opened 2022-02-15
Real-time highlighting of the search results

yousfiSaad:
## Description
This is an implementation of the Incremental search and highlights features.
Issues talking about this feature are : https://github.com/philc/vimium/issues/3469 , https://github.com/philc/vimium/issues/591 , https://github.com/philc/vimium/issues/3100

### DIFF
--- a/content_scripts/mode_find.js
+++ b/content_scripts/mode_find.js
@@ -39,6 +39,98 @@ var newPostFindMode = function() {
   return new PostFindMode();
 }
 
+// This class manages the highlighting
+// Logic of this feature is from the extension https://github.com/rogershen/chrome-regex-search
+class Highlightings {
+  //TODO: Understand the oo aspect in the project
+  //TODO: Add tests for this new feature
+  /*** CONSTANTS USED FOR HIGHLIGHTING ***/
+  static ELEMENT_NODE_TYPE = 1;
+  static DEFAULT_HIGHLIGHT_COLOR = '#FFFF00';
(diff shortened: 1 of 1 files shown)


gdh1995:
As said in https://github.com/philc/vimium/issues/3469#issuecomment-577964931, this PR creates new DOM elements to add styles to matched text, but this behavior may break some pages .

I suggest that we may use `element.getBoundingClientRect()` and create many `<div style="left:...;top:...;width:...;height:...">` to show bounding boxes (without background color), like what `Ctrl+Shift+J/K` do in my Vimium C (https://github.com/gdh1995/vimium-c).

chrisgrieser:
I'd love to have the feature merged into vimium 🙏

MathiasSven:
As others have expressed, even if the implementation as it currently is might break in some pages, the benefits of the feature might outweigh the potential issues, what would be the harm in having this as an optional experimental feature under the advanced options with a disclamer?

ivanwfr:
Comments from anyone interested to check my version are welcome ● [Download ZIP](https://github.com/ivanwfr/vimium/archive/refs/heads/master.zip)
Despite being a user of vi ... vim ... gvim since 1982, I discovered vimium only a few weeks ago!
I am still currently exploring the code, experimenting with some feature of my own.
When that works, it means I got it right, if not, then I learn something...
I am willing to work on any kind of improvements and suggestions you may have...

PR #4024, opened 2022-03-11
Filter suggestions if certain symbols are in the query terms.

ewollesen:
This implements a few filters based on the article here:
https://support.mozilla.org/en-US/kb/address-bar-autocomplete-firefox#w_changing-results-on-the-fly

The article mentions filtering based on bookmark tags, and that's not
possible due to them not being exposed via the API at this time.

## Description

### DIFF
--- a/background_scripts/completion.js
+++ b/background_scripts/completion.js
@@ -33,7 +33,8 @@ class Suggestion {
     this.insertText = null;
     // @deDuplicate controls whether this suggestion is a candidate for deduplication.
     this.deDuplicate = true;
-
+    // @filterFunction can remove the suggestion based on the query terms.
+    this.filterFunction = null;
     // Other options set by individual completers include:
     // - tabId (TabCompleter)
     // - isSearchSuggestion, customSearchMode (SearchEngineCompleter)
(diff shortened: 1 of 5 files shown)


gdh1995:
I think there should be a way to disable such filtering - or disable it by default unless an option is added/enabled.

ewollesen:
> I think there should be a way to disable such filtering - or disable it by default unless an option is added/enabled.

That makes good sense. I'll see if I can have a toggle. Do you think it should default to on, or off?

ewollesen:
I've added a checkbox to toggle the feature on/off (default: off). I considered adding more details to the help text, but it wasn't immediately obvious how the captions performed their vertical alignment, and so things overran each other and looked bad.

philc:
I didn't know about these filters in Firefox until now, and this is a high-quality implementation. Nice work fleshing out this idea @ewollesen.

I think the use case for this is too niche for mainline Vimium: this is only a known-convention in Firefox, and probably not that well known. Also, the main use cases they provide are possible with existing vimium shortcuts (scope search to open tabs by hitting t; scope to bookmarks by hitting b), albeit not as efficiently as your implementation.

It's also a substantial amount of (well-tested) code to maintain going forward.

PR #4028, opened 2022-03-18
Adds amazon smile search

ewollesen:
## Description

Just a tweak to the amazon search regular expression to match smile.amazon.com.


philc:
Thanks!

PR #4038, opened 2022-03-31
Link hints for role="radio"

ember-rose:
## Description
Add support for the ARIA `radio` role in link hints detection. This provides parity with `<input type="radio" />` radio buttons, which currently receive link hints.


philc:
Thanks!

PR #4039, opened 2022-04-01
Fix the messed up coordinates in getLocalHints function

janis-github:
## Description

Bug manifested itself as not being able to detect input with label located some distance from the input


philc:
I don't have a clear repro case, but it sure looks like a transposition typo, introduced in #2251. Thank you for the PR.

PR #4108, opened 2022-08-13
Fix scroll in extension popup on Firefox

sigseg5:
Fixed annoying unnecessary scroll in extension popup

https://user-images.githubusercontent.com/36568961/184510511-f3eb0ee5-4a75-4314-9cf7-e660afbec03b.mp4

### DIFF
--- a/pages/options.css
+++ b/pages/options.css
@@ -4,6 +4,7 @@ body {
   font: 14px "DejaVu Sans", "Arial", sans-serif;
   color: #303942;
   margin: 0 auto;
+  padding: 0 6px;
 }
 a, a:visited { color: #15c; }
 a:active { color: #052577; }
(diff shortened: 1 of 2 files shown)


philc:
Nice video; thanks for posting. That is annoying. I don't see this on my platform (OSX, Chrome). What is your platform?

The way this was fixed (using a negative margin) is a bit unintuitive and makes the buttons no longer aligned with anything -- previously they were aligned with the right edge of the dialog. While more work, a more robust solution is to transition this to using a flexbox layout.

Before:

image

After:

image

gdh1995:
This issue is caused by `#state { padding-left: 5px; width:  100%; }` in `pages/popup.html`:
https://github.com/philc/vimium/blob/5023057b052387a7b420bfec5af0da71053c06d0/pages/popup.html#L24-L27

And during my test on Firefox 103/105 on Win11, it works if just add `box-sizing: border-box;` to `#state`:
image

philc:
Thanks @sigseg5 for taking a look. I've rewritten this exclusions UI in v2.0, so this should no longer be an issue.

PR #4118, opened 2022-09-07
Chrome: fix state icon

alanoliveira:
## Description
This PR fixes extension icon on Google Chrome to use **ENABLED_ICON** if there is no partial rule for the current page.


philc:
Yes, this fixes an error I introduced in ffdd38f62054d82d893cbbdc6401def3c3864d85. The icon has been bothering me ever since! Thank you @alanoliveira .

PR #4166, opened 2022-11-07
avoid triggering popup blocker on some old versions of firefox 

### LINKED ISSUE #4165: Firefox can't open links in new windows
**Describe the bug**
When I press `F` to get hints that open in a new tab, they no longer open (from a couple of weeks ago, but the 1.67.3 release didn't fix this when it fixed other stuff). Typing `f` for hints and entering the hints in uppercase to get them in a new window also does nothing. Does seem to still work in Chrome though.

**To Reproduce**
0. Use Firefox with Vimium installed
1. Go to https://vimium.github.io/
2. Press `F`. Choose any link, this should open in a new browser tab.

Firefox: 94.0
OS: Linux 5.14.18-100.fc33.x86_64 #1 SMP Fri Nov 12 17:38:44 UTC 2021
Vimium: 1.67.3

gdh1995:
The old code in #4000 ignores Firefox versions which are older than 96 while don't belong to 91. So this PR adds it.

Should fix #4165 .

### DIFF
--- a/lib/dom_utils.js
+++ b/lib/dom_utils.js
@@ -320,16 +320,25 @@ var DomUtils = {
     const eventSequence = ["mouseover", "mousedown", "mouseup", "click"];
     const result = [];
     for (let event of eventSequence) {
-      // In firefox prior to 96, simulating a click on an element would be blocked. In 96+,
-      // extensions can trigger native click listeners on elements. See #3985.
-      const mayBeBlocked = event === "click" && Utils.isFirefox() && parseInt(Utils.firefoxVersion()) < 96
-          && /^91\.[0-5](\.|$)/.test(Utils.firefoxVersion())
+      // In firefox prior to 96, simulating a click on an element would be blocked under some conditions (#2602, #2960).
+      // In 96+, extensions can trigger native click listeners on elements (#3985).
+      // While 91.6.0esr also merged the patch, so it's necessary to check the minor version (#4000).
+      // The root difference is a logic change imported in https://bugzilla.mozilla.org/show_bug.cgi?id=1739929 .
+      let mayBeBlocked = false
+      if (event === "click" && Utils.isFirefox()) {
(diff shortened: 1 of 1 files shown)


tim-goto:
To circumvent the issue that is being fixed by this PR I installed the latest version of the plugin locally and it does not have this issue besides the fact that this PR is not yet merged. Maybe I am running into something differently though..

lib/dom_utils.js
```diff
@@ -323,7 +323,7 @@ var DomUtils = {
       // In firefox prior to 96, simulating a click on an element would be blocked. In 96+,
       // extensions can trigger native click listeners on elements. See #3985.
       const mayBeBlocked = event === "click" && Utils.isFirefox() && parseInt(Utils.firefoxVersion()) < 96
-          && /^91\.[0-5](\.|$)/.test(Utils.firefoxVersion())
+          && /^(?!91\.)|^91\.[0-5](\.|$)/.test(Utils.firefoxVersion())
```
philc, line 326: This regexp is now pretty complicated to look at, and the comment doesn't indicate why we're checking for Firefox 91.

I think this would be much easier to work with if 1) we added firefoxMajorVersion and firefoxMinorVersion to Utils and used those for comparisons, so we don't have to use parseInt on a string and regexps to parse those out here, and 2) updated the comment to describe why we're checking against Firefox 91 in this section.

gdh1995: Hello Philc, I've updated the code to make it clear.

gdh1995:
And as I've updated in https://github.com/philc/vimium/pull/3985#issue-1101757110 , Firefox 91.6+ still reports "popup-blocked" on a plain `click` event on `<a target=_blank>`, so I also add a check in this commit to fix it.

Just now I tested the commit cb78fa5 on Firefox 90, 91.0esr, 91.6esr and 96, and it works perfectly whether Firefox's `Block pop-up windows` (aka. `dom.disable_open_during_load`) is enabled or not.

philc:
Thanks @gdh1995

PR #4175, opened 2022-11-18
Fix shebang line error during build

smallming:
Fix error running `./make.js`.
## Description
Fix error `/usr/bin/env: ‘deno run --allow-read --allow-write --allow-env --allow-net --allow-run --unstable’: No such file or directory` when building via `./make.js package`.


philc:
What OS are you using? The build works on my machine, but this seems like a safe change to make.

smallming:
I'm using ubuntu 22.04, with deno 1.28.1. Was following the example from https://deno.land/manual@v1.28.2/examples/hashbang#example.

PR #4178, opened 2022-11-20
Fix text overflowing button container

mcp292:
image

image

I am not set up to test this change, so I don't know if the width increase is sufficient.

Notice in the image above that there is no 's' in "Show Advanced Options". Despite this, [the code in question](https://github.com/philc/vimium/blob/master/pages/options.js#L282-L291) seems correct. 

Although "show" and "hide" are both 4 characters, they are not equally spaced; therefore, I determine width to be the cause of this too. The proposed fix should fix both issues.

### DIFF
--- a/pages/options.css
+++ b/pages/options.css
@@ -137,7 +137,7 @@ input#searchUrl {
 }
 #buttonsPanel { width: 100%; }
 #advancedOptions { display: none; }
-#advancedOptionsButton { width: 170px; }
+#advancedOptionsButton { width: 200px; }
 .help {
   position: absolute;
   right: -320px;


philc:
What browser and OS are you using? The button has ample room for me.

image
image

mcp292:
Odd.

Firefox 107.0 (64-bit)
Arch Linux

gdh1995:
Usually it's because the system fonts of you two are different, or Firefox has a too large "min font size".

philc:
Thanks for looking into this @mcp292. This part of the settings Ui has been removed in v2.0 -- the settings page now shows all options, including advanced, so we don't have to worry about this UI sizing issue.

Issue #4173, opened 2022-11-17
yy shortcut is not working


Huray-hub:
The yy shortcut shows the popup "yanked <url>" but when I paste there is nothing on the clipboard

1. go to any URL
2. press yy
3. paste wherever you want

Vimium: 1.67.2

Browser: 
Brave 1.45.127, Chromium: 107.0.5304.110 (Official Build) (64-bit)

OS: 
1. Arch Linux
2. Windows 11 Version 22H2 (Build 22621.819)

jwilkeson:
Related to https://github.com/philc/vimium/issues/4149

mb720:
The issue I'm seeing is that although yanking the current URL works for HTTPS sites, it doesn't for HTTP (non-encrypted) sites. The debugger shows this after pressing `yy` on an HTTP site:

    Uncaught (in promise) TypeError: navigator.clipboard is undefined
        copy moz-extension://3c472571-c4dd-455e-81b3-f282ddd0880b/lib/clipboard.js:18
        copyToClipboard moz-extension://3c472571-c4dd-455e-81b3-f282ddd0880b/pages/hud.js:147
        setTimeout handler*setTimeout moz-extension://3c472571-c4dd-455e-81b3-f282ddd0880b/lib/utils.js:278
        copyToClipboard moz-extension://3c472571-c4dd-455e-81b3-f282ddd0880b/pages/hud.js:145
        <anonymous> moz-extension://3c472571-c4dd-455e-81b3-f282ddd0880b/pages/hud.js:174
        onmessage moz-extension://3c472571-c4dd-455e-81b3-f282ddd0880b/pages/ui_component_server.js:23
        portOpen moz-extension://3c472571-c4dd-455e-81b3-f282ddd0880b/pages/ui_component_server.js:21
        registerPort moz-extension://3c472571-c4dd-455e-81b3-f282ddd0880b/pages/ui_component_server.js:9

I did not encounter the symptoms described [here](https://github.com/philc/vimium/issues/4149): blurring and losing focus.

Vimium 1.67.3, Mozilla Firefox 107.0, Arch Linux 6.0.8.

Thanks for Vimium, it's awesome!

gdh1995:
`navigator.clipboard is undefined` has been reproduced on Firefox 106 for Windows x64, but not on Chrome 107.

So, I think it's a bug of Firefox - Chrome or Firefox tends to apply too strict limits on web-extensions unless too many developers "dare to" leave them.

[...]

Issue #4170, opened 2022-11-10
After "yf" command, no other commands work


ericauld:
After yanking a url, vimium will stop functioning on that page.

To reproduce
1. Go to any page with links, press "yf", and tap in a link code.
2. Press any vimium command ("j", for example) and note that it doesn't work.
3. Note that vimium **is** still working on other tabs which are open in the same window.

Chrome version 107.0.5304.110 (Official Build) (arm64)
Vimium version 1.67.2

MacBook Pro, 13-inch, M1, late 2020

kenny-evitt:
I've noticed `yf` not working but didn't consider the issue you describe being the culprit.

Interestingly, for a page in a tab (also in Chrome) for which `yf` isn't working, the command continued to NOT work after reloading the page or duplicating the tab (and trying the command in the new duplicated tab).

I *was* able to get the command to work for the same page by copying the page URL to the clipboard (without using Vimium) and then opening that URL in a new tab.

(Maybe) related:

* [yf not working · Issue #4158 · philc/vimium](https://github.com/philc/vimium/issues/4158)
* [yank url and open clipboard url commands not working · Issue #4153 · philc/vimium](https://github.com/philc/vimium/issues/4153)
* [Copying got broken in Visual mode · Issue #4147 · philc/vimium](https://github.com/philc/vimium/issues/4147) (the comments cover some issues with 'yank' commands not specific to visual mode)

aucker:
got the same issue on windows 11
chrome version: Version 107.0.5304.107 (Official Build) (64-bit)
vimium version: 1.67.2

philc:
Checking in again -- this doesn't repro for me on the latest version of Vimium (v1.67.6) in Chrome on Mac. Does this repro for anyone else?

b1az:
Still repros on 1.67.6, installed from source. The error as seen in `brave://extensions/?errors=<an_id>`:

```
Uncaught (in promise) SecurityError: Must be handling a user gesture to use custom clipboard

Stacktrace: 
pages/hud.js:152 (anonymous function)
```
![Untitled](https://user-images.githubusercontent.com/9404656/211134125-78312bf0-b7c8-4849-8a89-4d327f65ecd2.png)

Clipboard setting (`brave://settings/content/clipboard`) is set to `"Sites can ask to see text and images on your clipboard"`. The issue is present with either 

A) no per-site exceptions, or
[...]

PR #4185, opened 2022-12-04
Add clipboardWrite back to manifest.json

WilliamWelsh:
## Description
On Chrome and FireFox, `yy` and `yf` do not work. Adding `clipboardWrite` back to the `manifest.json` solves this. The current version doesn’t work for me on either Chrome or FireFox dev edition, after updating the manifest and installing on Chrome/FireFox, `yy` and `yf` work. Solves #4170 and #4173.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -34,6 +34,7 @@
     "sessions",
     "notifications",
     "webNavigation",
+    "clipboardWrite",
     "<all_urls>"
   ],
   "content_scripts": [


AetherUnbound:
I did not realize how much I use/depend on this feature regularly until this broke 😅 hopefully this can be merged soon!

philc:
This permission needs to be conditionally added for Firefox -- Chrome doesn't need it. Closing this in favor of #4186.

PR #4189, opened 2022-12-07
Added StartPage completion

matthewblott:
StartPage is a popular search engine which should be included.

### DIFF
--- a/background_scripts/completion_engines.js
+++ b/background_scripts/completion_engines.js
@@ -228,6 +228,23 @@ class UpToDate extends BaseEngine {
   parse(xhr) { return JSON.parse(xhr.responseText).data.searchTerms; }
 }
 
+class StartPage extends BaseEngine {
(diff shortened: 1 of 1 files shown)


philc:
Hey @matthewblott,

Completions got reworked a bit, and the parse function needs to be updated to this:

```
parse(text) {
  return Array.from(JSON.parse(text)).map((suggestion) => suggestion.phrase);
}
```

More importantly, I couldn't get the suggestions API to work with the URL in this PR. It seems the website adds a random token (sc) to the query string, and that is required to get results. E.g.:

https://www.startpage.com/suggestions?q=hey&sc=mErAnqplBxxn00&lui=english

I'm not sure if that's a session token or something. It seems to me that their suggestions API isn't stable, or it isn't meant to be used programmatically like this.

philc:
Closing due to inactivity.

PR #4238, opened 2023-03-28
Add RuneScape Wiki and OldSchool RuneScape Wiki

Ivy-Panda:
Search strings copied from https://runescape.wiki/w/RuneScape:Finding_the_wikis_with_ease

### DIFF
--- a/background_scripts/completion_engines.js
+++ b/background_scripts/completion_engines.js
@@ -228,6 +228,36 @@ class UpToDate extends BaseEngine {
   parse(xhr) { return JSON.parse(xhr.responseText).data.searchTerms; }
 }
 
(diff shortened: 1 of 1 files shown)


philc:
Thanks @Ivy-Panda for the submission!

I think this is too niche to live in the main repo -- I don't use the Runescape wiki and don't want to maintain the parsing code).

I haven't figured out a way to make search engine completions easily extensible with user-code for more niche search engines.

rileywilddog:
`JSON.parse(xhr.responseText)[1]` doesn't seem like a particularly onerous maintenance burden, and this seems like the search suggestion response format basically everything (including [google](https://www.google.com/complete/search?client=opera&q=lotus)) supports now...

Issue #4133, opened 2022-10-01
Copy link URL and text as Markdown hyperlink


jbriales:
Currently there exist features for copying a link URL (`LinkHints.activateModeToCopyLinkUrl`) OR text (`LinkHints.activateMode action=copy-text`), but AFAIK no alternative to copy *both* URL and text in a common format as e.g. a Markdown link (i.e. an output like `[Link text Here](https://link-url-here.org)`).

Would this feature be doable and welcome? Any thoughts or concerns before implementing it?

gdh1995:
You may try my customized version of Vimium, named Vimium C (https://github.com/gdh1995/vimium-c), and it support a parameter of `format`:
```
map xxx LinkHints.activateCopyLinkText access="[${innerText}](${href})"
```

Issue #2474, opened 2017-04-14
Feature Request:Copy link title


JonahWong:
Like 'yf' key to copy link address, adding shortcut to copy link title would be a nice touch.

sherwinWatson:
I would love to see that implemented too

SimonAlling:
Please make this happen!

smblott-github:
One way to do this might be:

    set yt copyCurrentUrl title

And possibly, for symmetry:

    set yt copyCurrentUrl url

lydell:
Just noting that in VimFx, you'd do this with `yv`. It let you copy the text of _any_ element, not just links.

mrmr1993:
@smblott-github won't that (confusingly) make bindings for both commands show up as bindings for a single command in the help dialog?

(Also, I think you may have meant `LinkHints.activateModeToCopyLinkUrl`, not `copyCurrentUrl`.)

knobhead:
As one of the VimFx users forced to migrate in the wake of FF57, I'd also love to see this feature. 

VimFx behaviour in `element_text_copy` mode is to also 'LinkHint' form fields (`<input>`, `<textarea`, even with their `disabled` attribute set...) and copy their content.

desigit:
This would be wonderful (and presumably quite quick) to implement - it's the only feature from VimFx I now miss, after yank URL features have been recently implemented on the Firefox Quantum version.

skating507:
really need  this  feature

edrex:
In qutebrowser `ym` copies the current URL/title as a markdown link, which is usually what I want.

knobhead:
Hi, any chance to get this feature into the development roadmap? Unfortunately I can't help with coding, but I'll be happy to help with testing or anything else within my reach. Also, I'd be happy to donate, as vimium-ff is my daily driver...

edrex:
@gdh1995 why are you forking instead of pushing improvements upstream?

gdh1995:
I have lots of ideas and there're too many ones not liked by Vimium's main developers. So currently most of my PRs to Vimium are only to fix bugs.

edrex:
There is a rejected patch that implements `ym`: https://github.com/philc/vimium/pull/2054

It seems like, with 69 open PRs and no commits in 6 months, this project isn't being actively maintained. Maybe the coffeescript legacy is dragging it down?

usfmohy:
I managed to overcome this by a custom keybinding:
[...]

Issue #889, opened 2013-07-25
Copy the page title to the clipboard


baskerville:
I'd like to be able to copy the page title to the clipboard.

vincentbernat:
@philc What better tools are available for that? This seems an useful feature for me.

smblott-github:
> What better tools are available for that?

If you're feeling really geeky, then perhaps [chromix](https://github.com/smblott-github/chromix). `chromix list` gives:

```
849 https://github.com/philc/vimium/issues/889 Copy the page title to the clipboard ? Issue #889 ? philc/vimium
```

> This seems an useful feature for me.

It might be useful, but I'm not sure it's sufficiently useful to warrant a new command.  An undiscoverable hack would be to have `2yy` yank the title.

mrmr1993:
@vincentbernat you could create a bookmarklet with the address `javascript:void(prompt("You may copy the page title below:", document.title));`, trigger that from the Vomnibar, and copy the title from the resulting prompt.

A resolution to #1392 would allow custom commands &mdash; so you, rather than the project maintainers, get the final say in which functionality you can have &mdash; but that's not a high priority issue at the moment.

vincentbernat:
@mrmr1993 Well, I would need to use the mouse. Even if you can put the text somewhere else, copy/pasting it would be rather less convenient than a two key shortcut.

@smblott-github Using Linux, I am currently using `xwininfo -root -children | grep Chromium`. I could do an alias. But this seems more natural to put just that in Vimium.

mrmr1993:
> I would need to use the mouse.

@vincentbernat  you shouldn't need to. If you create the bookmarklet and give it a name like `titleText`, you should be able to access it with `b titleText`, maybe tab a few times to select the bookmarklet, and then ctrl-c when the prompt comes up. It's not ideal, but it should work.

vincentbernat:
@mrmr1993 Oh, yes, sorry, I missed the comma before `document.title`. That seems a good workaround. Thanks!

ymyke:
@philc Here's my use case why I think something like `yp` or `ys` to yank the page title would be useful: I often send around links to web pages to friends or work colleagues. Most of the time, I want to send not only the link but also the page title, as it usually contains,e.g., the title of an article and helps recipients better understand what the link is about.

Today, I need to painstakingly select and copy an article's title manually.

Edge allows me to copy link+title in one, however a) other browsers don't and b) I'd like to have the two things separate.

PR #4240, opened 2023-03-30
Add yank (copy) markdown link of current tab

johanna-a:
This adds a new command, copyCurrentMarkdown, with a default keybinding `ym`, to copy a link to the current tab in markdown format, i.e. "[title](url)".

## Description
Firstly, this has been suggested and implemented before, see PR #2054.
However, that pull request was never merged.
Having a way to copy a link in markdown format is implemented in other projects for keyboard-based browsing, for example qutebrowser. It is typically something a user expects to be able to do easily.
It relates to issues #131 #889 #2474 #3611 #4133

### DIFF
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -197,6 +197,7 @@ const Commands = {
       "scrollToRight",
       "reload",
       "copyCurrentUrl",
+      "copyCurrentMarkdown",
       "openCopiedUrlInCurrentTab",
       "openCopiedUrlInNewTab",
       "goUp",
@@ -297,6 +298,7 @@ const defaultKeyMappings = {
   "u": "scrollPageUp",
   "r": "reload",
   "yy": "copyCurrentUrl",
+  "ym": "copyCurrentMarkdown",
   "p": "openCopiedUrlInCurrentTab",
   "P": "openCopiedUrlInNewTab",
   "gi": "focusInput",
@@ -386,6 +388,7 @@ const commandDescriptions = {
   toggleViewSource: ["View page source", { noRepeat: true }],
 
   copyCurrentUrl: ["Copy the current URL to the clipboard", { noRepeat: true }],
+  copyCurrentMarkdown: ["Copy the current title and URL as markdown link to the clipboard", { noRepeat: true }],
   openCopiedUrlInCurrentTab: ["Open the clipboard's URL in the current tab", { noRepeat: true }],
   openCopiedUrlInNewTab: ["Open the clipboard's URL in a new tab", { repeatLimit: 20 }],
 
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -633,6 +633,7 @@ var sendRequestHandlers = {
   // getCurrentTabUrl is used by the content scripts to get their full URL, because window.location cannot help
   // with Chrome-specific URLs like "view-source:http:..".
   getCurrentTabUrl({tab}) { return tab.url; },
+  getCurrentTabMarkdown({tab}) { return `[${tab.title}](${tab.url})` },
   openUrlInNewTab: mkRepeatCommand((request, callback) => TabOperations.openUrlInNewTab(request, callback)),
   openUrlInNewWindow(request) { return TabOperations.openUrlInNewWindow(request); },
   openUrlInIncognito(request) { return chrome.windows.create({incognito: true, url: Utils.convertToUrl(request.url)}); },
--- a/content_scripts/mode_normal.js
+++ b/content_scripts/mode_normal.js
@@ -119,6 +119,15 @@ var NormalModeCommands = {
     });
   },
 
+  copyCurrentMarkdown() {
+    chrome.runtime.sendMessage({ handler: "getCurrentTabMarkdown" }, function(url) {
+      HUD.copyToClipboard(url);
+      if (28 < url.length)
+        url = url.slice(0, 26) + "....";
+      HUD.showForDuration(`Yanked ${url}`, 2000);
+    });
+  },
+
   openCopiedUrlInNewTab(count) {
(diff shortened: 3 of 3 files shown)


ooloth:
I'd love to see this `ym` PR merged! It would be a big help for my note-taking workflow.

scotho3 (approved):
This looks fantastic! Nice simple addition.

philc:
Could everyone who's interested in this elaborate on your use cases?

1. Are you pasting this information into your markdown files, into Github, into social media sites?
2. Why just Markdown? Why not HTML -- or rich text, so the link can be copied into editors like Google Docs or Slack?
3. Do you ever want the title and not the URL? Many of the older bugs which were referenced were requesting a shortcut to copy the title of the page, not the combination of title and URL.
4. Why is this needed for the current page only, and not for link hints? Note that there is an existing command `yf` (`LinkHints.activateModeToCopyLinkUrl`)which copies a link URL to the clipboard. Should there be a similar command to copy the link+title to the clipboard?

I'm concerned about adding a feature which is too narrow.

Note that @gdh1995 [has an implementation](https://github.com/philc/vimium/issues/4133#issuecomment-1272559987) where the format of the copied text / URL can be customized, which allows it to be generic. If we went with a similar approach, then the command should probably be called `copyCurrentUrlAndTitle` (even if the default formatter is Markdown), rather than `copyCurrentMarkdown`.

ooloth:
> Could everyone who's interested in this elaborate on your use cases?

1. Pasting into markdown notes (Obsidian)
2. Additional powers sound fun (though would be used rarely)
3. No
4. Copying markdown links for link hints instead of having to visit those pages first would be great

boesi:
> Could everyone who's interested in this elaborate on your use cases?
> 
>     1. Are you pasting this information into your markdown files, into Github, into social media sites?
>     2. Why just Markdown? Why not HTML -- or rich text, so the link can be copied into editors like Google Docs or Slack?

Right now I use a simple Word file for notes - so HTML would be great. But for my workflow, the clipboard needs to have the content type text/html.

>     3. Do you ever want the title and not the URL? Many of the older bugs which were referenced were requesting a shortcut to copy the title of the page, not the combination of title and URL.

I use the command copyCurrentTitle to create the link in Word by hand. I think I wouldn't need it, if there was a command copyCurrentInfo

>     4. Why is this needed for the current page only, and not for link hints? Note that there is an existing command `yf` (`LinkHints.activateModeToCopyLinkUrl`) which copies a link URL to the clipboard. Should there be a similar command to copy the link+title to the clipboard?

I need this for the current page and link hints.

PS: I know Word is terrible in every way, but for now this is the easiest solution for me.

glostis:
> Could everyone who's interested in this elaborate on your use cases?

1. I'm often pasting links text editors that supports Markdown syntax, and I feel that using the page title is more readable than just pasting the link.
2. Markdown-compatible text editors are quite common: Github issues/PRs, Slack (you need to [activate an option](https://slack.com/intl/en-gb/help/articles/360039953113-Set-your-message-formatting-preference) for this), Google Docs (you need to [activate an option](e!rje65wP%%qr4) for this), JIRA. I generally feel more "in control" when using plain text as opposed to rich text.
3. No, when I paste a page title, I always want the URL with it.
4. Having it for link hints would be great as well!

80avin:
I also have to paste it into slack or rich text documents (google doc/sheet, confluence, etc) or markdown editors.

I think better way is to copy rich text (`text/html`) instead of the markdown as `text/plain`.
Most of the markdown editors support pasting rich text links, and automatically convert it into markdown link or whatever form they support.

Using native clipboard API, I can do it easily as
```javascript
const data = [new ClipboardItem({
    'text/html': new Blob([`<a href="${document.location.href}">${document.title}</a>`], {type: 'text/html'}),
    'text/plain': new Blob([document.title], {type: 'text/plain'})
})]
navigator.clipboard.write(data)
```

Pasting this clipboard item into most of markdown/rich-text editors works and they render it as they like. It further has the advantage of using  <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd>  to paste as plain text.

zzeitt:
Use VimiumC like this:
```
map ye copyWindowInfo format="[[${url}][${title}]]" type="tab"
```
Ref: https://github.com/gdh1995/vimium-c/wiki/Substitute-URLs-and-text-during-commands#copy-information-of-tabs

Note: Change format as per your need.

srackham:
This would be a great feature. I like the generic approach taken by VimiumC: for example by allowing a `format` argument for the `copyCurrentUrl` command. This would allow the creation of a custom mapping with something like e.g.

       map ym copyCurrentUrl format='[%t](%u)'

camoz:
> Could everyone who's interested in this elaborate on your use cases?
> 
>     1. Are you pasting this information into your markdown files, into Github, into social media sites?
> 
>     2. Why just Markdown? Why not HTML -- or rich text, so the link can be copied into editors like Google Docs or Slack?
> 
>     3. Do you ever want the title and not the URL? Many of the older bugs which were referenced were requesting a shortcut to copy the title of the page, not the combination of title and URL.
> 
>     4. Why is this needed for the current page only, and not for link hints? Note that there is an existing command `yf` (`LinkHints.activateModeToCopyLinkUrl`) which copies a link URL to the clipboard. Should there be a similar command to copy the link+title to the clipboard?

1. Mostly into my markdown files (a lot), also sometimes into markdown online editors like Github or other ones.
2. Yes, other markup formats would be very nice, too. Like org-mode, for example.
3. Rarely.
4. This would also be nice. Often I want to look at a link before copying it (to reference it elsewhere), so that's why the current page use-case is popular.

> I'm concerned about adding a feature which is too narrow.

Yes, a generic interface with a `format='...'` option would be ideal, IMO.

PR #4252, opened 2023-05-07
Move cursor at end of text on focused one-line textareas

### LINKED ISSUE #4247: gi: "focus the first (or n-th) text input box on the page" does not work as before
If I type "gi", the cursor does not go to the end of the text form where some text has already been filled in, but instead goes to the head.

This issue has been occurring since around April 2023. I would like "gi" to behave as it did before.

*using Chrome ver.112.0.5615.121

### LINKED ISSUE #4251: focusInput not working correctly on Google search page
**Describe the bug**
This issue first appeared about a month ago. Normally when focusInput is used on any page the cursor goes to the end of the text if there is text there already. However, when using Google search the cursor now appears at the start of the already input text if there is one. Note that this only happens on the Google search page not on www.google.com, that is the problem comes into place if you search for something on google.com or in your browser URL field and then want to amend the search term by adding something to it, usually to the end of the search term. 

**To Reproduce**
1. Go to URL 'www.google.com'
2. Click on 'search for something'
3. On the results page use focusinput to focus the input you will see that the cursor goes to the start of the input instead of the end

Microsoft Edge
Version 112.0.1722.64 (Official build) (64-bit)
Vimium version 1.67.4

MarcoLucidi01:
When we `simulateSelect` an input and the cursor is at the start of the text, we move it to the end of text for all `isSelectable` elements *except* `textarea`s to avoid bad UX for multiline inputs (e.g. the end of input could happen to be out of the viewport). However some sites may use `<textarea>` for what really is a single-line input for 99% times of the usage (e.g. Google search bar which recently has become a `<textarea>`) and the cursor won't go to the end of text in such cases.

This commit changes `simulateSelect` to skip `textarea` cursor reposition only if the `textarea` has *more than one* line of text treating single-line `textarea`s as regular `isSelectable` elements.

fixes #4247
fixes #4251

### DIFF
--- a/lib/dom_utils.js
+++ b/lib/dom_utils.js
@@ -299,11 +299,15 @@ var DomUtils = {
       return handlerStack.bubbleEvent("click", {target: element});
     } else {
       element.focus();
-      if (element.tagName.toLowerCase() !== "textarea") {
-        // If the cursor is at the start of the (non-textarea) element's contents, send it to the end. Motivation:
+      if ((element.tagName.toLowerCase() !== "textarea") || (element.value.indexOf("\n") < 0)) {
+        // If the cursor is at the start of the (non-multiline-textarea) element's contents, send it to the end.
+        // Motivation:
         // * the end is a more useful place to focus than the start,
         // * this way preserves the last used position (except when it's at the beginning), so the user can
         //   'resume where they left off'.
+        // This works well for single-line inputs, however, the UX is *bad* for multiline inputs (such as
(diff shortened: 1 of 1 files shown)


zoriya (approved):

jonagull:
Can someone merge this? :)))

0xatulpatil:
bumping this PR, please someone merge this

MarcoLucidi01:
hi @philc, sorry to bother you with a direct ping. can we have some feedback on this issue? it's been a while now and this thing is really annoying :S

thank you for the amazing plugin!

PR #4254, opened 2023-05-09
set .idea directory in .gitignore.

yangwansu:
## Description

When contributing using the IDEA, I wish to not include the .idea directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ dist
 jscoverage.json
 tags
 .cake_task_cache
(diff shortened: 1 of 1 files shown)


astery:
There are plenty of different editors and tools that can create some garbage in the project folder used by different developers. Consider to ignore such files not per project e.g. `<project>/.gitignore`, but in your global gitignore in a home folder `~/.gitignore`

philc:
Agreed.

PR #4280, opened 2023-07-21
Contributing: Update puppeteer version

rileywilddog:
The currently-listed puppeteer version did not work for me (hung indefinitely).
You may also need to update the deno lock, until I ran `deno cache --lock=deno.lock --lock-write make.js` I got:
```
$ ./make.js test
error: The source code is invalid, as it does not match the expected hash in the lock file.
  Specifier: https://esm.sh/jszip@3.5.0
  Lock file: .../vimium/deno.lock
```


philc:
Nice; thanks! I've updated deno.lock in faeb5bbc9cb215d15aa0faa390f6ca5da3ec17e9; hopefully that resolves the issue.

PR #4298, opened 2023-08-24
Correct rawQueryList initialization

wjianbo:
## Description

This pull request proposes a modification to the code logic to fix a problem I encountered while using the find-mode.

### Problem
The original code didn't cover all conditions when initializing the rawQueryList variable, resulting in an error when saving the query and finding the next match.

### Solution
In this pull request, I've restructured the logic to offer a clearer and more reliable initialization process for the rawQueryList variable.

Please review and provide feedback. Thank you!


philc:
Thank you @wjianbo !

Issue #4261, opened 2023-06-08
Find in page with regular expressions doesn't work in Chrome on MacOS


andrea-berling:
**Describe the bug**
Find in page using  regular expressions doesn't work in Chrome on MacOS, neither when using \r at the end of the search query, nor by enabling "Treat find queries as JavaScript regular expressions". The count of matches seems to be correct, but when using n to navigate to the next match, it goes to the next "regular" match, instead of the next regex match. See "To Reproduce" for an example

**To Reproduce**
1. Go to URL [https://cloud.google.com/monitoring/mql/reference](https://cloud.google.com/monitoring/mql/reference)
2. Click `/` to start a search
3. Write `val` in the search query. You should get something similar to this: ![image](https://github.com/philc/vimium/assets/25068541/37451fb9-7ce3-4667-aeb8-16a85cd217c7)
4. Now write `\b` after `val` to match the string `val` followed by a blank character. You should get something similar to this (note that the number of matches decreases, but the highlighted match is still for just `val`): ![image](https://github.com/philc/vimium/assets/25068541/aa4579f5-11f0-4a0c-83a4-b80871045631)
5. Click `Enter` and `n`, you should get something like this (note that the next highlighted match is still just for `val`): ![image](https://github.com/philc/vimium/assets/25068541/0930e066-2391-4eed-a336-99501c040005)

**Browser and Vimium version**

Chrome version: `114.0.5735.90 (Official Build) (arm64)`
OS Version: `macOS Version 13.4 (Build 22F66)`
Vimum version: `1.67.4`

wjianbo:
I reproduced the behavior in Chrome on MacOS.
Additionally, I found the metacharacters `^` and `$` are not working either.
I tried to improve it by myself for a couple of days, And I want to create a pull request later.

andrea-berling:
@wjianbo that is great to hear, looking forward to see the PR

wjianbo:
@andrea-berling I've made the PR but I am not sure whether it would be merged.
Maybe you can try out yourself if you really need this feature.

andrea-berling:
@wjianbo Nice job mate, it works but I have the following feedback:
[...]

Issue #4291, opened 2023-08-04
how to make the search window appears on the lower left foot  like vim instead of the lower right foot 


edte:
May I ask how to use CSS to modify, so that the search window that appears after pressing /  appears on the lower left foot like vim. instead of the lower right foot 

this: 
<img width="394" alt="截屏2023-08-04 17 31 27" src="https://github.com/philc/vimium/assets/50194671/13e52e32-8fe0-470f-b880-cd56be7db650">

wjianbo:
Hope this could help.
![image](https://github.com/philc/vimium/assets/2434388/73d5503e-4055-42b3-92f6-6bd4a86538eb)

edte:
it works,thanks.

PR #4299, opened 2023-08-31
Improve regular expression support

wjianbo:
## Description
Find in page with regular expressions doesn't work well as discussed in issue #4261.
This PR suggests an improvement on this problem.

Fixes: #4261

### DIFF
--- a/content_scripts/mode_find.js
+++ b/content_scripts/mode_find.js
@@ -145,7 +145,7 @@ class FindMode extends Mode {
     // characters. See #1434.
     this.restoreSelection();
     query = FindMode.query.isRegex
-      ? FindMode.getNextQueryFromRegexMatches(0)
+      ? FindMode.getQueryFromRegexMatches()
       : FindMode.query.parsedQuery;
     FindMode.query.hasResults = FindMode.execute(query, options);
   }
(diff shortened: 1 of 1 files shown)


philc (changes requested):
Hey @wjianbo, thanks for looking at this deeply. This looks promising. I wasn't able to try it out because it had some merge conflicts, but I've given it some initial comments.

Can you rebase this onto the head of master so it will merge cleanly, run `deno fmt` in the root of the project to format your code, and address the comments? I'll then take another pass, focusing on the UX.

wjianbo:
> Hey @wjianbo, thanks for looking at this deeply. This looks promising. I wasn't able to try it out because it had some merge conflicts, but I've given it some initial comments.
> 
> Can you rebase this onto the head of master so it will merge cleanly, run `deno fmt` in the root of the project to format your code, and address the comments? I'll then take another pass, focusing on the UX.

Thanks for your kind review. To address some of your comments:
 
In order to get  `^` and `$` to work, I had to treat the text on a page as separate lines (elements) instead of one text string. Since `window.find` does not support regular expressions, as you mentioned, I wrote my own window.find function. In this function I needed to first locate the line, then find the match itself. That's the reason why I use a 2d Array. The inner arrays in the 2d array represent match results of each line, so the length of each inner array is may vary. That's why I use a `for` loop to compute row and col.

wjianbo:
Hi @philc, I've rebased it so you can try it out. 
I'm working on removing the window.find code path, but it will take some time to make it work perfectly.

content_scripts/mode_find.js
```diff
@@ -194,34 +192,42 @@ class FindMode extends Mode {
...
+    this.query.regexMatches = regexMatches;
+    this.query.regexPattern = pattern;
+    this.query.regexMatchedNodes = matchedNodes;
+    this.query.activeRegexIndices = [0, 0];
 
-    return this.query.matchCount = regexMatches != null ? regexMatches.length : null;
+    return this.query.matchCount = regexMatches != null ? regexMatches.flat().length : null;
   }
 
   static getNextQueryFromRegexMatches(stepSize) {
```
philc, line 208: I had to look at the caller to understand what stepSize was supposed to be, and why it wouldn't always be "1". Instead let's use the parameter `backwards` to make this function match the rest of the code, and calculate stepSize inside of `getNextQueryFromRegexMatches` based on the value of `backwards`.

content_scripts/mode_find.js
```diff
@@ -194,34 +192,42 @@ class FindMode extends Mode {
...
     // find()ing an empty query always returns false
-    if (!this.query.regexMatches) {
+    if (!this.query.regexMatches || !this.query.regexMatches.length) {
```
philc, line 210: This can be `!this.query.regexMatches?.length`

content_scripts/mode_find.js
```diff
@@ -262,17 +264,16 @@ class FindMode extends Mode {
...
-        query,
-        options.caseSensitive,
-        options.backwards,
-        true,
-        false,
-        false,
-        false,
-      );
-    } catch { /* swallow */ } // Failed searches throw on Firefox.
+    if (this.query.regexMatches && this.query.regexMatches.length) {
```
philc, line 267: Use optional chaining

content_scripts/mode_find.js
```diff
@@ -401,5 +402,59 @@ const selectFoundInputElement = function () {
...
+  const rect = textNode.parentElement.getBoundingClientRect();
+  if (rect.top < 0 || rect.bottom > window.innerHeight) {
+    textNode.parentElement.scrollIntoView({ block: "center" });
+  }
+
+  return true;
+};
+
+const getAllTextNodes = () => {
+  function getAllTextNodes(node, textNodes) {
```
philc, line 442: This doesn't need to take `textNodes` as a parameter; it can just use the textNodes declared in the surrounding scope. For clarity, move the textNodes declaration on the line above this function declaration.

philc (changes requested):
I tried this and it works great. Nice work! Two questions:

1. Can you confirm this works on Firefox?
2. The original bug #4291 mentioned that regexp support didn't work on MacOS, but based on the code changes, I presume it didn't work on any OS?

wjianbo:
@philc

Thank you for trying it out.  and here are my responses to your questions:

> 1. Can you confirm this works on Firefox?

- Just confirmed it works as expected on Firefox.

> 2. The original bug [how to make the search window appears on the lower left foot  like vim instead of the lower right foot  #4291](https://github.com/philc/vimium/issues/4291) mentioned that regexp support didn't work on MacOS, but based on the code changes, I presume it didn't work on any OS?

- I believe you mean #4261 and you are correct. It didn't work on Windows or my Arch Linux either.

I've also made another change so that we can start a new search from the same position as the latest selection, just as the way window.find worked. Please give it a review. Thanks for your comments; I've learned a lot.

philc:
This is excellent work. Thanks for the contribution @wjianbo.

andrea-berling:
So excited about this 😃 great job @wjianbo and thanks @philc for reviewing, giving feedback, and ultimately merging the change. I just have one question for @philc, and pardon my impertinence/impatience, but when do you think you're going to make a new release to the Chrome Web Store? Is the next release going to be 2.0, or are you planning to release something in the middle before that? Like 1.68.0 or something

philc:
Next release will be 2.0 and I plan to ship it in the next few weeks.

PR #4310, opened 2023-09-23
Fix z-index of link hints

### LINKED ISSUE #4068: No hints on cookie banner
**Describe the bug**
Cookie banner containing one link and two buttons pops up, but vimium doesn't hint those controls so I can't dismiss the banner and use the site. I assumed this was just people using span tags or some nonsense, but it's a site I control and it has a proper hyperlink and button tags. Is there a way I can make this available to me? I am not sure if the popup is loading after vimium, or something else?

**To Reproduce**
1. Go to URL https://www.onetrust.com/cookie-banner-gallery/ (without cookies set for this site)
2. Generate hints. The cookie banner in the footer is what I'm looking at.

Firefox 94.0
Vimium 1.67.1

This definitely could be more of a question than a bug report, either way I'm hoping there's an easy fix!
Screenshot 2022-06-01 at 12-21-02 Cookie Banner Gallery OneTrust

JulianDeal:
Fixes #4068

## Description

This pull request addresses the issue where cookie banners or other DOM elements have a higher z-index than Vimium's link hints, causing them to be obscured. The code changes introduced in this PR focus on resolving the z-index conflict when these elements are obscured by their parent element.

## Discussion

An alternative to patching the z-index as done in this implementation would be reordering the link hints in the DOM as mentioned in [this comment](https://github.com/philc/vimium/issues/4068#issuecomment-1732490163) and set Vimium's elements z-index to the max value.

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -38,6 +38,7 @@ class LocalHint {
   element; // The clickable element.
   image; // When element is an <area> (image map), `image` is its associated image.
   rect; // The rectangle where the hint should shown, to avoid overlapping with other hints.
+  zIndex; // Tracking z-Index of parent
   linkText; // Used in FilterHints.
   showLinkText; // Used in FilterHints.
   // The reason that an element has a link hint when the reason isn't obvious, e.g. the body of a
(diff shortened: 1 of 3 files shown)


JulianDeal:
Will be replaced by new implementation.

PR #4312, opened 2023-09-24
Fix forward word/vimword behaviors in caret mode

twio142:


philc:
Hey @twio142, it looks like this fixes the issue where "w" does not move forward a word in caret mode. I haven't looked at the code in mode_visual.js -- it's in need of cleanup. From what I can tell from your diff, it looks like "move forward" was simply never implemented? Why does move backward work without the need to branch on alterMethod=="move"?

twio142:
@philc 

> Why does move backward work without the need to branch on alterMethod=="move"?

That's because for forward moving was this character-by-character strategy implemented.

While for backward moving it's just using this universal method: `this.selection.modify(this.alterMethod, direction, granularity);`

mistog4n:
Just wondering if this could be merged at some point?

PR #4320, opened 2023-10-02
Update default dark theme css to include vimiumHUD styling

0x0003:
## Description

Current default dark theme (applied by `prefers-color-scheme: dark`) doesn't style Vimium HUD, 
which clashes against other elements that utilize the dark theme, namely vomnibar.

This PR adds missing styles.


philc:
Sorry for the delay on this, and thank you for the PR. This was fixed in 3a9c4bf091546cbde74e8ff2064f94f33af96fea.

PR #4330, opened 2023-10-11
feat: open with browser default search engine

### LINKED ISSUE #2598: Feature request: respect default search engine in vomnibar
(I'm one of the VimFx refugees, so apologies if this request is too Firefox-specific.)

When I search with the Vomnibar, that search is executed in Google. My default search engine (in Firefox, but I believe you can set on in Chrome too?) is not Google, however. It would be nice if Vimium could find out which search engine I've set as default and use that for searches.

ayn2op:
Closes #2598

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ dist
 jscoverage.json
 tags
 .cake_task_cache
+
+# Visual Studio Code
+.vscode/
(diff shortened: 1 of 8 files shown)


philc:
Hey @ayn2op this looks pretty good -- it nicely allows us to remove an option from Vimium. Why did you close the PR?

Note that I think we'll need to add another permission to manifest.json for this, which unfortunately is usually a headache for users.

PR #4339, opened 2023-10-17
Fix: vomnibar does not use UserSearchEngine when it is given

### LINKED ISSUE #4337: After the upgrading of vimium, I can't search a keyword in specific website by 'search-engine shortcuts'.
As the title describes.
My chrome version is:
Version 118.0.5993.71 (Official Build) (64-bit)
The vimium version which can not effect: 2.0.3, 2.0.2, 2.0.1, 2.0.
But the version 1.67.7 and other version under 1.67.7 can still take effect.

### LINKED ISSUE #4340: Custom search engine with custom key binding broken
This might be a duplicate of #4337, but it's not clear.

**Describe the bug**

I have this custom search engine...

```
g: http://www.google.com/search?q=%s Google
```
and this key binding...

```
map sg Vomnibar.activate keyword=g
```

... which should open the Vomnibar with the Google custom search engine already activated.

As of the latest update (2.0.3) this no longer works.  Basically, the leading `g` to enable the search engine seems to be dropped.

Note... the search engine does still work if activated manually.  So the issue seems to be to do with the Vomnibar activation, not the custom search engine itself.

**To Reproduce**

Add the custom search engine and key mapping above and type `sg`.  We should be dropped into the custom Google search, but you're instead dropped into the generic Vomnibar.

Chrome: 117.0.5938.157
Vimium: 2.0.3

Thanks for your help.

NI57721:
## Description
fix #4337
fix #4340 
VomnibarUI comes to lack a method that reads a UserSearchEngine if it is given, through the upgrading for v2.0.

### DIFF
--- a/pages/vomnibar.js
+++ b/pages/vomnibar.js
@@ -30,6 +30,7 @@ const Vomnibar = {
     this.vomnibarUI.setInitialSelectionValue(options.selectFirst ? 0 : -1);
     this.vomnibarUI.setForceNewTab(options.newTab);
     this.vomnibarUI.setQuery(options.query);
+    this.vomnibarUI.setActiveUserSearchEngine(UserSearchEngines.keywordToEngine[options.keyword]);
     this.vomnibarUI.update();
   },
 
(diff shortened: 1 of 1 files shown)


philc:
Great, thanks for taking a look @NI57721.

PR #4343, opened 2023-10-19
Fix: find mode does not seek to display the match

### LINKED ISSUE #4341: Find mode jumps to middle of the document (on first match) and find-next/previous does not seek to display the match
Apologies in advance, if I'm holding this tool incorrectly...
Chrome 118.0.5993.70 on MacOS
How to reproduce:
1) GMail mailing list digest email, launch the digest in it's own tab, hamburger -> "Show Original"
2) Enable Vimium permission to read the page
3) Search for "Subject: " enter-key (with trailing space)
4) Display jumps to the middle of the document approximately 50%
5) n-key or shift-n-key navigation does not update the location displayed in the page
6) Manual navigation within the page to a section where the match occurs shows that the text is found and highlighted, and n-key and shift-n-key shows that the match is moving to the next occurance and will loop around properly

I have disabled all other extensions to try and rule out something that I'm causing.
I have tried the available options to see if I can make it work as it used to, with no success.
I have been using Vimium for more than several years in this manner without issue before 10/16/2023.

wjianbo:
## Description

This pull request addresses an issue where the selected text range was not being centered on the screen when the containing element was larger than the viewport. To solve this problem, I modified the code to calculate and scroll to the center of the selected range.

fix #4341

### DIFF
--- a/content_scripts/mode_find.js
+++ b/content_scripts/mode_find.js
@@ -465,9 +465,13 @@ const highlight = (textNode, startIndex, length) => {
   selection.addRange(range);
 
   // Ensure the highlighted element is visible within the viewport.
-  const rect = textNode.parentElement.getBoundingClientRect();
+  const rect = range.getBoundingClientRect();
   if (rect.top < 0 || rect.bottom > window.innerHeight) {
-    textNode.parentElement.scrollIntoView({ block: "center" });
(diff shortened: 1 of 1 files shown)


philc:
Looks good to me. Thanks @wjianbo!

PR #4358, opened 2023-11-06
Optional chaining for `link.innerText...` in prev/next link logic (fixes GitHub)

Strajk:
## Description

**On GitHub, prev/next logic does not work.**

```
const links = DomUtils.evaluateXPath(linksXPath, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE);
```

matches also svgs with `button` in class

image

and svgs have `undefined` innerText #TIL, which throws in the matching logic 

Screen 2023-11-06 at 20 54 55

---

Added optional chaining which fixes the issue.

(thx for amazing work btw, I'm unable to control browser without vimium)

### DIFF
--- a/content_scripts/mode_normal.js
+++ b/content_scripts/mode_normal.js
@@ -417,7 +417,7 @@ const findAndFollowLink = function (linkStrings) {
 
     let linkMatches = false;
     for (const linkString of linkStrings) {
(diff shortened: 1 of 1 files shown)


philc:
Nice find @Strajk! I've merged in your commit.

Issue #4375, opened 2023-11-28
Vimium doesn't seem to exclude keys on https://aniwave.to


cherrynoize:
**Describe the bug**

Vimium doesn't seem to exclude any keys on https://aniwave.to.

My pattern is `https?://*(flix|anime|stream|video|watch|tube|movie|film|play|game|aniwave)*`, exclude keys are `tmhjklfv1234567890<>` and yet no keys seem to be excluded. If I add a new rule for `https?://aniwave.*` (all keys) it works but only until I reload the page. These patterns are showing in the Vimium popup, it just isn't applied correctly.

**To Reproduce**

1. Go to URL '[https://aniwave.to](https://aniwave.to)'
2. Done

**Browser and Vimium version**

FF: 120.0
Vimium: 2.0.6

eguneys:
Not just videoes, but html canvas games on itch.io, god, fix this awful plugin what is this.

Issue #3960, opened 2021-11-19
Vimium isn't disabled in cross-origin subframes of excluded documents.


kashav:
**Describe the bug**
Vimium isn't disabled in cross-origin subframes of excluded documents.

**To Reproduce**
1. Add `https://developer.mozilla.org/*` to the exclude list in Vimium options.
2. Go to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe. Type <kbd>i</kbd> (or <kbd>?</kbd>, or anything else to test if Vimium is enabled). Notice that Vimium isn't enabled, which is as expected.
3. Focus either of the iframes in the first demo (both the container and the map are embedded on different sites, so either works).
2. Type <kbd>i</kbd> and notice that Vimium is still enabled in the frame.

**Browser and Vimium version**
Firefox 96.0a1, but this reproduces in both Firefox and Chrome. Vimium 1.67.

gdh1995:
In this case Vimium works as intended - exclusion rules are computed on every frame, but not on a tab.

kashav:
Okay, I see. Any reason to not exclude Vimium in a subframe if it's excluded in the embedder?

gdh1995:
um, maybe in case both an outer frame and an inner frame have matched rules.



---Original---
From: "kashav ***@***.***&gt;
Date: Sat, Nov 20, 2021 11:31 AM
To: ***@***.***&gt;;
Cc: "Dahan ***@***.******@***.***&gt;;
Subject: Re: [philc/vimium] Vimium isn't disabled in cross-origin subframes of excluded documents. (Issue #3960)




 
Okay, I see. Any reason to not exclude Vimium in a subframe if it's excluded in the embedder?
 
—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

PR #4402, opened 2024-01-13
Exclusions: exclude keys based on the top frame's URL, not subframe URLs

### LINKED ISSUE #3254: Cannot disable for Google Slides

philc:
The exclusion keys in Vimium apply on a per-frame basis. So, if you're visiting a site `example.com`, and it has an iframe showing content from `sub.example.com`, your exclusion keys won't work if the iframe has the focus. This is very surprising, and I think hard for people to debug and grok. It appears that Vimium's exclusion keys are simply not working. It's also hard to workaround -- how does one know what the domain of the iframe is, so that they can set exclusion rules for that domain? You have to open the browser's developer tools and poke around.

I'm not sure where this per-frame exclusion behavior originally came from, and whether there was a specific reason for it. It was discussed early on in #1453.

This PR changes it so that the exclusion rules and passkeys from the top frame's URL are applied to all subframes on that page. I think this is a better UX.

This fixes #3254, #3960, #4375, and possibly other open issues.

### DIFF
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -560,8 +560,9 @@ const sendRequestHandlers = {
   },
 
   async initializeFrame(request, sender) {
-    const tabId = sender.tab.id;
-    const enabledState = Exclusions.isEnabledForUrl(request.topFrameUrl);
+    // Check whether the extension is enabled for the top frame's URL, rather than the URL of the
+    // specific frame that sent this request.
+    const enabledState = Exclusions.isEnabledForUrl(sender.tab.url);
 
     const isTopFrame = sender.frameId == 0;
     if (isTopFrame) {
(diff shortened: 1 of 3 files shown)


gdh1995:
In a cross-domain iframe, this PR may cause
```
Uncaught DOMException: Failed to read a named property 'toString' from 'Location': Blocked a frame with origin "https://passport.zongheng.com" from accessing a cross-origin frame.
```

philc:
Great catch @gdh1995. I've changed this to use the tab's URL, rather than having the content script itself pull the top frame's URL. This now works on cross origin iframes in my tests.

PR #4407, opened 2024-01-17
Clarify that f and F show visible links

gitressa:
## Description

Great add-on, it will make browsing so much faster, thanks!

After installing the add-on, I was reading the help page (`?`) but could not see how to find links. I normally use "search for links in text" via the `'` character in Firefox.

https://support.mozilla.org/en-US/kb/search-contents-current-page-text-or-links#w_search-links-only

It was only by mistake that I pressed `f` and saw the links on the page highlighted with shortcuts, which is an even better solution for finding links.

So would it make sense to clarify that `f` and `F` show links on the help page?

PS. I also fixed a few instances where a line starts with a capital letter, since it looks like the style is to only start with small letters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ Navigating the current page:
     G       scroll to bottom of the page
     d       scroll down half a page
     u       scroll up half a page
-    f       open a link in the current tab
-    F       open a link in a new tab
+    f       show links and open in the current tab
+    F       show links and open in a new tab
     r       reload
     gs      view source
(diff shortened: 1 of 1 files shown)


allamedo:
@gitressa I think the descriptions at `vimium/background_scripts/commands.js` should also be updated.

gitressa:
Thanks @allamedo, I updated the `background_scripts/commands.js` file as well.

VincentBarkman (approved):
No apparent issues from what I can see

*Vincent A. Barkman*

PR #4417, opened 2024-02-07
feat: move tab to an existing window by orientation

IdoKendo:
## Description

Following the PR at #3925 -- this PR adds a new feature, to move the current tab to an existing window.

This solution takes a different approach, and merges the tab to an existing window, by the relative direction, in order to provide a quicker and seamless motions, following the hjkl paradigm:
- `wh` will move the tab to a window on the left
- `wl` will move the tab to a window on the right
- `wk` will move the tab to a window above
- `wj` will move the tab to a window below

An example use case would be when a user moves a tab to a new window using the existing `W` binding, and now wants to merge to back to the previous window, this feature adds this option, by allowing them to merge it using `wh` as an example, if the previous window was to the left of the new one.

This also allows moving tabs between existing windows in order to rearrange the tabs between windows without using the mouse.


YoavBaavour:
very interested in this feature!

philc:
Thanks @IdoKendo for the PR.

* What happens if the Chrome window is behind the current one? E.g. you have two chrome windows, both maximized on the same monitor. This is a common case for me.
* I presume this works as-is even if a Chrome window is minimized, or are minimized windows filtered out from consideration?

More generally, I think the UX for this feature needs more discussion and finalization before we implement it. See https://github.com/philc/vimium/issues/4640.

landorg:
I really like that approach so I installed your fork (merged with current upstream master).
For me there are a couple of problems when I use this with my i3 setup.

It seems to not always find the right window. I most of the time have multiple windows on different desktops. The direction seems to be detected by desktop number. This in my case might not be right order in my case. I have very often desktop 2,3 on my left monitor and 0,1 on the right one.
Also the handling of stacked windows isn't really working in my case. It moves tabs to other windows on other desktops most of the time. 
Sometimes to desktops that are hidden.

PR #4431, opened 2024-03-02
Use the tab.lastAccessed timestamp to sort tabs by recency in Vomnibar

### LINKED ISSUE #4368: 'Vomnibar.activateTabSelection' no longer ordering tab list in chronological order (of most recent)
**Description**

This has been an issue for the last year or so; my tab history using ⇧ T no longer is in chronological order of last open. Have no idea how the ordering is currently working. But I loved having this shortcut to quickly switch between most-recent, just like ⌘⇥ switches between most-recent applications.

I tried looking in the chrome extension settings and couldn't find anything that might be preventing vimium from configuring the action correctly, with chronologically-ordered tab history, as I had before.

**To Reproduce**

Vomnibar.activateTabSelection – which I have currently set to ⇧ T

Chrome Version: 117.0.5938.92 (Official Build) (arm64)

Vimium Version: 1.67.7

philc:
tab.lastAccessed has been available in Firefox since Firefox 56, and was just added to Chrome 121.

We were previously tracking this property ourselves. Our implementation was flawed because the tab last accessed timestamps would be garbage collected every time the extension's service worker gets unloaded, as of our migration to manifest V3.

Now we can get rid of our own state tracking in the background service worker, and as a result, this should fix #4368.

### DIFF
--- a/background_scripts/bg_utils.js
+++ b/background_scripts/bg_utils.js
@@ -1,80 +1,4 @@
-const TIME_DELTA = 500; // Milliseconds.
-
-// TabRecency associates a logical timestamp with each tab id. These are used to provide an initial
-// recency-based ordering in the tabs vomnibar (which allows jumping quickly between
-// recently-visited tabs).
-class TabRecency {
-  constructor() {
-    this.timestamp = 1;
-    this.current = -1;
-    this.cache = {};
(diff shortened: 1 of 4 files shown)


philc:
Closing in favor of #4443.

PR #4432, opened 2024-03-03
Fixes for Orion Browser #3610

tyirvine:
## Description

This pull request has some reference to issue #3610. But mainly this pull request is so we can run Vimium in [Orion](https://kagi.com/orion/). For those that don't know, Orion is a WebKit based browser that is attempting to support both Chrome and Firefox extensions alike.

I believe Vimium was working previously for the most part but then something to do with settings parsing broke it I believe. The issue I found was that WebKit's support for `JSON.parse()` seems to have a few issues.

1. There was a `^` symbol in `vimium.css` that WebKit's `JSON.parse()` didn't know how to handle when iterating and parsing through the settings keys in `settings.js: migrateSettingsIfNecessary(settings)`.

2. WebKit's `JSON.parse()` struggled again in the same iterator when given an item that was not a stringified JSON object. So a simple validation was added in so only stringified JSON objects would be parsed.

After these fixes were made I tested it with [Orion Version 0.99.126.4.1-beta (WebKit 618.1.2)](https://orionfeedback.org/d/6797-orion-macos-0991264).

_A small formatting fix was also made._

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -156,12 +156,23 @@ const Settings = {
     const shouldMigrate = settings["settingsVersion"] == null;
     if (!shouldMigrate) return settings;
 
+    // Safari struggles to parse anything that isn't a stringified JSON object when using JSON.parse().
+    // This ensures that only stringified JSON objects get passed into the parser.
+    const isValidJSON = (str) => {
+      try {
+        JSON.parse(str);
+        return true;
+      } catch (e) {
+        return false;
+      }
+    };
+
     // Migration for v2.0.0: decode all values so that they're not JSON string encoded.
     const newSettings = {};
     for (const [k, v] of Object.entries(settings)) {
       // Most pre-2.0 settings were strings, but the global marks were stored as native values. See
       // #4323. So check the setting value's type before migrating.
-      if (typeof v === "string") {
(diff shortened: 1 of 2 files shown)


tyirvine:
Hang on, sorry I just found an additional bug when using the `/` search command of Vimium. I'll work on fixing this.
```
TypeError: node.checkVisibility is not a function. (In 'node.checkVisibility()', 'node.checkVisibility' is undefined)
```

brunoti:
Any news on this one? I want to help! I think I have some more bugs and I want to see if I can contribute.

lib/settings.js
```diff
@@ -156,12 +156,23 @@ const Settings = {
...
+      }
+    };
+
     // Migration for v2.0.0: decode all values so that they're not JSON string encoded.
     const newSettings = {};
     for (const [k, v] of Object.entries(settings)) {
       // Most pre-2.0 settings were strings, but the global marks were stored as native values. See
       // #4323. So check the setting value's type before migrating.
-      if (typeof v === "string") {
+      if (typeof v === "string" && isValidJSON(v)) {
```
philg-dev, line 175: This code will call `JSON.parse(str)` twice on every successful call instead of putting the try-catch directly where it's necessary.

Issue #4408, opened 2024-01-19
Can't click github popups


Enzodtz:
**Describe the bug**

The f shortcut (and probably the other clickable ones), doesn't work on github popups. For example, if you want to approve a pull request you could click on "Review Changes" but not on the buttons that this first one opens.

<img width="646" alt="image" src="https://github.com/philc/vimium/assets/7780770/f89b3552-db4f-43bf-b2db-d0630186c0e8">

**To Reproduce**

1. Go to URL '...'

Include a screenshot if applicable.

**Browser and Vimium version**

Chrome: 120.0.6099.234 (Official Build) (arm64) 
Vimium: 2.0.5

gdh1995:
Duplicate of #4369

keithamus:
Hey friends, I've deployed a fix on GitHub.com, this should no longer be an issue.

Enzodtz:
@keithamus that sounds awesome, thank you. Unfortunately though that is not working for me yet, despite someone in the other thread saying that it worked for them 🤔

philc:
It appears this is not a duplicate of #4369. That issue reported that clicking a Vimium link hint for the "Review changes" button did nothing. This is now fixed (thank you @keithamus!).

Instead, this issue is reporting that link hints do not show for the content inside the Review Changes popup once it's opened. This is not yet resolved. I looked into it briefly and the link hints *do* appear, but they appear *under* the popup. I didn't look further into how this is possible. Vimium's link hints should have the max z-index.

I detected this by hiding the code review popup. In the screenshot below, you can see the link hints are present if you set the popover to 0.5 opacity.

<img width="716" alt="image" src="https://github.com/philc/vimium/assets/8598/acd24a09-48d3-4fb8-ab8f-58276e8423cf">

keithamus:
This will be because popovers are enabled on the browser you’re running (either via `dom.elements.popover.enabled` in `about:config` or by running the 122 beta I think). 

The solution in this case is to make vimiums overlay also use popover.

dudicoco:
@keithamus is that an easy fix? I'm also struggling with this issue.

dudicoco:
thanks @keithamus and @philc!

PR #4436, opened 2024-03-14
use popovers for hint markers

keithamus:
## Description

This PR offers a bit of a starting point for fixing https://github.com/philc/vimium/issues/4408. The basic principle is that all of the HintMarkers need to use [`popover`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover) in order to sit on the ["top layer"](https://developer.mozilla.org/en-US/docs/Glossary/Top_layer) so that they appear above all other content. 

I don't have any time to work on this PR further, but I thought it could serve as a bit of a starting point to help with #4408.

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -459,6 +459,7 @@ class LinkHintsMode {
       // Each hint marker is assigned a different z-index.
       el.style.zIndex = this.getNextZIndex();
       el.className = "vimiumReset internalVimiumHintMarker vimiumHintMarker";
+      el.popover = "manual";
       Object.assign(marker, {
         element: el,
         localHint,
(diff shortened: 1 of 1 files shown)


dudicoco:
Hey @keithamus. Can you clarify what else needs to be done on the PR?

keithamus:
I don't know. I didn't test this, I'm not sure it works or what more needs doing to it. I hope it (plus the links to the docs) provides some guidance as to where to go from here.

So I'd say the first step is to take this code, actually run it, see what breaks, see if it changes anything. It might be all that's needed, but I doubt it.

philc:
Thank you @keithamus. We'll take a look.

philc:
This didn't quite work as is, as the positioning of the hint markers needed to be adjusted. In  412cc37621a578ee260925f31e0722539af33f39 I took the approach of applying the popover attribute to the entire hint marker container, which looks like it works well. Thanks @keithamus for getting this started!

shea-parkes:
Thank you very much for getting this in!  GitHub approval popup now works again with vimium.  I very much appreciate it.

PR #4439, opened 2024-03-16
feat: open with browser default search engine

### LINKED ISSUE #2598: Feature request: respect default search engine in vomnibar

ayn2op:
Closes #2598

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ dist
 jscoverage.json
 tags
 .cake_task_cache
+
+# Visual Studio Code
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,3 @@ dist
 jscoverage.json
 tags
 .cake_task_cache
-
-# Visual Studio Code
-.vscode/
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -536,6 +536,10 @@ const sendRequestHandlers = {
     });
   },
 
+  launch({ query, openInNewTab }) {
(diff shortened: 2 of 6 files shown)


philc:
This looks pretty good to me. I'll merge it soon. Note that introduces the need for a new chrome permission, which will cause some friction on the next update.

.gitignore
```diff
@@ -9,3 +9,6 @@ dist
 jscoverage.json
 tags
 .cake_task_cache
+
+# Visual Studio Code
+.vscode/
```
philc, line 14: This shouldn't go in here -- non-repo specific ignores should be handled by the user's global git ignore. No need to repush; I can remove this when I merge.

(The other editor-specific entries in Vimium's .gitignore are there by historical accident and are due to be removed).

philc:
This is excellent. Nice work @ayn2op!

philc:
On further testing, this had a few edge cases. I think I've resolved them with bfe6dc53b4d6820515fc5684ad46bcaca30874d2 and 0fec0888f20c3104a5196c774faa1d043106f560.

philc:
That's fine. Confirming that I'm experiencing this too and it's super annoying; I just haven't had a chance to look at it yet.

philc:
I'm working on this and will push a fix in the next couple of days.

philc:
This should now be fixed.

PR #4443, opened 2024-03-26
Store tab recency information in session storage

### LINKED ISSUE #4368: 'Vomnibar.activateTabSelection' no longer ordering tab list in chronological order (of most recent)

philc:
Currently, the tab order shown by the `Vomnibar.activateTabSelection` command gets lost if the extension's background page is unloaded by the browser.

Here, we save the order into session storage so it persists.

This should fix #4368.

#4431 was another approach at this, but turned out to be non-viable.

### DIFF
--- a/background_scripts/bg_utils.js
+++ b/background_scripts/bg_utils.js
@@ -19,6 +19,8 @@ const BgUtils = {
   },
 };
 
+BgUtils.tabRecency.init();
+
 Object.assign(globalThis, {
   BgUtils,
 });
--- a/background_scripts/tab_recency.js
+++ b/background_scripts/tab_recency.js
@@ -1,10 +1,22 @@
 // TabRecency associates an integer with each tab id representing how recently it has been accessed.
-// These are used to provide an initial recency-based ordering in the tabs vomnibar.
+// The order of tabs as tracked by TabRecency is used to provide a recency-based ordering in the
+// tabs vomnibar.
+//
+// The values are persisted to chrome.storage.session so that they're not lost when the extension's
+// background page is unloaded.
+//
+// In theory, the browser's tab.lastAccessed timestamp field should allow us to sort tabs by
+// recency, but in practice it does not work across several edge cases. See the comments on #4368.
 class TabRecency {
   constructor() {
     this.counter = 1;
     this.tabIdToCounter = {};
+  }
 
+  // Add listeners to chrome.tabs, and load the index from session storage.
+  // If tabs are accessed before we finish loading chrome.storage.session, the in-memory stage and
+  // the session state is merged.
+  init() {
     chrome.tabs.onActivated.addListener((activeInfo) => this.register(activeInfo.tabId));
     chrome.tabs.onRemoved.addListener((tabId) => this.deregister(tabId));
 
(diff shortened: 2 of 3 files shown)


tobez:
Hi @philc , while this is a vast improvement, `visitPreviousTab` still jumps to something random if not used for a while (I am assuming - without verification - when the background service worker gets discarded).   It **keeps** jumping to something random if I use it repeatedly, until I actually care to invoke `Vomnibar.activateTabSelection` again, after which `visitPreviousTab` starts working flawlessly again - until next time it does not.   So all in all, still a regression over v1.67.1...

Baggydawg:
@philc this seems to have fixed all issues on my end! still on same specs as mentioned in #4368 

Thank you so much for working so hard to fix this issue. It's been a dearly missed feature! ❤️

Baggydawg:
@philc I'm really sorry – it's unfortunately still not working too well. I need to learn to properly QA before sending comments. @tobez summarised the issue well in his comment above.

When jumping between tabs, it seems they don't update to the (most recent) chronological ordering.

e.g.
Moving from:
tab 1 → tab 2 → tab 3
Then opening:
tab 4
Then moving back to:
tab 2
––
tab 2 seems to only be able to see tab 1 as the most recent last tab accessed.

philc:
Alright guys, thanks for bearing with me. Based on your description @tobez and @Baggydawg, I could see the issue by inspecting the code: there is a logic error when synchronizing the stored state and the in-memory state, which is narrowly fixed by 8b01abe9fa41fd71a43ec04fe73d41457f324cc0.

However, I reworked the logic (again) to be more robust still in c83e0b952f8bf7ea7a52789271549198d210c25c. Please take Vimium's master for a spin and let me know if you spot anything else. I'll include this fix in the next release.

Baggydawg:
@philc ok I have now done a bit more rigorous QA over the past hour, and this seems to have truly fixed it.

I have done the same test mentioned in my earlier comment several times now. Also tested with rebooting Chrome, and literally hopping between multiple each second, still works.

So @philc it looks like your latest state management / logic commit has fixed it! Thank you!!!!!!!! ❤️

Baggydawg:
@tobez would be great to know if it's been fixed on your end too 👍🏼

philc:
I've included these fixes in v2.1.2 and submitted to the stores.

tobez:
@philc It is perfect now, thanks a lot for getting to the bottom of it

Baggydawg:
@tobez that's great to hear, thanks for updating me.

@philc – just wanted to say another ENORMOUS thank you for fixing this!

It has completely (re-)revolutionised (once again) my web browsing experience!!!! Can't express enough my appreciation in words. ❤️

Issue #4444, opened 2024-03-28
Feature request: Add "R" keybinding for hard reload


tribhuwan-kumar:
Whenever I work on my django project, I have to always press `ctrl+shift+r` which is so annoying, I'd like to add a keybinding for hard reload like `ctrl+shift+r`.  We can use "R" for it.

philc:
This was completed in #4445.

philc:
This is a duplicate of #3735.

Issue #3735, opened 2020-12-24
Feature Request: R (big r) to empty cache and reload


mikevare:
In Chrome, there is only the option to do a hard reload using a shortcut CMD+SHIFT+R
If its possible, it would be awesome to simply press R to empty your cache and fully reload.
Normally, you need to open dev tools (r-click -> inspect) and r-click on the reload button, which is terrible.

I just found you extension, it's great, thank you. :-)

philc:
Great idea. This would be a great feature. It looks like this is possible with the tabs API ( https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/reload ) :

`browser.tabs.reload(2, {bypassCache: true});`

Anyone who wants to take a stab at it, I'd be happy to merge a PR.

gdh1995:
&nbsp;There has been such an option, named "hard".
https://github.com/philc/vimium/blob/bdf654aebe6f570f427c5f7bc9592cad86e642b5/background_scripts/main.js#L336



---Original---

mikevare:
Thanks gdh1995, so it will work after adding this line to your options
```map R reload hard```

Maybe someone can add this to the default key map, it makes total sense for webdevs :)

philc:
This was added as an advanced command under the keybinding "R" in #4445, which should make it more discoverable.

PR #4445, opened 2024-03-28
feat: add 'R' for hard reload

tribhuwan-kumar:
Added an advance command "R" in `background_scripts/commands.js` & a function for hard reloading in `background_scripts/main.js`. I updated the `README.md` as well. 
Checked in Google Chrome in Arch Linux, Everything is working fine!!

#4444

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,1 @@
-*.un~
-*.swo
-*.swp
-*.crx
-*.sublime*
-options/*.js
-node_modules/*
 dist
-jscoverage.json
-tags
-.cake_task_cache
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,20 @@
+2.1.0 (2024-03-27)
+
+- Fix Vomnibar doesn't always list tabs by recency.
+  ([#4368](https://github.com/philc/vimium/issues/4368))
+- Better domain detection in the Vomnibar ([#3268](https://github.com/philc/vimium/issues/3268))
+- Exclude keys based on the top frame URL, not a subframe's URL. This fixes many cases where the
+  excluded keys feature didn't seem to work. ([#4402](https://github.com/philc/vimium/issues/4402))
+- After selecting a link, if ESC is pressed, mouse out of the link. With this, Wikipedia's and
+  Github's link preview popups can be dismissed after following a link.
+  ([#3073](https://github.com/philc/vimium/issues/3073))
+- Fix link hints do not appear for links inside of github's popups. This fix is available on Chrome
+  114+, and soon Firefox. ([#4408](https://github.com/philc/vimium/issues/4408))
+
 2.0.5, 2.0.6 (2023-11-06)
 
 - Fix bug where "esc" wouldn't unfocus a textarea like it should.
-  ([#4336](https://github.com/philc/vimium/issues/4336)),
+  ([#4336](https://github.com/philc/vimium/issues/4336))
 - Fix passNextKey command.
 
 2.0.4 (2023-10-19)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+2.1.2 (2024-04-03)
+
+- Better fix for Vomnibar doesn't always list tabs by recency.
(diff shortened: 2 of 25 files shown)


VincentBarkman:
### Refactor `hardReload` Function

#### Changes:
- Replaced the nested function with `findIndex` for immediate tab location.
- Removed array reordering; implemented modulo looping for tab cycling.
- Added exit condition for non-existent `tabId` to improve error handling.

#### Code Comparison:
**Old Implementation**:
```javascript
function hardReload({ count, tabId, tab: { windowId } }) {
  const bypassCache = true;
  return chrome.tabs.query({ windowId }, function (tabs) {
    const position = (function () {
      for (let index = 0; index < tabs.length; index++) {
        const tab = tabs[index];
        if (tab.id === tabId) return index;
      }
    })();
    tabs = [...tabs.slice(position), ...tabs.slice(0, position)];
    count = Math.min(count, tabs.length);
    for (const tab of tabs.slice(0, count)) {
      chrome.tabs.reload(tab.id, { bypassCache });
    }
  });
}
```

**New Implementation**:
```javascript
function hardReload({ count, tabId, tab: { windowId } }) {
  const bypassCache = true;
  return chrome.tabs.query({ windowId }, function (tabs) {
    let startIndex = tabs.findIndex(tab => tab.id === tabId);
    if (startIndex === -1) return; // Exit if tabId is not found
    let processed = 0;
    while (processed < count) {
      chrome.tabs.reload(tabs[startIndex].id, { bypassCache });
      processed++;
      startIndex = (startIndex + 1) % tabs.length; // Wrap around if needed
    }
  });
}
```

*Vincent A. Barkman*

UncleSnail:
@tribhuwan-kumar I think this is a good feature, and your implementation is very similar to that of the regular reload. However, we now have a `forCountTabs` method that would be perfect for both of these. If @philc is okay with it, I propose we reimplement both regular reload and hard reload to use the `forCountTabs method instead of reimplementing the count logic, and get this merged. If you want to do the re-implementation @tribhuwan-kumar, go for it. Otherwise, I'd be happy to do it for you and make a PR to your fork, or make my own PR, whichever you prefer.
Just let me know what you think. This change could greatly simplify the implementation of both methods and make them more similar to the rest of the code. It also means they will work with Firefox hidden tabs, or any other future needs we build into the `forCountTabs` method.

philc:
@UncleSnail SGTM

tribhuwan-kumar:
@UncleSnail, It seems `forCountTabs` will be a great implementation. I liked your re-implementation idea using `forCountTabs`. Well, you can make a PR in any fork you want. It'll be your choice!

UncleSnail:
@tribhuwan-kumar I made a PR to your repo a few weeks ago and just wanted to make sure you saw it. If you would prefer, I can make the PR directly to the main repository if you would like.

UncleSnail:
@philc this should be ready to merge now, unless we want to consider changing the "R" explanation messages. Perhaps change the "ie ctrl+shift+r" to something like "(skip the cache)"?

philc:
Looks good. Thanks for the contribution!

I made a few small cleanups in 5fc9348aa3866c7bd3f1fd40664abd5eb166ff6e.

philc:
I just realized that we already have "hard reload" implemented. See #3735. It's just not mapped to a key by default.

The problem with the existing solution (adding the "hard" option to the reload command) is that it's not particularly discoverable. Even I'd forgotten about it. Although it is [documented in the wiki](https://github.com/philc/vimium/wiki/Tips-and-Tricks#hard-reload).

I think it's a useful enough to have its own key by default (R) as this PR implements. What do you think?

UncleSnail:
> I think it's a useful enough to have its own key by default (R) as this PR implements. What do you think?

I agree, I think having "R" mapped to hard reload by default is a good idea. However, we don't want duplicate functionality in the codebase, so we should probably only have one correct way to implement or map/unmap hard reload. I can make a PR that tries to clean up any duplicate functionality if you would like.

philc:
@UncleSnail That would be great.

UncleSnail:
> @UncleSnail That would be great.

I fixed this issue in [this PR](https://github.com/philc/vimium/pull/4518).

Issue #4447, opened 2024-03-29
[Regression Bug] v2.1.0 Key exclusion popup


wenfangdu:
Vimium: 2.1.0

Repro Steps:
1. Exclude keys for a website and save changes
2. Visit the website and evoke Vimium popup
3. It shows no exclusion was configured

![image](https://github.com/philc/vimium/assets/28700378/06a4195b-0a9c-4e5e-b2d9-db16f8b41a81)

@philc

devusr1x:
Same issue. Major bug. How did this get pushed into the store? I'm having to load the old version from source on multiple profiles, this is extremely annoying.

philc:
My mistake. It was introduced by the last commit before v2.1.0, and a hard bug to notice (the popup looks like it's still working). Note that existing exclusion rules still apply, and they're still editable on the Vimium Options page, but the popup itself does not function.

I've pushed v2.1.1 to the store which will fix this.

devusr1x:
I don't think that's the case. I can't give you debug info right now and I'm not sure/don't remember exactly what was happening, but something odd was going on and was not consistent with previous versions. I found out about this bug when I started messing e-mails by accident on Gmail, where I had some keys excluded.

tammersaleh:
> Note that existing exclusion rules still apply, and they're still editable on the Vimium Options page, but the popup itself does not function.

I think it's worse.  Vimium (which I LOVE, btw) is active in Google Docs even though I have that url excluded.

philc:
@tammersaleh I confirmed on 2.1.0 that existing exclusion rules will still be applied even though the dialog isn't working, and in particular on docs.google.com. Can you post more details? What is the exclusion rule that you have, and what is the URL of your google docs site?

devusr1x:
I just played around with 2.0.6, 2.1.0 and 2.1.1 to see what exactly was going on and couldn't figure out the logic of what was happening. It seems like some keys that weren't on the exclusion list were being excluded. As I said, it happened on Gmail with me, but I doubt it's site-specific.

Regardless of that, 2.1.1 from git is working as before (both exclusion functionality and popup).

PR #4448, opened 2024-03-29
fix(#4447): Fix Key Exlusion Bug

sudo-nick16:
Load tab_recency module before loading bg_utils
in action page.

## Description
bg_utils depends on tab_recency, since tab_recenty wasn't being loaded, bg_utils was throwing error
which caused the bug #4447 
So loading tab_recency fixes the issue.


philc:
Thanks for the quick investigation. I’ll merge this shortly and submit to
the store.

Issue #4446, opened 2024-03-28
Hint overlays do not appear on GitHub Enterprise


bent0b0x:
**Describe the bug**

Pressing `f` to trigger link/button hints does not yield any visible overlays on GitHub Enterprise. This is probably due to https://github.com/philc/vimium/commit/412cc37621a578ee260925f31e0722539af33f39.   

**To Reproduce**

1. Go to any github enterprise page
2. Press `f`

I would expect the overlays to appear so that I can press character keys to open links, but they do not appear.

**Browser and Vimium version**
Vimium: 2.1.0
Chrome: Version 123.0.6312.87 (Official Build) (arm64)

philc:
I'm surprised -- I'm not sure what would be so different about github enterprise vs. the regular site that would break link hints.

Unfortunately, I don't have access to any github enterprise instances. Could someone with access to it take a closer look?

bent0b0x:
It looks like the hint markers are all being rendered, but they are not visible due to a css rule that gets included in GHE (at least the version my org is on, which is `GitHub Enterprise Server 3.9.4`):

```
@supports not selector([popover]:open)
[popover]:not(.\:open) {
    display: none;
}
```

My `css` chops are a bit stale, but it looks like this is ultimately saying "hide all popover elements unless they are explicitly open"?

I think adding `display: 'block'` [here](https://github.com/philc/vimium/blob/9780226bf472eae93f49ee9d8984e6f2f12668ed/content_scripts/link_hints.js#L418-L425) would fix the issue, I can PR that to discuss

changmin-ahn:
Same for `GitHub Enterprise Server 3.9.10` 😢

roachsinai:
> I'm surprised -- I'm not sure what would be so different about github enterprise vs. the regular site that would break link hints.
> 
> Unfortunately, I don't have access to any github enterprise instances. Could someone with access to one take a closer look?

No. I cannot use `/` to search a word in regular site now. Type `/` and after type target word, vimium show it found matches, but can't jump to it by press Enter.

bent0b0x:
FYI, here's a temporary config-level fix until a new version of Vimium is released w/ a patch

https://github.com/philc/vimium/pull/4451#issuecomment-2034349648

roachsinai:
Thanks for quick reply!

philc:
v2.1.2 includes this fix and I've submitted it to the stores today.

PR #4451, opened 2024-03-30
Add display style attribute to prevent popover content from being hidden on GitHub Enterprise (and maybe other sites)

bent0b0x:
## Description

This PR was spun out of [this issue](https://github.com/philc/vimium/issues/4446), where it was reported that hint markers were no longer visible on GitHub Enterprise pages since [this commit](https://github.com/philc/vimium/commit/412cc37621a578ee260925f31e0722539af33f39). 

It looks like GHE pages have some styling rules that prevent popovers that are not technically "open" from being displayed. If we explicitly set our marker container's `display` style attribute that will take priority over those rules. 

## Verification
Testing locally, the issue is resolved on GHE

ghewin

I was not able to run the tests locally; it looks like there are some restrictions on my machine that interfere with puppeteer's installation:

>error trying to connect: invalid peer certificate: UnknownIssuer

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -419,6 +419,7 @@ class LinkHintsMode {
         top: 0,
         left: 0,
         position: "absolute",
+        display: 'block',
         width: "100%",
         height: "100%",
         overflow: "visible",


content_scripts/link_hints.js
```diff
@@ -419,6 +419,7 @@ class LinkHintsMode {
         top: 0,
         left: 0,
         position: "absolute",
+        display: 'block',
```
bent0b0x, line 422: Without the GHE styling interfering, this element otherwise has a computed display of `block`

soldonii:
I'm really looking forward to this PR be approved. Facing same issue using github enterprise. Great PR 👍🏻

philc:
Thank you @bent0b0x for the diagnosis and fix!

printercu:
If anybody is looking for a temporary fix, add this to styles in vimium config
```
 #vimiumHintMarkerContainer {
   display: block;
 }
```

PR #4461, opened 2024-04-13
Make `focusInput` focus all the input filelds in the whold page

### LINKED ISSUE #4046: Focus the first text input on the page - lack of scroll to active input field
If you type "gi" to focus the first text input on the page, in case scenario when the input field is not visible your screen at the moment of typing keystroke would be nice to implement automatic scroll to activated input filed position of a page.

In version 1.67  after type "gi" you must set scroll position manually (if input filed is outside the currently displayed user's view) or typing text into input field blindfold - you can't see what you typed.

YunfangHou:
## Description

close #4046 

This change makes `focusInput` focus all the input fields in the whold page (including the fields out of view).

## Reasons why applying this feature
1. For most webpages, there is a search bar at the top of the page, which is the most used input field. However, at the time when we want to search something else, we are very likely have been at the bottom of the page, and thus present *foucs input* function can not be used because the search bar is at top.
2. For most webpages, there are not much input fields. So focusing all input fields will not spend much time of user to switch to the aimed input field.

### DIFF
--- a/content_scripts/mode_normal.js
+++ b/content_scripts/mode_normal.js
@@ -256,7 +256,7 @@ const NormalModeCommands = {
 
     for (let i = 0, end = resultSet.snapshotLength; i < end; i++) {
       element = resultSet.snapshotItem(i);
-      if (!element.getClientRects(element, true)) {
+      if (!element.getClientRects()[0]) {
         continue;
       }
       visibleInputs.push({ element, index: i, rect: Rect.copy(element.getBoundingClientRect()) });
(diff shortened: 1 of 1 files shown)


kailowbi:
I have tested this feature.

https://github.com/user-attachments/assets/2ee8235e-3819-4b5d-951a-b9244c21382f

This improvement resolves the limitation where focus was only possible on text fields visible on the screen.
By removing the restrictions on focusInput, it becomes a more practical and user-friendly feature.
I would appreciate if this could be merged.

PR #4463, opened 2024-04-25
Select next scrollable

tumenbaev:
## Description

This adds a feature to select the next scrollable div. It's useful to scroll sidebars without focusing it manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,8 @@
 
 2.0.1 (2023-10-04)
 
-- Fix exception when migrating some pre-v2.0 settings. ([#4323](https://github.com/philc/vimium/issues/4323))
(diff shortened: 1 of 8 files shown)


tumenbaev:
Hey @philc what do you think about this MR? Does this feature sound like it's worth adding?

VincentBarkman (changes requested):
Here are a few concerns about the `nextScrollable()` function:

- **Performance**: Using `document.querySelectorAll("*")` may degrade performance on complex web pages.
- **Error Handling**: Absence of null checks for `activatedElement` could cause runtime errors.
- **Global Variable Use**: The global handling of `activatedElement` might lead to bugs and reduce reusability.
- **Accessibility Concerns**: The function could interfere with built-in accessibility tools.

*/Vincent A. Barkman*

PR #4467, opened 2024-05-07
fix: browser info loading condition

MonsterPi13:


philc:
Thank you; nice catch!

Issue #3593, opened 2020-06-08
J and K do not work well together with Panorama Tab Groups extension (hidden tabs)


silviubogan:
**Describe the bug**
J and K do not work well together with [this extension](https://github.com/projectdelphai/panorama-tab-groups).

Taken from the description of Vimium:
```
J, gT go one tab left
K, gt go one tab right
```

**To Reproduce**
1. Install the [Panorama Tab Groups](https://github.com/projectdelphai/panorama-tab-groups) extension.
2. Make 2 tab groups with new tabs created one after another in random groups.
3. Switch to a group.
4. Try to use J and K keys to go through the list of tabs.
5. It goes through the hidden tabs ignoring the separation between tab groups.

**Browser and Vimium version**
Firefox: 77.0.1
SO: Linux 5.4.0-33-generic (Ubuntu 20.04)
Vimium: 1.66

Thank you.

gdh1995:
Confirmed. This is because in Firefox [a tab has a `hidden` status](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/Tab), which doesn't exist on Chrome, and Vimium haven't support it.

gkaklas:
I believe this is a duplicate of #2998 :thinking:

010penetrator:
I have the same problem. Have to revert to hitting <C-Tab> and <C-S-Tab>

philc:
Closing as a duplicate of #2998.

Issue #3180, opened 2018-11-06
Do not select hidden tab when moving to adjacent tab


gagarine:
Using the shortcut `J, gT` or `K, gt ` let you select the right or left tab of the current tab.

Firefox web extension API let extension hide tab (see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/hide) . For example with https://addons.mozilla.org/en-US/firefox/addon/simple-tab-groups/ you can create tab group, only the tab from the active group are visible the other are hidden.

But vimium seem to select those tab, instead of moving to the adjacent tab, it select the hidden tab an activate the group of this tab hiding all my other tab. Using simple tab groups make this feature of vimium pretty much unusable.

Vimium should not select hidden tabs and only move trough active tabs.

0xf0cd:
Just discovered this extension the other day, but **this** is the only thing annoying me so far. I use firefox and simple-tab-groups and vimium's _go one tab left/right_ doesn't work correctly, it jumps around tab groups. Guessing it could be fixed if go one tab left/right functioned like firefox cycle tabs ```CTRL + [SHIFT] + TAB```

gdh1995:
Hello, you may try my customized version of Vimium, named Vimium C, and I've fixed quite a few issues on Firefox.

philc:
This is a duplicate of #2998.

Issue #2998, opened 2018-03-27
[Firefox] Play nicely with hidden tabs


donbex:
Firefox version: 59.0.1
Vimium version: 1.63.2

I'm organising my tabs in different [containers](https://support.mozilla.org/en-US/kb/containers) and I use Conex to keep visible only one at a time — it hides the other tabs through an API introduced in Firefox 59.

The problem is that changing tab with `gt` and `gT` crosses the container boundary. To give an example, suppose that I have two containers, C1 and C2; then I open a tab **t1** at the end of C1, then switch to C2 and open a tab **t2** at the end of C2, then switch back to C1 and open **t3** at then end of C1.

Now, if **t1** is selected and I press `gt` I would expect to select **t3**, while instead Vimium switches to C2 and selects **t2**. Note that switching tab with `ctrl + pgdown` (the default Firefox shortcut) works as expected.

    C1 (visible):    t1      t3
    C2 (hidden):         t2

I'm not sure what you would consider the appropriate behaviour here, but it would be nice to at least have an option to limit tab switching to visible tabs (which I think is the behaviour someone would expect). It would also be nice to have an option/command to switch to the next tab and open a new tab in the current container, even when the others aren't hidden.

smblott-github:
I don't think there's an [API](https://developer.chrome.com/extensions/tabs#type-Tab) for containers.  We'd need that to look at playing nice.

smblott-github:
On further investigation, it looks like this probably isn't to do with containers, but rather [tab hiding](https://wiki.mozilla.org/WebExtensions/TabHiding).  Tab hiding looks to be pretty experimental.  The one thing we would need...

![image](https://user-images.githubusercontent.com/2641335/38021552-12dfd6b4-3275-11e8-9d16-b6bf4b746c7d.png)

... doesn't appear to be implemented.

It looks like we cannot deal with this until they agree on and publish an API.

donbex:
I'm not sure if one can directly use this with `tabs.query()`, but the `tabs.Tabs` type [has the (optional) attribute](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/Tab) `hidden` for hidden tabs. Would something like this work?

```coffeescript
selectTab = (direction, {count, tab}) ->
  chrome.tabs.query { currentWindow: true, hidden: false }, (tabs) ->
  [...]
```

Or maybe does CoffeeScript offer a way to lazily filter the output of `tabs.query()`?

[...]

PR #4477, opened 2024-05-22
Correctly handle Firefox hidden tabs.

UncleSnail:
## Description

This PR solves issues  #2998, #3180, and #3593 by correctly handling Firefox hidden tabs. In Firefox there may be many hidden tabs between visible tabs in the main interface. The expected behavior for these tabs is that they would not be considered while performing operations. This PR makes Vimium behave the same as Firefox's default functionality for switching tabs (ctrl+tab) so that it will *feel more native* for Firefox users. Without these fixes, hidden tabs make Vimium almost unusable in Firefox.

For example, many add-ons hide tabs to allow for quick context switching. This is the result of console logging my current visible tabs:

```
0: Object { id: 3, index: 0, windowId: 1, … }
1: Object { id: 4, index: 1, windowId: 1, … }
2: Object { id: 5, index: 2, windowId: 1, … }
3: Object { id: 6, index: 3, windowId: 1, … }
4: Object { id: 7, index: 4, windowId: 1, … }
5: Object { id: 459, index: 5, windowId: 1, … }
6: Object { id: 458, index: 6, windowId: 1, … }
7: Object { id: 453, index: 7, windowId: 1, … }
8: Object { id: 455, index: 8, windowId: 1, … }
9: Object { id: 457, index: 9, windowId: 1, … }
10: Object { id: 468, index: 10, windowId: 1, … }
11: Object { id: 462, index: 11, windowId: 1, … }
12: Object { id: 464, index: 12, windowId: 1, … }
13: Object { id: 465, index: 13, windowId: 1, … }
14: Object { id: 466, index: 14, windowId: 1, … }
15: Object { id: 463, index: 15, windowId: 1, … }
16: Object { id: 2, index: 430, windowId: 1, … }
17: Object { id: 422, index: 431, windowId: 1, … }
18: Object { id: 423, index: 432, windowId: 1, … }
19: Object { id: 478, index: 433, windowId: 1, … }
20: Object { id: 451, index: 457, windowId: 1, … }
21: Object { id: 452, index: 458, windowId: 1, … }
22: Object { id: 490, index: 459, windowId: 1, … }
23: Object { id: 492, index: 460, windowId: 1, … }
```

Neither the id nor the index is directly related to the tab's "real" position. If I am in tab id 463 and send a command to select `nextTab`, I would expect the tab with id 2 to become selected.

## Implementation

I tried to implement these changes to:

1. Have the least impact on Chrome users.
2. Use sensible, easy code.
3. Cover all Firefox use cases.
4. Impact performance as little as possible.

I did this by adding two utility functions to the existing `bg_utils.js` file which can be used in all existing and future functions. These two functions are both implemented so that on Chrome and in basic cases (if the users has no hidden tabs), the functionality of Vimium will not change. These two functions are then used in all the necessary places.

In the case that we must find the index of a tab within the visible tab array, we must perform a search over the visible tabs. Since the indices are always ordered, I implemented a binary search to improve performance when many tabs are open. This makes the performance **significantly** better than using `findIndex` when the user has many tabs open. Even if the user has 100,000 tabs, the search will take at most 17 iterations, so it should not significantly impact performance.

The operations that I implemented were:

* remove tabs
* select tabs
* move tabs

I have also tried to update all relevant functions so that all operations will work in the future.

## Testing

I have manually tested all of this functionality with hidden tabs and it appears to work correctly. This includes moving tabs to new windows and removing tabs relatively. All manual tests were also done with counts, such as "5 >>" across hidden tab gaps.

I ran the Chrome unit tests and they still pass.

## Conclusion

If you see anything that is wrong or have any improvements to my code or method of implementation, please make a suggestion or fix the issue. I have been using my own fork with a few of these changes for a few years, but since some more people have been asking for this feature I wanted to clean up a good implementation and help other Firefox users be able to use Vimium with other tab-hiding extensions.

### DIFF
--- a/background_scripts/bg_utils.js
+++ b/background_scripts/bg_utils.js
@@ -17,7 +17,7 @@ const BgUtils = {
   // Get a query dictionary for `chrome.tabs.query` that will only return the visible tabs.
   visibleTabs() {
     const visibleTabsQuery = {
-      currentWindow: true
+      currentWindow: true,
     };
     // Only Firefox supports hidden tabs
     if (this.isFirefox()) {
@@ -36,11 +36,11 @@ const BgUtils = {
       // Since we know that all indices are in order, we can do a binary search.
       let l = 0;
       let r = tabs.length - 1;
-      while(l <= r) {
+      while (l <= r) {
         let m = (l + r) >> 1;
         if (tabs[m].index < tab.index) {
           l = m + 1;
-        } else if(tabs[m].index > tab.index) {
+        } else if (tabs[m].index > tab.index) {
           r = m - 1;
         } else {
           return m;
--- a/background_scripts/bg_utils.js
+++ b/background_scripts/bg_utils.js
@@ -14,38 +14,16 @@ const BgUtils = {
       : false;
   },
 
-  // Get a query dictionary for `chrome.tabs.query` that will only return the visible tabs.
-  visibleTabs() {
-    const visibleTabsQuery = {
-      currentWindow: true,
-    };
-    // Only Firefox supports hidden tabs
-    if (this.isFirefox()) {
-      visibleTabsQuery.hidden = false;
-    }
-    return visibleTabsQuery;
-  },
-
   // Find the real tab index in a given tab array.
+  // In Firefox there may be hidden tabs,
+  // so `tab.index` may not be the index in the visible tabs array.
   tabIndex(tab, tabs) {
     // First check if the tab is where we expect it (to save processing).
     if (tabs.length > tab.index && tabs[tab.index].index === tab.index) {
       return tab.index;
     } else {
       // If it's not where we expect, find its real position.
-      // Since we know that all indices are in order, we can do a binary search.
-      let l = 0;
-      let r = tabs.length - 1;
-      while (l <= r) {
-        let m = (l + r) >> 1;
-        if (tabs[m].index < tab.index) {
-          l = m + 1;
-        } else if (tabs[m].index > tab.index) {
-          r = m - 1;
-        } else {
-          return m;
-        }
-      }
+      return tabs.findIndex((t) => t.index === tab.index);
     }
   },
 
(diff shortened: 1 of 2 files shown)


philc:
Hey UncleSnail, this is a great fix. I don't use hidden tabs myself but I can see that this will be appreciated by users who do.

I'll leave some comments on the code and then we can merge it.

background_scripts/bg_utils.js
```diff
@@ -14,6 +14,41 @@ const BgUtils = {
...
+    };
+    // Only Firefox supports hidden tabs
+    if (this.isFirefox()) {
+      visibleTabsQuery.hidden = false;
+    }
+    return visibleTabsQuery;
+  },
+
+  // Find the real tab index in a given tab array.
+  tabIndex(tab, tabs) {
```
philc, line 30: We should document why we need to do this. E.g. "in the case of hidden tabs in Firefox browsers, tab.index may not correspond to the actual index in the visible tabs returned by tabs.query." Otherwise, in the code, it's pretty implicit that our tab logic needs to handle hidden tabs, and this will probably get regressed over time. Let's note it explicitly.

Also can callers use `tabs.findIndex` (or `Array.from(tabs).findIndex` if necessary, although that incurs a full array copy) instead of having this separate utility method?

background_scripts/bg_utils.js
```diff
@@ -14,6 +14,41 @@ const BgUtils = {
       : false;
   },
 
+  // Get a query dictionary for `chrome.tabs.query` that will only return the visible tabs.
+  visibleTabs() {
```
philc, line 18: Since this object isn't yet used anywhere other than background_scripts/main.js, I think it should live in there. We can declare it when the file is initialized:

```javascript
const visibleTabsQueryArgs = { currentWindow: true };
if (BgUtils.isFirefox()) {
  // Only Firefox supports hidden tabs.
  visibleTabsQueryArgs.hidden = false;
}
```

philc:
Nice work. Thanks for your contribution!

Issue #4161, opened 2022-11-04
[Feature request] Map zoom in/out


alexcg1:
It would be great to have keys for zooming. I previously used [Vimvixen](https://github.com/ueokande/vim-vixen) where it's mapped to:

`zi` - zoom in
`zo` - zoom out
`zz` - zoom reset

gdh1995:
Duplicate of #1866 .

PR #4488, opened 2024-06-05
Zoom Commands

UncleSnail:
## Description

Resolves the issues with [chjj's](https://github.com/chjj) original [PR](https://github.com/philc/vimium/pull/4367) for zoom commands.

This implementation fixes the following issues:

- Uses Chrome's zoom levels instead of a set increment.
- Resets to the user's default zoom level instead of always 1.
- Uses promises instead of callbacks.

It also uses a more similar code style to that of this extension. As expected, I have tested this manually in both Chrome and Firefox, ran the unit tests, and ran the formatter. If there are any suggestions for better implementations or code style, let me know.

Note that it would be very easy to add in an array of the Firefox default zoom levels and switch between the two based on the current browser. This might make the functionality more expected for Firefox users. If you would like me to do this, just let me know. I didn't do it here yet since in the original PR you mentioned to just use Chrome's values for now.

From the original PR:

> Implements zoom commands in a manner similar to cVim (which I recently migrated away from as it seems unmaintained).
> 
> The three new commands are zoomIn, zoomOut, and zoomReset. These are combined with a configurable zoomStep option which defaults to 0.10 (10%).
> 
> The default key mappings for the above commands are zi, zo, and z0 (cVim defaults) -- may be worth bikeshedding over.
> 
> Fixes https://github.com/philc/vimium/issues/1866 and https://github.com/philc/vimium/issues/4161 (also see https://github.com/philc/vimium/pull/2978).

### DIFF
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -351,6 +351,9 @@ const Commands = {
       "closeOtherTabs",
       "moveTabLeft",
       "moveTabRight",
+      "zoomIn",
+      "zoomOut",
+      "zoomReset",
     ],
     misc: ["showHelp", "toggleViewSource"],
   },
@@ -383,6 +386,9 @@ const Commands = {
     "toggleViewSource",
     "passNextKey",
     "hardReload",
+    "zoomIn",
+    "zoomOut",
+    "zoomReset",
   ],
 };
 
@@ -457,6 +463,9 @@ const defaultKeyMappings = {
   "X": "restoreTab",
   "<a-p>": "togglePinTab",
   "<a-m>": "toggleMuteTab",
+  "zi": "zoomIn",
+  "zo": "zoomOut",
+  "z0": "zoomReset",
 
   // Marks
   "m": "Marks.activateCreateMode",
@@ -550,6 +559,10 @@ const commandDescriptions = {
 
   moveTabLeft: ["Move tab to the left", { background: true }],
   moveTabRight: ["Move tab to the right", { background: true }],
+
+  zoomIn: ["Increase zoom on current tab", { background: true }],
+  zoomOut: ["Decrease zoom on current tab", { background: true }],
+  zoomReset: ["Reset zoom on current tab", { background: true }],
 
   "Vomnibar.activate": ["Open URL, bookmark or history entry", { topFrame: true }],
   "Vomnibar.activateInNewTab": ["Open URL, bookmark or history entry in a new tab", {
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -562,10 +562,10 @@ const commandDescriptions = {
   moveTabLeft: ["Move tab to the left", { background: true }],
   moveTabRight: ["Move tab to the right", { background: true }],
 
-  setZoom: ["Set zoom to the value given after the command, like setZoom 2.25", { background: true }],
-  zoomIn: ["Increase zoom on current tab", { background: true }],
-  zoomOut: ["Decrease zoom on current tab", { background: true }],
-  zoomReset: ["Reset zoom on current tab", { background: true }],
+  setZoom: ["Set zoom level to a given value. E.g. map zz setZoom 1.5", { background: true }],
+  zoomIn: ["Increase zoom", { background: true }],
+  zoomOut: ["Decrease zoom", { background: true }],
+  zoomReset: ["Reset zoom", { background: true }],
 
   "Vomnibar.activate": ["Open URL, bookmark or history entry", { topFrame: true }],
   "Vomnibar.activateInNewTab": ["Open URL, bookmark or history entry in a new tab", {
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -351,6 +351,7 @@ const Commands = {
       "closeOtherTabs",
       "moveTabLeft",
       "moveTabRight",
+      "setZoom",
       "zoomIn",
       "zoomOut",
       "zoomReset",
@@ -386,6 +387,7 @@ const Commands = {
     "toggleViewSource",
     "passNextKey",
     "hardReload",
+    "setZoom",
     "zoomIn",
     "zoomOut",
     "zoomReset",
@@ -560,6 +562,7 @@ const commandDescriptions = {
   moveTabLeft: ["Move tab to the left", { background: true }],
   moveTabRight: ["Move tab to the right", { background: true }],
 
+  setZoom: ["Set zoom to the value given after the command, like setZoom 2.25", { background: true }],
   zoomIn: ["Increase zoom on current tab", { background: true }],
   zoomOut: ["Decrease zoom on current tab", { background: true }],
   zoomReset: ["Reset zoom on current tab", { background: true }],
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -182,6 +182,14 @@ const mkRepeatCommand = (command) => (function (request) {
   }
 });
 
+const setZoom = (tabId, callback) => {
+  chrome.tabs.getZoom(tabId, (factor) => {
+    chrome.tabs.setZoomSettings(tabId, { scope: "per-tab" }, () => {
+      chrome.tabs.setZoom(tabId, callback(factor));
+    });
+  });
+};
+
 // These are commands which are bound to keystrokes which must be handled by the background page.
 // They are mapped in commands.js.
 const BackgroundCommands = {
@@ -294,6 +302,17 @@ const BackgroundCommands = {
   toggleMuteTab,
   moveTabLeft: moveTab,
   moveTabRight: moveTab,
+  zoomIn({ tabId, count }) {
+    const step = Settings.get("zoomStep");
+    setZoom(tabId, (factor) => factor + step * count);
+  },
+  zoomOut({ tabId, count }) {
+    const step = Settings.get("zoomStep");
+    setZoom(tabId, (factor) => factor - step * count);
+  },
+  zoomReset({ tabId }) {
+    setZoom(tabId, () => 1.0);
+  },
 
   async nextFrame({ count, tabId }) {
     // We're assuming that these frames are returned in the order that they appear on the page. This
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -183,7 +183,6 @@ const mkRepeatCommand = (command) => (function (request) {
 });
 
 async function setZoom(direction, { count, tabId }) {
-  console.log(direction, count, tabId);
   const zoomLevels = [
     0.25,
     0.33,
@@ -207,12 +206,12 @@ async function setZoom(direction, { count, tabId }) {
   switch (direction) {
     case "in":
       // Round down to the nearest zoom index.
-      const floorIndex = zoomLevels.findIndex(level => level > currentZoom) - 1;
+      const floorIndex = zoomLevels.findIndex((level) => level > currentZoom) - 1;
       newZoom = zoomLevels[Math.min(zoomLevels.length - 1, floorIndex + count)];
       break;
     case "out":
       // Round up to the nearest zoom index.
-      const ceilIndex = zoomLevels.findIndex(level => level >= currentZoom);
+      const ceilIndex = zoomLevels.findIndex((level) => level >= currentZoom);
       newZoom = zoomLevels[Math.max(0, ceilIndex - count)];
       break;
     case "reset":
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -326,7 +326,6 @@ const BackgroundCommands = {
   async setZoom({ tabId, registryEntry }) {
     const zoomLevel = registryEntry.optionList[0] ?? 1;
     const newZoom = parseFloat(zoomLevel);
-    // Chrome breaks on values below 0.01 and 10 or above.
     if (!isNaN(newZoom)) {
       chrome.tabs.setZoom(tabId, newZoom);
     }
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -182,13 +182,46 @@ const mkRepeatCommand = (command) => (function (request) {
   }
 });
 
-const setZoom = (tabId, callback) => {
-  chrome.tabs.getZoom(tabId, (factor) => {
-    chrome.tabs.setZoomSettings(tabId, { scope: "per-tab" }, () => {
-      chrome.tabs.setZoom(tabId, callback(factor));
-    });
-  });
-};
+async function setZoom(direction, { count, tabId }) {
+  console.log(direction, count, tabId);
+  const zoomLevels = [
+    0.25,
+    0.33,
+    0.50,
+    0.75,
+    0.80,
+    0.90,
+    1.00,
+    1.10,
+    1.25,
+    1.50,
+    1.75,
+    2.00,
+    2.50,
+    3.00,
+    4.00,
+    5.00,
+  ];
+  const currentZoom = await chrome.tabs.getZoom(tabId);
+  let newZoom;
+  switch (direction) {
+    case "in":
+      // Round down to the nearest zoom index.
+      const floorIndex = zoomLevels.findIndex(level => level > currentZoom) - 1;
+      newZoom = zoomLevels[Math.min(zoomLevels.length - 1, floorIndex + count)];
+      break;
+    case "out":
+      // Round up to the nearest zoom index.
+      const ceilIndex = zoomLevels.findIndex(level => level >= currentZoom);
+      newZoom = zoomLevels[Math.max(0, ceilIndex - count)];
+      break;
+    case "reset":
+      const zoomSettings = await chrome.tabs.getZoomSettings(tabId);
+      newZoom = zoomSettings?.defaultZoomFactor ?? 1.00;
+      break;
+  }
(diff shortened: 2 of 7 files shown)


UncleSnail:
Continuing the discussion about possible other default keyboard shortcut options, a friend recommended using "zz" as reset zoom, which is easier to press and has a president from some other programs like the open source Zathura document viewer (note that by default, Zathura supports both zz and z0). It also follows the pattern of double pressing the same key having a function. We can consider this, although z0 might be easier to remember for the general public.

background_scripts/main.js
```diff
@@ -169,6 +169,46 @@ const mkRepeatCommand = (command) => (function (request) {
...
+    3.00,
+    4.00,
+    5.00,
+  ];
+  const currentZoom = await chrome.tabs.getZoom(tabId);
+  let newZoom;
+  switch (direction) {
+    case "in":
+      // Round down to the nearest zoom index.
+      const floorIndex = zoomLevels.findIndex((level) => level > currentZoom) - 1;
```
UncleSnail, line 196: These lines are this way so that it handles zooming from between values correctly. For example, if we are at 225% and zoom in, we should go to 250%. Or if we zoom out, we should go to 200%. But, we also want, if we are at 200% and zoom out to go to 175% and in to go to 250%.
This implementation does that and seems reasonable, but if you have a more understandable way to achieve this, please suggest it.

background_scripts/main.js
```diff
@@ -169,6 +169,46 @@ const mkRepeatCommand = (command) => (function (request) {
   }
 });
 
+async function setZoom(direction, { count, tabId }) {
```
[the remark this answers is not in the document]
UncleSnail: That sounds good. I was originally modeling this style off of the `selectTab` function which has a similar method of picking next/previous/first/last tabs, but you are correct about unit tests so your way is better. I will make an implementation and unit tests to cover all the edge cases.

background_scripts/main.js
```diff
@@ -169,6 +169,46 @@ const mkRepeatCommand = (command) => (function (request) {
...
+  const zoomLevels = [
```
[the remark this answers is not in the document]
UncleSnail: Did we want to also use Firefox values on Firefox? If you don't have a problem with it, I think I will add correct Chrome/Firefox zoom level switching, and account for this in the unit tests.

UncleSnail:
Okay @philc, I implemented all the requested changes *and I found some more things that you should look at*.

1. I refactored the zoom commands to use **`nextZoomLevel`**.
2. I added **unit tests** to `main_test.js` in the `Next zoom level` context. Please let me know if there was a different place you wanted these.
3. I implemented the correct **zoom levels for Firefox** so that it matches Firefox users' expectations.
4. I performed much more thorough **manual testing in Chrome**, finding a few Chrome specific bugs.
5. Chrome allows having **zoom set to below and above the maximum**, so I fixed the code to properly handle zooming from these cases and added unit tests for this.
6. Chrome returns a zoom value with **floating point errors** in some cases, like 33% and 175%, making it impossible to zoom in past these values. I fixed the code to handle this and added unit tests.
7. Added **`setZoom value` command**. Now that we properly handle the edge cases in Chrome and Firefox, it is useful to have a way to let users set their own zoom values, allowing them to, for example, set z1, z2, z3, z4... to their preferred values. This is a feature that I personally want along with the zoom implementation and to me it made sense to round out the zoom feature set. If you would prefer it be in a different PR, or you don't like the feature/implementation, it can easily be removed. Thank you for your patience in reviewing my features. The implementation is simple and I added checks to avoid abuse/breaking when reasonable. If we accept this feature, I will add an **FAQ about using it in the wiki**.
8. I realized that `tabs.setZoom` has a "hidden" built-in behavior to set the zoom level to that tab's default zoom level by setting zoom to 0, making the `resetZoom` implementation *simpler and more performant*.

That summarizes the changes.

As usual, I have tried to do extensive testing in both Firefox and Chrome (previously I clearly hadn't tested thoroughly enough in Chrome). The unit tests all pass. I have formatted the code with `deno fmt` (and only committed the changes it suggested to the files I worked on. Note that there are suggested changes in many other files).

background_scripts/main.js
```diff
@@ -868,6 +951,7 @@ Object.assign(globalThis, {
   HintCoordinator,
   BackgroundCommands,
   majorVersionHasIncreased,
+  nextZoomLevel,
```
UncleSnail, line 954: I saw that this was how a previous coder had exposed `majorVersonHasIncreased` for testing purposes, so I did the same for `nextZoomLevel`. If you have a better preferred method, I'd love to use it.

tests/unit_tests/main_test.js
```diff
@@ -40,6 +40,114 @@ context("HintCoordinator", () => {
...
+  });
+
+  should("Zoom out from below the minimum", async () => {
+    const count = -1;
+    const currentZoom = 0.01; // lowest non-broken Chrome zoom level
+    const nextZoom = await nextZoomLevel(currentZoom, count);
+    assert.equal(0.25, nextZoom);
+  });
+
+  should("Test Chrome 33% zoom in with float error", async () => {
```
UncleSnail, line 136: These two floating point error tests only apply to Chrome. Firefox does not seem to have the floating point error issues.

tests/unit_tests/main_test.js
```diff
@@ -40,6 +40,114 @@ context("HintCoordinator", () => {
...
+  });
+
+  should("Zoom out past the minimum", async () => {
+    const count = -15;
+    const currentZoom = 1.00;
+    const nextZoom = await nextZoomLevel(currentZoom, count);
+    assert.equal(0.25, nextZoom);
+  });
+
+  should("Zoom in from below the minimum", async () => {
```
UncleSnail, line 108: These next four tests ensure zooming works when we zoom from outside of Chromes min/max zoom range, because Chrome (not Firefox) allows the zoom number to be outside the default accepted range. Supporting this should help avoid issues with other Chromium based browsers that allow different zoom values.

tests/unit_tests/main_test.js
```diff
@@ -40,6 +40,114 @@ context("HintCoordinator", () => {
...
+  });
+
+  should("Zoom out from between values", async () => {
+    const count = -1;
+    const currentZoom = 1.05;
+    const nextZoom = await nextZoomLevel(currentZoom, count);
+    assert.equal(1.00, nextZoom);
+  });
+
+  should("Zoom in past the maximum", async () => {
```
UncleSnail, line 94: Zooming should stop on the min/max boundary.

tests/unit_tests/main_test.js
```diff
@@ -40,6 +40,114 @@ context("HintCoordinator", () => {
...
+  });
+
+  should("Zoom out 2", async () => {
+    const count = -2;
+    const currentZoom = 1.00;
+    const nextZoom = await nextZoomLevel(currentZoom, count);
+    assert.equal(0.80, nextZoom);
+  });
+
+  should("Zoom in from between values", async () => {
```
UncleSnail, line 80: Zooming from between values should zoom up/down to the nearest value, not skipping.

tests/unit_tests/main_test.js
```diff
@@ -40,6 +40,114 @@ context("HintCoordinator", () => {
   });
 });
 
+context("Next zoom level", () => {
+  // NOTE: All these tests use the Chrome zoom levels, which are the default!
+  should("Zoom in 0 times", async () => {
```
UncleSnail, line 45: The first 5 tests are for basic zooming with counts in the normal "happy" cases.

background_scripts/main.js
```diff
@@ -295,6 +356,28 @@ const BackgroundCommands = {
   moveTabLeft: moveTab,
   moveTabRight: moveTab,
 
+  async setZoom({ tabId, registryEntry }) {
+    const zoomLevel = registryEntry.optionList[0] ?? 1;
+    const newZoom = parseFloat(zoomLevel);
+    // Chrome breaks on values below 0.01 and 10 or above.
```
UncleSnail, line 362: Note that Firefox just prints an error to the console when we try to zoom to a value outside of its range (0.3 to 5.0). Everything keeps working fine, and Firefox simply ignores the zoom. We can add a special case to change our handling of this if you like.

UncleSnail, line 362: Chrome allows setting the zoom between 1% and 999%, but it will not actually change the visual size of the page beyond 0.25 and 5.00. I decided to allow setting to any value Chrome allows, which seems the most sensible and should help with support for other Chrome based browsers.

background_scripts/main.js
```diff
@@ -295,6 +356,28 @@ const BackgroundCommands = {
...
+    if (!isNaN(newZoom) && newZoom >= 0.01 && newZoom < 10) {
+      chrome.tabs.setZoom(tabId, newZoom);
```
UncleSnail, line 364: Only try a zoom if we found a reasonable number.
We could also support parsing percents if you would like, so that users who try to enter:
`map z1 setZoom 10%`
will still have their config parse. That would not be difficult, but I think it's probably the right option might be to only have one correct method to set values.

background_scripts/main.js
```diff
@@ -182,6 +182,67 @@ const mkRepeatCommand = (command) => (function (request) {
...
+  let zoomLevels = chromeLevels; // Chrome by default
+  if (BgUtils.isFirefox()) {
+    zoomLevels = firefoxLevels;
+  }
+
+  if (steps === 0) { // Nothing
+    return currentZoom;
+  } else if (steps > 0) { // In
+    // Chrome sometimes returns values with floating point errors.
+    currentZoom += 0.0000001; // This is needed to solve floating point bugs in Chrome.
```
UncleSnail, line 231: We need to see if adding a tiny value will boost the number up to a value in our list to cover floating point errors.

background_scripts/main.js
```diff
@@ -182,6 +182,67 @@ const mkRepeatCommand = (command) => (function (request) {
...
+    const nextIndex = zoomLevels.findIndex((level) => level > currentZoom);
+    // Properly handle index not found (we want not found to be after array, not -1).
+    const floorIndex = nextIndex < 0 ? zoomLevels.length : nextIndex - 1;
```
UncleSnail, line 234: If we didn't find the value in our list, IE, it was bigger than all of them, we want to say the index is past the end, not -1. This helps handle several edge cases.

background_scripts/commands.js
```diff
@@ -551,6 +562,11 @@ const commandDescriptions = {
   moveTabLeft: ["Move tab to the left", { background: true }],
   moveTabRight: ["Move tab to the right", { background: true }],
 
+  setZoom: ["Set zoom to the value given after the command, like setZoom 2.25", { background: true }],
```
UncleSnail, line 565: I wanted to add an example here to make the usage more clear, but if there is a better place for this or a better way to make the usage clear, we should remove it here.

UncleSnail (commented):
Here are a few code explanations and question/suggestions. They are all also mostly covered in the summary message.

background_scripts/main.js
```diff
@@ -182,6 +182,67 @@ const mkRepeatCommand = (command) => (function (request) {
   }
 });
 
+function nextZoomLevel(currentZoom, steps) {
+  const chromeLevels = [ // Chrome's default zoom levels.
```
philc, line 186: Let's put the comment above the variable declaration. Also, these zoom levels can be fit all on one line if you remove the trailing zeroes. The list becomes less readable, but the values are uninteresting, so putting them on separate lines is not worth the giving them 30 lines of this source file.

background_scripts/main.js
philc: What does floating point error mean here?  isNaN?

background_scripts/main.js
```diff
@@ -182,6 +182,67 @@ const mkRepeatCommand = (command) => (function (request) {
...
+    zoomLevels = firefoxLevels;
+  }
+
+  if (steps === 0) { // Nothing
+    return currentZoom;
+  } else if (steps > 0) { // In
+    // Chrome sometimes returns values with floating point errors.
+    currentZoom += 0.0000001; // This is needed to solve floating point bugs in Chrome.
+    const nextIndex = zoomLevels.findIndex((level) => level > currentZoom);
+    // Properly handle index not found (we want not found to be after array, not -1).
```
philc, line 233: I think this "properly handle" comment is unnecessary if you make the condition on the next line `nextIndex == -1` (`nextIndex < 0` is actually less clear/specific than `nextIndex == -1`), and removing it from both branches is a nice DRY cleanup.

background_scripts/main.js
philc: Agreed.

tests/unit_tests/main_test.js
```diff
@@ -40,6 +40,114 @@ context("HintCoordinator", () => {
...
+  should("Zoom out from above the maximum", async () => {
+    const count = -1;
+    const currentZoom = 9.99; // highest non-broken Chrome zoom level
+    const nextZoom = await nextZoomLevel(currentZoom, count);
+    assert.equal(5.00, nextZoom);
+  });
+
+  should("Zoom in from above the maximum", async () => {
+    const count = 1;
+    const currentZoom = 9.99; // highest non-broken Chrome zoom level
```
philc, line 124: What does non-broken mean? What happens, exactly, if it's > 9.99?

background_scripts/commands.js
philc: Assuming the zoom value must be given as part of the command definition, then documenting that here would be unusual, as this string is shown in the help dialog.

Where to document a command's options is an unsolved issue. I think the #2827 has the most substantial discussion about it. The [tips and tricks wiki page](https://github.com/philc/vimium/wiki/Tips-and-Tricks) contains various command options organized by use case.

For now, I'd suggest this docstring: "Set zoom level to a given value. E.g. `map zz setZoom 1.5`"

background_scripts/commands.js
```diff
@@ -551,6 +562,11 @@ const commandDescriptions = {
...
+  zoomIn: ["Increase zoom on current tab", { background: true }],
```
philc, line 566: I think we can drop "on current tab." Our zoom works the same as Chrome's and the Chrome UI doesn't make that distinction.

Although now that I think about it, in Chrome when changing the zoom of a site, Chrome will also update all open tabs which are showing that same site. I haven't yet tested this PR, but presumably our implementation doesn't do that. That would be a nice future change (outside of this PR) to make the behavior consistent.

philc (changes requested):
Looking good. I plan to merge this once you've had a chance to look at my comments.

background_scripts/main.js
UncleSnail: Good point! Originally I needed to catch both -1 and -2, but that issue should be fixed now because I moved the subtraction (now `nextIndex - 1`) to this line instead.

tests/unit_tests/main_test.js
UncleSnail: It still scales the page correctly to 500%, but it does not parse the number correctly and shows odd % numbers at the top of the page. Note that in the Chrome API documentation for setZoom it does not specify a range of values to be used. I determined the values it handles well from my own testing.
Firefox DOES specify that the value should be between 0.3 and 5, but it gracefully fails for numbers outside of this range. Chrome isn't quite so graceful. That's why I decided to add a limit where Chrome begins to act oddly, but didn't do anything for Firefox's range.

background_scripts/commands.js
UncleSnail: I agree, that wording was left from the original implementation, which tried to limit the zoom to the current tab. It is actually **incorrect** now, as I had *already changed the implementation to match the default behavior*. The behavior of all of these zoom commands should match the browser default, and it **does** zoom all pages of the current domain when you use `zi` or any other of these commands (I just tested in Chromium to make sure it is still working).

background_scripts/main.js
UncleSnail: I add an example into the comments to make this meaning clear. There are also unit tests to cover the known values for which Chrome does not return the precise number.
For further clarification, if we run `tabs.setZoom(tabId, 0.33)` Chrome will show 33% in the UI. However, when we run `tabs.getZoom(tabId)` Chrome will give us the number `0.32999999999999996`

tests/unit_tests/main_test.js
UncleSnail: Okay, I don't know why I thought it was parsing the number incorrectly.
I think it **does** work correctly for bigger values, it's just that there isn't enough space to show the number, so it shows "1,0..." because the "1,000" gets truncated. And it looks like for smaller values it just does nothing, similar to Firefox.
I must have been thinking incorrectly before.
So this bounds checking may be unnecessary. I am going to remove it.

philc (approved):
Nice work. Looks great. Let's ship it!

Could you rebase this onto master (it looks like there's a merge conflict) and I'll merge it in.

UncleSnail:
> Nice work. Looks great. Let's ship it!
> 
> Could you rebase this onto master (it looks like there's a merge conflict) and I'll merge it in.

That's odd, I merged master in and don't see any conflicts. I will try to perform a rebase though. Give me a few minutes to see what I can do.

UncleSnail:
Okay @philc I performed a rebase onto master and force pushed. If there are still issues, let me know.

PR #4493, opened 2024-06-09
Fix find mode not searching inside elements with `display` property set to `contents`

### LINKED ISSUE #4492: Find mode doesn't work inside elements with display property set as contents
**Describe the bug**

When using find mode (the "/" command) the search doesn't display any results inside elements with the
`display` property set to the value `contents`. I have checked the code and debugged and found that the
method `element.checkVisibility()` returns `false` when `display` property is `contents`, however, even
though these elements are hidden, its children are visible. This is extremely bad as I've noticed this
in a website with a `div` with style `display: contents;` as a parent to the whole page.

**To Reproduce**

1. Go to URL 'https://kit.svelte.dev/docs/creating-a-project'
2. Type "/cd\<enter\>"
3. Notice how the exact word "cd" inside the codeblock isn't being highlighted.

I'm using Brave version 1.66.118 with Chromium: 125.0.6422.147 (Official Version) 64 bits

I'm using Vimium version 2.1.2

victoragcosta:
## Description

Resolves #4492.

The [`checkVisibility` method](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility) considers elements with `display` set to `contents` as not visible in the page and it is true that the element is not visible, but its children are completely visible. So I simply made the node traversal algorithm also visit the children of elements with `node.style.display === 'contents'` and it worked.

### DIFF
--- a/content_scripts/mode_find.js
+++ b/content_scripts/mode_find.js
@@ -483,7 +483,10 @@ const getAllTextNodes = () => {
   function getAllTextNodes(node) {
     if (node.nodeType === Node.TEXT_NODE) {
       textNodes.push(node);
-    } else if (node.nodeType === Node.ELEMENT_NODE && node.checkVisibility()) {
+    } else if (
+      node.nodeType === Node.ELEMENT_NODE &&
+      (node.checkVisibility() || node.style.display === "contents")
+    ) {
       const children = node.childNodes;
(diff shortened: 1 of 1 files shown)


philc:
Yep, seems reasonable and correct. Thanks!

Issue #4476, opened 2024-05-22
Outdated Instructions for Installing From Source in CONTRIBUTING.md


moonmaster36:
**Describe the bug**

In the [Installing From Source](https://github.com/philc/vimium/blob/445486e20f410fcfbbe8b0a186242bc9f4368511/CONTRIBUTING.md#installing-from-source) instructions, I noticed that some of the instructions are outdated and unclear.

For example, the original Firefox instructions are to navigate to
about:debugging in your browser. However, the actual URL to navigate to is
about:debugging#/runtime/this-firefox. 

**To Reproduce**

1. Go to [Installing From Source](https://github.com/philc/vimium/blob/445486e20f410fcfbbe8b0a186242bc9f4368511/CONTRIBUTING.md#installing-from-source)
2. Navigate to the `On Firefox` section of Installing From Source
3. Enter `about:debugging` in the URL bar
4. Try to click `Load Temporary Add-on`, but there is no such button at `about:debugging`
![image](https://github.com/philc/vimium/assets/76980875/7543aa74-12ab-42b5-bf65-961f102f715b)
Instead, you will find the `Load Temporary Add-on` button after clicking on `This Firefox` on the left hand side.
![image](https://github.com/philc/vimium/assets/76980875/4225de14-27d8-48de-8383-2c41d5f3ce0b)

**Browser and Vimium version**

* Firefox
  * Firefox Version: 126.0
  * OS: Windows_NT 10.0 19045

**Browser and Vimium version**

* Firefox
  * Firefox Version: 126.0
  * OS: Windows_NT 10.0 19045

philc:
Thank you. Fixed in 76fea7addfb22be61f0fce80b65a8449f56ed501.

PR #4495, opened 2024-06-12
Moonmaster36 patch 1

moonmaster36:
### Rationale

* Current `Installing From Source` instructions were a little outdated.
* Made instructions more clear by adding screenshots
* Made instructions more clear by making it easy to copy/paste command line arguments.

### Corresponding Issue

* Corresponding Issue: https://github.com/philc/vimium/issues/4476


philc:
Hey @moonmaster36, thanks for putting in the effort to make these instructions clearer. However, I think this is an unnecessarily high level of handholding, and the screenshot approach doesn't age well as the browser UIs slowly evolve. My understanding is that most people install the extension from the stores, and those who want to install from source are able to figure out from the sparse instructions (including how to check out the repo from Github).

moonmaster36:
That is totally fine, I was just making the PR as part of a class
assignment.

PR #4505, opened 2024-06-23
add role="menu" to link hints

metiftikci:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -1111,6 +1111,7 @@ const LocalHints = {
         "tab",
         "link",
         "checkbox",
+        "menu",
         "menuitem",
         "menuitemcheckbox",
         "menuitemradio",


UncleSnail:
Should the menu itself be clickable, or only the menu items? It sounds like (from a quick skim) the "menu" role is supposed to be a container for menu items, and is not a clickable option itself. If you look at the [example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role#example_1_navigation_menu), the `role="menu"` element itself should not be clickable (ie, should not have link hints). There is a clickable button that opens the menu, and there are menu items in the menu that should be clickable, but the menu itself does not seem like it needs hints.
Correct me if I am misunderstanding.
Do you have an example website where this would fix the website's functionality? My guess is that some websites use the `role="menu"` incorrectly, but we do not want to try to cover "incorrect use" cases.

metiftikci:
@UncleSnail thanks for comment, then we can close this

PR #4506, opened 2024-06-24
horizontally resizable search engine textfield

davidk55:
## Description
It is still a bit wonky since everything will be moved along, but with the boundaries it should look okay.

## Rationale
URLs are often very long, and it is hard to edit them in a small textarea field.

### DIFF
--- a/pages/options.css
+++ b/pages/options.css
@@ -128,6 +128,11 @@ textarea#userDefinedLinkHintCss, textarea#keyMappings, textarea#searchEngines {
   min-height: 140px;
   white-space: pre;
 }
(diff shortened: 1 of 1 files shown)


philc:
Hey @davidk55, thanks for the PR and for bringing this to my attention. You're right, it is kind of ridiculous to edit URLs in such narrow, scrunched textareas.

I've done a small revamp on the options page in 61bcaeb44311870343d3c0c6abbd728c2b1d912d such that the whole page is now resizable (up to point) rather than fixed-width, and all of the textareas are now larger. Give it a try to see if it helps.

PR #4511, opened 2024-06-30
Icon maintenance 

leoheitmannruiz:
1. Remove three unused icons

2. Update the icon to https://github.com/vimium/vimium.github.com/blob/master/vimium_logo.svg
```
inkscape -w 16  -h 16  icon.svg -o icon16.png
inkscape -w 48  -h 48  icon.svg -o icon48.png
inkscape -w 128 -h 128 icon.svg -o icon128.png
inkscape -w 295 -h 295 icon.svg -o vimium.png

zopflipng -y icon16.png  icon16.png
zopflipng -y icon48.png  icon48.png
zopflipng -y icon128.png icon128.png
zopflipng -y vimium.png  vimium.png
```

Is vimium.png needed? Or can it be deleted and icon128.png used instead? Why is it 295px?
It's only used here:
```
./background_scripts/main.js:761:      iconUrl: chrome.runtime.getURL("icons/vimium.png"),
```

3. And a tid bit in make.js, so that Firefox uses the SVG directly.


philc:
Looks good, thank you @leoheitmannruiz. I'll test it soon and then merge.

philc:
I looked into vimium.png. It might've been used as a high-res icon for the chrome store at some point, but today, the store icon asset is supposed to be 128x128, so icon128.png can work for that, and for the notification use case. I've removed it in 7ff8cb6fa225134b9dfd189734c3db027b35be46.

PR #4518, opened 2024-07-16
Fix DRY and Styling Issues (hardReload)

UncleSnail:
## Description

Removes duplicate `hardReload` functionality. I decided to go with removing the `hardReload` function entirely and mapping `R` to `reload hard`. The other obvious option would be to keep the `hardReload` function and have it call `reload({count: count, tabId: tabId, registryEntry: {options: {hard: true}}})`. This fixes the one main downside of my chosen implementation: it means the key mapping appears like this: "`R`, `r`  Reload the page" which is not intuitive and does not make it clear that `R` is hard reload and `r` is regular reload.

Perhaps this calls for a change to how the commands are rendered in the help menu. If you would like, I can do that in this PR.

Since this is a refactor pull request, I used it to make two other changes that have been outstanding:

1. Rename `mkRepeateCommands` to `createRepeatCommands`.
2. Resolve the other `deno fmt` corrections.

I looked through the rest of the `background_scripts/main.js` file and there didn't seem to be any more obvious DRY violations or other easy refactors that should be done. If you have any more reactors you have been wanting done, let me know and I can add them to this PR.

The tests pass and with quick testing, the addon seems to still work correctly in Chromium and Firefox.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,8 @@
 
 2.0.1 (2023-10-04)
 
-- Fix exception when migrating some pre-v2.0 settings. ([#4323](https://github.com/philc/vimium/issues/4323))
+- Fix exception when migrating some pre-v2.0 settings.
+  ([#4323](https://github.com/philc/vimium/issues/4323))
 
 2.0.0 (2023-09-28 -- partially rolled out to users on the Chrome store)
 
@@ -62,14 +63,14 @@
 - Allow find mode to work when using only private windows.
   ([#3614](https://github.com/philc/vimium/issues/3614))
 - Add a count option to closeTabsOnLeft and closeTabsOnRight commands, to allow binding a key to
-  "close just 1 tab on the left/right" rather than closing all tabs, as is the default. E.g. `map cl
-  closeTabsOnLeft count=1`. ([#4296](https://github.com/philc/vimium/pull/4296))
+  "close just 1 tab on the left/right" rather than closing all tabs, as is the default. E.g.
+  `map cl closeTabsOnLeft count=1`. ([#4296](https://github.com/philc/vimium/pull/4296))
 - Add search completions for Brave Search. ([#3851](https://github.com/philc/vimium/pull/3851))
 - Make regular expressions in find mode work again; other find mode improvements.
   ([#4261](https://github.com/philc/vimium/issues/4261))
 - Bug fixes. ([#3944](https://github.com/philc/vimium/pull/3944),
-[#3752](https://github.com/philc/vimium/pull/3752),
-[#3675](https://github.com/philc/vimium/pull/3675))
+  [#3752](https://github.com/philc/vimium/pull/3752),
+  [#3675](https://github.com/philc/vimium/pull/3675))
 
 1.67.7 (2023-07-12)
 
--- a/README.md
+++ b/README.md
@@ -5,9 +5,11 @@ spirit of the Vim editor.
 
 **Installation instructions:**
 
-* Chrome: [Chrome web store](https://chrome.google.com/extensions/detail/dbepggeogbaibhgnhhndojpepiihcmeb)
-* Edge: [Edge Add-ons](https://microsoftedge.microsoft.com/addons/detail/vimium/djmieaghokpkpjfbpelnlkfgfjapaopa)
-* Firefox: [Firefox Add-ons](https://addons.mozilla.org/en-GB/firefox/addon/vimium-ff/)
+- Chrome:
+  [Chrome web store](https://chrome.google.com/extensions/detail/dbepggeogbaibhgnhhndojpepiihcmeb)
+- Edge:
+  [Edge Add-ons](https://microsoftedge.microsoft.com/addons/detail/vimium/djmieaghokpkpjfbpelnlkfgfjapaopa)
+- Firefox: [Firefox Add-ons](https://addons.mozilla.org/en-GB/firefox/addon/vimium-ff/)
 
 To install from source, see [here](CONTRIBUTING.md#installing-from-source).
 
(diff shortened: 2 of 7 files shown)


philc:
All of the non-reload changes are fine and can be merged; perhaps split the reload stuff out into a separate PR?

I agree that the UX of the help dialog is poor. I didn't realize that was an issue with the "reload hard" approach. If we maintain this approach, what is your suggested UX for the help dialog?

I'm surprised in commands.js specifying the command as "reload hard" actually works, and everything flows through Vimium's data structures correctly. This is the first example of that in the codebase. I'm not yet sure how I feel about encoding both the command name and its options as an unstructured string in the code, but it struck me as odd and I thought I'd mention the concern. I realize this is how the user's keymap configuration data is loaded and represented.

UncleSnail:
> I agree that the UX of the help dialog is poor. I didn't realize that was an issue with the "reload hard" approach. If we maintain this approach, what is your suggested UX for the help dialog?

Since it only requires minimal changes, I would just suggest separating out commands with options as different commands. This would also fix the UI for the new zoom commands.

So currently if we set commands for many zoom levels, like z1, z2, z3 etc, it shows up in the help menu as `z1, z2, z3, z4, z5, z6, z7, z8, z9 Set zoom level to a given value. E.g. map zz setZoom 1.5 `

I would suggest just changing it to include the command in some way, like:

`z1 setZoom 0.3 Set zoom level to a given value. E.g. map zz setZoom 1.5`
`z2 setZoom 0.75 Set zoom level to a given value. E.g. map zz setZoom 1.5`

This would be pretty easy and would make minimal changes to the existing experience. If the multiple keys are mapped to the exact same command, we could just add them with commas, and we wouldn't need to show the commands that run. The better but more difficult to implement and maintain option would be to change the help message based on the command and the params. This would also mean that we would have different messages in the help menu (sometimes) from what is in the settings "all commands" menu. So the messages would appear as:

`z1 Set zoom level to 0.3.`
`z2 Set zoom level to 0.75.`
`R Hard reload the page`

We would need to have a system to select which message should be shown. This could be implemented to also help us define the valid options for commands on the "all commands" help list. I would propose allowing arrays of options/message tuples in the `commandDescriptions` list. If there is no array of accepted options, or if none of the options match, then we display the default message. If there is an options array, the first match is selected. It would be like a switch statement. We would need to make a format very clear, but that shouldn't be hard. Here is an example of how that might look in the `commands.js` file.

```
  showHelp: ["Show help", { topFrame: true, noRepeat: true }], // A normal command with one help message.
  reload: ["Reload the page", { background: true, options: [["hard", "Hard reload the page"]] }], // One opt matches to the literal "hard"
  setZoom: ["Set zoom level to a given value", { background: true, options: [["$level:float", "Set zoom level to $level"]] }], // $level's value will replace the help string. Float will be used in the "all commands" help section that specifies available options.
  example: ["Multiple arg opts", { options: [["literal", "Match literal first"], ["$var:bool", "Match var to $var second"]] }], // We could allow for multiple types of options, or for commands with multiple args.
```

This method is pretty easy to understand and implement. It also helps solve the "where do we list available options" issue because we can easily extract the options and types from the options lists. In this example, I included a type for non-literal options, which I think would be helpful to have in an options list. Note that I chose to use a list of "tuples" (lists) here instead of a dictionary because we need to decide which option will be matched first somehow. Lists allow us a convenient way to do this. There are other ways, but this seems like the most straight forward to me from both understandability and coding.

We can implement this for both the help messages and the settings page command options by:

### Help Page:

```
If there are no options set on this shortcut's command, give the default message.
Otherwise, if there is an options list for the command, loop through it.
    If the option at `i` is a literal and it matches, put this help message.
    Otherwise, move on. If the option at `i` is variable (right now I think we only have floats and URLs) always match it (unless we want to do type matching, which is fine) and insert it into the help message as specified.
If we end the list and nothing matched, use the default message.
```

### Settings Page:
```
If there is no options list, add no options to the list of available commands.
If there is an options list, loop through it.
    Add literal options as they are.
    Add variable options, noting their type if specified.
```

Hopefully that proposal all makes sense and is stated clearly enough. I believe I could implement this pretty easily, it should have **no effect** on existing commands that we don't change. It solves the help messages problems (that multiple shortcuts doing different things because of different options are grouped together and have the same message, and that we don't have help messages for commands like `reset hard`), and gives us the framework to solve the "unknown options" issue.

philc:
Riffing on your proposal:

**Showing a command's arguments in the help UI**

We could show the arguments passed to the command, if present. This won't add much clutter, because none of the default key mappings provide arguments to their commands (except recently, for the reload hard command), so in the default configuration, the help dialog will look as it does today. If a person creates their own mapping to a command with arguments, then the arguments will be shown.

The help dialog text could look like this:

    t Create new tab
    ;t0 Create new tab (url=https://...)
    R Reload the page (hard)
    z2 Zoom the page (zoom=1.2)

One difficulty with showing the argument values in help dialog UI is that the UI has to be robust to strings of varying lengths. E.g. `createTab` can have a URL parameter which can be very long. The standard solution for this is to ellipsize aggressively, and show the full command string on hover.

Second, with this approach it's possible for the help dialog to have many more rows than it does currently. But I think this is already handled well enough: today if the help dialog is too tall, the page will get a scroll bar (not the dialog itself) and one can scroll the page to see the full help dialog.

**Parameterizing a command's help message**

That's a fancy proposal, but adds a lot of complexity, and doesn't add much UX value if we instead show the arguments passed to the command.

**Data model for a command's possible options**

If it's not strictly required for our UX work, then I think it's out of scope and should be tackled apart from this hard reload / UX dialog work.

However, I think it's very valuable to do (see also #2827). It's annoying that one can't easily trace which options a command accepts when browsing commands.js.

UncleSnail:
The code in this pull request and the suggested changes/new features have been broken (as requested by @philc) into two new PRs: #4554 and #4553 

This PR is no longer relevant (except for discussion reference) since all changes now live in other places.

PR #4519, opened 2024-07-17
Remove unused icons from build

leoheitmannruiz:
Now that the Firefox version uses SVGs directly, the PNGs can be excluded for that build (saves ~17K)

### DIFF
--- a/make.js
+++ b/make.js
@@ -125,7 +125,7 @@ async function buildStorePackage() {
 
   const firefoxManifest = createFirefoxManifest(chromeManifest);
   await writeDistManifest(firefoxManifest);
-  await shell("bash", ["-c", `${zipCommand} ../firefox/vimium-firefox-${version}.zip .`]);
+  await shell("bash", ["-c", `${zipCommand} ../firefox/vimium-firefox-${version}.zip . -x icons/*.png`]);
 
   // Build the Chrome Store package.
   await writeDistManifest(chromeManifest);


leoheitmannruiz:
I was wrong. The Firefox build does use one PNG icon, for the update notification on major releases:

```
background_scripts/main.js:828:      iconUrl: chrome.runtime.getURL("icons/icon128.png"),
```

I'll hold off on sending a patch as I'm unsure how to go about this. FWIW, I do enjoy the unneeded icons being excluded. The Firefox build lost around a third of its size between the 2.1.2 and 2.2.1 releases: https://addons.mozilla.org/addon/vimium-ff/versions/ 

Sorry for my oversight!

philc:
The ~90K decrease in file size from 2.1.2 to 2.2.1 is unlikely due to just the icons. The total size of all pngs in the icons folder is only 16k. Also, why do you care about a few kilobytes in an extension?

ISTM we should do either:

1. Include all pngs in the Firefox build by reverting 4b1491f, so we don't have a divergent build based on browser vendor, which can lead to subtle bugs like this one. Note that we can also further compress the png icons. I ran `pngquant 128 *.png` which reduced the total bytes from 16k to 8k, and I couldn't tell the difference between the originals and the quantized versions. 
2. Continue to exclude pngs from the Firefox build, and use icon.svg in the call to `chrome.notifications.create` in Firefox. I haven't tested it on any platform, but MDN says the URL in Firefox can be an SVG.

leoheitmannruiz:
> The ~90K decrease in file size from 2.1.2 to 2.2.1 is unlikely due to just the icons. The total size of all pngs in the icons folder is only 16k.

You're right, of course!

> Also, why do you care about a few kilobytes in an extension?

I'm not sure. I do appreciate a smaller file size but it is probably not really what I enjoy about excluding the unused icons. I think it is rather some feeling of "correctness".

> ISTM we should do either:
> 
> 1. Include all pngs in the Firefox build by reverting [4b1491f](https://github.com/philc/vimium/commit/4b1491f0172223b750f22fa9f251b789a9123138), so we don't have a divergent build based on browser vendor, which can lead to subtle bugs like this one. Note that we can also further compress the png icons. I ran `pngquant 128 *.png` which reduced the total bytes from 16k to 8k, and I couldn't tell the difference between the originals and the quantized versions.
> 
> 2. Continue to exclude pngs from the Firefox build, and use icon.svg in the call to `chrome.notifications.create` in Firefox. I haven't tested it on any platform, but MDN says the URL in Firefox can be an SVG.

It seems the second option would require adding height and width attributes to icon.svg, as these are currently not present: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/notifications/NotificationOptions#iconurl

I do prefer option two, though I'm hesitant to take part in this decision as I don't want to cause further troubles.

In case you decide to try using icon.svg I'm happy to help test.

PR #4524, opened 2024-07-29
Final icon maintenance

leoheitmannruiz:
I thought I'd finish #4511 off with updating the action icons :)

The V is the same as in https://github.com/philc/vimium/blob/master/icons/icon.svg. The blue is #3680f0, that is, the color in the middle of the Chromium logo. I found it a little strong at first, though started to enjoy it super quickly. Hope you like it! Though, of course, feel free to complain and I'll see what I can do :)

PS: I'm quite happy about how well the XML turned out hihi


philc:
The icons are looking good. Thanks @leoheitmannruiz !

PR #4533, opened 2024-08-11
✨ feat(completions): add tab favicon for better image quality

abelfubu:
## Description

image

As many of you know, a significant number of users apply custom CSS to enhance their Vimium experience. While creating my own custom styling, I encountered a limitation with the Vomnibar: the favicon is always fetched at a 16px size. This fixed size makes resizing the icon challenging, as it can result in a blurry appearance.

image 

This PR addresses this issue by fetching the tab's favicon URL directly when available, allowing for a higher-quality icon. If the favicon isn't available, the PR provides a fallback with an improved size, ensuring the icon looks sharp and visually appealing even when resized.

### DIFF
--- a/background_scripts/completion.js
+++ b/background_scripts/completion.js
@@ -47,6 +47,7 @@ class Suggestion {
   // The generated HTML string for showing this suggestion in the Vomnibar.
   html;
   searchUrl;
+  favIconUrl;
 
   constructor(options) {
     Object.seal(this);
@@ -77,8 +78,9 @@ class Suggestion {
     if (this.description === "tab" && !BgUtils.isFirefox()) {
       const faviconUrl = new URL(chrome.runtime.getURL("/_favicon/"));
       faviconUrl.searchParams.set("pageUrl", this.url);
-      faviconUrl.searchParams.set("size", "16");
-      faviconHtml = `<img class="vomnibarIcon" src="${faviconUrl.toString()}" />`;
+      faviconUrl.searchParams.set("size", "64");
+      const src = this.favIconUrl.startsWith("http") ? this.favIconUrl : faviconUrl.toString();
+      faviconHtml = `<img class="vomnibarIcon" src="${src}" />`;
     }
     if (this.isCustomSearch) {
       this.html = `\
@@ -489,6 +491,7 @@ class TabCompleter {
           queryTerms,
           description: "tab",
           url: tab.url,
+          favIconUrl: tab.favIconUrl,
           title: tab.title,
           tabId: tab.id,
           deDuplicate: false,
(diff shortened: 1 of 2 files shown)


UncleSnail:
I can see the benefit of this.

I might be misunderstanding the code, but it looks like you have pushed the problem back by setting a new size of 64px. Imagining into the future, is there a potential that someone else will make a PR to change the new 64px size to something higher? In other words, is 64px also just an arbitrary size? If so, maybe we can find a way to get the size the CSS is actually trying to use.

Also, it looks like these changes might mean a 64px image is always being pulled now, even for users who use a 16px thumbnail, which would slow down the experience of the majority of existing users.

I completely agree that we should fix this issue so that we can have larger icons appear sharp for the users who want it, but I'm wondering if I'm understanding the code correctly and if so, if we can improve some of the downsides.

Thank you for your contribution.

background_scripts/completion.js
```diff
@@ -77,8 +78,9 @@ class Suggestion {
     if (this.description === "tab" && !BgUtils.isFirefox()) {
       const faviconUrl = new URL(chrome.runtime.getURL("/_favicon/"));
       faviconUrl.searchParams.set("pageUrl", this.url);
-      faviconUrl.searchParams.set("size", "16");
-      faviconHtml = `<img class="vomnibarIcon" src="${faviconUrl.toString()}" />`;
+      faviconUrl.searchParams.set("size", "64");
+      const src = this.favIconUrl?.startsWith("http") ? this.favIconUrl : faviconUrl.toString();
```
philc, line 82: If `this.faviconUrl` is valid and present, then for clarity, we shouldn't bother constructing and initializing the faviconUrl URL object in the preceding lines.

What caused you to add a `startsWith("http")` check here?

background_scripts/completion.js
```diff
@@ -489,6 +491,7 @@ class TabCompleter {
           queryTerms,
           description: "tab",
           url: tab.url,
+          favIconUrl: tab.favIconUrl,
```
philc, line 494: It seems we're passing this along because tab.favIconUrl is sometimes a higher quality image than what can be fetched from chrome's favicon API? Is so, we should mention that here in a comment, with a reference. [This](https://issues.chromium.org/issues/40856652) is the Chromium bug I found about it, and some [related discussion](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/qS1rVpQVl8o/m/qmg1M13wBAAJ?pli=1).

philc (changes requested):
Good change. I tested it and it works well. I'm fine with the approach of using 64px icons always. I left a few comments.

Issue #4546, opened 2024-09-30
Enhance Find Mode Performance and Responsiveness


darpham:
### Problem Statement:
When using `/` find mode on large web pages, searching is slow to the point of unusable.
Reproducible via [https://nix-community.github.io/home-manager/options.xhtml](https://nix-community.github.io/home-manager/options.xhtml)

### Planned Improvements:
- [ ] Delay / debounce `FindMode.findInPlace()` to reduce the frequency of searches during typing.
- [ ] Implement a caching mechanism for text nodes, potentially using `SimpleCache` from `./lib/utils.js`.
- [ ] Offload text searching to a Web Worker or run asynchronously, possibly utilizing `AsyncDataFetcher` from `./lib/utils.js`.

### Expected Benefits:
- Improved responsiveness during rapid typing in find mode.
- Faster search performance, especially on large web pages.
- Reduced strain on system resources during find operations.

UncleSnail:
The first task should be sufficient. The second two are good ideas, but are probably unneeded and would introduce unnecessary complexity.
I like the idea for the debounce though, and this is a real issue. It also freezes on your example page on my machine. I'm going to take a look at your PR.

PR #4547, opened 2024-09-30
Add debounce to `HUD.search()` method

darpham:
## Description

This PR adds a debounce mechanism to the `HUD.search()` method in the `content_scripts/hud.js` file. The purpose of this change is to improve performance and reduce unnecessary searches when users are typing quickly in the HUD search field.

The implementation uses a `setTimeout` to delay the search execution by 300ms, clearing any previous timeout if a new search is initiated before the delay expires. This approach helps to minimize the number of searches performed while the user is still typing, potentially reducing strain on system resources and improving the overall user experience.

Related to issue: https://github.com/philc/vimium/issues/4546

### Changes made:

1. Added a `searchTimeout` property to the `HUD` object to store the timeout ID.
2. Implemented a debounce mechanism in the `search` method using `Utils.setTimeout`.
3. Moved the existing search logic inside the debounce callback.

### Rationale:

This change aligns with Vimium's design principles, particularly:

1. **Reliable**: By reducing the frequency of searches, we ensure smoother performance on various websites.
2. **Feels native**: The debounce mechanism is a common pattern in web applications, making Vimium feel more like a native part of the browser.
3. **Simple**: The implementation is straightforward and doesn't add significant complexity to the codebase.

### Additional context:

The PR introduces a small change (less than 50 LOC) that should be beneficial for many Vimium users, especially those who frequently use the HUD search feature on content-heavy pages.

### Testing:

I've run the test suite, and all tests have passed:
```bash
❯ ./make.js test
run: started
test-unit: started
Pass (192/192)
test-unit: finished (149ms)
test-dom: started
Listening on http://localhost:35500/
Running dom_tests.html
Pass (109/109)
Running vomnibar_test.html
Pass (3/3)
Unstable API 'Deno.Server.shutdown'. The --unstable flag must be provided.
```
Additionally, I've manually tested the changes by installing from source on Firefox Browser. The debounce functionality works as expected, reducing the number of searches performed during rapid typing.
Here's a screenshot of the Developer Tools, showing the Debugger Tab and Console (I didn't keep the `console.debug()`):
debug

### Follow-up / Open questions
- Upgrading Deno
- Add debounce test(s) using [FakeTime](https://jsr.io/@std/testing/doc/time/~/FakeTime)
  - I took a stab at it, but failed to integrate with `shoulda`, some documentation on how to add tests would be helpful.

### DIFF
--- a/content_scripts/hud.js
+++ b/content_scripts/hud.js
@@ -6,6 +6,7 @@ const HUD = {
   tween: null,
   hudUI: null,
   findMode: null,
+  searchTimeout: null,
   abandon() {
     if (this.hudUI) {
       this.hudUI.hide(false);
@@ -86,16 +87,22 @@ const HUD = {
   },
 
   search(data) {
-    // NOTE(mrmr1993): On Firefox, window.find moves the window focus away from the HUD. We use
-    // postFindFocus to put it back, so the user can continue typing.
-    this.findMode.findInPlace(data.query, {
-      "postFindFocus": this.hudUI.iframeElement.contentWindow,
-    });
+    if (this.searchTimeout) {
+      clearTimeout(this.searchTimeout);
+    }
 
-    // Show the number of matches in the HUD UI.
-    const matchCount = FindMode.query.parsedQuery.length > 0 ? FindMode.query.matchCount : 0;
-    const showMatchText = FindMode.query.rawQuery.length > 0;
-    this.hudUI.postMessage({ name: "updateMatchesCount", matchCount, showMatchText });
+    this.searchTimeout = Utils.setTimeout(300, () => {
+      // NOTE(mrmr1993): On Firefox, window.find moves the window focus away from the HUD. We use
+      // postFindFocus to put it back, so the user can continue typing.
+      this.findMode.findInPlace(data.query, {
+        "postFindFocus": this.hudUI.iframeElement.contentWindow,
+      });
+
+      // Show the number of matches in the HUD UI.
+      const matchCount = FindMode.query.parsedQuery.length > 0 ? FindMode.query.matchCount : 0;
+      const showMatchText = FindMode.query.rawQuery.length > 0;
+      this.hudUI.postMessage({ name: "updateMatchesCount", matchCount, showMatchText });
+    });
   },
 
   // Hide the HUD.


darpham:
@philc @UncleSnail can I have a review when you have time please 🙏 

@leoheitmannruiz as an active contributer, I'd appreciate any review or feedback as well.

UncleSnail (changes requested):
I love the idea of debouncing this as there is a real issue when on huge pages. There are two things I would like to see changed.
Firstly, I do not believe this implementation is correctly debouncing. The previously queued callback functions will still run after 300ms, even if another letter was typed. In a debounce, if another letter is typed within 300ms, the original callback should not run. I would suggest looking at reference JS debounce implementations and putting one in the utils next to the `setTimeout`.
Secondly, in 99.9% of cases, this is not an issue. I don't really want to make every user's experience "more laggy" all the time for a 0.1% use case. Even on a huge example page, the only potential for freezing is when there are LOTS of matches, which typically only occurs when the search input is small (about 3 or fewer letters). Once we have more letters, we don't really need the debouncing, since the number of results has drastically lowered. I would just only debounce when the search term is 3 or fewer letters, and not debounce otherwise. This is an easy way to make the experience snappy for most cases, but also ensure it will not freeze on large pages.
Another (arguably better) option would be to figure out how much of the slow down is from finding and counting the matches, and how much is from rendering the yellow boxes. I bet most is from the yellow boxes. As a way to ensure we always debounce only when we need, we could count the matches immediately all the time, then debounce the box drawing if there is a large number. That is assuming we can verify that will fix the freezing.

Either of those options will introduce some complexity and should come with a comment explaining what it is done that way. @philc will need to approve any PRs and will have helpful thoughts on the complexity of implementations and their benefit. I can, however, help you get the code working correctly so that his job is easier when/if he does decide to review your PR.
Thank you again for finding this issue and making a fix. I hope we can get something merged in to avoid freezes and crashes in the future.

darpham:
> Firstly, I do not believe this implementation is correctly debouncing. The previously queued callback functions will still run after 300ms, even if another letter was typed. In a debounce, if another letter is typed within 300ms, the original callback should not run. I would suggest looking at reference JS debounce implementations and putting one in the utils next to the `setTimeout`.

Admittedly, I'm not a software developer:-) and I appreciate the feedback and guidance. I will review and reference the JS debounce implementation(s) and docs, furthermore I'll create a `Utils.debounce()` method per your reccommendation.

> Secondly, in 99.9% of cases, this is not an issue. I don't really want to make every user's experience "more laggy" all the time for a 0.1% use case. Even on a huge example page, the only potential for freezing is when there are LOTS of matches, which typically only occurs when the search input is small (about 3 or fewer letters).

I agree and adding an if statement to bypass debounce where input `data.query` has length > 3 is feasible. I'd be curious on your thoughts on how (and if) this would interact when the input query is a regular expression or when the option `backwards` is enabled.

> Another (arguably better) option would be to figure out how much of the slow down is from finding and counting the matches, and how much is from rendering the yellow boxes.

I didn't mention in this PR or the linked [GH Issue](https://github.com/philc/vimium/issues/4546), however my findings with Firefox Profiler indicated the bottleneck to be within the `mode_find.js` as part of finding the matches, specifically the function `getAllTextNodes`.
This PR serves to provide a robust "band-aid solution" wherein future follow-ups to better identify and address the root cause.

> I can, however, help you get the code working correctly so that his job is easier when/if he does decide to review your PR.
> Thank you again for finding this issue and making a fix. I hope we can get something merged in to avoid freezes and crashes in the future.

Thank you @UncleSnail, I hope to also have something merged in the near future. I'll fixup the PR to properly debounce while bypassing if search input > 3 letters. If I'm unable to properly implement, I would appreciate some of your time to resolve. I hope contribute more in the future as this is a great FOSS extension.

philc:
I don't have a strong opinion about this yet, but it is conspicuous that on my machine (Macbook Pro M2 Max), the [case study page](https://nix-community.github.io/home-manager/options.xhtml) has no slowdown when doing any find query, short or long, whereas in Firefox, it is noticeably very slow to type when the match count is high.

It would be nice to better understand why there's a difference in responsiveness, because a debounce does degrade the responsiveness of the UX and does add some complexity. I agree that debouncing when query.length <= 3 is a good way to mitigate the UX impact, and would be required if we ever do more work per search, like highlighting all matches on the page as Chrome's native search UX does (but not Firefox's).

UncleSnail:
> I don't have a strong opinion about this yet, but it is conspicuous that on my machine (Macbook Pro M2 Max), the [case study page](https://nix-community.github.io/home-manager/options.xhtml) has no slowdown when doing any find query, short or long, whereas in Firefox, it is noticeably very slow to type when the match count is high.

Without extensively looking into it, I'm guessing Chrome put rendering of the page into another thread than rendering of the addon's text box, and Firefox didn't. Testing on my machine Chrome vs Firefox, on Firefox the whole page and text box freezes, but on Chrome, the page word highlighting still takes a half-second to update (as if the search is de-bounced) and the textbox still responds. It seems like allowing the page drawing to be async from addon drawing (in Chrome) would be one explanation explanation for this behavior.
In other words, on Firefox, the order of events goes:
1) I type "ni" on the keyboard.
2) I see "n" in the textbox.
3) I see all "n"s highlighted on the page.
4) I see "ni" in the textbox.
5) I see all "ni"s highlighted on the page.

Each of these steps has some lag in between. On Chrome, the order of events is:
1) I type "ni" on the keyboard.
2) I see "ni" in the textbox.
3) I see all "n"s highlighted on the page.
5) I see all "ni"s highlighted on the page.

This implies to me that Chrome is keeping a queue of render updates to be performed asynchronously on the page. That may not be correct, but it seems to explain the behavior and why there aren't "noticeable" performance issues on Chrome.

PR #4549, opened 2024-10-02
add commands for set proxy in FF

klim-iv:
Commands:
  'sn' -> none proxy
  'ss' -> system proxy
  'sa' -> autodetect proxy
  'sm' -> manual proxy

## Description


philc:
I'm curious, what's your use case for having these commands exposed in Vimium? How often do you switch proxies?

klim-iv:
> I'm curious, what's your use case for having these commands exposed in Vimium? How often do you switch proxies?

due to government restrictions in our country some sites can't be opened without foreign proxies, that's why I want to use quick switch for open such sites and update info in tabs. And 'yes' - I do it often :-) , especially to news sites

philg-dev:
Wouldn't it make much more sense to use an addon specifically designed for proxy switching, e.g. FoxyProxy, instead and configure URL patterns accordingly? That way you could have FoxyProxy automatically select the proxy you need for any frequently used site. I know it's not exactly the same as having a "shortcut" for open an arbitrary URL with a given proxy, but in the past when I had to use a bunch of proxies for work related stuff FoxyProxy came in extremely handy for like 99% of what I did.

klim-iv:
> Wouldn't it make much more sense to use an addon specifically designed for proxy switching, e.g. FoxyProxy, instead and configure URL patterns accordingly? That way you could have FoxyProxy automatically select the proxy you need for any frequently used site. I know it's not exactly the same as having a "shortcut" for open an arbitrary URL with a given proxy, but in the past when I had to use a bunch of proxies for work related stuff FoxyProxy came in extremely handy for like 99% of what I did.

Yes,  I used FoxyProxy, but this is one more plugin with additional configs and etc. As for me shortcuts for swithing is more simplest.  Ok, if you  think that such shortcut is excess - I'll close request and will use it privately :-)

philg-dev:
> Ok, if you think that such shortcut is excess - I'll close request and will use it privately :-)

That's not necessarily what I wanted to say. I was just curious, if FoxyProxy would solve your use case entirely already or if there's further reason to have proxy shortcuts like you've described. I was just trying to add to the discussion and see if you're familiar with proxy switchers.

I'm not the Vimium Owner, so it's not for me to say if it's "excess" or not. Maybe the maintainers like your changes after all :+1:  I don't have a strong opinion on it. I think you should leave the pull request open until more people give their opinions / feedback.

philc:
Nice PR. I think ultimately this is niche for inclusion in Vimium, especially given that the proxy switching is only supported in Firefox.

PR #4553, opened 2024-10-08
Fix Deno Format and simple TODOs

UncleSnail:
## Description

This PR pushes the `deno fmt` changes to all files to clean up the formatting, and will also cover fixing/removing small TODOs that have been hanging around the project. The changes are the same as from [this](https://github.com/philc/vimium/pull/4518) previous PR but do not include the hard reload DRY changes, as requested.

PR #4554, opened 2024-10-08
Command options (in help message)

UncleSnail:
## Description

Show command options in the help message dialogue as specified in [this comment](https://github.com/philc/vimium/pull/4518#issuecomment-2268204745) and the associated PR discussion.

Handles truncating based on the full length of the line, not just the length of the options, and shows the full options on hover.

image

Always indicates the presence of options even on lines with very long descriptions.

image

Correctly handles flags (without showing them as flag=true)

image

Handles commands with multiple options.

image

This PR also switches the new `hardReload` for the old `reload hard` to avoid repeated code/functionality.

## Implementation

There could be cleaner implementations, so let me know if you have a better idea, but I believe I have found a good method that requires minimal changes and fits very well into the existing mechanism. I will describe the implementation and logic more in comments on the code.

I have formatted all new code (see #4553 for the other changes from `deno fmt`) and all tests pass.

### DIFF
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -304,7 +304,6 @@ const Commands = {
       "scrollToLeft",
       "scrollToRight",
       "reload",
-      "hardReload",
       "copyCurrentUrl",
       "openCopiedUrlInCurrentTab",
       "openCopiedUrlInNewTab",
@@ -393,7 +392,6 @@ const Commands = {
     "enterVisualLineMode",
     "toggleViewSource",
     "passNextKey",
-    "hardReload",
     "setZoom",
     "zoomIn",
     "zoomOut",
@@ -416,7 +414,7 @@ const defaultKeyMappings = {
   "d": "scrollPageDown",
   "u": "scrollPageUp",
   "r": "reload",
-  "R": "hardReload",
+  "R": "reload hard",
   "yy": "copyCurrentUrl",
   "p": "openCopiedUrlInCurrentTab",
   "P": "openCopiedUrlInNewTab",
@@ -506,7 +504,6 @@ const commandDescriptions = {
   scrollFullPageUp: ["Scroll a full page up"],
 
   reload: ["Reload the page", { background: true }],
-  hardReload: ["Hard reload the page", { background: true }],
   toggleViewSource: ["View page source", { noRepeat: true }],
 
   copyCurrentUrl: ["Copy the current URL to the clipboard", { noRepeat: true }],
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -257,32 +257,44 @@ const Commands = {
   // Build the "helpPageData" data structure which the help page needs and place it in Chrome
   // storage.
   prepareHelpPageData() {
-    const commandToKey = {};
+    /*
+      Map of commands to option sets to keys to trigger that command option set.
+      Commands with no options will have the empty string options set.
+      Example:
+      {
+        "zoomReset": {
+          "": ["z0", "zz"] // No options, with two key maps, ie: `map zz zoomReset`
+        },
+        "setZoom": {
+          "1.1": ["z1"], // `map z1 setZoom 1.1`
+          "1.2": ["z2"], // `map z2 setZoom 1.2`
+        }
+      }
+    */
+    const commandToOptionsToKeys = {};
     for (const key of Object.keys(this.keyToRegistryEntry || {})) {
       const registryEntry = this.keyToRegistryEntry[key];
-      const optionString = registryEntry.optionList?.length > 0
-        ? registryEntry.optionList?.join(" ")
-        : "";
-      if (commandToKey[registryEntry.command] == null) {
-        commandToKey[registryEntry.command] = {};
-      }
-      (commandToKey[registryEntry.command][optionString] != null
-        ? commandToKey[registryEntry.command][optionString]
-        : (commandToKey[registryEntry.command][optionString] = [])).push(key);
+      const optionString = registryEntry.optionList?.join(" ") || "";
+      commandToOptionsToKeys[registryEntry.command] ||= {};
+      commandToOptionsToKeys[registryEntry.command][optionString] ||= [];
+      commandToOptionsToKeys[registryEntry.command][optionString].push(key);
     }
     const commandGroups = {};
     for (const group of Object.keys(this.commandGroups || {})) {
       const commands = this.commandGroups[group];
       commandGroups[group] = [];
       for (const command of commands) {
-        for (const [options, keys] of Object.entries(commandToKey[command] ?? { "": [] })) {
+        // Default to base command has no keys for "show available commands" menu.
+        const optionsToKeys = commandToOptionsToKeys[command] ?? { "": [] };
+        for (const [options, keys] of Object.entries(optionsToKeys)) {
+          const advanced = this.advancedCommands.includes(command) ||
+            this.advancedCommands.includes(`${command} ${options}`);
           commandGroups[group].push({
             command,
             description: this.availableCommands[command].description,
-            keys: keys,
-            advanced: this.advancedCommands.includes(command) ||
-              this.advancedCommands.includes(`${command} ${options}`),
-            options: options,
+            keys,
+            advanced,
+            options,
           });
         }
       }
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -64,8 +64,8 @@ const Commands = {
 
     for (const line of configLines) {
       const tokens = line.split(/\s+/);
-      const command = tokens[0].toLowerCase();
-      switch (command) {
+      const action = tokens[0].toLowerCase();
+      switch (action) {
         case "map":
           if (tokens.length >= 3) {
             const [_, key, command, ...optionList] = tokens;
@@ -118,7 +118,7 @@ const Commands = {
           }
           break;
         default:
-          logWarning(`"${command}" is not a valid config command in line:`, line);
+          logWarning(`"${action}" is not a valid config command in line:`, line);
       }
     }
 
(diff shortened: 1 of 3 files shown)


pages/help_dialog.js
```diff
@@ -124,7 +124,17 @@ const HelpDialog = {
             });
           }
 
-          $$(descriptionElement, ".vimiumHelpDescription").textContent = command.description;
+          const MAX_LENGTH = 50;
+          const desiredOptionsLength = Math.max(0, MAX_LENGTH - command.description.length - 3);
```
UncleSnail, line 128: Find the length that the options list should be to make the full description the right length. Don't ever let it be less than 0, and leave room for the ellipsis "...".

Since this might not be clear, I will explain why I implement it this correct way. It bothers me how many people implement this wrong. Leaving room means that we fit as long lines as we can. For example, if we want a line fewer than 10 characters, not leaving room (not having the `- 3`) would mean we need to truncate anything over 7 characters so that the ... does not make the line too long. So we end up with:

"truncated!" -> "truncat..."

There is no reason to do this since the word could have fit on the line (word is 10 chars and we are fine with 10 chars). The other common wrong implementation is to truncate anything >= 10 and add the "..." giving:

"truncated!" -> "truncated!..."

This gives us a line that is **longer** than 10 chars!

The correct implementation (like this where we leave space) means only lines that need to be truncated are, and we never make a line too long:

"truncated!" -> "truncated!" (10 chars)
"even longer" -> "even lo..." (10 chars)

I hope this explains what the `-3` is for. I thought about adding a quick comment in the code there since it might not be clear, and I will do so if you like.

pages/help_dialog.js
```diff
@@ -124,7 +124,17 @@ const HelpDialog = {
...
+          const optionsTruncated = command.options.substring(0, desiredOptionsLength);
+          let ellipis = "";
+          if ((command.description.length + command.options.length) > MAX_LENGTH) {
```
UncleSnail, line 131: If we needed to truncate, add a set of ellipsis and set the title to the full options list so that we can see it on hover.

pages/help_dialog.js
```diff
@@ -124,7 +124,17 @@ const HelpDialog = {
...
+            ellipis = "...";
+            $$(descriptionElement, ".vimiumHelpDescription").title = command.options;
+          }
+          const optionsString = command.options ? ` (${optionsTruncated}${ellipis})` : "";
+          const fullDescription = `${command.description}${optionsString}`;
+          $$(descriptionElement, ".vimiumHelpDescription").textContent = fullDescription;
```
UncleSnail, line 137: Note that this implementation works exactly as you described in your comment on #4518 where the options list is truncated, but not the ending ")" or the command description itself.

I tried the other methods of truncation, and I like this one the most. I agree with your design, it works very well.

UncleSnail (commented):
I worked through a few different methods for implementing this, and have tried to explain why I believe this is the best given the existing code. If you have any more questions about why I chose this implementation, please ask!

UncleSnail:
One more improvement that I think I should look into is ensuring the lines for the same command with different options inside of the help menu are reasonably ordered. Right now they aren't entirely. It's probably rare that it would be relevant, but for my `setZoom` commands that I use it is noticeable.
image
I could sort alphanumerically, but I think it makes the most sense to sort by the order the user has defined the commands in their mappings config. I plan to look into this, but I need to see if that is even feasible and it could be saved for a future PR.

UncleSnail:
I also just pushed a commit to allow for a command + options set to be considered advanced. This way the `reload hard` command can be considered advanced and be hidden by default.

If you do not like this feature, I can easily remove this commit. It seems like hard reloading should be an advanced command, so I wanted to let us mark it as such, but it does add some complexity and the potential for the advanced command list to explode with advanced option choices.

Another implementation is to consider any command with options to be advanced, but then user mappings of simple commands (like new tab) with options are considered advanced and can't be seen in help without showing advanced options. I can imagine many users want to see their custom mappings without having "see advanced" checked, so I went with the implementation above. But, all options commands being advanced is a simpler solution to the problem with less scope-creep and still lets `reload hard` be advanced.

background_scripts/commands.js
```diff
@@ -260,21 +260,31 @@ const Commands = {
...
-          keys: commandToKey[command] != null ? commandToKey[command] : [],
-          advanced: this.advancedCommands.includes(command),
-        });
+        for (const [options, keys] of Object.entries(commandToKey[command] ?? { "": [] })) {
+          commandGroups[group].push({
+            command,
+            description: this.availableCommands[command].description,
+            keys: keys,
+            advanced: this.advancedCommands.includes(command) ||
+              this.advancedCommands.includes(`${command} ${options}`),
```
UncleSnail, line 284: This is the line for allowing command + options sets to be considered advanced (like `reload hard`). We just also check if the advancedCommands list includes the command + options set.

background_scripts/commands.js
```diff
@@ -386,11 +395,11 @@ const Commands = {
     "enterVisualLineMode",
     "toggleViewSource",
     "passNextKey",
-    "hardReload",
```
UncleSnail, line 389: Previously (last PR), hardReload was considered an advanced command, but reload is not.

UncleSnail:
@philc Let me know if there is anything you would like me to fix. Also, if all of the comment explanations/justifications are more overwhelming than helpful I can try to limit them in the future; just let me know your preference.

philc:
@UncleSnail Will do. Thanks for posting this. I'll return to the project and review it soon!

background_scripts/commands.js
```diff
@@ -260,21 +260,31 @@ const Commands = {
     const commandToKey = {};
     for (const key of Object.keys(this.keyToRegistryEntry || {})) {
       const registryEntry = this.keyToRegistryEntry[key];
-      (commandToKey[registryEntry.command] != null
-        ? commandToKey[registryEntry.command]
-        : (commandToKey[registryEntry.command] = [])).push(key);
+      const optionString = registryEntry.optionList?.length > 0
+        ? registryEntry.optionList?.join(" ")
```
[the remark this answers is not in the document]
philc: The previous (ugly) conditional was generated by the automatic coffeescript->js conversion awhile back and was never touched up. Your new version looks good.

This can be made a bit shorter in either of these two ways:

    const optionString = (registryEntry.optionsList || []).join(" ")

    const optionString = registryEntry.optionsList?.join(" ") || "";

Calling join on an empty list produces the empty string.

background_scripts/commands.js
```diff
@@ -260,21 +260,30 @@ const Commands = {
...
+        : "";
+      if (commandToKey[registryEntry.command] == null) {
+        commandToKey[registryEntry.command] = {};
+      }
+      (commandToKey[registryEntry.command][optionString] != null
+        ? commandToKey[registryEntry.command][optionString]
+        : (commandToKey[registryEntry.command][optionString] = [])).push(key);
```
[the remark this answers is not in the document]
philc: Given the new structure, the current name is actively misleading. The map should be called `commandToOptionStringToKeys`, or `commandToOptionsToKeys`. A shorter name if possible would be nice, too.

And it would be nice to include your illustrative example data structure in a comment, which makes the name of this variable less important, because after reading your comment, it's easy to hold the shape of the data structure in one's head as one reads the code. The example should indicate that a command with no options will have an empty string as its key.

The use of the ternary operator here for control flow is an artifact of the coffeescript->js conversion process awhile ago. Let's change it to less dense and easier to read statements:

    commandToKey[registryEntry.command][optionString] ||= [];
    commandToKey[registryEntry.command][optionString].push(key);

background_scripts/commands.js
```diff
@@ -260,21 +260,31 @@ const Commands = {
...
       const commands = this.commandGroups[group];
       commandGroups[group] = [];
       for (const command of commands) {
-        commandGroups[group].push({
-          command,
-          description: this.availableCommands[command].description,
-          keys: commandToKey[command] != null ? commandToKey[command] : [],
-          advanced: this.advancedCommands.includes(command),
-        });
+        for (const [options, keys] of Object.entries(commandToKey[command] ?? { "": [] })) {
```
philc, line 278: I find the `?? { "": [] }))` bit difficult to read on this already long line.

background_scripts/commands.js
philc: Minor, and debatable, but I would pull this into a separate variable so there's less logic happening within the push statement.

    const advanced = this.advancedCommands.includes(command) || this.advancedCommands.includes(`${command} ${options}`)
    ...
    commandGroups[group].push({
      command,
      description: this.availableCommands[command].description,
      keys,
      advanced,
    ...

pages/help_dialog.js
philc: This was clear to me as someone who's ellipsized display strings before, but I think it's helpful to have a short comment saying `- 3 because 3 is the length of the ellipsis string.`

pages/help_dialog.js
```diff
@@ -124,7 +124,17 @@ const HelpDialog = {
...
           }
 
-          $$(descriptionElement, ".vimiumHelpDescription").textContent = command.description;
+          const MAX_LENGTH = 50;
+          const desiredOptionsLength = Math.max(0, MAX_LENGTH - command.description.length - 3);
+          const optionsTruncated = command.options.substring(0, desiredOptionsLength);
+          let ellipis = "";
+          if ((command.description.length + command.options.length) > MAX_LENGTH) {
+            ellipis = "...";
+            $$(descriptionElement, ".vimiumHelpDescription").title = command.options;
```
philc, line 133: It would be helpful to have a comment above this line saying "The non-ellipsized option list will be visible on-hover."

pages/help_dialog.js
```diff
@@ -124,7 +124,17 @@ const HelpDialog = {
...
+          }
+          const optionsString = command.options ? ` (${optionsTruncated}${ellipis})` : "";
+          const fullDescription = `${command.description}${optionsString}`;
```
philc, line 136: This confused me because it looks like here we're setting the textContent to be a truncated options string, but it may have not been truncated.

I think this can be made clearer by just mutating the display string we're going to show. Roughly:

    let optionDisplayString = command.options.substring;

    if (we need to truncate) {
      optionDisplayString = command.options.substring(0, desiredOptionsLength) + "...";
      $$(descriptionElement, ".vimiumHelpDescription").title = command.options;
    }

    const optionsString = command.options ? ` (${optionDisplayString})` : "";
    const fullDescription = `${command.description}${optionsString}`;

philc (changes requested):
This is going to be great! Overall it looks good; I've left some comments.

pages/help_dialog.js
UncleSnail: I agree, without thoroughly reading the code it does look like the string is always being truncated. I'll change it to be more intuitive.

UncleSnail (commented):
I have fixed all the suggested changes and added a few comments where needed to help understandability.
I reran the formatter to ensure I didn't introduce any new formatting mistakes and tested the extension lightly in Chrome. Everything still appears to function as it should/did in my previous PR testing.

philc:
Excellent!

philc:
There were some broken tests from this PR which I couldn't figure out how to make work, and so I deleted them in 19159d75d1cda3636526053de944def2f6b051c7.

PR #4566, opened 2024-10-29
Fix clickable elements with class 'btn'

### LINKED ISSUE #3550: Make class="btn" clickable / Or support clickable css rule extention
Some button without link could not be clicked : 

image
image

I think we can add `<button>` , ` <xxx class="btn xxxx"></xxx>`,  ` <xxx class="button  xxx"></xxx>`  to clickable list .

Futher, we may support different framework's clickable element, such as element with attribute `data-toggle="xxx"`  or someting else . This can be added as extention of vimium .

ScottWilliamAnderson:
Fixes #3550

Update `content_scripts/link_hints.js` to make elements with class 'btn' clickable.

* Add logic to `getLocalHintsForElement` to check for class 'btn' and mark elements as clickable.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/philc/vimium/issues/3550?shareId=c288d7af-145c-4168-b305-1f3dbc8f7f3b).

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -1,3 +1,18 @@
+//
+// This implements link hinting. Typing "F" will enter link-hinting mode, where all clickable items
+// on the page have a hint marker displayed containing a sequence of letters. Typing those letters
+// will select a link.
+//
+// In our 'default' mode, the characters we use to show link hints are a user-configurable option.
+// By default they're the home row. The CSS which is used on the link hints is also a configurable
+// option.
+//
+// In 'filter' mode, our link hints are numbers, and the user can narrow down the range of
+// possibilities by typing the text of the link itself.
+//
+
+// A DOM element that sits on top of a link, showing the key the user should type to select the
+// link.
 class HintMarker {
   hintDescriptor;
   localHint;
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -1,18 +1,3 @@
-//
-// This implements link hinting. Typing "F" will enter link-hinting mode, where all clickable items
-// on the page have a hint marker displayed containing a sequence of letters. Typing those letters
-// will select a link.
-//
-// In our 'default' mode, the characters we use to show link hints are a user-configurable option.
-// By default they're the home row. The CSS which is used on the link hints is also a configurable
-// option.
-//
-// In 'filter' mode, our link hints are numbers, and the user can narrow down the range of
-// possibilities by typing the text of the link itself.
-//
-
-// A DOM element that sits on top of a link, showing the key the user should type to select the
-// link.
 class HintMarker {
   hintDescriptor;
   localHint;
@@ -1220,6 +1205,12 @@ const LocalHints = {
     if (!isClickable && !(tabIndex < 0) && !isNaN(tabIndex)) {
       isClickable = true;
       onlyHasTabIndex = true;
+    }
+
+    // Check for elements with class 'btn'
+    if (!isClickable && className?.toLowerCase().includes("btn")) {
+      isClickable = true;
+      possibleFalsePositive = true;
     }
 
     if (isClickable) {
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -1,3 +1,18 @@
+//
+// This implements link hinting. Typing "F" will enter link-hinting mode, where all clickable items
+// on the page have a hint marker displayed containing a sequence of letters. Typing those letters
+// will select a link.
+//
+// In our 'default' mode, the characters we use to show link hints are a user-configurable option.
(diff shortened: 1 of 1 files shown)


content_scripts/link_hints.js
```diff
@@ -1213,6 +1213,11 @@ const LocalHints = {
       possibleFalsePositive = true;
     }
 
+    if (!isClickable && className?.toLowerCase().includes("btn")) {
+      isClickable = true;
+      possibleFalsePositive = true;
+    }
```
UncleSnail, line 1219: This could be part of the same if statement by doing something like:
`if (!isClickable && (className?.toLowerCase().includes("button") || className?.toLowerCase().includes("btn")))`

UncleSnail (changes requested):

content_scripts/link_hints.js
ScottWilliamAnderson: Agreed, I've made a change to address this.

ScottWilliamAnderson:
@philc I think this PR is ready for review, let me know what you think

philc:
Thanks for the PR. Link hint detection is the main hot/expensive path in Vimium, so we should carefully understand and introduce each change there. Can you provide a few examples of websites which use btn as the class name for elements which aren't already detected by Vimium?

content_scripts/link_hints.js
```diff
@@ -1204,11 +1204,11 @@ const LocalHints = {
     // # Detect elements with "click" listeners installed with `addEventListener()`.
     // isClickable ||= element.hasAttribute "_vimium-has-onclick-listener"
 
-    // An element with a class name containing the text "button" might be clickable. However, real
+    // An element with a class name containing the text "button" or "btn" might be clickable. However, real
     // clickables are often wrapped in elements with such class names. So, when we find clickables
     // based only on their class name, we mark them as unreliable.
     const className = element.getAttribute("class");
-    if (!isClickable && className?.toLowerCase().includes("button")) {
+    if (!isClickable && (className?.toLowerCase().includes("button") || className?.toLowerCase().includes("btn"))) {
```
philc, line 1211: We should extract the className and call toLowerCase() on it only once.

ScottWilliamAnderson:
> Thanks for the PR. Link hint detection is the main hot/expensive path in Vimium, so we should carefully understand and introduce each change there. Can you provide a few examples of websites which use btn as the class name for elements which aren't already detected by Vimium?

Thanks Phil. I came across this when using the default bootstrap toggles during development:

https://getbootstrap.com/docs/5.3/forms/checks-radios/#toggle-buttons

How expensive is the detection, have you been able to profile it before?

philc:
On the page you linked, I couldn't find the Bootstrap controls which aren't showing hints -- all of them (except the disabled ones) show hints. Was it for an older version of Bootstrap?

Benchmarking requirers running the link hint code on a page several times in a loop (see examples in #3063, #2410). I'm sure your change wouldn't show up. The cost of this change is having more rules, and the documentation for why they're there, in a critical part of the code.

philc:
Checking in to note that the two reported cases of this issue (this, and #3550) both reference Bootstrap toggle controls, which uses a label with class "btn" as the control's underlying element.

In my tests on the Bootstrap's docs page for that control, this portion of link_hints.js is correctly detecting that the label is clickable:

        isClickable ||= (element.control != null) &&
          !element.control.disabled &&
          ((this.getLocalHintsForElement(element.control)).length === 0);

I'm not against merging in this change, but we should have at least one current, reproducible example of an element with a "btn" class in the wild that think should be fixed, as part of making this change, for documentation purposes.

ScottWilliamAnderson:
Thanks Phil, turns out the missing links on the bootstrap docs pages were caused by another extension. I had installed vimium-c on a lightweight Linux machine for testing, and my settings sync downloaded it on my main machine as well. That extension is the one with this bug :D

I can see the buttons highlighted using vimium's main branch, so I think this PR is not needed anymore. I'm not sure if #3550 refers to any other unseen issues but I don't see it on my end.

Thanks again for your help!

philc:
Excellent. Closing

gdh1995:
Sorry to disturb you, but I have a somehow related question about this PR - Could you give a snapshot about that Vimium C lacks (or causes Vimium lacks) hint markers for some buttons on https://getbootstrap.com/docs/5.3/forms/checks-radios/#toggle-buttons ? I'm the author of Vimium C and I want to fix this compatibility issue, while I didn't reproduce it just now.

northern-mindset:
For those who are interested…
There is another way of making Vimium detect clickable elements. It is to use Tempermonkey or the like.
Here is an example of a script that makes it detect an "expand control" in Confluence: https://codeberg.org/akd/tampermonkey-scripts/src/branch/main/vimium/confluence.user.js.

Inspired by https://stackoverflow.com/a/79758021/3082178.

Issue #3333, opened 2019-05-24
why  my chrome can not use it following is the message 


anewcoder1992:
Your browser does not run web extensions like Vimium on certain pages, usually for security reasons.  Unless your browser's developers change their policy, then unfortunately it is not possible to make Vimium (or any other web extension, for that matter) work on this page

gdh1995:
Chrome doesn't allow Vimium to run on URLs of `chrome://`, `https://chrome.google.com/*` and `about:*`.

Hafizur046:
mine is showing this error even on FB quite strange it didn't use to show it there and it shouldn't

dantehemerson:
How I can enable it?

lucidBrot:
Same message in vimium encountered rn on https://linux.die.net/man/1/ed ... no idea why but reloading the page fixed it.

adeora7:
Maybe you recently installed the extension and those pages were already open and the extension couldn't inject any javascript magic that it does when a page loads. That was the issue in my case.

bsdson:
This happened to me on site "stackoverflow".
A reload fixes it.

USAMAWIZARD:
@dantehemerson @gdh1995 @anewcoder1992 please checkout this pr if you want to run it on chrome:// pages Thank  you

philc:
For why it doesn't run on chrome:// pages, this is documented [here in the FAQ](https://github.com/philc/vimium/wiki/FAQ#why-doesnt-vimium-work-on-the-chrome-store-or-urls-starting-with-chrome).

PR #4583, opened 2024-12-05
allow extension to run on chrome://pages

USAMAWIZARD:
## Description
its very annoying when I am on a home tab and cannot user vim binding so this pr adds permission to allow chrome to run extension on chrome:// pages https://github.com/philc/vimium/issues/3333

you will need to enable extension chrome://flags/#extensions-on-chrome-urls

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -16,7 +16,7 @@
     "browser_style": false,
     "open_in_tab": true
   },
-  "host_permissions": ["<all_urls>"],
+  "host_permissions": ["chrome://*/*","<all_urls>"],
   "permissions": [
     "tabs",
     "bookmarks",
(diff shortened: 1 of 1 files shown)


philc:
Thanks for documenting this workaround.

I think this PR should probably be added if Chrome's extensions-on-chrome-urls flag was true by default.

Presumably Firefox doesn't have this feature?

Do you happen to know what the permissions behavior is when the user installs an extension with "chrome://*/*" in the host_permissions when the user doesn't have extension-on-cchrome-urls enabled? Are they still prompted to grant the permission on install?

Requesting a permission that most users won't make use of (because they wouldn't bother to enable chrome://flags/#extensions-on-chrome-urls) isn't worth the added UX friction.

Note that there will probably be some fragmentation in how chrome URLs will be treated by chromium
browsers; see [this discussion about Brave](https://www.reddit.com/r/brave_browser/comments/1bwk773/allow_extensions_on_chrome_urls_how_dead_flag/).

PR #4591, opened 2024-12-22
Support query keyword in activateBookmark.

mijoharas:
## Description

I would like to create a shortcut that lets me browse a specific folder in my bookmarks. To do this I would like to prefill a query into the `Vomnibar.activateBookmark` command.

What I want is to have:
```
map wtf Vomnibar.activateBookmarks query=/links
```
To allow me to browse the links folder of my bookmarks.

It turns out that those options already get parsed into the keyregistry, and the activate command already accepts these options, so the only thing to needed would be to add this extra argument to the content script (similarly to the `Vomnibar.activate` above) as is done in the PR.

### DIFF
--- a/content_scripts/vomnibar.js
+++ b/content_scripts/vomnibar.js
@@ -24,19 +24,23 @@ const Vomnibar = {
     });
   },
 
-  activateBookmarks(sourceFrameId) {
-    this.open(sourceFrameId, {
-      completer: "bookmarks",
-      selectFirst: true,
-    });
+  activateBookmarks(sourceFrameId, registryEntry) {
+    let options = Object.assign({}, registryEntry.options,
+      {
+        completer: "bookmarks",
+        selectFirst: true,
+      });
+    this.open(sourceFrameId, options);
   },
 
-  activateBookmarksInNewTab(sourceFrameId) {
-    this.open(sourceFrameId, {
-      completer: "bookmarks",
-      selectFirst: true,
-      newTab: true,
-    });
+  activateBookmarksInNewTab(sourceFrameId, registryEntry) {
+    let options = Object.assign({}, registryEntry.options,
+      {
+        completer: "bookmarks",
+        selectFirst: true,
+        newTab: true,
+      });
+    this.open(sourceFrameId, options);
   },
 
   activateEditUrl(sourceFrameId) {


philc:
This is a cool feature. I've tested your PR and it works.

However, I'm realizing that one can't pass a query with a space in it using the `map` keyword in one's Vimium config, unless I'm mistakenly missing some syntax that's available. Neither %20 nor + works.

This feature is still useful, because the query `/one` will prefix match a folder named "one two". But it's not ideal: if a user has a space somewhere in the folder chain of their bookmarks, they can't precisely query for that folder, only the prefix prior to the space.

Before merging this and documenting the feature, I think we'll need a syntax for encoding spaces in options passed to commands with `map`.

FYI @UncleSnail since you've been working on command syntax recently.

mijoharas:
Good point, I'd not noticed that.

So, if we wanted to just do a quick substitution we could do the following for `%20` going to space.
[background_scripts/commands.js:210](https://github.com/philc/vimium/blob/feature-query-keyword-in-bookmarks-tab/background_scripts/commands.js#L209)
```
      options[parse[0]] = parse.length === 1 ? true : parse[1].replace("%20", " ");

```

 Also, while messing around with that, I noticed that the unit tests are broken. I fixed them in a commit, which I'm gonna push to this branch (might be cleaner to split it off into it's own PR, but I'm not sure how you want to do things for this)
 
 Actually while I"m at it  I'm also gonna push a commit and test for changing `%20` to space. Feel free to cherry pick this stuff and/or change the approach (we obvs didn't settle on a syntax that was wanted) ofc, just figure I'll throw it up here for comment while it's on my machine.

philc:
@mijoharas Ack, sorry about the unit test breakage. That was a subtle breakage (make.js still exits successfully...). I've pushed a fix in bf0bdda96066cba9c9fce86fdc34c1a90143ed69.

I'm not sure %20 is a good solution for how to handle spaces; it's just one thing that I tried. If we decode %20 in option values, does that mean we decode every other escape sequence? Supporting some kind of quoting around option values is another solution. I haven't used Vim in years, but I think .vimrc supports triple backslash for escaping spaces, and also surrounding option values with single quotes, for instance.

Let's see if anyone else chimes in with an opinion.

mijoharas:
yeah, I don't particularly think that `%20` is a very nice solution either, (mostly just pointing out where we could do a quick fix if we wanted) nor do I use vim so I don't have much to suggest, so let's wait for someone else to chime in.

UncleSnail:
> Supporting some kind of quoting around option values is another solution. I haven't used Vim in years, but I think .vimrc supports triple backslash for escaping spaces, and also surrounding option values with single quotes, for instance.
> 
> Let's see if anyone else chimes in with an opinion.

Thank you for pinging me on this @philc. I was thinking the same as you suggested here. I need to escape spaces in bash nearly every day, and while I occasionally write the bash command (if it's short) like:

`command -d option\ with\ spaces`

It is usually easier to use:

`command -d "option with spaces"`

For the average user, I believe this would be the easiest and most understandable solution. If we want an option with spaces, just write it in double quotes:

`map asdf Vomnibar.activateBookmarks query="/path with spaces/to your/bookmarks"`

None of the options that are currently supported are likely to have double quotes in them, except potentially URLs, which are likely to have the encoding instead, so I don't think there would be a big issue with breaking existing configs. I personally think it's better to solve this by escaping double quotes like `"option with a \" in it"` than to escape spaces.

At first glance, it seems like we could just ignore spaces until we reach an "optionName=" but since we support unnamed options like `reload hard` and `setZoom 1.2`, we cannot do that. **However**, we can sidestep the issue the same way these two commands do if we don't want to come up with an implementation right now. As long as the command (`activateBookmarks`) doesn't have any other options it can take, with the current implementation it is easy to hijack the system and just use the full options array joined together as the path. So without any new functionality, we could implement this command like:

`map asdf Vomnibar.activateBookmarks query=/path with spaces/to your/bookmarks`

This would make the options list become `["/path", "with", "spaces/to", "your/bookmarks"]`, so you could just write `options.join(" ")` to get the full path back.

Again, I personally prefer the "quote multi-word options" design, but there are other options.

UncleSnail:
Also, if you do decide to go with quoting out multi-word options, I'd be willing to write the parser changes if you would like.

mijoharas:
I removed the other two commits (fixing the tests, and using `%20` encoding for spaces) and left this Pr with just the change to support passing options through to `activateBookmarks` as it was originally.

It feels to me that this, and the support for spaces are somewhat orthogonal? should we merge this one now and wait for the extra enhancement to allow us to have spaces in the options later @philc ?

(also, I think quoted spaces sounds like a nice enough solution from a user perspective imo @UncleSnail )

philc:
Thanks for the discussion guys. I've added a command listing recently which touches a lot of these areas, and I'll come back to this PR soon.

mijoharas:
Nice work on the documentation for the options (all of this would have simplified things a lot when figuring out what was going on here before! great work!). Not sure whether you'd prefer this as a new issue or PR, so just putting it as a comment:

It looks like your new feature is removing the query and keyword options from `Vomnibar.activateInNewTab` when they were previously available. On master that command is treated the same as `Vomnibar.activate` [here](https://github.com/philc/vimium/blob/master/content_scripts/vomnibar.js#L15-L18) but in the command listing is missing the same options as `activate` [here](https://github.com/philc/vimium/blob/master/background_scripts/all_commands.js#L294-L298) (as you can see [here](https://github.com/philc/vimium/blob/master/background_scripts/all_commands.js#L303-L308)).

philc:
@mijoharas That omission was unintended. It's a bit difficult to find which options are supported where, which is part of the goal for formalizing them. I'll take care of fixing Vomnibar.activateInNewTab.

I still plan to return to this PR shortly.

philc:
Alright folks, on master, option values can now optionally be surrounded with quotes, and `Vomnibar.activateBookmark` accepts a query argument. Take it for a spin:

`map wtf Vomnibar.activateBookmarks query="/hello world"`

mijoharas:
Can confirm, it works as intended! Thanks again.

mijoharas:
Hey, sorry to post on this old issue (didn't think it was worth opening another one), but I just wanted to ask @philc if you're planning to make a release any time soon? it'd be good to use the new functionality.

philc:
Yes, I plan to release soon to the store. The holdup is that I've been tracking down a few subtle and tricky regressions related to the new features in this release.

PR #4627, opened 2025-02-10
refactor: remove Selection.type polyfill

### LINKED ISSUE #4626: Suggestion: use native .type() method on the Selection object
Hello!

I was wondering if it would be warranted to remove the polyfill implementation of `getSelectionType(selection)` since selection.type is [available on Firefox as well](https://developer.mozilla.org/en-US/docs/Web/API/Selection/type#browser_compatibility):

https://github.com/philc/vimium/blob/b3cc736015d9a81a4ece46eb9b58ddc883d06f43/lib/dom_utils.js#L542-L553

peterdanwan:
## Description

This potentially closes #4626, which aims to reduce some code that can be replaced by simply using `selection.type`.

### DIFF
--- a/content_scripts/mode_find.js
+++ b/content_scripts/mode_find.js
@@ -396,13 +396,13 @@ FindMode.restoreDefaultSelectionHighlight = forTrusted(() =>
 
 const getCurrentRange = function () {
   const selection = getSelection();
-  if (DomUtils.getSelectionType(selection) === "None") {
+  if (selection.type === "None") {
     const range = document.createRange();
     range.setStart(document.body, 0);
     range.setEnd(document.body, 0);
     return range;
   } else {
-    if (DomUtils.getSelectionType(selection) === "Range") {
+    if (selection.type === "Range") {
       selection.collapseToStart();
     }
     return selection.getRangeAt(0);
--- a/content_scripts/mode_find.js
+++ b/content_scripts/mode_find.js
@@ -8,7 +8,7 @@ class SuppressPrintable extends Mode {
     super.init(options);
     const handler = (event) =>
       KeyboardUtils.isPrintable(event) ? this.suppressEvent : this.continueBubbling;
-    const type = DomUtils.getSelectionType();
+    const initialType = globalThis.getSelection().type;
 
     // We use unshift here, so we see events after normal mode, so we only see unmapped keys.
     this.unshift({
@@ -19,7 +19,7 @@ class SuppressPrintable extends Mode {
         // If the selection type has changed (usually, no longer "Range"), then the user is
         // interacting with the input element, so we get out of the way. See discussion of option 5c
         // from #1415.
-        if (DomUtils.getSelectionType() !== type) {
+        if (globalThis.getSelection().type !== initialType) {
           return this.exit();
         }
       },
@@ -401,12 +401,13 @@ const getCurrentRange = function () {
     range.setStart(document.body, 0);
     range.setEnd(document.body, 0);
     return range;
-  } else {
-    if (selection.type === "Range") {
-      selection.collapseToStart();
-    }
-    return selection.getRangeAt(0);
+  } 
+
+  if (selection.type === "Range") {
+    selection.collapseToStart();
   }
+
+  return selection.getRangeAt(0);
 };
 
 const getLinkFromSelection = function () {
--- a/lib/dom_utils.js
+++ b/lib/dom_utils.js
@@ -322,15 +322,15 @@ const DomUtils = {
     if (element.isContentEditable) {
       const node = selection.anchorNode;
       return node && this.isDOMDescendant(element, node);
+    } 
+
+    if ((selection.type === "Range") && selection.isCollapsed) {
+      // The selection is inside the Shadow DOM of a node. We can check the node it registers as
+      // being before, since this represents the node whose Shadow DOM it's inside.
+      const containerNode = selection.anchorNode.childNodes[selection.anchorOffset];
+      return element === containerNode; // True if the selection is inside the Shadow DOM of our element.
     } else {
-      if ((selection.type === "Range") && selection.isCollapsed) {
-        // The selection is inside the Shadow DOM of a node. We can check the node it registers as
-        // being before, since this represents the node whose Shadow DOM it's inside.
-        const containerNode = selection.anchorNode.childNodes[selection.anchorOffset];
-        return element === containerNode; // True if the selection is inside the Shadow DOM of our element.
-      } else {
-        return false;
-      }
+      return false;
     }
   },
 
--- a/lib/dom_utils.js
+++ b/lib/dom_utils.js
@@ -323,7 +323,7 @@ const DomUtils = {
       const node = selection.anchorNode;
       return node && this.isDOMDescendant(element, node);
     } else {
-      if ((DomUtils.getSelectionType(selection) === "Range") && selection.isCollapsed) {
+      if ((selection.type === "Range") && selection.isCollapsed) {
         // The selection is inside the Shadow DOM of a node. We can check the node it registers as
         // being before, since this represents the node whose Shadow DOM it's inside.
         const containerNode = selection.anchorNode.childNodes[selection.anchorOffset];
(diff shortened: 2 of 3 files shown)


lib/dom_utils.js
```diff
@@ -323,7 +323,7 @@ const DomUtils = {
       const node = selection.anchorNode;
       return node && this.isDOMDescendant(element, node);
     } else {
```
philg-dev, line 325: Same as above. Just a suggestion - I'm not sure how @philc prefers to handle such refactoring changes:

While we're at it, we might as well get rid of the unnecessary level of indentation.

```js
    if (element.isContentEditable) {
      const node = selection.anchorNode;
      return node && this.isDOMDescendant(element, node);
    } else {
      if ((selection.type === "Range") && selection.isCollapsed) {
        // The selection is inside the Shadow DOM of a node. We can check the node it registers as
        // being before, since this represents the node whose Shadow DOM it's inside.
        const containerNode = selection.anchorNode.childNodes[selection.anchorOffset];
        return element === containerNode; // True if the selection is inside the Shadow DOM of our element.
      } else {
        return false;
      }
    }
```

Since we return at the end of the first if block, there's no need for an else at all. And get rid of the unnecessarily verbose `else {return false;}` by just returning `false` after all other conditions didn't return:

```js
    if (element.isContentEditable) {
      const node = selection.anchorNode;
      return node && this.isDOMDescendant(element, node);
    }

    if ((selection.type === "Range") && selection.isCollapsed) {
      // The selection is inside the Shadow DOM of a node. We can check the node it registers as
      // being before, since this represents the node whose Shadow DOM it's inside.
      const containerNode = selection.anchorNode.childNodes[selection.anchorOffset];
      return element === containerNode; // True if the selection is inside the Shadow DOM of our element.
    }
    return false;
```

content_scripts/mode_find.js
```diff
@@ -396,13 +396,13 @@ FindMode.restoreDefaultSelectionHighlight = forTrusted(() =>
 
 const getCurrentRange = function () {
   const selection = getSelection();
-  if (DomUtils.getSelectionType(selection) === "None") {
+  if (selection.type === "None") {
     const range = document.createRange();
     range.setStart(document.body, 0);
     range.setEnd(document.body, 0);
     return range;
   } else {
```
philg-dev, line 404: Just a suggestion - I'm not sure how @philc prefers to handle such refactoring changes:

While we're at it, we might as well get rid of the unnecessary level of indentation.

```js
  if (selection.type === "None") {
    const range = document.createRange();
    range.setStart(document.body, 0);
    range.setEnd(document.body, 0);
    return range;
  } else {
    if (selection.type === "Range") {
      selection.collapseToStart();
    }
    return selection.getRangeAt(0);
  }
```

Since we return at the end of the first `if` block, there's no need for an `else` at all.
The resulting single-statement `if` might also be written without the `{curly braces}` of course. Not sure about the style guide, if there is one.

```js
  if (selection.type === "None") {
    const range = document.createRange();
    range.setStart(document.body, 0);
    range.setEnd(document.body, 0);
    return range;
  }

  if (selection.type === "Range") {
    selection.collapseToStart();
  }
  return selection.getRangeAt(0);
```

philg-dev (changes requested):
# Remaining / Unaccounted references of `getSelectionType` from `lib/dom_utils.js`:

This function also has 7 references in `content_scripts/mode_visual.js` which are unaccounted for in this pull request. Those would completely fail due to this function being removed.

This function also has 2 additional references without the `selection` parameter in `content_scripts/mode_find.js`. Those would also fail completely due to this function being removed.

# My analysis to try to confirm the method is otherwise obsolete:

Aside from that I've taken a look at all the places where the function is called with a parameter and tried to trace back the parameter's origin. To me it looked like all of them originated from either `document.getSelection()`, `globalThis.getSelection()` or `getSelection()`. The latter is equal to `window.getSelection()` which again is a shorthand for `document.getSelection()` according to the note here: https://developer.mozilla.org/en-US/docs/Web/API/Document/getSelection#related_objects

The `selection` reference in objects of the `VisualMode` class are just a reference to the `selection` property `Movement` class which also originate from `globalThis.getSelection()`.

So all of the parameters should have nothing special about them and therefore the behavior should be exactly the same as called without the `selection` parameter because in that case the `getSelectionType(selection)` function determines the `selection` on it's own with `selection = document.getSelection()`. That currently leads to an invariant where the only branch in the function that is ever executed is:
```
    if (selection.type) {
      return selection.type;
    }
```

So the function indeed appears to be obsolete, if I didn't miss any weird side effects.

peterdanwan:
Hello @philg-dev and thank you for your advice!

Following your advice, I have removed the other getSelectionType() function calls as well as removed the nested if-else conditions on the areas that you pointed out.

Please let me know if there's anything else you'd like changed in this PR - I'm happy to make any adjustments.

PS: while I was looking at mode_visual.js, I found a couple of functions with the `return` keyword despite these functions returning `undefined`, for example:

```js
  collapseSelectionToAnchor() {
    if (this.selection.toString().length > 0) {
      return this.selection
        [this.getDirection() === backward ? "collapseToEnd" : "collapseToStart"]();
    }
  }

  collapseSelectionToFocus() {
    if (this.selection.toString().length > 0) {
      return this.selection
        [this.getDirection() === forward ? "collapseToEnd" : "collapseToStart"]();
    }
  }
  ```
  
  I'd be happy to remove these return statements if it's warranted (and in a new issue/PR pair if needed).

File: lib/dom_utils.js [hunk 3]
```diff
@@ -322,15 +322,15 @@ const DomUtils = {
...
       const node = selection.anchorNode;
       return node && this.isDOMDescendant(element, node);
+    } 
+
+    if ((selection.type === "Range") && selection.isCollapsed) {
+      // The selection is inside the Shadow DOM of a node. We can check the node it registers as
+      // being before, since this represents the node whose Shadow DOM it's inside.
+      const containerNode = selection.anchorNode.childNodes[selection.anchorOffset];
+      return element === containerNode; // True if the selection is inside the Shadow DOM of our element.
     } else {
```
philg-dev, line 332: There's still the unnecessarily verbose

```js
} else {
  return false;
}
```

in there, instead of just skipping the `else` and simply having `return false` outside of the block.

content_scripts/mode_find.js
```diff
@@ -8,7 +8,7 @@ class SuppressPrintable extends Mode {
     super.init(options);
     const handler = (event) =>
       KeyboardUtils.isPrintable(event) ? this.suppressEvent : this.continueBubbling;
-    const type = DomUtils.getSelectionType();
+    const initialType = globalThis.getSelection().type;
```
philg-dev, line 11: Renaming the variable makes is a little easier to understand this code fragment - good catch!

philg-dev (changes requested):
Looks much better now - thanks! Just the `else` block remains.

If you get rid of that, I'll see if I can find some time to test the PR at runtime. So far I only looked at the code.

lib/dom_utils.js
peterdanwan: Hi @philg-dev and thank you for looking into this! There's a few more functions in dom_utils.js where I can remove the nested if-else conditions. For example:

- `simulateSelect(element)`
- `getViewportTopLeft()`
- ... etc.

Would you like to me to fix the remaining nested if-else blocks as well?

philg-dev: I'm really not sure - I feel like it's fine for the sections that were touched by your changes anyway... I think for now we should leave it at that and see what philc says about the PR. He might just do those minor refactorings himself if he sees fit.

peterdanwan: Got it, thanks for letting me know! If needed, please let me know if you'd like me to squash my changes as well.

philg-dev (approved):
I've tested the changes locally in Firefox and Chromium now by playing around in Visual Mode, Caret Mode and Find Mode a bit. Everything seems to work as before. So from what I've seen this should be fine to merge.

----

I did notice some **unrelated SyntaxErrors**, that have been there for a long time. I've created an issue #4629 and pull request #4630 for this old bug. For more details take a look at those.

TL;DR for testing:

Navigating in Visual Mode or Caret Mode by using `w` produces this error **in Firefox**... Chromium silently ignores this:

```
SyntaxError: Selection.modify: The granularity argument must be one of: "character", "word", "line", or "lineboundary"
```

PR #4630, opened 2025-02-12
fixes #4629 SyntaxError in Selection.modify

philg-dev:
- fixes #4629  SyntaxError on Selection.modify() due to invalid granularity parameter 'vimword' due to missing return statements

## Bug analysis:

- just for reference - valid parameters for `Selection.modify()`:
  - https://developer.mozilla.org/en-US/docs/Web/API/Selection/modify#granularity
  - side note: Firefox supports fewer options than Chromium, but it doesn't matter in this case
  - side note: Chromium will never throw a `SyntaxError` and instead will fail / return silently when an invalid `granularity` is given
- Vimium differentiates `word` and `vimword`
  - `forward word` is triggered with `e` to **jump forward** to the **end of a word**
  - `backward word` is triggered by `b` to **jump back** to the **start of a word**
  - `forward vimword` is triggered with `w` to jump forward right before the **start of the next word** (therefore including non-word characters after the original word's end)
  - I think there's **no way to trigger `backward vimword`**, but there is a code section that handles that case (and does the same as `backward word`)
- missing `return` statements in several places in `content_scripts/mode_visual.js` in function `Movement.runMovement` lead to the `SyntaxError`
  - the section handling `vimword` functionality runs properly
    - `forward vimword` advances char-by-char and thus doesn't rely on `modify`'s `granularity` parameter
    - `backward vimword` has no binding, but would work fine, since it replaces the `granularity` value by `word` when handing it over as a parameter to `Selection.modify()`
    - neither of those code sections have `return` statements after their work is done
    - this leads to the `else` section at the very end being called for the `vimword` cases which have already been handled

## Implementation notes

- I've added `return` statements to the `vimword` code sections
- I've also added `return` statements to the remaining `if else` blocks to avoid this problem in case the function gets extended in the future
- I've added some short comments to highlight the differences in the two code blocks for `vimword` and `word` since they look confusingly similar at the first glance
- optionally we could completely get rid of the `else` sections and have the code as consecutive `if() { doSomething(); return; }` blocks, since each block is meant to `return` when done
- in one of the cases the existing code returned the result of `.modify()` despite the function not having a return type, so I moved the `return` to a separate line instead

PR #4632, opened 2025-02-12
Fix caret mode w e

### LINKED ISSUE #3075: firefox - w not working currectly in visual mode before selection.
In the visual mode (after pressing 'v') before starting the selection by pressing 'v' again, pressing 'w' takes the cursor one letter ahead instead of one word(same behavior as 'l') however it works fine after starting the selection.

### LINKED ISSUE #3772: Unexpected behaviour of `w` and `e` in caret mode
**Describe the bug**
In vim, `w` and `e` are word motions, but currently in caret mode (just c, not v), both `w` and `e` move the cursor one character forward. This only happens on firefox and not on chromium for me. 

**To Reproduce**
Enter caret mode on any website using firefox and try `e` and `w`.

* Firefox 85.0.2 (64-bit)
* OS: Darwin 19.6.0 Darwin Kernel Version 19.6.0: Tue Jan 12 22:13:05 PST 2021; root:xnu-6153.141.16~1/RELEASE_X86_64
* Vimium 1.66

### LINKED ISSUE #4615: `b` works on caret mode but `w` does not.
**Describe the bug**

Not sure if this is intended behaviour, but it doesn't seem like it. After entering caret mode, `b` works normally. It moves the whole single character long selection back one word at a time. But `w` does something different. It expands the end of the selection to the end of the current word, as if you were in visual mode. After this, subsequent presses of `w` don't do anything at all.

**To Reproduce**

1. Enter caret mode
2. Press `w` and `b`
3. Observe behaviour

```
Vimium: 	2.1.2

Name  	 	Firefox
Version 	134.0.2
Build ID 	20250120135430
OS 	        Windows_NT 10.0 19045
```

### LINKED ISSUE #4629: Firefox: SyntaxError in Selection.modify() when using 'w' to advance Visual Mode and Caret Mode
I've noticed `SyntaxErrors` in the Firefox browser console, when I was testing another Vimium PR.
This is a purely cosmetic problem that has no influence on the functionality - it just clutters the browser console.

```
03:21:43.544 SyntaxError: Selection.modify: The granularity argument must be one of: "character", "word", "line", or "lineboundary" mode_visual.js:93
    runMovement moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/content_scripts/mode_visual.js:93
    commandHandler moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/content_scripts/mode_visual.js:354
    handleKeyChar moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/content_scripts/mode_key_handler.js:140
    onKeydown moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/content_scripts/mode_key_handler.js:75
    keydown moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/content_scripts/mode.js:68
    alwaysSuppressPropagation moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/lib/handler_stack.js:129
    keydown moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/content_scripts/mode.js:66
    bubbleEvent moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/lib/handler_stack.js:61
    installListeners moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/content_scripts/vimium_frontend.js:299
    installListener moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/content_scripts/vimium_frontend.js:281
    forTrusted moz-extension://792162ca-50cd-416e-a4b0-6c8d954d80c6/lib/utils.js:6
```

I've already implemented a fix for this, the pull request will follow shortly.
I'll include a bug analysis and notes on my changes there.

## Steps to reproduce the behavior:

Happens only on Firefox, since Chromium silently ignores invalid `granularity` parameters for the `selection.modify()` function.

1. enter Visual Mode (default: `v`)
2. optional: enter Caret Mode (default: 'c')
3. hit `w` to advance to the start of the next word (make sure the caret has a suitable position)
4. the `SyntaxError` appears in the console

Firefox 135.0
Vimium 2.1.2

philg-dev:
## Description

> Note: I've decided to close PR #4630 and create this PR instead which fixes more issues that are in the exact same place in the code and thus would only create merge conflicts if I were to commit them separately based on the master. Should be much easier to review this way too.

Fix SyntaxError on Selection.modify() due to invalid granularity parameter 'vimword' due to missing return statements
- fixes #4629 SyntaxError in Selection.modify

Fix 'w' and 'e' not working in Caret Mode
- fixes #3075 
- fixes #3772 
- fixes #4615 

Since forward movements for 'word' and 'vimword' use char-by-char custom logic which extends the selection range, we need to collapse the selection to the end, if we're in Caret Mode instead of Visual Mode.

Also removed dead code section for 'backward vimword', as there is no binding that ever uses that and the implemented logic is equal to 'backward word' ('b' key).

## Bug analysis regarding the SyntaxError (commit c0c6f49448de4e063da2ae23ca9f79d65a6cc3a9):

- just for reference - valid parameters for `Selection.modify()`:
  - https://developer.mozilla.org/en-US/docs/Web/API/Selection/modify#granularity
  - side note: Firefox supports fewer options than Chromium, but it doesn't matter in this case
  - side note: Chromium will never throw a `SyntaxError` and instead will fail / return silently when an invalid `granularity` is given
- Vimium differentiates `word` and `vimword`
  - `forward word` is triggered with `e` to **jump forward** to the **end of a word**
  - `backward word` is triggered by `b` to **jump back** to the **start of a word**
  - `forward vimword` is triggered with `w` to jump forward right before the **start of the next word** (therefore including non-word characters after the original word's end)
  - I think there's **no way to trigger `backward vimword`**, but there is a code section that handles that case (and does the same as `backward word`)
- missing `return` statements in several places in `content_scripts/mode_visual.js` in function `Movement.runMovement` lead to the `SyntaxError`
  - the section handling `vimword` functionality runs properly
    - `forward vimword` advances char-by-char and thus doesn't rely on `modify`'s `granularity` parameter
    - `backward vimword` has no binding, but would work fine, since it replaces the `granularity` value by `word` when handing it over as a parameter to `Selection.modify()`
    - neither of those code sections have `return` statements after their work is done
    - this leads to the `else` section at the very end being called for the `vimword` cases which have already been handled

## Implementation notes regarding the SyntaxError

- I've added `return` statements to the `vimword` code sections
- I've also added `return` statements to the remaining `if else` blocks to avoid this problem in case the function gets extended in the future
- I've added some short comments to highlight the differences in the two code blocks for `vimword` and `word` since they look confusingly similar at the first glance
- optionally we could completely get rid of the `else` sections and have the code as consecutive `if() { doSomething(); return; }` blocks, since each block is meant to `return` when done
- in one of the cases the existing code returned the result of `.modify()` despite the function not having a return type, so I moved the `return` to a separate line instead

### DIFF
--- a/content_scripts/mode_visual.js
+++ b/content_scripts/mode_visual.js
@@ -74,9 +74,10 @@ class Movement {
           return;
         }
       }
-      return;
-    } else if (granularity === vimword) {
-      this.selection.modify(this.alterMethod, backward, word);
+      // in Caret Mode collapse selection to the end
+      if (this.alterMethod === "move") {
+        this.selection.collapseToEnd();
+      }
       return;
     }
 
@@ -94,6 +95,10 @@ class Movement {
         if (this.extendByOneCharacter(forward) === 0) {
           return;
         }
+      }
+      // in Caret Mode collapse selection to the end
+      if (this.alterMethod === "move") {
+        this.selection.collapseToEnd();
       }
       return;
     } else {
--- a/content_scripts/mode_visual.js
+++ b/content_scripts/mode_visual.js
@@ -62,35 +62,43 @@ class Movement {
     // Native word movements behave differently on Linux and Windows, see #1441. So we implement
     // some of them character-by-character.
     if ((granularity === vimword) && (direction === forward)) {
+      // extent selection to the end of the 'vimword'
       while (this.nextCharacterIsWordCharacter()) {
         if (this.extendByOneCharacter(forward) === 0) {
           return;
         }
       }
+      // extend selection after the 'vimword' to position before next word
       while (this.getNextForwardCharacter() && !this.nextCharacterIsWordCharacter()) {
         if (this.extendByOneCharacter(forward) === 0) {
           return;
         }
       }
+      return;
     } else if (granularity === vimword) {
       this.selection.modify(this.alterMethod, backward, word);
+      return;
     }
 
     // As above, we implement this character-by-character to get consistent behavior on Windows and
     // Linux.
     if ((granularity === word) && (direction === forward)) {
+      // extent selection to the start of the next 'word' (non-word characters, e.g. whitespace)
       while (this.getNextForwardCharacter() && !this.nextCharacterIsWordCharacter()) {
         if (this.extendByOneCharacter(forward) === 0) {
           return;
         }
       }
+      // extend selection to the end of the 'word'
       while (this.nextCharacterIsWordCharacter()) {
         if (this.extendByOneCharacter(forward) === 0) {
           return;
         }
       }
+      return;
     } else {
-      return this.selection.modify(this.alterMethod, direction, granularity);
+      this.selection.modify(this.alterMethod, direction, granularity);
+      return;
     }
   }
 


philg-dev:
I've noticed a different kind of error that Firefox throws on `granularity` values that would generally be valid, but are unsupported in Firefox. For that I would rather do a separate PR at some other time.

More details can be found in #4633

philc (changes requested):
Generally LGTM. When writing comments, uppercase the first letter of your sentence, and put a period at the end. This is documented in the coding style section of CONTRIBUTING.md.

Also fix this typo: "extent" is used in a few places where "extend" is meant.

philc:
Excellent, thanks!

As a nice followup, it would be good to re-enable the commented-out visual mode related tests in dom_tests.js and get them working. They were disabled sometime ago because they were failing, but we do need the coverage.

PR #3666, opened 2020-09-21
Document Mac modifier key issue in readme

linesarefuzzy:
Adding a simple note to the readme because it took me forever to find this fix:
https://github.com/philc/vimium/issues/3197#issuecomment-614829140

PR #4643, opened 2025-03-07
macOptionFix: Added support for Mac OS Option + Key presses (Fixes #3197, #3666, #3662, #3284)

### LINKED ISSUE #3197: Vimium Mac Alt-based hotkeys don't work
I am using Vimium in Firefox Dev edition 65.0b1 - on Mac OS X. The `a-` hotkey is not working with the alt/option key.

### LINKED ISSUE #3284: Alt keys don't work on mac
`map <a-e> visitPreviousTab` doesn't work on chrome 73 macos.
As well as any other alt key combos.

### LINKED ISSUE #3662: Trigger <a-f> on Mac?
Hey guys, 

How do you trigger the above on a Mac?

I tried Command as well as alt (option) key but neither work for me.

Thanks!

alextorma:
## Description

This PR adds in support for <a-{key}> shortcuts in Vimium on MacOS. On MacOS, when the setting is enabled, Vimium will treat any \<a> and <a-{key}> presses as if keyboard layout is ignored, thereby avoiding Mac's øπ†îøñ ß¥µ∫ø¬ß (aka option symbols).

My apologies if the style isn't up to snuff - I'm somewhat scrappy and not exactly the most familiar with AirBnB style, but I'm happy to make revisions if you let me know what needs to be addressed!

Should fix #3197
Should fix #3666
Should fix #3662
Should fix #3284

### DIFF
--- a/lib/keyboard_utils.js
+++ b/lib/keyboard_utils.js
@@ -36,11 +36,8 @@ const KeyboardUtils = {
   getKeyChar(event) {
     let key;
     if (
-      (!Settings.get("ignoreKeyboardLayout") && this.platform !== "Mac") ||
-      (!Settings.get("ignoreKeyboardLayout") &&
-        (this.platform === "Mac" && !Settings.get("macOptionKeyFix"))) ||
-      (!Settings.get("ignoreKeyboardLayout") &&
-        (this.platform === "Mac" && Settings.get("macOptionKeyFix") && !event.altKey))
+      !Settings.get("ignoreKeyboardLayout") &&
+      !(this.platform === "Mac" && Settings.get("macOptionKeyFix") && event.altKey)
     ) {
       key = event.key;
     } else if (!event.code) {
--- a/lib/keyboard_utils.js
+++ b/lib/keyboard_utils.js
@@ -36,9 +36,11 @@ const KeyboardUtils = {
   getKeyChar(event) {
     let key;
     if (
-      !Settings.get("ignoreKeyboardLayout") && !Settings.get("macOptionKeyFix") ||
-      !Settings.get("ignoreKeyboardLayout") &&
-        (this.platform === "Mac" && Settings.get("macOptionKeyFix") && !event.altKey)
+      (!Settings.get("ignoreKeyboardLayout") && this.platform !== "Mac") ||
+      (!Settings.get("ignoreKeyboardLayout") &&
+        (this.platform === "Mac" && !Settings.get("macOptionKeyFix"))) ||
+      (!Settings.get("ignoreKeyboardLayout") &&
+        (this.platform === "Mac" && Settings.get("macOptionKeyFix") && !event.altKey))
     ) {
       key = event.key;
     } else if (!event.code) {
--- a/lib/keyboard_utils.js
+++ b/lib/keyboard_utils.js
@@ -31,10 +31,83 @@ const KeyboardUtils = {
     }
   },
 
+  // Adds support for Mac Option + Keypress. See #3197.
+  // (Vimium will ignore the symbols and treat them
+  // like letters, at least on a US Laptop keyboard).
+  // I only have a laptop with a US keyboard so I can only get the keypresses that are
+  // available to me, and this list can likely be expanded. I have no idea if Option +
+  // key in different regions produces different symbols or not.
+  //
+  // I suppose another way of handling this would be to simply treat any Option + (Shift? +)
+  // key press as if "Ignore Keyboard Layout" and only use the event.code value for those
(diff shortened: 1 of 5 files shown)


philc:
Nicely done. I'd love to get this fixed.

I haven't tested this on alternate keyboard layouts, but I think it's safe to say that the approach in this PR only work for English keyboard layouts.

I don't think it's scalable to expand the support for this by hard coding the expected key for every available keyboard layout.

I'm leaning towards the approach of always ignoring keyboard layout on Mac when the alt key is pressed. Ignoring the keyboard layout for alt key presses on all platforms would probably break some existing configurations, but it shouldn't break existing configurations, because presumably they're already broken given that Alt keybindings currently don't work.

Let's leave this open for a bit to see if anyone else wants to chime in.

torma616:
@philc, I hear you loud and clear. Try this one on for size! It's confirmed to work on both Mac and Linux (and by "work on Linux" I mean there is no behavior change on that platform as far as I can tell).

I've also been considering the potential for breaking existing configurations too - seeing as it was impossible to have both <a-{letter}> and <a-{symbol}> maps in a configuration in the past (at least, not with both working) I think we might be in the clear. There's probably a more in-depth way to handle translating the user maps on the fly rather than which character was read, but any solution that involves the option symbols will need to be locaized.

What I've done here is to simply treat any `Alt` and `Alt + <key>` press on macOS as keyboard-layout independent. I've also included a toggle on the options page (that only displays on macOS) to enable/disable this behaviour. It is enabled by default for the sake of new users, and the language it uses "Treat Option as Alt/Meta" should be familiar to Vimium's userbase (who I'm assuming are at least familiar with the very similarly named settings toggle in most terminal applications). Still, the new behavior can be disabled if someone so chooses.

With regards to setting a new (sane) default, my $0.02:
>these issues have been reported going back years, and on all the threads I read, I never saw anybody mentioning `<a-{option + key}>` as being a solution/workaround until I mentioned it last week so how many configurations (of people who don't share information back to the community) are we worried about breaking here?

torma616:
Sorry about that double commit just then... the stupid IDE in question with the diff:
image

lib/keyboard_utils.js
```diff
@@ -31,9 +31,17 @@ const KeyboardUtils = {
     }
   },
 
+  // Adds support for Mac Option + Keypress. See #3197.
+  // (Vimium will ignore the symbols and treat them like letters, at least on a US Laptop keyboard).
   getKeyChar(event) {
     let key;
-    if (!Settings.get("ignoreKeyboardLayout")) {
+    if (
+      (!Settings.get("ignoreKeyboardLayout") && this.platform !== "Mac") ||
```
UncleSnail, line 39: In my opinion, this could be made more understandable through the use of grouping (and simplification), like this:
```
(!Settings.get("ignoreKeyboardLayout") &&
(this.platform !== "Mac" || !Settings.get("macOptionKeyFix") || !event.altKey))
```
Note that we can remove a lot of checks because we already know some facts by the time we get to that point in the statement.
So if the `this.platform !== "Mac"` statement fails and we continue checking the `||` we know that `this.platform === "Mac"` we don't need to check it. Same with `!Settings.get("macOptionKeyFix")`. Some might argue checking visibly is more readable though, so it's preference.

It should be something like that, although I might have gotten a detail wrong. In the end, this block is interpreting the key's character value, not literally/ignoring keyboard layout. When we do that is: We haven't selected to ignore keyboard layout and we are either not on a mac, not fixing mac bug, or not using the alt key.
In other words, we do want to ignore keyboard layout if we are on a mac, fixing the mac bug, and pushing alt (negate the expression above).

UncleSnail (changes requested):
This generally looks like a good fix, although the condition can be simplified. Good job.

lib/keyboard_utils.js
UncleSnail: Note that actually doing the negation might be even more understandable to make: ```
(!if (Settings.get("ignoreKeyboardLayout") &&
!(this.platform === "Mac" && Settings.get("macOptionKeyFix") && event.altKey))) {
```
I think this is probably best.

alextorma: I'd definitely agree with you. It was one of those nights, I guess.

Changes incorporated, tests passed (and confirmed myself in Mac/Linux).

UncleSnail (approved):
Everything looks good to me.

philc:
Hey @alextorma, great work on this. In the end I decided that we can skip exposing this as a user-facing option and just ignore keyboard layout on MacOS alt keypresses. Since the help dialog has changed a lot since you wrote this, I couldn't easily rebase your PR, and ended up merging in your narrow change to keyboard_utils.js.

Cheers!

Issue #4661, opened 2025-03-27
Scroll instantly to next result in Find mode


cBournhonesque:
When I'm looking for a term via `/`, and then press `n` to cycle to the next result, vimium scrolls quite slowly to the result.

That's different from the behaviour in Chrome with CTRL+F where the browsers scrolls instantly to the next location.

I've disabled the `Use Smooth Scrolling` option but it didn't change anything

philg-dev:
I found the culprit and implemented a fix for this.

With my fix the scroll behavior will respect the "Use smooth scrolling" in the Vimium options when jumping back and forward between search results in find mode.

philc:
I see that in both Chrome and Firefox, without Vimium, they both scroll immediately to the next search result, even though `page up` and `page down` scroll the page smoothly.

Scrolling immediately to the next find result seems like a reasonable UX choice to me. I don't see why Vimium needs to do it differently. @philg-dev How about we just change Vimium's behave to match the browsers? So: jump immediately to the next find result, not matter the value of smooth scroll.

philg-dev:
> Scrolling immediately to the next find result seems like a reasonable UX choice to me.

I agree - it's probably best to have the same behavior as the native browser search in that regard. Otherwise we would "force" people to either use smooth or instant scrolling for everything, which might not be what most people want for their normal Vimium movements. And splitting the setting to provide a separate flag just to change the search scroll behavior individually seems to be a bit overkill.

PR #4662, opened 2025-03-27
Fixes Scroll instantly to next result in Find mode #4661

philg-dev:
Navigating between search results in find mode now respects the Vimium setting "Use smooth scrolling".

## Description

This is a simple fix for #4661 

In find mode until now the [window.scrollTo()](https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollTo) method was always called with hard-coded `behavior: "smooth"`. With this change it respects the `Settings.smoothScroll` flag when jumping between search results.

I've successfully tested the changes on...
Firefox Version 136.0.3 (64-bit) (Arch Linux)
and
Chromium Version 134.0.6998.165 (Official Build) Arch Linux (64-bit)

## Off-Topic / vaguely related:

I checked the rest of the code base for occurrences of `"smooth"` and `"instant"` and could only find one more hard-coded usage of `"instant"` in `scroller.js` in the `performScroll` function. It might be worth taking a look at that separate from this pull request, as I've noticed some other fairly "strange" behavior with that.

https://github.com/philc/vimium/blob/1d0281f8ce2194e17235f06455579b2275193c0f/content_scripts/scroller.js#L66-L77

The off topic strange behavior I've noticed was:
When using `d` to scroll down half a page, the `performScroll` function was triggered a total of 3 times with scroll `amount` of `1`, `1`, `1`, `10` and `359` pixels in my test case for some reason. I got different values depending on the page I tested it on, . So it might be worth investigating why it's triggering 5 scroll actions instead of just one. I suspect it has to do with scrollable elements somehow - I'm not very familiar with how the scroller is implemented in that regard. On [Wikipedia about JavaScript](https://en.wikipedia.org/wiki/JavaScript) I got 9x 1 pixel scrolls, a 17 pixel and a 344 pixel scroll. Triggering a total of 11 `scrollBy()` calls surely doesn't seem very efficient for a simple `d` press. These tests were done on Firefox.

I tested to implement the same logic to use `"smooth"` or `"instant"` depending on the `Settings.smoothScroll` flag as an experimental change, however using smooth scrolling in that context kinda broke the entire functionality, so that might be the reason why the `"instant"` behavior is hard-coded in that code segment. I've only tested it with `d` and `u` functions, but since those were broken already I didn't bother testing any of the others for that experimental change.

I ran into this issue with the experimental change on Chromium and didn't bother testing on Firefox for now, since it was off-topic anyway.

**The experimental changes are obviously NOT part of this PR.**

### DIFF
--- a/content_scripts/mode_find.js
+++ b/content_scripts/mode_find.js
@@ -475,9 +475,10 @@ const highlight = (textNode, startIndex, length) => {
   const rect = range.getBoundingClientRect();
   if (rect.top < 0 || rect.bottom > globalThis.innerHeight) {
     const screenHeight = globalThis.innerHeight;
+    const scrollBehavior = Settings.get("smoothScroll") ? "smooth" : "instant";
     globalThis.scrollTo({
       top: globalThis.scrollY + rect.top + rect.height / 2 - screenHeight / 2,
-      behavior: "smooth",
+      behavior: scrollBehavior,
     });
   }
 


philg-dev:
The "strange stuff" that I've noticed with the scroller implementation might be related to PR #4328 and it's related issues / previous PRs.

philg-dev:
(Rebased onto the current master)

PR #4673, opened 2025-04-09
Fixes Multiple TODO's in Code

Govi2020:
## Description

All major changes made were made according to TODO's that was in the comments in the code.

- pages/options.js --> Now if there is a error then the window will scroll into view to show the error and input in validation
- content_scripts/mode_visual --> has unneeded closure, which was changed into plain for loop
- content_scripts/linkhints.js --> css was hard coded for popups which was changed just adding a classname 
- content_scripts/vimium.css -> moved the hard coded css to this file as recommend by the dev 

Auto Format code was a pain in the neck since i am using astro vim.Formatting created a lot of unnecessary changes which you can ignore and you can focus your attention on the actual changes that was made.
I have tried my best to follow the contributing guidelines . But I am new to contributing to open source projects , so i would be very grateful if you would point out the problems in my code for me to fix. which i will get working on immediately.

### DIFF
--- a/background_scripts/exclusions.js
+++ b/background_scripts/exclusions.js
@@ -25,7 +25,6 @@ const ExclusionRegexpCache = {
     }
   },
 };
-
 // Make RegexpCache, which is required on the page popup, accessible via the Exclusions object.
 const RegexpCache = ExclusionRegexpCache;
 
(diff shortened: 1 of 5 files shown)


philc:
Hi @Govi2020, thanks for looking into these. There is too much noise in the diffs for me to find the changes you're referring to, and it makes this difficult to merge. You can run `deno fmt` which will format the code base according to our style rules.

Second, probably at least some of these should be broken out into their own PRs so they're easier to review and merge, rather than bundled together.

Govi2020:
I have addressed the problems and did a push request https://github.com/philc/vimium/pull/4678

PR #4678, opened 2025-04-10
Fixes Multiple TODO's in Code

Govi2020:
## Description

All changes made were made according to TODO's that was in the comments in the code.

- pages/options.js --> Now if there is a error then the window will scroll into view to show the error and input in validation
- content_scripts/mode_visual --> has unneeded closure, which was changed into plain for loop
- content_scripts/linkhints.js --> css was hard coded for popups which was changed just adding a classname
- content_scripts/vimium.css -> moved the hard coded css to this file as recommend by the dev

I made originally made changes in [Pull Request](https://github.com/philc/vimium/pull/4673) where due to formatting issues it created a lot of garbage changes which made it difficult to understand the real changes. 

I thought it was better to discard all of that changes and edit from the latest version of vimium. So i have only made some small of incredibly simple changes to the code. This is my first time contributing to a open source project and i intent to make some more changes for this project. Thank you :)

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -424,23 +424,11 @@ class LinkHintsMode {
       this.containerEl.appendChild(el);
     }
 
-    // TODO(philc): 2024-03-27 Remove this hasPopoverSupport check once Firefox has popover support.
-    // Also move this CSS into vimium.css.
     const hasPopoverSupport = this.containerEl.showPopover != null;
     if (hasPopoverSupport) {
       this.containerEl.popover = "manual";
       this.containerEl.showPopover();
-      Object.assign(this.containerEl.style, {
-        top: 0,
-        left: 0,
-        position: "absolute",
-        // This display: block is required to override Github Enterprise's CSS circa 2024-04-01. See
-        // #4446.
-        display: "block",
-        width: "100%",
-        height: "100%",
-        overflow: "visible",
-      });
+      this.containerEl.classList.add("vimiumPopover");
     }
 
     this.setIndicator();
--- a/content_scripts/mode_visual.js
+++ b/content_scripts/mode_visual.js
@@ -536,15 +536,12 @@ class VisualLineMode extends VisualMode {
 
   extendSelection() {
     const initialDirection = this.movement.getDirection();
-    // TODO(philc): Reformat this to be a plain loop rather than a closure.
-    return (() => {
-      const result = [];
-      for (const direction of [initialDirection, this.movement.opposite[initialDirection]]) {
-        this.movement.runMovement(direction, lineboundary);
-        result.push(this.movement.reverseSelection());
-      }
-      return result;
-    })();
+    const result = [];
+    for (const direction of [initialDirection, this.movement.opposite[initialDirection]]) {
+      this.movement.runMovement(direction, lineboundary);
+      result.push(this.movement.reverseSelection());
+    }
+    return result;
   }
 }
 
--- a/content_scripts/vimium.css
+++ b/content_scripts/vimium.css
@@ -196,6 +196,16 @@ iframe.vimium-hud-frame {
   opacity: 0;
 }
 
+.vimiumPopover {
+  top: 0 !important;
+  left: 0 !important;
+  position: absolute !important;
+  display: block !important;
+  width: 100% !important;
+  height: 100% !important;
+  overflow: visible !important;
+}
+
 body.vimium-find-mode ::selection {
   background: #ff9632;
 }
--- a/pages/options.js
+++ b/pages/options.js
@@ -248,7 +248,7 @@ const OptionsPage = {
     if (hasErrors) {
       const error = document.querySelector(".validation-error");
 
-      error?.scrollIntoView({ smooth: true });
+      error?.scrollIntoView();
 
       return;
     }
--- a/pages/options.js
+++ b/pages/options.js
@@ -246,8 +246,10 @@ const OptionsPage = {
   async saveOptions() {
     const hasErrors = this.showValidationErrors();
     if (hasErrors) {
-      // TODO(philc): If no fields with validation errors are in view, scroll one of them into view
-      // so it's clear what the issue is.
+      const error = document.querySelector(".validation-error");
+
+      error?.scrollIntoView({ smooth: true });
+
       return;
     }
 
--- a/background_scripts/exclusions.js
+++ b/background_scripts/exclusions.js
@@ -46,8 +46,6 @@ function getRule(url, rules) {
   }
   // Strip whitespace from all matching passKeys strings, and join them together.
   const passKeys = matchingRules.map((r) => r.passKeys.split(/\s+/).join("")).join("");
-  // TODO(philc): Remove this commented out code.
-  // passKeys = (rule.passKeys.split(/\s+/).join "" for rule in matchingRules).join ""
   if (matchingRules.length > 0) {
     return { passKeys: Utils.distinctCharacters(passKeys) };
   } else {


content_scripts/link_hints.js
```diff
@@ -424,23 +424,11 @@ class LinkHintsMode {
       this.containerEl.appendChild(el);
     }
 
-    // TODO(philc): 2024-03-27 Remove this hasPopoverSupport check once Firefox has popover support.
```
philg-dev, line 427: This TODO comment has just been removed without taking the proper action that it mentions.

Firefox supports the Popover API since January 2025, i.e. Version 125 as can be seen here:
https://developer.mozilla.org/en-US/docs/Web/API/Popover_API#browser_compatibility

Therefore the `hasPopoverSupport` flag can be removed from this piece of code and the if condition removed (obviously while preserving the statements inside the if-block).

Furthermore I think this will also have an impact on the required minimum Version of Firefox that Vimium supports. This is mentioned in the [CONTRIBUTING.md section Coding Style](https://github.com/philc/vimium/blob/master/CONTRIBUTING.md#coding-style).

and needs to be changed in the [manifest.json](https://github.com/philc/vimium/blob/master/manifest.json) as far as I know. From what I've seen the minimum Version for Chromium browsers is also outdated for this piece of functionality.

content_scripts/link_hints.js
```diff
@@ -424,23 +424,11 @@ class LinkHintsMode {
...
-        left: 0,
-        position: "absolute",
-        // This display: block is required to override Github Enterprise's CSS circa 2024-04-01. See
-        // #4446.
-        display: "block",
-        width: "100%",
-        height: "100%",
-        overflow: "visible",
-      });
+      this.containerEl.classList.add("vimiumPopover");
```
philg-dev, line 431: I'm pretty sure @philc just changed all CSS classes to be `dash-case` instead of `camelCase`, so this new class should adhere to the new style. The only `camelCase` classes that he kept were preserved in order to avoid breaking custom CSS for link hints. Since the `vimiumPopover` class didn't exist until you just added it, we can safely use `dash-case` without breaking custom CSS.

See also his large comment about it in vimium.css from [this commit](https://github.com/philc/vimium/commit/f97e139e434e9ae77e16b9971435af030c1fa366#diff-b7bc4b3cd228233b80f4b8bcf87672b0b7b8ecb627073b526d91b5e534345d6d).

content_scripts/vimium.css
```diff
@@ -196,6 +196,16 @@ iframe.vimium-hud-frame {
   opacity: 0;
 }
 
+.vimiumPopover {
```
philg-dev, line 199: Are you sure the `!important` flag is really needed on all of those properties? Generally it's fairly bad practice to use the flag. The MDN mentions some [Best Practices](https://developer.mozilla.org/en-US/docs/Web/CSS/important#best_practices), but that's probably going a bit too far right now.

If you could test the behavior without the `!important` flags and only add them on properties that otherwise get overridden, that would be a good start.

pages/options.js
```diff
@@ -246,8 +246,10 @@ const OptionsPage = {
   async saveOptions() {
     const hasErrors = this.showValidationErrors();
     if (hasErrors) {
-      // TODO(philc): If no fields with validation errors are in view, scroll one of them into view
-      // so it's clear what the issue is.
+      const error = document.querySelector(".validation-error");
```
philg-dev, line 249: I would keep the comment in the code for documentation of this piece of code. Just remove the `TODO(philc):` part.

Aside from that, I think your implementation will always scroll to the first validation error, even if there's more than one and even if one of the fields with a validation error is already in view. So I'm not exactly sure if that's the behavior we want from this.

philg-dev (changes requested):

content_scripts/link_hints.js
Govi2020: but that means people with outdated versions cannot use it.
According to Caniuse (https://caniuse.com/mdn-api_htmlelement_popover) **90.15%** seems to have popover support.
I checked the default version of firefox which comes with kali , and it checks out too.
should we perform the change. or should we write code to not use popover if it does not supports it and write our custom popover element?
what you think is the best decision, i think both are great options  considering 90.15% support of popover

content_scripts/vimium.css
Govi2020: I checked and it seems that  div.vimium-reset was overriding it (due to increased specificity than .vimiumPopover). i have changed .vimiumPopover to  div.vimium-popover. I am not sure if it is the best way to go. but i will commit for you to review the changes

pages/options.js
Govi2020: You are right .I missed "If no fields with validation errors are in view" part.
I have fixed this in the new commit

content_scripts/link_hints.js
philg-dev: Well, I'm not the one to make that decision. I'm just another recent contributor and I've had an eye on most things that have been going on with the repo in the past couple of months.

I guess my main point was: **either** you don't remove the flag & check and leave the TODO just as-is, because nothing has been done about that part of the TODO **or** you follow the instructions of the TODO and remove it. In your change, you just removed the TODO comment and didn't change the related code.

In this case it's probably best to ask @philc for his opinion. Judging from the TODO comment, I would assume he's fine with removing the check. I'm not so sure what he thinks about the changes in minimum supported browser versions though.

pages/options.js
```diff
@@ -243,12 +243,34 @@ const OptionsPage = {
     return result;
   },
 
+  isElementInView(element) {
```
philg-dev, line 246: Hmm... I've seen this or very similar functionality used in different spots around the project. Maybe this would be a good chance to centralize it?

@philc do you think this function would be better suited to be put into `dom_utils.js`?

I've seen **similar but not identical** usages in:
- [`scroller.js#checkVisibility`](https://github.com/philc/vimium/blob/385ee8e1b436c315760205faee601fe28e3faddd/content_scripts/scroller.js#L163)
- [`mode_find.js`](https://github.com/philc/vimium/blob/385ee8e1b436c315760205faee601fe28e3faddd/content_scripts/mode_find.js#L474)
- [`mode_visual.js`](https://github.com/philc/vimium/blob/385ee8e1b436c315760205faee601fe28e3faddd/content_scripts/mode_visual.js#L341)
- [`visibility_test.html`](https://github.com/philc/vimium/blob/385ee8e1b436c315760205faee601fe28e3faddd/test_harnesses/visibility_test.html#L35)

It might be worth considering consolidating at least some of those implementations.

pages/options.js
```diff
@@ -243,12 +243,34 @@ const OptionsPage = {
...
 
-      error?.scrollIntoView();
+      let isAnyElementInView = errors.some((element) => this.isElementInView(element) == true);
+
+      if (isAnyElementInView) {
+        return;
+      }
+
+      errors[0]?.scrollIntoView({
+        behavior: "smooth",
```
philg-dev, line 271: It's a very minor thing, but I'm not sure if we want to respect Vimium's `smoothScroll` flag from the Settings in this case.

I just recently looked into the scroll behavior and found that the `smoothScroll` flag seems to be respected in almost all cases. One exception was `find_mode`, where we now explicitly decided to use `instant` scroll behavior instead of the previously hard-coded `smooth` scroll. See also #4661 for more details on the discussion.

I do think that hard-coded smooth scroll in this use-case could be a reasonable decision, since it is a potential scrolling action that is not explicitly triggered by the user. When the user hits the "Save" button in the options, they don't necessarily expect the viewport to suddenly scroll elsewhere. So smooth might be better in terms of visual feedback to the user.

philg-dev (commented):
Overall the PR consists of a bunch of changes that are otherwise unrelated to each other. As you can see from the comments, even some tiny changes can sometimes lead to further discussions / considerations... So in the future it might be better to split your PRs into **self-contained units that semantically belong together**. 

See https://github.com/philc/vimium/pull/4673#issuecomment-2791573380
> Second, probably at least some of these should be broken out into their own PRs so they're easier to review and merge, rather than bundled together.

pages/options.js
Govi2020: Yes i do think hard coding scroll behaviour in this particular instance is good.if it is necessary i can use the Settings 👍🏻

content_scripts/link_hints.js
philc: I'd probably leave the TODO for a few more months until Firefox 125+ adoption creeps up higher, and then remove it.

philg-dev (changes requested):
As per my comments on some of the changes, I'm leaving this code review as a partial review. I've only gone over these files and then stopped for reasons mentioned in my comment on `content_scripts/mode_visual.js:L334`.

- `background_scripts/exclusions.js`
- `content_scripts/link_hints.js`
- `content_scripts/mode_visual.js`

**I did not review the changes in the remaining files.**

content_scripts/mode_visual.js
```diff
@@ -331,18 +331,8 @@ class VisualMode extends KeyHandlerMode {
...
-            selectionRect = { top: 0, bottom: 0, left: 0, right: 0, width: 0, height: 0 };
-          }
-        }
-        selectionRect = Rect.intersect(
-          selectionRect,
-          Rect.create(0, 0, globalThis.innerWidth, globalThis.innerHeight),
-        );
-        if ((selectionRect.height >= 0) && (selectionRect.width >= 0)) {
+
+        if (DomUtils.isElementVisible(this.selection.getRangeAt(0), { partialCheck: true })) {
```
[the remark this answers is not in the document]
Govi2020: Yea, you are right i was just too fast and excited to jump in and make change. I'll revert the changes for now and let it stay as it is when i can code.
I am very sorry for the inconvenience as i said i am new to contributing to open projects. I mostly worked on personal projects or work that i maintained. It was my mistake

content_scripts/link_hints.js
Govi2020: I have added it

pages/options.js
Govi2020: @philc what do u think is the right move here

philc:
Hey @Govi2020, I'm going to close this. None of these are high priority and they're relatively expensive for me to carefully review and test. Generally, a good practice (and what has generally been the case in Vimium) is to fix the low-priority TODOs when working in that part of the codebase on a nearby feature or bug fix. That way it's not a complete context switch for me, the reviewer. Thank you for the contribution!

Issue #4675, opened 2025-04-09
Highlight scroll hints with another color


Krasav4k1:
Is it possible to add some functionality to switch scroll if the website has multiple?
I know that when you call LinkHints by 'f' it also select shows you window with scroll which you can switch, but on some websites there are a lot of hints and difficult to find the hint which is responsible to switching the scroll.
With using `Use the link's name and characters for link-hint filtering` it's easier to switch scroll, because it show hint with name `scroll`, but if I don't want to enable link-hint filtering.

I would expect hints which related to scroll show in different color, to make them noticeable.
Or maybe automatically assign number for scroll, if you don't use link-hint filtering.

danielcovill:
> With using Use the link's name and characters for link-hint filtering it's easier to switch scroll, because it show hint with name scroll, but if I don't want to enable link-hint filtering.
This is something that's bothered me as well. I never understood why you have to use link-hint mode to get "Frame:" and "Scroll:" appended to the link. If we add those tips to the markers in default mode, I think it would solve this problem, but it seems like there might be a reason that's not already done. If there's no objection, I might try to implement that.

danielcovill:
I can't find a good reason why the solutions for Alphabet hints and Filter hints are different. Showing the "Scroll." and "Frame." tips is really helpful and with a minor tweak seems like it can be available to both. I'll put in a PR for this and hopefully it'll either either be accepted or we'll find out what I'm missing.

PR #4689, opened 2025-04-27
Add linkText to AlphabetHints 

danielcovill:
This should make it easier to discern scroll/frame markers from links. It's a feature that has been only available with Filter hints enabled. I may be missing something, but I see no reason why it shouldn't be visible when using the default Alphabet hints as well. 

https://github.com/philc/vimium/issues/4675


itpropro:
@philc any feedback?

philc:
I'll look at this soon.

philc:
UX comments:

Adding "scroll" and "frame" to the hint text can add a lot of clutter to the page and obscure nearby hints. The UX rationale for only having it in filtered hints has been because the clutter in that mode is less of a problem, because you can very quickly type a few characters that you see in the anchor, or in the hint itself, to hide the many other hints on the page. With alphabet hints, you can't quickly filter out the clutter like this.

Showing "scroll" and "frame" in filtered hints isn't my favorite because it's not self-explanatory. However, at least that mode has the convention of showing the name of some links in the hints (e.g. for form controls). In alphabet hints, having the word "scroll" appear next to a hint feels even less self-explanatory and more random, because I expect the content of the hints to only be characters that I should type to navigate to the link.

I think a change in color might work better, or some other UX. Any thoughts?

Feedback about this PR:

1. Which page are you primarily testing this on?
2. If this PR, there should only be link text shown for frames and scrollables. I noticed the values of the form controls on the Vimium options page are also showing link text. See screenshot.
Screenshot 2025-06-28 at 6 36 07 PM

Also, I found an adjacent bug which is unrelated to this PR. I tested this change on test_harnesses/iframe.html in the Vimium repo. On that page, after hitting space to toggle the hints, typing S navigates to the first link starting with an S, but instead it should wait for further characters to be typed. Update: I believe I fixed this in 544b0ce7ad6be1946cc90d6fdfcbc9b20e568d20.

content_scripts/link_hints.js
```diff
@@ -792,18 +792,21 @@ class AlphabetHints {
     this.hintKeystrokeQueue = [];
   }
 
+  renderMarker(marker) {
+    let linkText = marker.linkText;
+    console.log(linkText);
```
philc, line 797: Log statement left in.

content_scripts/link_hints.js
```diff
@@ -792,18 +792,21 @@ class AlphabetHints {
...
+    const caption = marker.hintString.toUpperCase()
+      + (marker.localHint.showLinkText ? ": " + linkText : "");
+    marker.element.innerHTML = spanWrap(caption);
+  }
+
   fillInMarkers(hintMarkers) {
     const hintStrings = this.hintStrings(hintMarkers.length);
-    if (hintMarkers.length != hintStrings.length) {
```
philc, line 797: Is this safety check no longer needed?

danielcovill:
- Thanks for the callouts on the review, both the log statement and safety check were sloppy mistakes. I've addressed those.
- I am now testing using test_harnesses/iframe.html and the Vimium options page.
- I revisited and changed the approach based on your comments. I've stopped appending the text and when a link is "Scroll." or "Frame." I'm changing the color of the hint-marker instead.
- Additionally, I found that the hint markers for scroll/frame frequently end up stacked behind other markers so I added a bit of code to check for that scenario and add a slight offset (2px down and right) to any markers that are in the exact same spot. I think it provides a nice indication that there are markers behind one another. 

Happy to remove that last bit if it exceeds the scope of this PR.

Krasav4k1:
Is there any approximate understanding about these changes are going to be merged and released with next release?

icalvo:
My vote for accepting this PR, telling apart scroll/frame markers is important in pages with lots of scrollable elements. Also, `gf` doesn't work on scrollable elements so `f` is the only way to activate and scroll them.

danielcovill:
Hi @philc - I'd love a review if possible, please. Thanks

PR #4710, opened 2025-06-08
Fix chrome store link in README

sandorex:
## Description
The original link does not work anymore so i just updated it to the one in sidebar


philc:
Great, thank you!

Issue #4491, opened 2024-06-07
No scrolling after quick j-k


oriongonza:
**Describe the bug**

No scrolling after pressing k j (stands still)
**To Reproduce**

1. press j
2. release j
3. press k

UncleSnail:
I cannot reproduce this issue. What browser/website are you using? Is it happening for you in all browsers or all websites?

UncleSnail:
> Chrome, all websites. It happens only in linux (nixos, wayland). In my windows machine there's a slight stutter of a fraction of a second when you do it and it keeps going. In my linux it just stops

Okay. I am also on Wayland Linux (Arch with Sway wm) and I cannot reproduce this issue in either Firefox or Chromium. I would guess it is likely an issue caused by something else in your setup, like another addon, or your key polling rate of your setup, etc. Some keyboards also do weird things with the key outputs instead of sending out the "direct" signals, so maybe try using a different keyboard and see if it's fixed.
You could also try running your Nix install script on another machine and seeing if you still have the problem or if it is just that machine being slow for some reason.
Unfortunately, unless I can find a consistent way to reproduce this I don't see how we can debug it to fix the problem, especially if the problem isn't actually in Vimium.

gdh1995:
How about changing screen DPI to 100%?

UncleSnail:
My dpi is 100% I believe. I would need to check to make sure, but I don't
know why this would effect anything. Does changing it make a difference for
you?

Drednaught608:
This seems to happen for me when I do so in this order:
press j
press k
release j
Or
press k
press j
release k

This order can happen whenever you're switching between scrolling with j & k in quick succession. The releasing of the opposite scroll key results in the scrolling stopping no matter the direction you're moving.

UncleSnail:
> This seems to happen for me when I do so in this order: press j press k release j Or press k press j release k
> 
> This order can happen whenever you're switching between scrolling with j & k in quick succession. The releasing of the opposite scroll key results in the scrolling stopping no matter the direction you're moving.

Yes, there is a short pause when switching directions, but I believe that is the correct functionality. It should stop until your key repeat timeout is up (how long your computer takes before holding a key makes it type that key a bunch of times quickly). For me everything seems to be working correctly. If for you the scrolling actually stops and stays stopped, make sure when you hold down a key that key eventually starts repeating (IE, you haven't turned off repeating on your system) and if repeating works and the scrolling stops, let me know.

philc:
[...]

Issue #2857, opened 2017-12-08
j/k Scrolling Inconsistency


rmanne:
Arrow key behavior:
1. Press and hold down arrow; Page starts scrolling down
2. Press and hold up arrow; Page starts scrolling up
3. Release down arrow; Page continues scrolling up

Hold j/k behavior:
1. Press and hold "j"; Page starts scrolling down
2. Press and hold "k"; Page starts scrolling up
3. Release "j"; Page stops scrolling, instead of continuing to scroll up

Chrome 62.0.3202.94 (Official Build) (64-bit)
Vimium 1.61.1
(Adding OS info in case it's the source of the inconsistency: macOS Sierra 10.12.6)

BTW:
(neovim) NVim 0.2.1 j/k follows above Arrow key behavior
(vanilla vim) VIM - Vi IMproved 7.4 j/k follows above Arrow key behavior

Will try this again as soon as I get home to my linux desktop (xorg + xmonad, no DE).

mrmr1993:
Generally, holding a scroll down key in while you're also holding the scroll up key (and intending to scroll up) is a strange thing to do. Is there a use case for this?

To support smooth scrolling, the scrolling algorithm keeps running until another key is pressed, a key is released, or the window loses focus. As far as I can see, there is no nice way to change this behaviour without managing a stack of active scrolling instances, all for little/no benefit.

rmanne:
So I mentioned the hold behavior but this is also true for close presses. If you press and hold "j" and just as you're about to release "k", you won't scroll all the way back up, but rather, a single scroll tick.

This is just a special case of the one described above, and while it's not a frequent thing, it does happen from time to time. I only really noticed and started figuring out how to reproduce it when I was playing around earlier.

smblott-github:
It's not obvious that there is a bug, or unreasonable Vimium behaviour here.

philc:
I've never run into this, but I can see how it can be annoying if one rolls from keypress to keypress. This is fixed by #4717.

PR #4717, opened 2025-06-16
fix scrolling gets stuck on keyup event of a different key + fix typo

NickKhalow:
## Description

Using vimium, I spotted an issue with scrolling. When you press 'k' or 'j' to scroll in opposite direction and keep holding previous key it starts to scroll normally, but if you unpress the previous key the scroll stops despite you are holding the new key.

Issue was caused by undistinguishment in `keyIsDown` property of `CoreScroller` that listens for every `keyup` event and it will cancel the scroll without checking what exact key has been unpressed.

The issue is fixed by memorizing `event.code` (which holds keycode of a key, for example `KeyJ`) in the `keydown` event and checking its equality later in the `keyup` event to avoid the case described in the current PR.

Also there is a minor fix of typo from `installCanceEventListener` to `installCancelEventListener`.

### DIFF
--- a/content_scripts/scroller.js
+++ b/content_scripts/scroller.js
@@ -188,11 +188,11 @@ const CoreScroller = {
   init() {
     this.time = 0;
     this.lastEvent = this.keyIsDown = null;
-    this.installCanceEventListener();
+    this.installCancelEventListener();
   },
 
   // This installs listeners for events which should cancel smooth scrolling.
-  installCanceEventListener() {
+  installCancelEventListener() {
     // NOTE(smblott) With extreme keyboard configurations, Chrome sometimes does not get a keyup
     // event for every keydown, in which case tapping "j" scrolls indefinitely. This appears to be a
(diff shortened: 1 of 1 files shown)


philg-dev:
At the first glance it looks promising to me. I don't have time to test this PR right now though.

Mostly just commenting because I had in mind that there were related issues to this:

This would fix #4491 as the issue is described well in [this comment](https://github.com/philc/vimium/issues/4491#issuecomment-2395164402).

And #2857 seems to describe the same issue.

NickKhalow:
Thank you for reply.

The linked issues seem to describe the exact bug the PR solves. 

I suggest to close them as solved when the PR gets merged.

NickKhalow:
Did someone have a chance to test the PR?

philc:
I merged this in 8b61539e27386f36848c9cdf3ed4e11d34645896. Great fix @NickKhalow, and thanks for the helpful repro case @philg-dev.

PR #4718, opened 2025-06-18
Add `goUp` options to pop URL anchor/fragment/hash & query string. Fixes #3763

### LINKED ISSUE #3763: goUp command should work with page anchors (#xxx), query strings (?xxx), and sub-domains (xxx.)
Currently, the `goUP` command behaves unaware of page anchors (#xxx), query strings (?xxx), and sub-domains (xxx.). When first using `goUP` command, anchor and query is removed at along with the last part of the directory. And after using the last `goUP` command, sub-domains are not removed and it stays on the same URL. Lack of these features is the primary reason why I still keep the separate [NavigUp extension](https://chrome.google.com/webstore/detail/navigup/eigojmmlmgigbldiclbpibmjnlbdpiaa) around:

> Navigate to parent address. It means removing page anchor (#xxx), query string (?xxx), file or the last directory (/xxx), port number (:xxx), sub-domain (xxx.) or secure layer (https -> http), whichever is first found (preserving mentioned order).

Removing the secure layer (https -> http) is probably unnecessary in 2021 (the extension is from 2013), but the rest I find very useful. For when you actually want to remove the achors and queries at once, the extension has additional shortcuts:

> Use Alt+Up (Shift = +1, Ctrl = +2 levels) or click the icon (many times = many levels).
> ...
> Use (Ctrl+)(Shift+)Alt+Up for it. Shift means going up one more level, Ctrl - additional two levels. Ctrl and Shift can be combined. If you click extension's icon (in the address bar) many times within one third of second, then you'll go up many levels at once.

Vimium already has this implemented in a more intuitive way (number and then the `goUP` command shortcut), so I would love to use that instead, if `goUP` is able to navigate in more parts of the URL like the NavigUp extension.

In the meanwhile, I think the `goToRoot` command should probably stay as it is, without stripping the subdomain, as stripping the subdomain would be achievable with a subsequent `goUP` command if desired.

(A portion of this potential feature was previously mentioned by one of the maintainers in #1344.)

ravindUwU:
## Description

This PR,

- Changes how `goUp` manipulates the URL, using a `URL` object to parse/get/set its components where possible.

- Adds 2 count-compatible Boolean options to `goUp`:
  - `popAnchor` removes the hash/anchor/fragment/hash from the URL.
  `example.com/a?b=c#d` → `example.com/a?b=c`.

  - `popQuery` removes query params and the hash from the URL.
  `example.com/a?b=c#d` → `example.com/a`.

## Usage

- `map gu goUp`  
  `example.com/a/b?c=d#e` → `example.com/a` → `example.com`

- `map gu goUp popAnchor`  
  `example.com/a/b?c=d#e` → `example.com/a/b?c=d` → `example.com/a` → `example.com`

- `map gu goUp popQuery`  
  `example.com/a/b?c=d#e` → `example.com/a/b` → `example.com/a` → `example.com`

- `map gu goUp popAnchor popQuery`  
  `example.com/a/b?c=d#e` → `example.com/a/b?c=d` → `example.com/a/b` → `example.com/a` → `example.com`

## Subdomains 🤔

#3763 also suggests popping subdomains, but it's unclear how/if port numbers should be manipulated when navigating up a subdomain. e.g., should `goUp` from `owo.example.com:1337` navigate to,
- `owo.example.com` (dropped port)?
- `example.com:1337` (dropped subdomain)?
- `example.com` (both dropped)?

IMO, there is no correct option here because it depends on what's at the URL, so would it be best to leave the host name as-is?

### DIFF
--- a/background_scripts/all_commands.js
+++ b/background_scripts/all_commands.js
@@ -132,6 +132,7 @@ const allCommands = [
     advanced: true,
     options: {
       popAnchor: "Remove the anchor/fragment/hash from the URL, if present.",
+      popQuery: "Remove query parameters from the URL, if present.",
     },
   },
 
--- a/background_scripts/all_commands.js
+++ b/background_scripts/all_commands.js
@@ -130,6 +130,9 @@ const allCommands = [
     desc: "Go up the URL hierarchy",
     group: "navigation",
     advanced: true,
+    options: {
+      popAnchor: "Remove the anchor/fragment/hash from the URL, if present.",
+    },
   },
 
   {
--- a/content_scripts/mode_normal.js
+++ b/content_scripts/mode_normal.js
@@ -135,16 +135,17 @@ const NormalModeCommands = {
 
   // Url manipulation.
   goUp(count) {
-    let url = globalThis.location.href;
-    if (url.endsWith("/")) {
-      url = url.substring(0, url.length - 1);
+    const url = new URL(globalThis.location.href);
+
+    // Pop path segments.
+    if (url.pathname != "/") {
+      url.pathname = url.pathname.split("/").slice(0, -count).join("/");
+      url.search = "";
+      url.hash = "";
     }
 
-    let urlsplit = url.split("/");
-    // make sure we haven't hit the base domain yet
-    if (urlsplit.length > 3) {
-      urlsplit = urlsplit.slice(0, Math.max(3, urlsplit.length - count));
-      globalThis.location.href = urlsplit.join("/");
+    if (globalThis.location.href !== url.toString()) {
+      globalThis.location.href = url.toString();
     }
   },
 
--- a/content_scripts/mode_normal.js
+++ b/content_scripts/mode_normal.js
@@ -134,14 +134,22 @@ const NormalModeCommands = {
   },
 
   // Url manipulation.
-  goUp(count) {
+  goUp(count, { registryEntry }) {
+    let c = count;
     const url = new URL(globalThis.location.href);
 
+    // Pop anchor.
+    if (c > 0 && registryEntry.options.popAnchor && url.hash !== "") {
+      url.hash = "";
+      --c;
+    }
+
     // Pop path segments.
-    if (url.pathname != "/") {
-      url.pathname = url.pathname.split("/").slice(0, -count).join("/");
+    if (c > 0 && url.pathname != "/") {
+      url.pathname = url.pathname.split("/").slice(0, -c).join("/");
       url.search = "";
       url.hash = "";
+      --c;
     }
 
     if (globalThis.location.href !== url.toString()) {
--- a/content_scripts/mode_normal.js
+++ b/content_scripts/mode_normal.js
@@ -144,6 +144,13 @@ const NormalModeCommands = {
       --c;
     }
 
+    // Pop query params.
+    if (c > 0 && registryEntry.options.popQuery && url.search !== "") {
+      url.search = "";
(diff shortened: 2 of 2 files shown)


philc:
Please expand on your use cases which requires popping the query string or anchor often enough for it to be a command or option.

For my own usage, I think the default behavior of `goUp` today is what I most generally want: go one path upward, and remove any query strings or anchors along the way, without needing several invocations. If accurate, then we shouldn't change that default behavior by making the first invocation remove the query string or hash anchor if present.

I find removing query strings and anchors to be rare. When I need to do it, I use `ge` to edit the current url, and ctrl-w on Mac (which deletes a whole word) to remove hunks of the URL. This is fast enough for me. If I wanted to copy and paste the URL without the anchor portion, I would copy it with `yy` and edit it after pasting.

ravindUwU:
Hi @philc 👋

> expand on your use cases which requires popping the query string or anchor

**Personally, removing the anchor is something I do quite often!** e.g., I'd follow a link to the documentation for the [`System.IO.DirectoryInfo.GetFiles(String, EnumerationOptions)`](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.getfiles?view=net-9.0#system-io-directoryinfo-getfiles(system-string-system-io-enumerationoptions)) .NET method overload. This link has the anchor `#system-io-directoryinfo-getfiles(system-string-system-io-enumerationoptions)`, which I'd like to remove to form the [link](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.getfiles?view=net-9.0) to the documentation of the .NET method group. My options are to,
1. Visit the link to the .NET method group from elsewhere on the same page (e.g., sidebar). _But, what if there is no sidebar, or it is collapsed or out of view by default?_
2. (Similar to what you suggested) <kbd>Ctrl</kbd>+<kbd>L</kbd> to focus the URL bar, <kbd>End</kbd> to go to the end of the URL, <kbd>Ctrl</kbd>+<kbd>Backspace</kbd> 9 times and then a final <kbd>Backspace</kbd> to delete word-by-word until the anchor and the `#` is removed, finally <kbd>Enter</kbd> to accept. _But, that's a _lot_ of key presses!_
3. Or if `goUp` can remove the anchor _if it exists_, simply, `goUp`.

~Popping the query string on the other hand isn't something I do often.~ (Edit because this just happened). **Popping the query string might be useful when**, e.g., I'm on the Discord server search page for a specific search term <https://discord.com/servers?query=huh> and I want to drop the search term (in this case `?query=huh`) to visit their server list, which is simply <https://discord.com/servers>. My options are to,
1. Scroll uppppppppppp to locate the "x" button in the search bar on the page, and press it (which in this case is possible with Vimium; but there are many websites with small, interactive elements that aren't marked as such using accessibility attributes (`role`, `tabindex`, etc.), and therefore aren't detected by Vimium).
2. <kbd>Ctrl</kbd>+<kbd>L</kbd>, <kbd>End</kbd>, <kbd>Ctrl</kbd>+<kbd>Backspace</kbd> twice, <kbd>Backspace</kbd> once more, <kbd>Enter</kbd>.
3. Or if `goUp` can remove the query string _if it exists_, simply, `goUp`.

> then we shouldn't change that default behavior ...

Agreed; the default behaviour should not change. This PR adds these as _opt-in_ options so they are disabled by default, and can be enabled by those who find them useful.

ravindUwU:
Hi @philc 👋, thoughts on this PR? 🤔

PR #4756, opened 2025-08-30
Center align "Save changes" button label in Options

breadtk:
## Description

Within the Options page, the `#save` button is displayed as "No changes"
(10 chars) on load. When `onUpdated()` is called in `pages/options.js`,
it turns into "Save changes" (12 chars). This commit ensures that the
update label is appropriately centered, even if that means less padding
on left/right, as opposed to having more padding on left/right in the
fixed size button.

### DIFF
--- a/pages/options.css
+++ b/pages/options.css
@@ -311,6 +311,9 @@ footer {
 
 #save,
 #exclusion-add-button {
+  align-items: center;
+  display: flex;
+  justify-content: center;
   white-space: nowrap;
   width: 110px;
 }


UncleSnail:
I think the preferred way to do this might be to remove the `width` property from the `#save` styles. That would mean that the button size can change though, so it depends on preference. However, this problem is likely to happen again in the future since the Add Rule button, and potentially other buttons, also share the `#save` style. It might be better just let them set their own width to avoid accidentally letting the text overflow. We could also use a `min-width` instead, and then future accidentally longer text wouldn't overflow.

breadtk:
I'm fine either way.

@ilya / @philc - Do you folks have a preference either way?

philc:
I think it's fine, and conceptually simpler, to have no fixed width on the save button. Maybe a fixed width was more important in previous versions of the Options page layout prior to flexbox.

I've made the change in 8b7ce14ff70a1fd1e90c6cf66231fbcfc7aac1b3. Thanks @breadtk.

PR #4764, opened 2025-09-19
Added settings keys sorting

cub-uanic:
I'm using Git to store my dotfiles and share them between several hosts.
Sometimes, on different hosts, I'm getting the keys in vimium-options.json in a different order.
This makes it difficult to understand exactly what was changed (while in reality, it could be just one line), and also makes diffs larger.
As a solution, I suggest always sort the keys of the settings object.


philc:
Nice find @cub-uanic. I slightly adapted your change in bb2641331abf861ae690d78dee5118988f096fb5.

Issue #4698, opened 2025-05-18
Title: Hint keys stack on top of each other, making it hard to select elements


phucleeuwu:
# Description:
When using the f key to trigger link hints in Vimium, the hint labels often overlap significantly, especially on pages with many tightly packed clickable elements. This makes it difficult or impossible to distinguish and select the correct element.

# Steps to reproduce:
	1.	Open a web page with a dense UI (e.g., a dashboard or toolbar with many buttons).
	2.	Press f to trigger link hints.
	3.	Observe how the hint labels are stacked on top of each other.

# Expected behavior:
Hint labels should be spaced in a way that keeps them legible and clearly associated with their corresponding elements.

# Actual behavior:
Hint labels overlap or stack, especially when buttons/icons are close together. This hinders usability.

# Screenshot:

![Image](https://github.com/user-attachments/assets/56122b3b-4b3a-4973-b587-b0ce974b105e)

# Environment:
	•	OS: macOS 
	•	Browser: Zen
	•	Vimium version: always latest

# Suggestions:
	•	Improve spacing algorithm for hint labels.
	•	Add an option to enable line/arrow connectors from label to target element.
	•	Allow user-defined padding between hint boxes.

philg-dev:
You can hit `<space>` when link hints are active to alternate the layering of the hints. Hope this helps.

phucleeuwu:
> You can hit `<space>` when link hints are active to alternate the layering of the hints. Hope this helps.

I mean the improve the hints generator

dithmer:
IMO the hint generator does exactly what it should there. It might be a nice option on first glance, to "improve" it by removing potential duplicates. But that also takes a bit of control away. The `<space>` option @philg-dev mentioned, was new to me as well, but should be everything one need here. 

Otherwise it might happen, that the improvements takes away a possible link option, I want to click, solely because it thought, that I do not need it. The decision should stay in the users hand.

seki-mei:
maybe adding some transparency to the hint boxes in your custom css might improve legibility?

Issue #3454, opened 2020-01-03
Feature Request: Option to Flatten Linkhints


alivefromupstairs:
![linkhints being redundant and on top of one another in the wild.](https://user-images.githubusercontent.com/46059092/71703994-09462800-2dd8-11ea-8aee-533d6b271b6a.png)

**Problem:** There are linkhints that are redundant (e.g., the fullscreen button in the bottom right of the player controls with five linkhints) and inaccessible due to being on top of one another (e.g., the play button, the entirety of the right side of the nav with some of them having three linkhints)

**proposal:** If the linkhints overlap, there is a pretty good chance they do the same thing. Can we get an option to detect overlap and flatten linkhints to a single one?

---

Another sensible thing to consider is always making sure the first character is visible.

gdh1995:
Just an AD: I have a customized version of Vimium - [Vimium C](https://github.com/gdh1995/vimium-c) - and it applys some aggressive rules to exclude some hints.

alivefromupstairs:
@gdh1995 Switched to your extension. However, I'll keep this issue open if the folks here want to add this.

innaterebel:
A workaround is that the overlapped hints are in fact accessible (make readable) by <kbd>Space</kbd> key.

PR #4766, opened 2025-09-21
feat: suppress overlapping hints

geekbrother:
## Description

Adds a new user setting to control whether overlapping link-hint markers are visually de-duplicated. 
When enabled, only one marker is shown per overlapping region to reduce clutter.

### Motivation

Overlapping hints can stack and obscure each other in dense UIs.
Most of the overlapping hints on modern websites point to the same link to support tapping on mobile devices, since it's limited by the finger width.
Users may prefer either cleaner visuals or full visibility of all hints; this makes it configurable.

### Changes

* New setting: suppressOverlappingHintMarkers (default: true) in lib/settings.js.
* Options UI: Checkbox “Hide overlapping link-hint markers” added to pages/options.html.
* Options binding: Registered the setting in pages/options.js so it saves/restores with other preferences.
* Behavior gating: Calls to suppressOverlappingMarkers() in content_scripts/link_hints.js now respect the new setting.

### Default Behavior

* Enabled by default: overlapping markers are hidden except for one per overlapping region.
* Users can disable this in the Options page to show all overlapping markers.

### User-facing UI

* Options → “Hide overlapping link-hint markers” (pages/options.html).

### Testing

* Verified no linter errors.
* Manual checks:
    * Toggle setting on/off and launch link hints on dense pages.
    * With setting ON: only one marker is visible where hints overlap; rotation still functions.
    * With setting OFF: all overlapping markers display.

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -445,6 +445,10 @@ class LinkHintsMode {
     }
 
     this.setIndicator();
+    // Ensure overlapping markers don't visually stack on top of each other.
+    if (Settings.get("suppressOverlappingHintMarkers")) {
+      this.suppressOverlappingMarkers();
+    }
   }
 
   setOpenLinkMode(mode, shouldPropagateToOtherFrames) {
@@ -607,6 +611,10 @@ class LinkHintsMode {
       for (const matched of linksMatched) {
         this.showMarker(matched, this.markerMatcher.hintKeystrokeQueue.length);
       }
+      // After visibility changes, suppress overlaps so only one marker per overlapping region shows.
+      if (Settings.get("suppressOverlappingHintMarkers")) {
+        this.suppressOverlappingMarkers();
+      }
     }
 
     return this.setIndicator();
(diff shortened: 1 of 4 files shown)


geekbrother:
cc @philc @philg-dev for a review 🙏

philg-dev:
Hi @geekbrother I'm not entirely sure why you specifically addressed me for a review, as I'm just another random contributor. So just letting you know that nothing of what I'm saying has anything to do with what the owner and the maintainers of the project think about it.

## Some issues regarding the same or strongly related topics:

- #4698 
- #3454 
- #757 

## Feedback regarding the changes in this PR:

1) You can simply hit `<space>` to cycle through overlapping hints, bringing a different one to the top each time.

2) I think enabling this new behavior by default would be a really bad idea because not all overlapping hints are just "useless duplicates". If I remember correctly there's even cases on Google, Reddit and GitHub where a distinction between some overlapping hints is regularly important to have (and using `<space>` to cycle through them provides a working interaction for those cases. Not only would that be a breaking feature on sites like that, but it would also suddenly disturb the user experience of the existing users which are used to the current behavior.

3) After taking a look at the code it seems to me like there should be a more efficient way to implement it, maybe on-the-fly during the `linkHints` generation process, I'm not sure. Adding an additional `O(n²)` pass over all the link hints seems a bit concerning to me, as on cluttered / complex sites there can be quite a lot of hints present.

Just as a side note: I **did not test** the PR on my end. So my feedback doesn't include a test of the runtime behavior.

Let's see what the others have to say about it though - as I said, I'm just a "nobody".

Issue #4769, opened 2025-09-24
Vomnibar not closing on launching an actual url


goodhippo:
**Describe the bug**

Vominibar when used to launch an url in new tab is not hiding/closing itself after pressing enter.

**To Reproduce**

1. Open Vomnibar with new tab option `O`.
2. Type an actual url. Eg `github.com`.
3. Close the new tab with github.
4. Notice that the vomnibar of previous tab is  still active.

goodhippo:
This is caused by not calling the hide method of VominibarUi when handling actual url. wiill submit my PR.

philc:
This was fixed by #4770.

PR #4770, opened 2025-09-24
Fix: Vomnibar not closing on launching an actual url

goodhippo:
## Description

Fix for [#4769](https://github.com/philc/vimium/issues/4769)


philc:
Nice find @goodhippo, and good fix.

Issue #2049, opened 2016-03-12
Better New tab/Current tab


agzam:
When you try to open a link, depending if Shift is being held or not it shows you a hint "Open link in a new tab" or "Open link in current tab". How difficult would it be to make it that it would slightly change link label's color? That way I think it would be more clear whether new tab would be created or current tab would be replaced with new content.

jaknb:
Yes! That would be great!

(I know how old this issue is but I guess it's still relevant)

JUSTIVE:
@agzam @jaknb 
Hi, It's a bit of an aged issue, but I made a PR just right for this. #4784 
Please have a look and feel free to comment on that.

PR #4784, opened 2025-11-05
feat: passing link mode.name to lint marker's classname, added .bg-ta…

### LINKED ISSUE #4783: Differentiate vimiumHintMarker styles between opening in a new tab and the current tab
Currently, there’s only one design option for vimiumHintMarker, so when I accidentally press F instead of f, it’s hard to tell immediately whether the link will open in a new tab or the current tab.
.
It would be great if there were more detailed class distinctions to allow different styles depending on whether the hint corresponds to a new-tab or same-tab action

This would make it much more intuitive to recognize which mode is active and reduce mistakes when using Vimium.

JUSTIVE:
Resolves #4783 #2049 

`open a link in a new tab` link hint is now visually distinguishable from `open in current tab`.

It can be achieved by exposing the mode's name directly to link_hint's classname.
Since all the mode names are kebab-cased, it's fully compatible with CSS classnames.
Also, other modes can be modified by adding additional styles in CSS.

I adjusted the hue slightly from the original color to choose a new one.
Please feel free to share any thoughts or preferences regarding the color.

|mode|before|after|
|---|---|---|
|open in current tab| image| image|
|open in new tab|image|image|

### DIFF
--- a/content_scripts/link_hints.js
+++ b/content_scripts/link_hints.js
@@ -380,7 +380,7 @@ class LinkHintsMode {
     // This count is used to rank equal-scoring hints when sorting, thereby making JavaScript's sort
     // stable.
     this.stableSortCount = 0;
-    this.hintMarkers = hintDescriptors.map((desc) => this.createMarkerFor(desc));
+    this.hintMarkers = hintDescriptors.map((desc) => this.createMarkerFor(desc, mode));
     this.markerMatcher = Settings.get("filterLinkHints") ? new FilterHints() : new AlphabetHints();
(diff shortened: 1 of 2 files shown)


UncleSnail:
Honestly, I like the flexibility of at least having a way for the user to add a custom style for new tab links. Whether the default style should be different may be a point of contention, but unless there is already another way to add a different style to new tab links, I think it would be beneficial to add one.

JUSTIVE:
If you feel the default styles for new tabs don't match the PR, I'll remove them. The core goal of this PR is to provide the ability to specify different tab styles for different link types.

Issue #4789, opened 2025-11-11
Vimium fails to load on Google Chrome 144.0.7512.1 dev


hyperair:
**Describe the bug**

Vimium fails to load on google-chrome-unstable 144. This seems to be due to the `globalThis.browser` object being introduced, causing the `getBrowserInfo` message handler in the service worker to throw an error that `getBrowserInfo` is undefined in `getFirefoxVersion()` and not respond at all.

From my investigations, I've found that:
```js
browser.runtime === chrome.runtime
browser.runtime.getBrowserInfo === undefined
```

This causes an exception to be thrown from https://github.com/philc/vimium/blob/10c725581311594a24fbb29515b2959fe91da09a/background_scripts/bg_utils.js#L17

**To Reproduce**

1. Install google-chrome-unstable on Ubuntu from the `Dev` channel.
2. Load a simple page, e.g. https://www.debian.org
3. Note that none of the vimium keybindings work
4. Click on the vimium extension icon and note that Vimium is displaying an error message

   <img width="598" height="245" alt="Image" src="https://github.com/user-attachments/assets/0e3cb828-6274-46c4-89ee-1f63dfd2be8c" />
5. Go to chrome://extensions and inspect the service worker
6. In the console, note that there is this error:
   ```
   bg_utils.js:17 Uncaught (in promise) TypeError: browser.runtime.getBrowserInfo is not a function
    at Module.getFirefoxVersion (bg_utils.js:17:54)
    at initializeFrame (main.js:704:37)
    at main.js:764:36
    at async utils.js:272:24
    ```

To workaround:
1. Go to chrome://extensions
2. Inspect the service worker
3. Add a breakpoint in `background_scripts/main.js:33` (early in the service worker)
4. Open the console and type `chrome.runtime.reload()` to trigger a reload of the service worker
5. At the breakpoint, issue `delete globalThis.browser` in the console
6. Continue execution
7. Refresh any tab that is supposed to work with vimium
8. Note that vimium loads now.

<img width="1273" height="1027" alt="Image" src="https://github.com/user-attachments/assets/55346f0d-09cd-454a-aec9-26b6feb2122e" />

**Browser and Vimium version**

If you're using Chrome, include the Chrome and OS version found at chrome://version. Also include
the Vimium version found at chrome://extensions.

```
Google Chrome	144.0.7512.1 (Official Build) dev (64-bit) 
[...]

PR #4788, opened 2025-11-11
Fix Firefox version retrieval logic by using isFirefox() check

yuezk:
## Description

In Chrome Dev 144.0.7512.1, `globalThis.browser` is an object instead of `undefined`, which breaks the Firefox detection logic.

image

This PR update the code to use the more reliable version the Firefox check.

Related issue: #4785

PR #4790, opened 2025-11-11
Fix getFirefoxVersion() throwing errors in chrome

hyperair:
## Description

Avoid calling `browser.runtime.getBrowserInfo()` on chrome by relying on `isFirefox()` isntead of the presence of `globalThis.browser`. This prevents this uncaught error from being thrown in the service worker:

```
bg_utils.js:17 Uncaught (in promise) TypeError: browser.runtime.getBrowserInfo is not a function
 at Module.getFirefoxVersion (bg_utils.js:17:54)
 at initializeFrame (main.js:704:37)
 at main.js:764:36
 at async utils.js:272:24
```

Fixes: #4789

### DIFF
--- a/background_scripts/bg_utils.js
+++ b/background_scripts/bg_utils.js
@@ -14,7 +14,7 @@ export function isFirefox() {
 }
 
 export async function getFirefoxVersion() {
-  return globalThis.browser ? (await browser.runtime.getBrowserInfo()).version : null;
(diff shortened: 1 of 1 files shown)


UncleSnail:
This PR is the same as #4788

hyperair:
Oh whoops, I'll close this then

PR #4798, opened 2025-11-24
fix typo noRepat -> noRepeat

otomist:
1 character change: fix typo noRepat -> noRepeat

Found a typo in the `closeOtherTabs` command in `all_commands.js` while working on my own feature. Thought I should put in a pr. 

All tests still pass with `./make.js test`

### DIFF
--- a/background_scripts/all_commands.js
+++ b/background_scripts/all_commands.js
@@ -541,7 +541,7 @@ const allCommands = [
     group: "tabs",
     advanced: true,
     background: true,
-    noRepat: true,
+    noRepeat: true,
   },
 
   {


philg-dev:
Looks good to me. I did not test this change locally, but from looking at the code base, it looks correct.

I assume in the current master with that typo it would be possible to repeat the command if prefixed with a number (`count`). As a reference, the related commands above don't have the `noRepeat` property and they **are** intended to be repeated, while for the `closeOtherTabs` repeating makes no sense. Since the outcome of applying the command once vs. multiple times is the same, there should not be any semantic difference from the fixed typo.

philc:
Thank you!

PR #4819, opened 2026-01-02
reformat manifest.json for make min diff after execute task 'write-fi…

klim-iv:
…refox-manifest'

changes:
* format arrays and lists definitions
* reformat some comments for developing in Firefox
* reorder some definitions

## Description


philc:
I like the idea -- it is annoying to have manifest.json be a changed file when working in Firefox.

However, this change removes what I think are helpful comments embedded in the manifest.json, which I think are more valuable than the smaller diff.

I've removed the unnecessary structural changes that are created when writing the firefox manifest in 5f60bcbf667576a8b93d820b487a27f9ccc95115.

PR #4820, opened 2026-01-02
Fix the display of enabled keys in the action popup.

### LINKED ISSUE #4733: Vimium popup falsely says "All Vimium keys are enabled on this page." when there are matching exclusion rules for the page
**Describe the bug**

Despite that exclusion rules are matching for a page, the Vimium popup seems to always say "All Vimium keys are enabled on this page." for any page.

When clicking the "Exclude Vimium keys on this page" button, a new rule is added in the resulting form, and all matching rules are shown, revealing the existing exclusion rules that are already active for the current page:

https://github.com/user-attachments/assets/f22f527c-7f8e-410e-8b9e-65507831a042

This seems to happen in both Incognito and Normal mode, and regardless of whether the existing exclusion rule is excluding certain keys or all keys for the page.

**To Reproduce**

1. Add an exclusion rule for a page, e.g. url `https?://www.google.com/*` and keys `/`
2. Visit google.com and enter a search e.g. "test", and click "Search"
3. De-focus the search input, then hit `/` and confirm that it focuses the search bar (because Vimium's `/` was successfully disabled)
4. Open the Vimium popup and observe that it falsely says "All Vimium keys are enabled on this page." and shows no matching rules

> If you're using Chrome, include the Chrome and OS version found at chrome://version. Also include
the Vimium version found at chrome://extensions.

Google Chrome	137.0.7151.122 (Official Build) (arm64) 
Revision	4c03e725fa2c140a1606dad6fef5158687287c83-refs/branch-heads/7151@{#2359}
OS	macOS Version 12.5 (Build 21G72)

Vimium: 2.3

ademuri:
`syncEnabledKeysCaption` was using an incorrect ID selector.

This also adds test coverage for this feature.

This fixes #4733. The tests pass, and I loaded this pull request into Chrome and checked that the feature works there as well.

### DIFF
--- a/pages/action.js
+++ b/pages/action.js
@@ -5,7 +5,7 @@ import "../lib/settings.js";
 import * as bgUtils from "../background_scripts/bg_utils.js";
 import { ExclusionRulesEditor } from "./exclusion_rules_editor.js";
 
-const ActionPage = {
+export const ActionPage = {
   async init() {
     // Is it possible for the current tab's URL to change while this action popup is open?
     const tabs = await chrome.tabs.query({ active: true, currentWindow: true });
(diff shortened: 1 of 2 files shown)


philc:
Hey @ademuri, great catch, and thanks for the PR.

I pushed a simple change in d604506b9dbdf21a68ac2d73cf216603fa3f0939 which fixes this, without a test case. I think the test case in this PR is a good idea, but overkill given the bug and the overall complexity of the action page.

PR #4821, opened 2026-01-03
Add a CI config to check formatting and run tests

ademuri:
This uses Github Actions to run `deno fmt --check` and then `./make.js test`. It also configures `deno fmt` to ignore the icon SVGs, and changes the DOM tests to launch Puppeteer with `--no-sandbox`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,36 @@
+name: CI
+
+on:
+  - pull_request
+  - push
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/puppeteer/puppeteer:23.11.1
+      # Run as root to fix permission issues with Github Actions workspace.
+      # SYS_ADMIN is recommended for running Chrome in Docker.
+      options: --user root --cap-add=SYS_ADMIN
+      env:
+        # Point Puppeteer to the Chrome binary installed in the pptruser home directory, since running as root changes $HOME.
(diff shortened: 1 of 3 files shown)


philc:
Thanks for the PR. What was your motivation for adding the workflow?

I'm concerned maintaining the configuration adds more moving parts than it's worth -- I haven't experienced much pain resulting from broken tests or unformatted code in PRs for this project.

Why is make.js reformatted in this PR? The version in master looks correctly formatted.

Great idea to exclude svgs from deno's fmt check.

ademuri:
I'm not sure why make.js was re-formatted - I reverted that change.

I read your contributing guide, and thought that since reviewer time is relatively scarce, this would be helpful. This also makes contributing easier, as for small changes authors don't necessarily need to install the build environment locally to get test results. The maintenance burden for this config is low, and I'm happy to own fixing any breakages. If anything breaks, it's easy to ignore or disable.

Checking that formatting is applied at each commit helps keep diffs/blame clean, rather than requiring specific formatting commits.  If you're amenable, I can submit a pull request to add a commit hook to run `deno fmt` on commit, so that formatting is automatically applied.

philc:
I've integrated your .svg exclusion from deno fmt in fae19529fa74de37a0de9d39d1d5881b6c1408d3

As for a pre-commit hook, I think if we did this, it should be a simple shell script which runs `deno fmt --check`. However, we will then need to instruct others to install those hooks when they clone the repo.

I don't think the need is particularly strong for this at the moment. There aren't that many PRs, and format fixup commits are relatively rare.

Issue #4847, opened 2026-02-16
Add a cheatsheet pdf, and latex file


sh1V1k:
With tools like these I like being able to print a cheat sheet that I can hang by my desk to make it so I don't have to reference the source for common things until it becomes muscle memory. Here I have provided a latex file generated using Claude code for people to modify to their liking as well as a pdf for people who just want one to print out quickly.

Kyo04200201:
I think button `?`  is good enough for most users.

PR #4848, opened 2026-02-16
Issue #4847 add cheatsheet

sh1V1k:
## Description

With tools like these I like being able to print a cheat sheet that I can hang by my desk to make it so I don't have to reference the source for common things until it becomes muscle memory. Here I have provided a latex file generated using Claude code for people to modify to their liking as well as a pdf for people who just want one to print out quickly.


ananthu-sk:
Not manageable, it should be automated to handle future option addition or removal.

seki-mei:
The `?` dialog already lists your custom keybinds and you can print it using your browser's print shortcut (usually ctrl-P)

Issue #3937, opened 2021-10-13
Feature Request: Chrome tab group control


alanlunspace:
Dear Vimium developers,

May I know will we implement commands for Chrome's tab group control please?
For example:
1. Pin all tabs in the same group
2. Close tab group
3. Move tab group.

Thank you very much,
Alan

---

A big `thank you` to Vimium developers.
Vimium is my most favourite and important Chrome plugin for daily works and browsing.
Stay healthy and happy 😊

gdh1995:
Currently those browser-level API to control tab groups are **not exposed** to extensions like Vimium which has `V2 manifest`. Chrome has just supported V3 manifest this year, but it requires some code changes to Vimium itself.

alanlunspace:
Thank you @gdh1995 for your info and time.
Looking forward to v3 support.

marziply:
v3 support would be great, currently sitting on >100 tabs in various groups, would be super convenient to manipulate them via keybinds

phye:
Bump up for this feature, again :)

manio:
Hi!
First of all really thank you for this great plugin! I really love Vimium! :)
Google Chrome tab groups control is the one feature I am missing most...
It would be also great if we could open a tab which is inside a tab group which is "collapsed" or move "over" it to another one. But I know this probably need some "cursor" and it is a plenty of work... And I don't even know if Google is exposing such interface for grouped tab manipulation...

freder:
an `add current page to group` command would be insanely useful already 👍

marcelarie:
really looking for this :)

jlevitt:
Would also love this feature!

p5a0u9l:
+1 please

cam5:
I'd be keen to work up a PR if someone more deeply _in_ the project could sketch a possible roadmap to realizing it.

I see that #4274 is merged, which gives support for V3 manifests. (The blocker mentioned above.)

cc: @philc, I'm browsing through PRs for a good reference regarding conventions for new command names, [groups](https://github.com/philc/vimium/blob/31af27bb9c362792bc531df8b0865dfb87afa36b/background_scripts/commands.js#L243) (`advancedCommands`?) and default bindings (if any) but would be indebted to you if you could point at a good example to read through.

Cheers!

grandmasteri:
Hi, is there any updates on this thread? It would be very useful to have native tab group manipulation control with Vimium C.

nikhilmaddirala:
+100 for this feature request! 😊

PR #4858, opened 2026-03-02
feat: add tab group support (skip collapsed groups, zc/zj/zk navigation)

### LINKED ISSUE #4693: Support for jump over tab groups
Current default behaviour of `J` and `K` keys is to scroll one tab left/right **including** tabs in tab groups.  

It would be nice to align this beviour with default `C-tab/C-shift-tab` Chrome keys, which **skip** tabs in tab groups if groups are collapsed.  
Or maybe provide a handler for this.

tusharmurali:
## Description

Fixes #4693
Related #3937, #4639

This PR adds tab group support to Vimium:

**1. Skip collapsed tab groups when navigating (J/K/gt/gT/g0/g$)**

Chrome and Firefox's native `Ctrl+Tab` skips collapsed tab groups, but Vimium's `J/K` currently navigates into them. This aligns Vimium's behavior with Chrome and Firefox's native UX.

**2. New commands for tab group navigation (Vim fold-style keybindings)**

| Key | Command | Description |
|-----|---------|-------------|
| `zc` | `collapseTabGroup` | Collapse current tab's group |
| `zj` | `previousTabGroup` | Switch to previous tab group |
| `zk` | `nextTabGroup` | Switch to next tab group |

These follow Vim's fold navigation conventions (`z` prefix, `j`/`k` for direction).

  Notes:
  - Adds `tabGroups` permission (no new user-facing permission warnings)
  - Works on any browser with tabGroups API support (Chrome, Firefox 140+); no-ops on older versions
  - Users can still access tabs in collapsed groups via `T` (Vomnibar tab search)

### DIFF
--- a/background_scripts/all_commands.js
+++ b/background_scripts/all_commands.js
@@ -494,6 +494,30 @@ const allCommands = [
   },
 
   {
+    name: "collapseTabGroup",
+    desc: "Collapse current tab's group",
+    group: "tabs",
+    background: true,
+    noRepeat: true,
+  },
+
+  {
+    name: "previousTabGroup",
+    desc: "Switch to previous tab group",
+    group: "tabs",
+    background: true,
+    noRepeat: true,
+  },
+
+  {
+    name: "nextTabGroup",
+    desc: "Switch to next tab group",
+    group: "tabs",
+    background: true,
+    noRepeat: true,
+  },
+
+  {
     name: "removeTab",
     desc: "Close current tab",
     group: "tabs",
--- a/background_scripts/commands.js
+++ b/background_scripts/commands.js
@@ -482,6 +482,9 @@ const defaultKeyMappings = {
   "zi": "zoomIn",
   "zo": "zoomOut",
   "z0": "zoomReset",
+  "zc": "collapseTabGroup",
+  "zj": "previousTabGroup",
+  "zk": "nextTabGroup",
 
   // Marks
   "m": "Marks.activateCreateMode",
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -340,6 +340,22 @@ const BackgroundCommands = {
     });
   },
   toggleMuteTab,
+  // Tab group commands (Chrome only; no-op on Firefox).
+  async collapseTabGroup({ tab }) {
+    if (bgUtils.isFirefox() || !chrome.tabGroups || tab.groupId == -1) return;
+    const tabs = await chrome.tabs.query({ currentWindow: true });
+    let nextTab = tabs.find((t) => t.index > tab.index && t.groupId != tab.groupId) ||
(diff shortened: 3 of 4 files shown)


imsodin:
Thanks for working on this!

> - Firefox: gracefully no-ops (Firefox 140+ will add tab group API support)

Just out of interest, not complaining in any way: As the code here is chrome only still and firefox does have tab groups for a while - did you check and find that these operations aren't available/possible in firefox or just only focused on chrome for now?

philg-dev:
> Thanks for working on this!
> 
> > * Firefox: gracefully no-ops (Firefox 140+ will add tab group API support)
> 
> Just out of interest, not complaining in any way: As the code here is chrome only still and firefox does have tab groups for a while - did you check and find that these operations aren't available/possible in firefox or just only focused on chrome for now?

According to the MDN compatibility table, the tab group related functionality that is used in this PR should also be available on Firefox:
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabGroups/TabGroup#browser_compatibility
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabGroups/update#browser_compatibility

So I personally think that there should be no distinction in Vimium regarding this feature.

Edit: the feature would obviously only work on newer Firefox versions, so it might be necessary to increase the minimum Firefox version in the `manifest.json` or probably less restrictive: some proper handling of cases where the API methods are unavailable (e.g. Firefox ESR versions).

tusharmurali:
> > Thanks for working on this!
> > > * Firefox: gracefully no-ops (Firefox 140+ will add tab group API support)
> > 
> > 
> > Just out of interest, not complaining in any way: As the code here is chrome only still and firefox does have tab groups for a while - did you check and find that these operations aren't available/possible in firefox or just only focused on chrome for now?
> 
> According to the MDN compatibility table, the tab group related functionality that is used in this PR should also be available on Firefox:
> 
> * https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabGroups/TabGroup#browser_compatibility
> * https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabGroups/update#browser_compatibility
> 
> So I personally think that there should be no distinction in Vimium regarding this feature.
> 
> Edit: the feature would obviously only work on newer Firefox versions, so it might be necessary to increase the minimum Firefox version in the `manifest.json` or probably less restrictive: some proper handling of cases where the API methods are unavailable (e.g. Firefox ESR versions).

Thanks for flagging, I was focusing only on Chrome. I've updated the code to use `chrome.tabGroups` feature detection (which actually simplifies the code) so tab group commands will work on Firefox 140+ automatically (and no-op on older versions).

stumash:
really interested in this feature so I'm gonna tag a maintainer @philc

PR #4876, opened 2026-03-18
Remove 'f' from default linkHintCharacters

Alireza-Majd:
## Summary
- Removes `f` from the default `linkHintCharacters` setting to prevent accidental link clicks when pressing `f` twice (once to activate hints, once matching a hint label starting with `f`).

## Test plan
- Press `f` to activate link hints and verify no hint labels contain the letter `f`
- Confirm all other hint characters still work as expected

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -11,7 +11,7 @@ const defaultOptions = {
   scrollStepSize: 60,
   smoothScroll: true,
   keyMappings: "# Insert your preferred key mappings here.",
-  linkHintCharacters: "sadfjklewcmpgh",
+  linkHintCharacters: "sadjklewcmpgh",
   linkHintNumbers: "0123456789",
   filterLinkHints: false,
   hideHud: false,


philc:
`f` is in the default link hints by design, since it's on the home row. I haven't ever mistakenly double-typed f when using link hints. Maybe your keyboard has key chatter on the `f` key?

Alireza-Majd:
Agree with you—and I’m not suggesting removing F from its main show/hide role in Vimium. Actually using it for link hints is a great choice and I love vimium for it.

I just removed F from the link hint characters. On link-heavy pages or slower Chromium setups, there’s sometimes a small delay between pressing F and the hints showing up. In that gap, I’ve caught myself pressing it again thinking it didn’t register, which can end up activating a link labeled F.

To avoid that, I remove F from the hint characters on each device. With this setup, double-pressing F works nicely as a quick show/hide toggle and an easy way to check hint coverage on more complex pages (like FB Marketplace modals).

ghost:
Would solve https://github.com/philc/vimium/issues/4921. Please reconsider!

Issue #4103, opened 2022-08-06
[Feature Request] Command-line mode for entering ex commands


mcp292:
I would love to be able to issue `:closeTabsOnRight` and similar.

philc:
This is a great idea. It's especially useful for the rarely-used commands that you don't need or want a binding for; closeTabsOnRight is a good example.

In Vim, it's named command line mode. In Emacs, it's called the M-x command prompt. In apps more generally, it's typically called a command palette.

IMO the UX should be big and centered (using the same UI as the Vomnibar), rather than at the bottom, like Vimium's mode HUD.

Issue #4481, opened 2024-05-28
No keybinding for commenting on Youtube


lephantriduc:
**Describe the bug**

There's no keybinding for commenting on Youtube

**To Reproduce**

1. Go to 'youtube.com'
2. Click on any video that enables commenting

![Screenshot 2024-05-28 at 21 28 41](https://github.com/philc/vimium/assets/87744086/ee0a18b8-d756-4dc7-b72d-3b4cac517066)

**Browser and Vimium version**

Chrome version: 125.0.6422.112 (Official Build) (arm64)
OS version: macOS Version 14.3 (Build 23D56)
Vimium version: 2.1.2

qrockz:
1 year and still an issue..... I'm with you bro. This still sucks.

philg-dev:
There you go... Let's see if and when my fix gets merged.

Since it's a trivial change in the code, chances are it might make it into the next release.

PR #4880, opened 2026-03-24
[feature] Command mode

ckardaris:
Introduce command mode in the omni bar.

Features:
- Launch with ':' (default key mapping).
- Support prefix counts.
- Show all commands (even mapped ones).
- Show all commands as specified (with any options) in user defined key mappings.
- Show keys for mapped commands, similarly to the help page.

Fix:
- Fix collection of unit tests in subdirectories.

Other:
- Revise 'noRepeat' attribute for all commands.

## Description

Good evening, this MR implements feature request #4103.

I have tried to implement this feature as cleanly as possible.
I understand that this is quite a big patch, so I expect some changes based on reviewer suggestions,
but I think that it is working nicely. 
I tried to add some tests, as well, but I am afraid I am not very familiar with `shoulda` and `puppeteer`,
so I am not sure if I have followed best practices and if I am inside the general spirit of the Vimium tests.

I am looking forward to your feedback and hope to be merging this soon.

### DIFF
--- a/background_scripts/all_commands.js
+++ b/background_scripts/all_commands.js
@@ -32,12 +32,14 @@ const allCommands = [
     name: "scrollToTop",
     desc: "Scroll to the top of the page",
     group: "navigation",
+    noRepeat: true,
   },
 
   {
     name: "scrollToBottom",
     desc: "Scroll to the bottom of the page",
     group: "navigation",
+    noRepeat: true,
   },
 
   {
@@ -82,12 +84,15 @@ const allCommands = [
     desc: "Scroll all the way to the left",
     group: "navigation",
     advanced: true,
+    noRepeat: true,
   },
 
   {
     name: "scrollToRight",
     desc: "Scroll all the way to the right",
     group: "navigation",
+    advanced: true,
+    noRepeat: true,
   },
 
   {
@@ -98,6 +103,7 @@ const allCommands = [
     options: {
       hard: "Perform a hard reload, forcing the browser to bypass its cache.",
     },
+    noRepeat: true,
   },
 
   {
@@ -137,6 +143,7 @@ const allCommands = [
     desc: "Go to the root of current URL hierarchy",
     group: "navigation",
     advanced: true,
+    noRepeat: true,
   },
 
   {
@@ -175,6 +182,7 @@ const allCommands = [
     name: "focusInput",
     desc: "Focus the first text input on the page",
     group: "navigation",
+    noRepeat: true,
   },
 
   {
@@ -303,6 +311,7 @@ const allCommands = [
     },
     group: "vomnibar",
     topFrame: true,
+    noRepeat: true,
   },
 
   {
@@ -315,6 +324,7 @@ const allCommands = [
         "section of the Vimium Options page. The Vomnibar will be scoped to use that search engine.",
     },
     topFrame: true,
+    noRepeat: true,
   },
 
   {
@@ -325,6 +335,7 @@ const allCommands = [
       query: "The text to prefill the Vomnibar with.",
     },
     topFrame: true,
+    noRepeat: true,
   },
 
   {
@@ -335,6 +346,15 @@ const allCommands = [
       query: "The text to prefill the Vomnibar with.",
     },
     topFrame: true,
+    noRepeat: true,
+  },
+
+  {
+    name: "Vomnibar.activateCommand",
+    desc: "Execute a Vimium command",
+    group: "vomnibar",
+    topFrame: true,
+    noRepeat: true,
   },
 
   {
@@ -342,6 +362,7 @@ const allCommands = [
     desc: "Search through your open tabs",
     group: "vomnibar",
     topFrame: true,
+    noRepeat: true,
   },
 
   {
@@ -349,6 +370,7 @@ const allCommands = [
     desc: "Edit the current URL",
     group: "vomnibar",
     topFrame: true,
+    noRepeat: true,
   },
 
   {
@@ -356,6 +378,7 @@ const allCommands = [
     desc: "Edit the current URL and open in a new tab",
     group: "vomnibar",
     topFrame: true,
+    noRepeat: true,
   },
 
   //
@@ -433,14 +456,14 @@ const allCommands = [
 
   {
     name: "previousTab",
-    desc: "Go one tab left",
+    desc: "Go one tab left (vertical: up)",
     group: "tabs",
     background: true,
   },
 
   {
     name: "nextTab",
-    desc: "Go one tab right",
+    desc: "Go one tab right (vertical: down)",
     group: "tabs",
     background: true,
   },
@@ -457,6 +480,7 @@ const allCommands = [
     desc: "Go to the first tab",
     group: "tabs",
     background: true,
+    noRepeat: true,
   },
 
   {
@@ -464,6 +488,7 @@ const allCommands = [
     desc: "Go to the last tab",
     group: "tabs",
     background: true,
+    noRepeat: true,
   },
 
   {
@@ -479,6 +504,7 @@ const allCommands = [
     desc: "Pin or unpin current tab",
     group: "tabs",
     background: true,
+    noRepeat: true,
   },
 
   {
@@ -517,11 +543,12 @@ const allCommands = [
     group: "tabs",
     advanced: true,
     background: true,
+    noRepeat: true,
   },
 
   {
     name: "closeTabsOnLeft",
-    desc: "Close tabs on the left",
+    desc: "Close tabs on the left (vertical: up)",
     group: "tabs",
     advanced: true,
     background: true,
@@ -529,7 +556,7 @@ const allCommands = [
 
   {
     name: "closeTabsOnRight",
-    desc: "Close tabs on the right",
+    desc: "Close tabs on the right (vertical: down)",
     group: "tabs",
     advanced: true,
     background: true,
@@ -546,7 +573,7 @@ const allCommands = [
 
   {
     name: "moveTabLeft",
-    desc: "Move tab to the left",
+    desc: "Move tab to the left (vertical: up)",
     group: "tabs",
     advanced: true,
     background: true,
@@ -554,7 +581,7 @@ const allCommands = [
 
   {
     name: "moveTabRight",
-    desc: "Move tab to the right",
+    desc: "Move tab to the right (vertical: down)",
     group: "tabs",
     advanced: true,
     background: true,
@@ -569,6 +596,7 @@ const allCommands = [
     options: {
       level: "The zoom level. This can be a range of [0.25, 5.0]. 1.0 is the default.",
     },
+    noRepeat: true,
   },
 
   {
@@ -593,6 +621,7 @@ const allCommands = [
     group: "tabs",
     advanced: true,
     background: true,
+    noRepeat: true,
   },
 
   //
--- a/background_scripts/all_commands.js
+++ b/background_scripts/all_commands.js
@@ -350,7 +350,7 @@ const allCommands = [
   },
 
   {
-    name: "Vomnibar.activateCommand",
+    name: "Vomnibar.activateCommandSelection",
     desc: "Execute a Vimium command",
     group: "vomnibar",
     topFrame: true,
--- a/background_scripts/all_commands.js
+++ b/background_scripts/all_commands.js
@@ -456,14 +456,14 @@ const allCommands = [
 
   {
     name: "previousTab",
-    desc: "Go one tab left (vertical: up)",
+    desc: "Go one tab left",
     group: "tabs",
     background: true,
   },
 
   {
     name: "nextTab",
-    desc: "Go one tab right (vertical: down)",
+    desc: "Go one tab right",
     group: "tabs",
     background: true,
   },
@@ -548,7 +548,7 @@ const allCommands = [
 
   {
     name: "closeTabsOnLeft",
-    desc: "Close tabs on the left (vertical: up)",
+    desc: "Close tabs on the left",
     group: "tabs",
     advanced: true,
     background: true,
@@ -556,7 +556,7 @@ const allCommands = [
 
   {
     name: "closeTabsOnRight",
-    desc: "Close tabs on the right (vertical: down)",
+    desc: "Close tabs on the right",
     group: "tabs",
     advanced: true,
     background: true,
@@ -573,7 +573,7 @@ const allCommands = [
 
   {
     name: "moveTabLeft",
-    desc: "Move tab to the left (vertical: up)",
+    desc: "Move tab to the left",
     group: "tabs",
     advanced: true,
     background: true,
@@ -581,7 +581,7 @@ const allCommands = [
 
   {
     name: "moveTabRight",
-    desc: "Move tab to the right (vertical: down)",
+    desc: "Move tab to the right",
     group: "tabs",
     advanced: true,
     background: true,
--- a/background_scripts/completion/completers.js
+++ b/background_scripts/completion/completers.js
@@ -15,6 +15,8 @@ import * as bgUtils from "./../bg_utils.js";
 import * as completionSearch from "./search_wrapper.js";
 import * as userSearchEngines from "../user_search_engines.js";
 import * as ranking from "./ranking.js";
+import { allCommands } from "../all_commands.js";
+import { Commands, RegistryEntry } from "../commands.js";
 import { RegexpCache } from "./ranking.js";
 
 // Set this to true to render relevancy when debugging the ranking scores.
@@ -47,6 +49,14 @@ export class Suggestion {
   tabId;
   // Whether this is a suggestion provided by a user's custom search engine.
   isCustomSearch;
+  // Suggestion in 'command' mode.
+  // command = {
+  //   // 'RegistryEntry' to execute the command in 'NormalMode.commandHandler'.
+  //   registryEntry: RegistryEntry,
+  //   // Key mapping to show in the omni bar suggestions
+  //   keys: Array[string]
+  // }
+  command;
   // Whether this is meant to be the first suggestion from the user's custom search engine which
   // represents their query as typed, verbatim.
   isPrimarySuggestion = false;
@@ -72,7 +82,8 @@ export class Suggestion {
   generateHtml() {
     if (this.html) return this.html;
     const relevancyHtml = showRelevancy
-      ? `<span class='relevancy'>${this.computeRelevancy()}</span>`
+      ? `
+    <span class='relevancy'>${this.computeRelevancy()}</span>`
       : "";
     const insertTextClass = this.insertText ? "" : "no-insert-text";
     const insertTextIndicator = "&#8618;"; // A right hooked arrow.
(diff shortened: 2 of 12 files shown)


philc:
Whoa, very exciting! Thank you for the PR. I'll take this for a spin in the next few days.

background_scripts/completion/completers.js
```diff
@@ -47,6 +49,14 @@ export class Suggestion {
   tabId;
   // Whether this is a suggestion provided by a user's custom search engine.
   isCustomSearch;
+  // Suggestion in 'command' mode.
```
philc: Use double quotes here, and throughout all the comments, to match the existing style.

background_scripts/completion/completers.js
```diff
@@ -47,6 +49,14 @@ export class Suggestion {
...
+  // command = {
+  //   // 'RegistryEntry' to execute the command in 'NormalMode.commandHandler'.
```
philc: This comment shouldn't be double commented.

background_scripts/completion/completers.js
```diff
@@ -72,7 +82,8 @@ export class Suggestion {
   generateHtml() {
     if (this.html) return this.html;
     const relevancyHtml = showRelevancy
-      ? `<span class='relevancy'>${this.computeRelevancy()}</span>`
+      ? `
+    <span class='relevancy'>${this.computeRelevancy()}</span>`
```
philc: Why is there a newline added here?

background_scripts/completion/completers.js
```diff
@@ -87,12 +98,32 @@ export class Suggestion {
       faviconHtml = `<img class="icon" src="${faviconUrl.toString()}" />`;
     }
     if (this.isCustomSearch) {
-      this.html = `\
-<div class="top-half">
-   <span class="source ${insertTextClass}">${insertTextIndicator}</span><span class="source">${this.description}</span>
-   <span class="title">${this.highlightQueryTerms(Utils.escapeHtml(this.title))}</span>
-   ${relevancyHtml}
- </div>\
+      this.html = `
```
philc: This looks like an unrelated whitespace change (indentation and additional line wrapping).

background_scripts/completion/completers.js
```diff
@@ -87,12 +98,32 @@ export class Suggestion {
...
+  <div class="top-half">
+    <span class="source ${insertTextClass}">${insertTextIndicator}</span><span class="source">${this.description}</span>
+    <span class="title">${
+        this.highlightQueryTerms(Utils.escapeHtml(this.title))
+      }</span>${relevancyHtml}
+  </div>
+`;
+    } else if (this.command) {
+      // Key mappings containing key-modifiers are represented in the form of '<modifier-key>'
```
philc: key modifiers should be two words.

background_scripts/completion/completers.js
```diff
@@ -87,12 +98,32 @@ export class Suggestion {
...
+      // (e.g <c-e>) and are parsed as HTML tags when used in a raw string. Escape them properly.
+      const escapeKeyForHtml = (key) => {
+        return key.replace(/</g, "&lt;").replace(/>/g, "&gt;");
+      };
+      const keybindings = this.command.keys.map((key) => `
+    <span class="key-block">
```
philc, line 114: Note that this section is related to help_dialog_page.js, which also generates keys using similar CSS class names, even though no code is shared.

background_scripts/completion/completers.js
```diff
@@ -87,12 +98,32 @@ export class Suggestion {
...
+      <span class="key">${escapeKeyForHtml(key)}</span>
+      <span class="comma">, </span>
+    </span>`).join("\n");
+      this.html = `
+  <div class="top-half">
+    <span class="source ${insertTextClass}">${insertTextIndicator}</span><span class="source">${this.description}</span>
+    <span class="title">${
+        this.highlightQueryTerms(`${this.title}`)
```
philc: I think this is more readable if the template string isn't padded with newlines.

background_scripts/completion/completers.js
```diff
@@ -103,9 +134,8 @@ export class Suggestion {
  <div class="bottom-half">
   <span class="source no-insert-text">${insertTextIndicator}</span>${faviconHtml}<span class="url">${
         this.highlightQueryTerms(Utils.escapeHtml(this.shortenUrl()))
-      }</span>
-  ${relevancyHtml}
-</div>\
+      }</span>${relevancyHtml}
```
philc: Why this whitespace change? Does it have any effect?

background_scripts/completion/completers.js
philc: I don't have a strong opinion on whether it's warranted here, but the more formatting work we do in Javascript, the more we would benefit from using an HTML template and populating it using DOM APIs, as the help dialog does.

background_scripts/completion/completers.js
```diff
@@ -353,6 +383,86 @@ export class BookmarkCompleter {
...
+        // If the default action is not bound, add the entry explicitly to the suggestions.
+        // This makes unbound commands accessible from the omni bar in 'command' mode.
+        if (!isDefaultBound) {
+          suggestions.push(
+            new Suggestion({
+              queryTerms,
+              description: "command",
+              title: command.desc,
+              url: command.name,
+              command: {
```
philc: It's weird that we have something in scope called `command`, and that it's different from the command object we're building here. We're overloading the word `command`. Readability can be improved by picking a better/more specific name for one of the two.

background_scripts/completion/completers.js
```diff
@@ -353,6 +383,86 @@ export class BookmarkCompleter {
...
+        topFrame: command.topFrame,
+        options: {},
+      });
+    };
+
+    // Option suffix to add to the suggestion entry based on the command options for a mapping.
+    // Used in two places:
+    // - title: to set the actual visible text in the omni bar
+    // - url: used as a key of difference during the clean-up of the suggestions.
+    const optionsSuffix = (option) => {
```
philc: It's unclear to me what this is based on the code. Can you add an example to the comment? I'll note that it's a bad smell that such a simple function needs 4 lines of comments to explain it. Ideally this function doesn't exist, or maybe it can be inlined where it's used so it's clearer what it's doing.

background_scripts/completion/completers.js
```diff
@@ -353,6 +383,86 @@ export class BookmarkCompleter {
...
+      return option ? ` (${option})` : "";
+    };
+
+    let suggestions = [];
+    allCommands.filter((command) => ranking.matches(queryTerms, command.desc))
+      .map((command) => {
```
philc: filter and map statements shouldn't have side effects. Either assign the result of the map to a variable and remove the `push` operations, or switch this to a for loop.

It's a bad smell that this map body is so large.

background_scripts/all_commands.js
```diff
@@ -335,27 +346,39 @@ const allCommands = [
       query: "The text to prefill the Vomnibar with.",
     },
     topFrame: true,
+    noRepeat: true,
+  },
+
+  {
+    name: "Vomnibar.activateCommand",
```
philc: IMO this should be called activateCommandSelection, to mirror activateTabSelection, and to convey it's a UI.

background_scripts/all_commands.js
```diff
@@ -433,14 +456,14 @@ const allCommands = [
 
   {
     name: "previousTab",
-    desc: "Go one tab left",
+    desc: "Go one tab left (vertical: up)",
```
philc: I get why you added this, but this doesn't belong in this PR, and there may be a more elegant solution, so please decouple it from this PR.

lib/types.js
```diff
@@ -11,4 +11,7 @@ globalThis.VomnibarShowOptions = {
   selectFirst: "boolean",
   // A keyword which will scope the search to a UserSearchEngine.
   keyword: "string",
+  // A prefix count number before launching the omni bar
+  // (used to propagate the count in 'command' mode).
+  omniCommandCount: "number",
```
philc: IMO this variable name isn't particularly clear/self-evident. I stared at it the first few times I saw it in the code.

tests/unit_tests/completion/completers_test.js
```diff
@@ -351,6 +354,195 @@ context("multi completer", () => {
...
+  should("return all commands with default options if no mappings are specified", async () => {
+    stub(chrome.storage.session, "get", async () => ({
+      commandToOptionsToKeys: {},
+    }));
+    stub(Commands, "keyToRegistryEntry", {});
+    const suggestions = await filterCompleter(commandCompleter, []);
+    assert.equal(allCommands.length, suggestions.length);
+    assert.isTrue(
+      suggestions.every((suggestion) =>
+        JSON.stringify(suggestion.command.registryEntry.options) === "{}"
```
philc: I don't follow what this is asserting. Either a more descriptive test name, a comment, or assigning this expression to a named intermediate variable would help make it clearer.

tests/unit_tests/completion/completers_test.js
```diff
@@ -351,6 +354,195 @@ context("multi completer", () => {
...
+      ),
+    );
+  });
+
+  should("return an empty list when the query is empty", async () => {
```
philc: Do we need a test for this given the other tests?

tests/unit_tests/completion/completers_test.js
```diff
@@ -351,6 +354,195 @@ context("multi completer", () => {
...
+    stub(chrome.storage.session, "get", async () => ({
+      commandToOptionsToKeys: {},
+    }));
+    stub(Commands, "keyToRegistryEntry", {});
+    assert.equal([], await filterCompleter(multiCompleter, []));
+  });
+
+  should("handle key bound commands with options", async () => {
```
philc: This test is way too long, and thus unmantainable. Make more narrow assertions about only the essential pieces of the data we care about.

I'm also not sure what this is testing differently from the other tests.

tests/unit_tests/ui_component_test.js
```diff
@@ -19,6 +20,7 @@ context("UIComponent", () => {
     // Which page we load doesn't matter; we just need any DOM.
     await testHelper.jsdomStub("pages/help_dialog_page.html");
     stub(Utils, "isFirefox", () => false);
+    await Settings.load();
```
philc: Why do we need this here? It may be valid to have this here, but no other test has needed it here until now, and so it's conspicuous that you're adding this.

If we did need to add this here for some structural reason, it needs a good comment explaining why.

tests/unit_tests/vomnibar_page_test.js
```diff
@@ -1,7 +1,14 @@
 import * as testHelper from "./test_helper.js";
 import "../../tests/unit_tests/test_chrome_stubs.js";
-import { Suggestion } from "../../background_scripts/completion/completers.js";
+import {
+  CommandCompleter,
+  MultiCompleter,
+  Suggestion,
+} from "../../background_scripts/completion/completers.js";
```
philc, line 7: Having each import on its own line does not match the style in the rest of the codebase.

tests/unit_tests/vomnibar_page_test.js
```diff
@@ -91,4 +98,85 @@ context("vomnibar page", () => {
     // The query should not be treated as a user search engine.
     assert.equal("constructor ", ui.input.value);
   });
+
+  should("fill the suggestions list with correct HTML", async () => {
```
philc: This test is too verbose, and it's too fragile -- changing any part of the HTML will break this test.

make.js
```diff
@@ -256,13 +256,27 @@ async function buildStorePackage() {
...
+    }
+  }
+  return results.sort();
+}
+
 async function runUnitTests() {
   // Import every test file.
   const dir = path.join(projectPath, "tests/unit_tests");
-  const files = Array.from(Deno.readDirSync(dir)).map((f) => f.name).sort();
+  const files = collectFiles(dir);
```
philc: Nice find. That seems like a bad error =)

But it's not related to this PR; please split out into a separate PR.

background_scripts/completion/completers.js
```diff
@@ -353,6 +383,86 @@ export class BookmarkCompleter {
...
+    // Example:
+    // "closeTabsOnRight" : {
+    //    "count=2": ["c2l", "c2k"],
+    //    "count=3": ["c3l", "c3k"],
+    // }
+    const commandToOptionsToKeys =
+      (await chrome.storage.session.get("commandToOptionsToKeys")).commandToOptionsToKeys;
+
+    // Create a RegistryEntry for the default action (no options specified) of a command.
+    const createUnboundRegistryEntry = (command) => {
```
philc, line 396: I haven't looked closely at why this is needed, but it would be nice if we didn't have to have this.

content_scripts/mode_normal.js
```diff
@@ -22,9 +22,20 @@ class NormalMode extends KeyHandlerMode {
         this.setKeyMapping(changes.normalModeKeyStateMapping.newValue);
       }
     });
+
+    // Listen and handle 'command' events coming from the background and
+    // originally sent from the omnibar page.
```
philc: This comment is wrapped too aggressively.

philc (changes requested):
Hey @ckardaris, I tried this out and it works well. I've left some initial comments on the code. For the UX:

* We should remove "command" as a prefix in the UI. One could argue it should be removed for bookmarks and tabs too when using their specific selection UIs, but that can be debated offline. Since commands can only be searched when using the command search UI, it just adds clutter to the UI.
* The styles of the keybindings in this dialog needs to be improved. The Vomnibar's background colors are different than the help dialog's, so it doesn't look good, at least in dark mode. I can fiddle with it post-merge.
* Hitting enter when there are no completions should do nothing, rather than submit the query to the default search engine. This is an existing bug, and also affects the bookmark and tabs listing.

make.js
ckardaris: Done. https://github.com/philc/vimium/pull/4882.

tests/unit_tests/ui_component_test.js
ckardaris: I have split this along with the other test related change in PR https://github.com/philc/vimium/pull/4882.

background_scripts/completion/completers.js
ckardaris: Done.

background_scripts/completion/completers.js
ckardaris: Done.

background_scripts/completion/completers.js
ckardaris: I have reverted. It was an attempt to unify the generated HTML and better format it, but I agree that maybe I should not touch unrelated code and keep the scope narrower.

background_scripts/completion/completers.js
ckardaris: Same. I have reverted to keep the scope narrower.

background_scripts/completion/completers.js
ckardaris: Done.

background_scripts/completion/completers.js
ckardaris: Done. I removed the unnecessary backticks inside the function call and the formatter worked as expected.

background_scripts/completion/completers.js
ckardaris: See above. I have reverted unnecessary changes for this PR.

background_scripts/all_commands.js
ckardaris: Done.

tests/unit_tests/vomnibar_page_test.js
ckardaris: I am afraid this is automatically formatted that way.

background_scripts/completion/completers.js
ckardaris: This is done this way because the main loop uses `commandsToOptionsToKeys` to gather available commands. This means that non-mapped commands are not collected even though they are available. `createUnboundRegistryEntry` fills this gap. 

To be honest this part of the implementation seemed quite obscure to me and it also made for some weird test programming that you have already commented on in the other threads.

I am open to suggestions if I am missing an obvious and easier way to do this.

background_scripts/all_commands.js
ckardaris: Done.

tests/unit_tests/completion/completers_test.js
ckardaris: I believe you are right. Removed.

content_scripts/mode_normal.js
ckardaris: Done.

ckardaris:
Hello @philc. Thanks for the review. I am in the process of going over it and fixing stuff based on your suggestions.

>     * We should remove "command" as a prefix in the UI. One could argue it should be removed for bookmarks and tabs too when using their specific selection UIs, but that can be debated offline. Since commands can only be searched when using the command search UI, it just adds clutter to the UI.

I also agree that it is "weird" to see the same "source" for non-omni modes. Would the proposal of #4481 be interesting? I think we could the best of both worlds considering that we would still have an indicator of the current mode of operation.

> 
>     * The styles of the keybindings in this dialog needs to be improved. The Vomnibar's background colors are different than the help dialog's, so it doesn't look good, at least in dark mode. I can fiddle with it post-merge.

Thanks. Would it maybe make sense to do a style update and unify the background with the help page, so that both modes work as expected and the CSS styling can keep being shared?

>     * Hitting enter when there are no completions should do nothing, rather than submit the query to the default search engine. This is an existing bug, and also affects the bookmark and tabs listing.
I have pushed a one-line fix for this. Let me know if you want to keep it in this PR or you would prefer to split it, since it is kind of a bigger (conceptually) change.

background_scripts/completion/completers.js
ckardaris: Done. As I explain in the other thread about `createUnboundRegistryEntry` I am not sure how to reduce the body size.

lib/types.js
ckardaris: Done. Let me know if the rename and the updated comment are OK.

tests/unit_tests/completion/completers_test.js
ckardaris: Done. I hope it's a little better now.

tests/unit_tests/completion/completers_test.js
ckardaris: Done. Let me know if this is better.

tests/unit_tests/vomnibar_page_test.js
ckardaris: I was mostly interested about testing the key block. I have simplified the test.

background_scripts/completion/completers.js
ckardaris: I have simplified the code. I was complicating more than necessary in order to provide a `url` which is not needed for commands, but the de-duplication code was needing it. So I have cleaned up the implementation a little. 

1. `command` suggestions do not provide a `url` and set the `deDuplicate` flag to `false`.
2. I have refactored a little the de-duplication code. There is little added value in my opinion to keep separate `count` and `seenUrls` variables, considering that the `dedupedSuggestions` array contains maximum 10 results.

background_scripts/completion/completers.js
ckardaris: I agree. I would prefer to leave this for some other PR though. You probably have a better idea of how to refactor this properly.

background_scripts/completion/completers.js
ckardaris: Done I have used `commandInfo`, similarly to other places in the code.

Some additional notes from my side. I had a lot of trouble understanding the existing code, because `command` and `registryEntry` are very closely connected, but not always corresponding to the same thing.

1. We have a simple `command` object that is part of the `allCommand` array and contains
``` js
{
  name: ...,
  desc: ...,
  group: ...
  ...
}
```

2. We have a `RegistryEntry` class that looks as follows
``` js
export class RegistryEntry {
  keySequence;
  command;
  noRepeat;
  repeatLimit;
  background;
  topFrame;
  options;
}
```
and integrates a lot of the information of the simple `command` objects from `all_commands`.

3. The `commandHandler` in `mode_normal.js` expects an object that is
``` js
{
  command: ...
  count: ...
}
```
but in this case `command` is actually a `RegistryEntry` object.
This means that

4. In `vomnibar_page.js` we need to create a request object like follows
``` js
        await chrome.runtime.sendMessage({
          handler: "runNormalModeCommand",
          command: completion.command.registryEntry,
          // Propagate "commandModePrefixCount" to the selected command.
          count: this.commandModePrefixCount,
        });
```
All these are pointing out some issues with the current naming IMO and I think a small refactor would probably be beneficial.

ckardaris:
Hello @philc. Thanks again for looking at this PR. I have finished addressing the first round of review. Let me know if you have any additional remarks.

philc:
Thanks @ckardaris. I've been doing some further cleanup work on this and will get it merged soon.

philc:
Ok, this is now merged as of 5e31dc54f77b341f10b504dad8abcebbc82e149c! There's still some UI polish left, but it's otherwise ready for use. Nice work @ckardaris!

ckardaris:
> Ok, this is now merged as of [5e31dc5](https://github.com/philc/vimium/commit/5e31dc54f77b341f10b504dad8abcebbc82e149c)! There's still some UI polish left, but it's otherwise ready for use. Nice work @ckardaris!

Thanks for merging this @philc. I hope it will be received well by the rest of the users and possibly attract more people :)

Issue #4509, opened 2024-06-28
Style the omnibar different when a new tab is going to be opened


ManuelTS:
When the omnibar is displayed by a shortcut for a new tab, render it visually different than the omnibar replacing the URL of the current  tab. Possible solutions from the top of my head:

- + as prefixed icon before the omnibar text or input
- t+ as prefixed icon before the omnibar text or input

As I think of that, you maybe to the exact opposite, render the replace-current-tab omnibar with as vim's `:ls` command displays its files:

- "a" as prefixed icon before the omnibar text or input to signal that the current **a**ctive tab will be **a**ffected by choosing an URL

UncleSnail:
I also like this idea. If @philc agrees that it is good, I can take over implementing it if @ManuelTS would like.
Another option if we don't want to add a prefix by default would be to add a class to the vomnibar for each action that opened it. This would allow custom styles to be added to each vomnibar type and would allow users to change the color of the input box or add a prefix with `:before` text.

ManuelTS:
I see you are definitely more in the code as me, feel free to take over and to take off ;)

PR #4881, opened 2026-03-26
[feature] New tab indicator in omni bar

ckardaris:
- Add omni bar mode indicator to the left of the input field.
- Add new tab indicator '+' if the omni bar action will be completed in
  a new tab.
- Provide shortcut (alt + n) to switch between current tab and new tab
  mode.
- Provide shortcut (alt + m / alt + M) to cycle over the available omni
  bar completers.

## Description

Good evening. This MR is a POC related to #4509.

The proposed design changes may be deemed too radical, but I am submitting them here for discussion.

**Change no. 1**
The omni bar "mode" is shown to the left of the input field. Shortcuts `Alt + m/M` are used to cycle through modes (the input text is preserved). This can be useful if you realize you need to search in another mode and don't want to lose text input or press `Esc` + `<correct shortcut>`.

image

**Change no. 2**
A plus sign `+` is added next to the name of the mode if the selected action will be opened in a new tab (not counting with any modifier shortcuts like `Shift + Enter`. The shortcut `Alt + n` is used to switch this current tab / new tab mode.

image

### DIFF
--- a/background_scripts/completion/all_completers.js
+++ b/background_scripts/completion/all_completers.js
@@ -1,0 +1,1 @@
+export const completerNames = ["omni", "bookmarks", "tabs"];
--- a/background_scripts/main.js
+++ b/background_scripts/main.js
@@ -49,7 +49,7 @@ const completionSources = {
   searchEngines: new SearchEngineCompleter(),
 };
 
-const completers = {
+export const completers = {
   omni: new MultiCompleter([
     completionSources.bookmarks,
     completionSources.history,
--- a/pages/vomnibar_page.css
+++ b/pages/vomnibar_page.css
@@ -22,23 +22,39 @@ ul {
   z-index: 2139999999;
 }
 
-#vomnibar input {
+#vomnibar-search-area * {
   font-size: 20px;
   height: 34px;
   margin-bottom: 0;
   padding: 4px;
-  background-color: white;
-  color: black;
   border-radius: 3px;
   border: 1px solid #e8e8e8;
   box-shadow: #444 0px 0px 1px;
-  width: 100%;
   outline: none;
   box-sizing: border-box;
 }
 
+#vomnibar #completer-name {
+  margin-right: 8px;
+  font-size: 12px;
+  display: flex;
+  align-items: center;
+  background: unset;
+  color: #990000;
+}
+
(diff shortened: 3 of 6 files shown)


philc:
One thing I don't love about this implementation is that the query's text input and the results aren't on the same vertical grid line.

As part of merging #4880, in 37e6c09256e827e52d6ed0501f4b6318608007c4, I used the placeholder text in the Vomnibar to show which completion mode you're in. This work well for tabs, bookmarks, commands. The default case (omni) shows no placeholder.

image

philc:
I also think the "+" sign is too opaque of an indicator of what's going to happen.

Here's a rough mock of a more self-evident alternative:

image

Nothing would be shown on the right when the result is going to be opened in the same tab. It could instead say "Open in new tab" to be even more self-evident.

ckardaris:
> One thing I don't love about this implementation is that the query's text input and the results aren't on the same vertical grid line.
> 
> As part of merging #4880, in [37e6c09](https://github.com/philc/vimium/commit/37e6c09256e827e52d6ed0501f4b6318608007c4), I used the placeholder text in the Vomnibar to show which completion mode you're in. This work well for tabs, bookmarks, commands. The default case (omni) shows no placeholder.
> image

I like that. It is elegant and avoid the flickering when the bar is launched.

> I also think the "+" sign is too opaque of an indicator of what's going to happen.
> 
> Here's a rough mock of a more self-evident alternative:
> image
> 
> Nothing would be shown on the right when the result is going to be opened in the same tab. It could instead say "Open in new tab" to be even more self-evident.

This suggestion also looks nice. I will take a crack at it.

@philc What do you think about the mode changing shortcuts? `Alt+m/M` and `Alt+n`?

philc:
I personally wouldn't find those mode changing shortcuts useful, and they're hard to decide on and document. Finally, I don't think Vim has a conceptual analog that we can use as a reference.

ckardaris:
> I personally wouldn't find those mode changing shortcuts useful, and they're hard to decide on and document. Finally, I don't think Vim has a conceptual analog that we can use as a reference.

Understood. Let's keep it simple.

ckardaris:
@philc I have pushed an attempt of implementing your suggestion. Let me know what you think.

Issue #3503, opened 2020-02-17
Feature Request: Hotkey to temporarily Disable vimium


mogwai:
It would be nice not to have to disable the extension with several clicks but to do so with a hotkey?

- press CTRL+ALT+V then vimium is disabled temporarily
- You use the sites build in keyboard shortcuts
- press CTRL+ALT+V enabled vimium again

mogwai:
You can do this in `chrome://extensions/shortcuts`

gdh1995:
Currently Vimium supports `enterInsertMode` and `passNextKey` commands. See https://github.com/philc/vimium/wiki/Disabling-Vimium#temporarily-disabling-vimium and https://github.com/philc/vimium/wiki/Tips-and-Tricks#using-the-escape-key-in-inputs .

If you want a shortcut to disable Vimium totally and re-enable it, then there's no command for it.

The below is an AD: my customized version of Vimium (named Vimium C) supports this feature, and its usage is here: https://github.com/gdh1995/vimium-c/issues/113#issuecomment-580055009 .

yoyeet961:
i also want it

MissingRoberto:
Upvoting this one. Basically the use case for me is to be able to use vimium in combination with whatever shortcuts a particular website has. 

The reason is that normally the tailored shortcuts for a website are often more powerful and faster than using Vimium to do the same thing, but they usually don't cover full keyboard navigation (e.g. Jira, Github, Pocket, Gmail, etc). 

For some of those sites I end up disabling Vimium completely.

peetzweg:
Count me in! Although disabling it by pressing a button sounds great, for me personally just a blacklist of website vimium should not run on, would be saving me some keystrokes. :)

robinmoussu:
I also would like such functionality. And ideally an option to automatically toggle off vimium when entering fullscreen.

lvwarren:
Need this. Lost a chunk of work today because of contention between vimium and the vim undo and Cntl-R commands.

Ipicon:
> Yes, please. This would be awesome with Jupyter notebooks.

Actually just came here because it was annoying in Jupyter notebook

vmindru:
Actually just came here because it is annoying in google docs where Ctrl+shift+v is uesed for special paste value  notebook

sjunepark:
+1 Please

timcharper:
[...]

PR #4888, opened 2026-04-02
Add global toggle to disable/enable Vimium via keyboard shortcut

karanmahato44:
- Adds a global toggle shortcut to disable/enable Vimium across all tabs.
- Quicker than per site exclusion rules when you just need Vimium out of the way temporarily.
- Extracted `getIconSet()` to deduplicate the inline icon set in `initializeFrame`.

image

image


iacgm:
Sounds like #3503

PR #4918, opened 2026-06-20
Change the google suggestquery completion url to use https

luddet:
## Description

search_engines.js use http for the google suggestqueries url. While it seems like browsers upgrade to TLS automatically, at least Firefox do, it would be better to be explicit about it. In case someone is running a browser which doesn't upgrade, unencrypted queries would leak.

Changed http to https in three places, for Google, Google Maps and Youtube queries.


philc:
Sounds good. Thanks!

PR #4933, opened 2026-07-22
Expose data-message-name on the HUD, so it can be styled with user CSS

### LINKED ISSUE #3112: How to hide the "open link in current tab" in the lower right corner?

philc:
This is a mechanism for one to hide certain messages that are too obstructive.

Fixes #3112.

Previous PR: #4758

Each link hints related message sets a different data-message-name attribute on the `#hud` element. Here's an example selector for a user's CSS:

```css
#hud[data-message-name="link-hints/open-in-current-tab"] { display: none} 
```

Names for the other HUD message types can be added later. This just populates the link hints ones.

### DIFF
--- a/content_scripts/hud.js
+++ b/content_scripts/hud.js
@@ -73,12 +73,14 @@ const HUD = {
   },
 
   // duration - if omitted, the message will show until dismissed.
-  async show(text, duration) {
+  // messageName - an optional identifier for this message, exposed as a data attribute so the HUD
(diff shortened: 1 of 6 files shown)


roboslone:
@philc, hi! This change is not yet pushed to Chrome Web Store, right?

philc:
Correct! I typically do a release once a season, bundling several changes larger together.

PR #4940, opened 2026-09-03
Fix find mode missing text in class-based `display: contents` elements

hj-ueda:
## Problem

Find mode (`/`) fails to find text on pages like ChatGPT, where the
conversation content is wrapped in elements that get `display: contents`
from a **CSS class** rather than an inline `style` attribute.

## Cause

`getAllTextNodes()` in `content_scripts/mode_find.js` recurses into an
element when `node.checkVisibility()` is true. `checkVisibility()` returns
`false` for `display: contents` elements because they generate no box,
even though their children are visible, so e304d0a2 added a fallback:

```js
(node.checkVisibility() || node.style.display === "contents")
```

But `node.style.display` only reads the **inline style attribute**.
ChatGPT applies `display: contents` through a Tailwind `.contents` class,
so `node.style.display` is `""`, the traversal bails at those wrapper
elements, and the entire thread's text nodes are excluded from search.
On a live ChatGPT page this drops the collected text nodes from ~451 to
~18 and the query matches nothing.

## Fix

Check the **computed** style instead:

```js
(node.checkVisibility() ||
  globalThis.getComputedStyle(node).display === "contents")
```

The `||` short-circuits, so `getComputedStyle` is only called when
`checkVisibility()` is already `false` — no perf regression on the common
path.

Follow-up to e304d0a2.

## Tests

Added `tests/unit_tests/mode_find_test.js` covering find-mode text-node
traversal into `display: contents` elements set both inline and via a CSS
class (the class case fails without this fix). `./make.js test` passes
(unit 223/223, dom 109/109); `deno fmt --check` and `deno lint` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hj-ueda:
Sorry for the noise.

PR #4943, opened 2026-09-05
refactor: refactor: replace legacy getSign helper with Math.sign

paultheman:
## Description

Replace the legacy getSign helper with the native Math.sign implementation in the scroller.
This removes obsolete compatibility code for older Chrome versions.


content_scripts/scroller.js
```diff
@@ -10,23 +10,7 @@ let activatedElement = null;
...
-    return 0;
-  } else {
-    if (val < 0) {
-      return -1;
-    } else {
-      return 1;
-    }
-  }
-};
+const getSign = Math.sign;
```
philg-dev: I think there's only two calls to `getSign` in the entire codebase, right? So it would make sense to get rid of the `getSign` entirely and replace the references with calls to `Math.sign` instead.

PR #4944, opened 2026-09-06
Fix passNextKey after entering fullscreen

### LINKED ISSUE #3922: Fullscreen breaks passNextKey
**Describe the bug**
When using passNextKey command to enter fullscreen mode (e.g. on YouTube), fullscreen is entered, but Pass Next Key mode is not exited.

**To Reproduce**
1. Add `map <c-]> passNextKey` mapping in the settings and save it
1. Go to [any youtube video](https://www.youtube.com/watch?v=0_LZxYxoT3M)
1. Use `<c-]> f`. Fullscreen is entered, however "Pass Next Key" mode remains active and can only be exited by refreshing the page.

Note: using `<c-]> t` works fine entering theatre mode on YouTube, as expected, so I suspect it's the fullscreen that messes it up.

Google Chrome | 93.0.4577.82 (Official Build) (x86_64)
OS | macOS Version 11.5.2 (Build 20G95)
Vimium | 1.67

### LINKED ISSUE #4250: passNextKey does not automatically exit when used to enter Youtube fullscreen mode
**Describe the bug**
passNextKey does not automatically exit when used to enter Youtube fullscreen mode. After entering the passNextKey mode and pressing F to enter fullscreen in a youtube video, the "Pass next key." prompt keeps showing on the bottom-right corner instead of disappearing and keystrokes keep on being passed just like in insert mode. No keys or key combinations can exit the passNextKey mode, including pressing ESC, or pressing the passNextKey hotkey again. (Not even exiting fullscreen mode and pressing ESC/passNextKey, only a refresh can clear the passNextKey state.)

**To Reproduce**
1. Open any youtube video
2. Press a key bound to passNextKey (I tried `<c-f>`, `<c-g>` and `g`, all with the same issue. I'm using macOS so `<c-f>` shouldn't collide with any browser hotkeys.)
3. Press F to enter fullscreen.
4. Observe that the "Pass next key." prompt keeps showing on the bottom-right corner, and any key press will still be passed on as raw inputs instead of being intercepted and interpreted as vimium commands.

image

**Chrome:** 112.0.5615.137
**OS:** macOS 10.15.7
**Vimium:** 1.67.4

### LINKED ISSUE #4389: The passNextKey dialog box does not disappear
**Describe the bug**

If you assign passNextKey with `map <s-space> passNextKey`, etc., and then `<s-space>j` (hold down), the dialog box in the lower right corner will remain visible and not disappear until you refresh the page or switch tabs.

**To Reproduce**

1. Assign passNextKey
2. Go to any URL you like
3. Enter the key assigned to passNextKey and press and hold down `f`, etc

Browser: Vivaldi 6.4.3160.44 (Stable channel) stable (64-bit)
Vimium: 2.0.5
OS: Gentoo Linux 2.14 (kernel version 6.1.67)

maignal:
## Description
This PR fixes #4250 (and also fixes #3922 and fixes #4389) where entering fullscreen, with a key pressed right after the passNextKey command was triggered, would keep the passNextKey mode open and keep passing keys to the page.

This fix exit the common one-key case before passing the key to the page.
I wrote a unit test and I'm currently using this feature daily.

If this fix doesn't suit you, I'm happy to search for an alternative :)


philg-dev:
Looks pretty good to me!

I assume, that "prematurely" exiting the `passNextKeyMode` can't possibly have any weird side effects?

With the fix, the corresponding `keyUp` event would always go to Vimium's Normal Mode, but I think Normal Mode doesn't care about `keyUp` events, so it'll end up being passed to the page anyway - correct? Because if my assumption is incorrect, then the fix would break the `passNextKeyMode` for cases where the `keyUp` event is the relevant event that the user wants to pass to the underlying page.

At least I couldn't find any references to `keyUp` in `mode_normal.js`. I don't have time to dig deeper right now, but I'm pretty sure it's fine the way the fix implemented.

maignal:
Correct. After PassNextKeyMode exits, the matching keyup goes through the handler stack again. NormalMode only installs a keydown handler, so it does not consume that keyup. It falls through to the page normally.

